PR #11, opened 2025-01-28
add the "generate_resource" for the istio agent as tool

peterj:


python/src/kagent/tools/istio/_istio_crds.py
```diff
@@ -0,0 +1,94 @@
...
+    return await _generate_crd(PEER_AUTHENTICATION_PROMPT, policy_description)
+
+
+async def _generate_virtual_service_crd(
+    policy_description: Annotated[str, "Detailed description of the VirtualService to generate YAML for"],
+) -> str:
+    return await _generate_crd(VIRTUAL_SERVICE_PROMPT, policy_description)
+
+
+class IstioResources(Enum):
```
EItanya: This is a nit but imo this type should be located in `prompts` or something related to the CRD gen. Anything exported from this package should be a tool

python/src/kagent/tools/istio/_istio_crds.py
```diff
@@ -0,0 +1,94 @@
...
+from autogen_core.models import SystemMessage, UserMessage
+from autogen_core.tools import FunctionTool
+from autogen_ext.models.openai import OpenAIChatCompletionClient
+
+from .prompts import AUTH_POLICY_PROMPT, GATEWAY_PROMPT, PEER_AUTHENTICATION_PROMPT, VIRTUAL_SERVICE_PROMPT
+
+
+def get_model_client():
+    # TODO: We should have a way to configure externally somehow.
+    return OpenAIChatCompletionClient(
```
EItanya, line 19: I have some ideas on how to do this, but I'm not worried about it for now, defaulting to this will work in 99% of our testing use-cases

python/src/kagent/tools/istio/prompts/base.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+
+
+class IstioResources(Enum):
```
npolshakova, line 4: Need to fix the import in the tools: https://github.com/kagent-dev/kagent/blob/953b07ab145edd78e69d4aaf924ecc2afb5fd5e3/python/src/kagent/tools/istio/_istio_crds.py#L13

PR #12, opened 2025-01-28
add Kube controller

ilackarms:
initializes the kube controller component of kagent

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -1,6 +1,6 @@
 module github.com/kagent-dev/kagent/go
 
-go 1.23.3
+go 1.23.5
 
 require (
 	github.com/onsi/ginkgo/v2 v2.21.0


controller/api/v1alpha1/autogenteam_types.go
```diff
@@ -0,0 +1,85 @@
...
+type SelectorTeamConfig struct {
+	SelectorPrompt string `json:"selectorPrompt"`
+	ModelConfig    string `json:"modelConfig"`
+}
+
+type TerminationCondition struct {
+	// ONEOF: maxMessageTermination, textMentionTermination, orTermination
+	MaxMessageTermination  *MaxMessageTermination  `json:"maxMessageTermination,omitempty"`
+	TextMentionTermination *TextMentionTermination `json:"textMentionTermination,omitempty"`
+	OrTermination          *OrTermination          `json:"orTermination,omitempty"`
```
yuval-k: including OrTermination here creates a recursive datastructure. is that the intent?

controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -0,0 +1,56 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AutogenModelConfigSpec defines the desired state of AutogenModelConfig.
+type AutogenModelConfigSpec struct {
+	Model           string `json:"model"`
+	APIKeySecret    string `json:"apiKeySecret"`
```
yuval-k: should this be a ref to a secret?

controller/api/v1alpha1/autogenteam_types.go
```diff
@@ -0,0 +1,85 @@
...
+	TerminationCondition TerminationCondition `json:"terminationCondition"`
+	MaxTurns             int64                `json:"maxTurns"`
+}
+
+type SelectorTeamConfig struct {
+	SelectorPrompt string `json:"selectorPrompt"`
+	ModelConfig    string `json:"modelConfig"`
+}
+
+type TerminationCondition struct {
```
npolshakova: Maybe "TerminationType" and add validation:
```
// +kubebuilder:validation:XValidation:message="There must one termination type set",rule="1 == (self.maxMessageTermination != null?1:0) + (self.textMentionTermination != null?1:0) + (self.orTermination != null?1:0)"

```

controller/api/v1alpha1/autogentool_types.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// AutogenToolSpec defines the desired state of AutogenTool.
+type AutogenToolSpec struct {
+	Description string `json:"description,omitempty"`
```
npolshakova: Are there additional things we need to define on the tool? https://github.com/microsoft/autogen/blob/2f1684b698e83418d47f54309b426ddfd022f3ee/python/packages/autogen-core/src/autogen_core/tools/_base.py#L77

controller/api/v1alpha1/autogentool_types.go
```diff
@@ -0,0 +1,57 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// AutogenTool is the Schema for the autogentools API.
+type AutogenTool struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   AutogenToolSpec   `json:"spec,omitempty"`
+	Status AutogenToolStatus `json:"status,omitempty"`
```
npolshakova: What would a tool status contain? (Which agents are using it?)

controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// AutogenAgentSpec defines the desired state of AutogenAgent.
+type AutogenAgentSpec struct {
+	Name          string   `json:"name,omitempty"`
+	Description   string   `json:"description,omitempty"`
+	SystemMessage string   `json:"systemMessage,omitempty"`
+	Tools         []string `json:"tools,omitempty"`
```
npolshakova: This would match the built-in tools name? (ie. `k8s.apply_manifest`, `istio.generate_resource`, etc.)

controller/api/v1alpha1/autogentool_types.go
ilackarms: tool is a placeholder for now; we can fill it out once we have some examples of user-created tools to support

controller/api/v1alpha1/autogenagent_types.go
ilackarms: yes or it could also refer to the name of a user-provided tool (defined as a crd)

go/controller/api/v1alpha1/autogenteam_types.go
```diff
@@ -0,0 +1,91 @@
...
+	Participants         []string             `json:"participants"`
+	Description          string               `json:"description"`
+	SelectorTeamConfig   SelectorTeamConfig   `json:"selectorTeamConfig"`
+	TerminationCondition TerminationCondition `json:"terminationCondition"`
+	MaxTurns             int64                `json:"maxTurns"`
+}
+
+type SelectorTeamConfig struct {
+	SelectorPrompt string `json:"selectorPrompt"`
+	ModelConfig    string `json:"modelConfig"`
```
EItanya, line 37: Is this a reference to a `ModelSpec`? I think this may be overcomplicated for the first go. Agreed that eventually it might be nice to have this defined in one spot, but in 99% of cases defining it inline here is probably easier for the user. What do you think?

EItanya: Also, will this be the model config used for the whole run, or just for the selector itself?

go/controller/api/v1alpha1/autogenteam_types.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// +kubebuilder:validation:XValidation:message="There must one termination type set",rule="1 == (self.maxMessageTermination != null?1:0) + (self.textMentionTermination != null?1:0) + (self.orTermination != null?1:0)"
```
EItanya, line 40: I think this API may be a tad overcomplicated, we don't need to fix it now but it may be worth simplifying it in the future. Realistically we can just list the termination conditions, and then have a `join_condition` or something like that which defaults to `or`, but can set to `and`.

yuval-k:
> Any reason we're not starting our controllers off with `krt`? It will handle a lot of the boiler plate in terms of finding the dependents etc @yuval-k

How many dependencies are we talking about?
My understanding is that we have very few, so my hunch is to stay the course and use controller-runtime.

go/controller/api/v1alpha1/autogenteam_types.go
ilackarms: its a reference to model config
i dont see why users would want to redefine this in every single team cr when the config will be the same when they are using a single llm / api key

go/controller/api/v1alpha1/autogenteam_types.go
ilackarms: it's modeled directly off of autogen. we can always change in the future but i think to keep it simple not overthink / redefine their api with new abstractions

PR #13, opened 2025-01-29
initial CLI. still work in progress, but basic list commands work

peterj:


peterj:
I wanted to get a preliminary review to make sure I am going in the right direction

peterj:
I can only assign one reviewer, but @ilackarms, @EItanya , @yuval-k  please take a look as well.

cli/internal/cli/team.go
```diff
@@ -0,0 +1,141 @@
...
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/kagent-dev/kagent/cli/internal/api"
+	"github.com/kagent-dev/kagent/cli/internal/config"
+	"github.com/spf13/cobra"
+)
+
+func newTeamCmd() *cobra.Command {
```
npolshakova, line 13: Is there a reason to have a cli tool to create a team if we have a CR representing teams: https://github.com/kagent-dev/kagent/blob/main/examples/1-team-resources.yaml#L2

npolshakova: Same question for list/get, why would you use:

```
kagent team list
```

Versus:
```
kubectl get AutogenTeam -A -oyaml
```

cli/internal/cli/team.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return err
+	}
+
+	teamID, err := strconv.Atoi(args[0])
+	if err != nil {
+		return fmt.Errorf("invalid team ID: %s", args[0])
+	}
+
+	client := api.NewClient(cfg.APIURL, cfg.WSURL)
```
npolshakova, line 118: For the status cli commands, do we want to check the status of the resource or reach out to the api server directly?

cli/internal/cli/run.go
```diff
@@ -0,0 +1,154 @@
...
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/kagent-dev/kagent/cli/internal/api"
+	"github.com/kagent-dev/kagent/cli/internal/config"
+	"github.com/spf13/cobra"
+)
+
+func newRunCmd() *cobra.Command {
```
npolshakova, line 15: Should the cli also be able to install the controller (like `istioctl install`) or do we prefer only having a helm way to install?

cli/internal/cli/root.go
```diff
@@ -0,0 +1,38 @@
...
+
+func NewRootCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "kagent",
+		Short: "kagent CLI",
+		Long:  `A CLI tool to interact with kagent`,
+	}
+
+	// Add global flags
+	cmd.PersistentFlags().String("api-url", "http://localhost:8081/api", "Backend API URL")
```
npolshakova, line 16: So the CLI can't really do anything until the server is up and running? We might want to add a check to help figure out what the api-url is from the config in the cluster (or even just check if the server is running)

cli/internal/api/run.go
```diff
@@ -0,0 +1,45 @@
+package api
+
+import "fmt"
+
+func (c *Client) CreateRun(req *CreateRunRequest) (*Run, error) {
```
npolshakova: I think there should be a better UX than having to get the session id in order to create a run? 
```
❯ ./bin/kagent run create
Error: accepts 1 arg(s), received 0
Usage:
  kagent run create [session-id] [flags]
```

Maybe we can save the last used session-id?

cli/internal/cli/team.go
peterj: probably not in the long run; once the controller is there,we don't need this (+ I haven't tested this yet :))

cli/internal/cli/team.go
peterj: once we have the API server running we'll probably go through the API server.

cli/internal/cli/run.go
peterj: good question - I am not sure which is the better option.

cli/internal/cli/root.go
peterj: agree, once we have the controller we can add the check

cli/internal/api/run.go
peterj: perhaps `kagent session set`? or something similar to how kube context is?

cli/internal/cli/team.go
EItanya: I'm not sure about list/get. This CLI could print them in some sort of prettier way, or something which is more appealing. idk just spitballing

cli/internal/cli/run.go
EItanya: In the future it should, that does not need to happen yet though

cli/internal/cli/root.go
EItanya: In general the url should actually be discovered by looking at `kube` unless explicitly set but that can be added later

cli/internal/api/run.go
EItanya: 100%, this is an implementation detail. IMO the CLI should create a new session each time it runs by default, unless there's some reason the user wants to explicitly tie it to an existing one which feels unlikely.

cli/internal/cli/team.go
peterj: if you agree, I'll leave this until we have the controller up and running. once that's ready, we can discuss if we want a nicer output or if we want to remove it completely.

cli/internal/cli/run.go
peterj: added a placeholder for now.

cli/bin/kagent
EItanya, line 1: This shouldn't be committed

PR #17, opened 2025-01-31
implement kagent run and team team

peterj:
With the PR we can create an team through cli (`kagent team create -f <file.json>`) (I included a sample json as well). 

Once created you can use `kagent run create <teamname> "How many pods are in my cluster"` to run the team. You'll get prompts you can respond to through command line.

### DIFF
--- a/examples/team-with-user.json
+++ b/examples/team-with-user.json
@@ -14,80 +14,82 @@
           "component_type": "agent",
           "version": 1,
           "component_version": 1,
-          "description": null,
           "config": {
-            "description": "An agent that provides assistance with ability to use tools.",
+            "name": "PlanningAgent",
             "model_client": {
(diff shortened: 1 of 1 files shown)


cli/internal/cli/team.go
```diff
@@ -75,7 +81,33 @@ func runTeamList(cmd *cobra.Command, args []string) error {
...
+	if err := json.Unmarshal(jsonFile, &team); err != nil {
+		return fmt.Errorf("error parsing YAML: %w", err)
+	}
+
+	cfg, err := config.Get()
+	if err != nil {
+		return err
+	}
+
+	client := api.NewClient(cfg.APIURL, cfg.WSURL)
```
EItanya, line 104: I know this isn't necessarily bound to this PR, but why do we need to pass in two URLs? won't it basically always be the same host, just different prefix/paths?

PR #23, opened 2025-02-11
Kube helm chart

ilackarms:


Makefile
```diff
@@ -0,0 +1,34 @@
+# Image configuration
+DOCKER_REGISTRY ?= gcr.io
+DOCKER_REPO ?= solo-io-kagent
+CONTROLLER_IMAGE_NAME ?= autogenstudio-controller
+WEB_IMAGE_NAME ?= autogenstudio-web
+VERSION ?= $(shell git describe --tags --always --dirty)
+CONTROLLER_IMAGE_TAG ?= $(VERSION)
+WEB_IMAGE_TAG ?= $(VERSION)
```
peterj, line 8: is this the python backend code that will be in WEB_IMAGE?

(we'll have another container for the UI, I created an issue to create a dockerfile + update the helm once this is merged)

peterj: or should we run both the UI and the backend in the same container?

helm/values.yaml
```diff
@@ -0,0 +1,52 @@
...
+      cpu: 100m
+      memory: 256Mi
+    limits:
+      cpu: 1000m
+      memory: 1Gi
+
+service:
+  type: ClusterIP
+  port: 80
+  targetPort: 8081  # Assuming the web UI runs on port 8081
```
peterj: the backend runs on 8081 and the UI runs on 8001

Makefile
ilackarms: there are 2 containers, same pod. this PR handles both

helm/values.yaml
ilackarms: what backend? not sure what you mean. autogenstudio-ui runs on 8081

peterj: the autogenstudio-ui only runs the python backend; we also have the frontend part that I merged this morning and that one runs on 8001

PR #26, opened 2025-02-12
Use hosted db for docs

npolshakova:
- Removes python scripts (@djannot we should move the typescript database creation to the kagent-dev project along with the README)
- Removes qdrant support (will initially only support sqlite vector db)
- Adds config to provide custom db path
- Adds an example with all docs (kubernetes, istio, argo)

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -19,6 +19,7 @@ dependencies = [
   "python-dotenv>=1.0.1",
   "pyyaml>=6.0.2",
   "mcp>=1.2.0",
+  "sqlite-vec>=0.1.0",
 ]
 
 [project.optional-dependencies]
--- a/python/src/kagent/tools/docs/query_documentation.py
+++ b/python/src/kagent/tools/docs/query_documentation.py
@@ -99,36 +99,6 @@ class SQLiteDownloader:
         return db_path
 
 
-class BaseTool(BaseTool[BaseModel, Any], Component[Config]):
-    """Base class for all Documentation search tools."""
-
-    component_type = "tool"
-    component_config_schema = Config
-
-    @property
-    def component_provider_override(self) -> str:
-        """Build the component provider path from the class name."""
-        return f"kagent.tools.docs.{self.__class__.__name__}"
-
-    def __init__(
-        self,
-        config: Config,
-        input_model: Type[BaseModel],
-        description: str,
-    ) -> None:
-        super().__init__(input_model, Any, self.__class__.__name__, description)
-        self.config = config
-
-    def _to_config(self) -> Config:
-        """Convert to config object."""
-        return self.config.copy()
-
-    @classmethod
-    def _from_config(cls, config: Config) -> "BaseTool":
-        """Create instance from config."""
-        raise NotImplementedError("Use specific tool implementations")
-
-
 class QueryInput(BaseModel):
     query: str = Field(
         description="The search query to use for finding relevant documentation. Be specific and include relevant keywords."
@@ -145,13 +115,15 @@ class QueryInput(BaseModel):
     )
 
 
-class QueryTool(BaseTool):
+class QueryTool(BaseTool, Component[Config]):
     """Tool for querying documentation for a specific product."""
 
-    description = """Searches a vector database for relevant documentation related to various software projects."""
+    component_type = "tool"
+    component_config_schema = Config
+    component_provider_override = "kagent.tools.docs.QueryTool"
+    _description = """Searches a vector database for relevant documentation related to various software projects."""
 
     def __init__(self, config: Config) -> None:
-        super().__init__(config=config, input_model=QueryInput, description=self.description)
         # Initialize SQLite downloader with base S3 URL and product mapping if override is not provided
         self.db_downloader = SQLiteDownloader(
             base_url=config.docs_download_url
@@ -160,6 +132,9 @@ class QueryTool(BaseTool):
         )
         # Initialize OpenAI with API key from config
         self.openai = OpenAI(api_key=config.openai_api_key)
+        self.config: Config = config
+
+        super().__init__(QueryInput, BaseModel, "query_tool", description=self._description)
 
     async def run(self, args: QueryInput, cancellation_token: CancellationToken) -> Any:
         db_path = self.config.docs_base_path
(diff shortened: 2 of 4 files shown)


python/notebooks/all_docs_agent copy.ipynb
EItanya, line 1: Why is this suffixed with `copy`?

python/src/kagent/tools/docs/query_documentation.py
```diff
@@ -19,66 +30,115 @@ def __init__(self, chunk_id, distance, content, url=None, **kwargs):
...
+            raise ValueError("Product name cannot be empty")
+        if product_name not in self.product_map:
+            raise ValueError(
+                f"Unsupported product: {product_name}. Supported products: {', '.join(self.product_map.keys())}"
+            )
+
+    def get_db_path(self, product_name: str) -> Path:
+        """Get local path for cached database."""
+        self.validate_product(product_name)
+        cache_dir = Path.home() / ".cache" / "doc-query"
```
EItanya: `.kagent/cache`?

python/pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
   "python-dotenv>=1.0.1",
   "pyyaml>=6.0.2",
   "mcp>=1.2.0",
+  "qdrant-client>=1.13.2",
```
peterj: should we remove this or are we still using it?

peterj: nevermind, looks like we are :)

npolshakova: Will remove now- got rid of the qdrant support for the initial tool

python/src/kagent/tools/docs/query_documentation.py
```diff
@@ -19,95 +43,230 @@ def __init__(self, chunk_id, distance, content, url=None, **kwargs):
...
+                raise
+
+        return db_path
+
+
+openai = OpenAI(api_key=os.environ["OPENAI_API_KEY"])
+
+# Initialize SQLite downloader with base S3 URL and product mapping
+db_downloader = SQLiteDownloader(
+    base_url=os.environ.get("DB_BASE_URL", "https://doc-sqlite-db.s3.sa-east-1.amazonaws.com"),
```
peterj: are we hardcoding the URL like this?

python/src/kagent/tools/docs/query_documentation.py
```diff
@@ -19,95 +43,230 @@ def __init__(self, chunk_id, distance, content, url=None, **kwargs):
...
+
+    conn = sqlite3.connect(str(db_path))
+    conn.enable_load_extension(True)
+    sqlite_vec.load(conn)
+    conn.enable_load_extension(False)
+
+    cursor = conn.cursor()
+
+    query = """
+        SELECT *, distance FROM vec_items WHERE embedding MATCH ?
```
peterj: use single `" "` as this creates a multiline string

python/src/kagent/tools/docs/query_documentation.py
npolshakova: The current behavior is if the db path isn't provided, we'll use the base url to download it. Open to changing it based on what we want the behavior to be. The two uses cases are:
1) Running locally in a jupyter notebook to test out the agent
2) Agents bundled with the controller should already have the databases setup

peterj: ooh, so the path the user can provide is a local path?and then if not available we default to the URL (which is owned by us, right?). 

Could the path user provide be a URL as well? I wonder if it would make sense to have the config item that's the URL instead? and then we default to downloading it (we'd have to download these files at some point anyway, right?)

python/src/kagent/tools/docs/query_documentation.py
```diff
@@ -19,95 +43,230 @@ def __init__(self, chunk_id, distance, content, url=None, **kwargs):
...
+openai = OpenAI(api_key=os.environ["OPENAI_API_KEY"])
+
+# Initialize SQLite downloader with base S3 URL and product mapping
+db_downloader = SQLiteDownloader(
+    base_url=os.environ.get("DB_BASE_URL", "https://doc-sqlite-db.s3.sa-east-1.amazonaws.com"),
+    product_map=PRODUCT_DB_MAP,
+)
+
+
+class BaseTool(BaseTool[BaseModel, Any], Component[Config]):
```
EItanya: Why is there a `BaseTool` here? You should just be able to use the upstream `BaseTool` impl

python/src/kagent/tools/docs/query_documentation.py
```diff
@@ -19,95 +43,230 @@ def __init__(self, chunk_id, distance, content, url=None, **kwargs):
...
-        """
-        params = [np.array(query_embedding, dtype=np.float32).tobytes()]
+            except requests.exceptions.RequestException as e:
+                logging.error(f"Error downloading database for {product_name}: {e}", file=sys.stderr)
+                raise
+
+        return db_path
+
+
+openai = OpenAI(api_key=os.environ["OPENAI_API_KEY"])
```
EItanya: I think it would make sense if this came from `config`?

python/src/kagent/tools/docs/query_documentation.py
npolshakova: Yeah good- point, let's make it an optional arg and set the s3 bucket as the default?

python/notebooks/all_docs_agent copy.ipynb
npolshakova: Ah, I removed the old individual doc search examples (istio_docs, kubernetes_docs) to combine them in one example. I copied the Istio example to create the file. Will remove.

python/src/kagent/tools/docs/query_documentation.py
npolshakova: Yep, can update!

python/src/kagent/tools/docs/query_documentation.py
npolshakova: This was similar to what @peterj was doing with the prometheus tool. Not sure if we really need it, but this part defines the component provider path from the class name:
```
    @property
    def component_provider_override(self) -> str:
        """Build the component provider path from the class name"""
        return f"kagent.tools.docs.{self.__class__.__name__}"
```

PR #27, opened 2025-02-12
Github Action to Test K8S Challenges

jmunozro:
There is a new action `Run PE Test`, including 24 custom kubernetes challenges, that will do the following steps:

1. `Run challenge` Execute some `scenario` from a file located in `.github/data/agent-framework/<scenario-name>/README.md`. An scenario is a markdown file that is supposed to be created by a human or an IA
2. `Test baseline` Use a test.js, located in `.github/data/agent-framework/<scenario-name>/test.js` file to test the scenario is valid. All dependencies for this file should be contained within the scenario folder. This is supposed to be always SUCCESSFUL
3. `Break environment` Using a challenge definition file, located in `.github/data/agent-framework/<scenario-name>/<challenge-name>.yaml`, the scenario is broken.
4. `Test again after breaking`. Same test, but this this it is expected to FAIL.
5. `Fix it with Agent Framework`. Using the prompt from the challenge definition file, solve the issue.
6. `Test final state`. Same test as before.

After all 24 challenges have been processed, a summary job is printing the results.

As an example, I ran these same 24 challenges with a GPT-4o agent [here](https://github.com/solo-io/product-excellence/actions/runs/13289929466) with the results:

1st try: 10/24 (40%) challenges solved
2nd try: (54%) 13/24
3rd try: (58%) 14/24

### DIFF
--- a/.github/data/agent-framework/.gitignore
+++ b/.github/data/agent-framework/.gitignore
@@ -1,1 +1,3 @@
 node_modules
+*.failure
+*.success
--- a/.github/data/agent-framework/run-agent-framework.test.sh
+++ b/.github/data/agent-framework/run-agent-framework.test.sh
@@ -10,8 +10,11 @@ CLUSTER1=cluster1
 test_dir=$(pwd)
 # Loop through each challenge defined in the .github/data/agent-framework directory
 for scenario_dir in *; do
+  if [ ! -d "$scenario_dir" ]; then
+    continue
+  fi
   pushd $scenario_dir
-  npm i
+  pnpm i || npm i
   echo "pwd=$(pwd)"
   for challenge_file in *.yaml; do
     # Extract the challenge name and description from the YAML metadata file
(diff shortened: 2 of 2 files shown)


jmunozro:
I added a file `.github/data/agent-framework/run-agent-framework.test.sh` to execute an equivalent of the GH action locally.

.github/data/agent-framework/scenario1/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "procgen",
+  "version": "0.0.1",
+  "description": "Solo Procedure Generator",
+  "main": "procgen.js",
+  "scripts": {
+    "cilium-intro-default": "node procgen.js -d cilium-intro -f cilium-intro/workshops/default.yaml -o dist/cilium-intro-default --overwrite",
```
peterj: why do wee need all this?

jmunozro: we dont. I've reduced the package.json to the minimum

PR #31, opened 2025-02-17
simplify UI, add tool selection

peterj:


ui/src/components/create/ToolsSection.tsx
```diff
@@ -2,23 +2,58 @@ import { Button } from "@/components/ui/button";
...
 export const ToolsSection = ({ allTools, selectedTools, setSelectedTools, isSubmitting }: ToolsSectionProps) => {
   const [showToolSelector, setShowToolSelector] = useState(false);
   const [searchTerm, setSearchTerm] = useState("");
   const [selectedInDialog, setSelectedInDialog] = useState<Tool[]>([]);
+  const [configTool, setConfigTool] = useState<Tool | null>(null);
+  const [showConfig, setShowConfig] = useState(false);
+  const [fieldErrors, setFieldErrors] = useState<Record<string, string>>({});
+
+  const getToolType = (provider: string): keyof typeof TOOL_CONFIGS | "unknown" => {
+    if (provider.startsWith("kagent.tools.prometheus")) return "kagent.tools.prometheus";
```
npolshakova, line 28: @peterj is this where we add the tools manually?

ui/src/lib/data.ts
```diff
@@ -6,47 +6,211 @@ export const AVAILABLE_MODELS: Model[] = [
...
+
+const defaultPrometheusConfig: PrometheusToolConfig = {
+  base_url: "http://localhost:9090/api/v1",
+  username: undefined,
+  password: undefined,
+}
+
+// Used to render the UI
+export const TOOL_CONFIGS = {
+  'kagent.tools.prometheus': createConfigMetadata<PrometheusToolConfig>(defaultPrometheusConfig)
```
npolshakova, line 49: Other tool types (istio, kubernetes, etc.) will be listed here?

ui/src/components/create/ToolsSection.tsx
peterj: yes here as well as in the data.ts

ui/src/lib/data.ts
peterj: exactly (for now). Ideally we'd either have backend where we can read these from OR expose an MCP server with these tools somehow (from the backend and/or controller somehow)

PR #35, opened 2025-02-18
more istio tools

peterj:


python/src/kagent/tools/istio/__init__.py
```diff
@@ -1,7 +1,33 @@
-from ._istio_crds import GenerateResource
+from ._istio_crds import Config, IstioCRDTool
```
EItanya: Can we prefix/suffix this `Config` to relate which type it belongs to

PR #36, opened 2025-02-19
add more k8s tools

peterj:


python/src/kagent/tools/k8s/_kubectl.py
```diff
@@ -1,14 +1,180 @@
...
-from autogen_core import CancellationToken, Component
-from autogen_core.tools import BaseTool, FunctionTool
-from pydantic import BaseModel
+from autogen_core.tools import FunctionTool
 
 from .._utils import create_typed_fn_tool
 from ..common.shell import run_command
 
 
+class KubernetesResource(Enum):
```
EItanya: I don't think we should enumerate the resource types here, there's no reason these commands shouldn't be able to handle custom resources and this would limit them.

PR #39, opened 2025-02-19
Swarm + magentic teams support

ilackarms:


go/autogen/api/types.go
```diff
@@ -77,10 +77,13 @@ type TeamConfig struct {
...
+	ModelClient *ModelComponent `json:"model_client,omitempty"`
 	// SelectorGroupChat specific fields
-	ModelClient          *ModelComponent `json:"model_client,omitempty"`
-	SelectorPrompt       string          `json:"selector_prompt"`
-	AllowRepeatedSpeaker bool            `json:"allow_repeated_speaker"`
+	SelectorPrompt       string `json:"selector_prompt"`
+	AllowRepeatedSpeaker bool   `json:"allow_repeated_speaker"`
+	// MagentoOneGroupChat specific fields
+	MaxStalls         int    `json:"max_stalls"`
+	FinalAnswerPrompt string `json:"final_answer_prompt"`
```
EItanya, line 86: Should we break these into sub-types so it's more clear?

go/controller/internal/autogen/planning-agent-system-prompt.txt
```diff
@@ -0,0 +1,288 @@
...
+- Resource optimization
+
+3. Technical Integration
+- Tool coordination
+- Workflow optimization
+- Interface management
+- Data flow control
+- Output consolidation
+
+Available Agent Types:
```
peterj, line 26: would we want to 'hardcode' the agent types in this way? Also, does it add any value? (same for the dev agents, API integration etc.).

go/controller/internal/autogen/planning-agent-system-prompt.txt
```diff
@@ -0,0 +1,288 @@
...
+- Model Development Expert
+- Data Engineering Expert
+- MLOps Specialist
+- Vector Database Expert
+- Training Pipeline Expert
+
+Planning Framework:
+
+1. Task Analysis
+```yaml
```
peterj, line 59: is this yaml used anywhere or is it just as an example of how the planning agent should structure the responses? (I've seen people do JSON as well and XML, not sure which one works better or if it even works better :))

go/autogen/api/types.go
ilackarms: this is the json structure autogen expects; we can use embedded structs to make it clearer to reader

go/controller/internal/autogen/planning-agent-system-prompt.txt
ilackarms: we can make this more dynamic

go/controller/internal/autogen/planning-agent-system-prompt.txt
ilackarms: i think it helps the LLM to output something structured but we can definitely iterate on this

PR #42, opened 2025-02-19
Support running kagent tools as MCP tools

EItanya:
This PR adds the ability to run kagent tools as MCP tools. Currently all of the tools are shipped in a single built CLI called `kagent`.  A user could use a tool like `uvx` to run `kagent istio` or `kagent k8s` to run the sets of tooling that we expose.


python/src/kagent/tools/prometheus/_prometheus.py
```diff
@@ -231,15 +231,8 @@ async def run(self, args: LabelValuesInput, cancellation_token: CancellationToke
...
-
-class TargetState(str, Enum):
-    ACTIVE = "active"
-    DROPPED = "dropped"
-    ANY = "any"
-
-
 class TargetsInput(BaseModel):
-    state: Optional[TargetState] = Field(default=None, description="Target state filter")
+    state: Optional[str] = Field(default=None, description="Target state filter, must be one of: active, dropped, any")
```
peterj, line 235: I am not 100% sure if the enums actually get translated to the OpenAI tools schema correctly, so it shows the possible values; that's why I used Enum. but I am fine removing it (I have it in my k8s tools as well)

PR #45, opened 2025-02-20
add a promql tool

peterj:


python/src/kagent/tools/prometheus/_promql.py
```diff
@@ -0,0 +1,211 @@
...
+        )
+
+    async def run(self, args: PromQLToolInput, cancellation_token: CancellationToken) -> BaseModel:
+        query_description = args.query_description
+
+        try:
+            if cancellation_token.is_cancelled():
+                raise Exception("Operation cancelled")
+
+            result = await self._model_client.create(
```
EItanya, line 198: Since this is a tool can we pass a schema to `openai` so that the results are consistent. That way anyone using this tool can expect a specific schema.

peterj: on a response you mean?

Issue #38, opened 2025-02-19
run the ui in the cluster


peterj:
add the docker image and make it run & connect to the backend running in the cluster

peterj:
Done.

Issue #49, opened 2025-02-20
publish the images and helm chart


peterj:

peterj:
Use ghcr.io and kagent-dev for the repository

EItanya:
Manually pushing works: https://github.com/orgs/kagent-dev/packages

Still need to add workflows for release

EItanya:
Pushing a tag of the format `v*.*.*` will now trigger a release flow.

For example https://github.com/kagent-dev/kagent/releases/tag/v0.0.3

Issue #25, opened 2025-02-12
add docker image for the UI, update the Helm chart


peterj:

EItanya:
Closed by #53

PR #53, opened 2025-02-22
making the UI and backend work in kubernetes

peterj:
This PR enables UI to run inside the cluster.

I had to combine the autogen backend and the UI into the same container, so they are behind a proxy (I gave up on trying making the requests from the server side, it just didn't work). 

The PR includes bunch of UI changes (mostly refactoring + fixing backend URLs based on whether we're running in dev (locally) or production (in the cluster)). 

The second part is the Dockerfile for the combined image (in the root folder), updated makefile and changed to Helm to add the container, mount the openai secret etc.

### DIFF
--- a/helm/templates/secret.yaml
+++ b/helm/templates/secret.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.openai.apiKey }}
 apiVersion: v1
 kind: Secret
 metadata:
@@ -7,3 +8,4 @@ metadata:
 type: Opaque
 data:
   {{ .Values.openai.secretKey }}: {{ .Values.openai.apiKey | b64enc }}
+{{- end }}
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -37,7 +37,7 @@ app:
       cpu: 1000m
       memory: 1Gi
 
-appService:
+service:
   type: ClusterIP
   port: 80
   targetPort: 80
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -46,7 +46,7 @@ spec:
                   key: {{ .Values.openai.secretKey }}
           ports:
             - name: http
-              containerPort: {{ .Values.appService.targetPort }}
+              containerPort: {{ .Values.service.targetPort }}
               protocol: TCP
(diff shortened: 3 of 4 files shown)


peterj:
Tagging you guys here for review @EItanya  + @ilackarms (since you can only assign one reviewer)

peterj:
Fixes #38, #25, #8

helm/values.yaml
```diff
@@ -15,11 +23,11 @@ controller:
       cpu: 500m
       memory: 512Mi
 
-web:
+app:
   image:
     registry: gcr.io
-    repository: solo-io-kagent/autogenstudio-web
-    tag: ""
+    repository: solo-io-kagent/kagent-app
```
EItanya, line 29: Doesn't need to happen now, but all of these should be `ghcr.io` and `kagent-dev`. We are going to host on GitHub

helm/templates/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
EItanya, line 2: nit: we should only create this secret if the user sets `openai.apikey`?

helm/templates/appService.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "autogenstudio.fullname" . }}
+  name: {{ include "autogenstudio.fullname" . }}-app
   labels:
     {{- include "autogenstudio.labels" . | nindent 4 }}
 spec:
-  type: {{ .Values.service.type }}
+  type: {{ .Values.appService.type }}
```
EItanya: Why this rename, I don't like it

Dockerfile
EItanya, line 1: Just to be clear, this Dockerfile runs the UI and the apiserver now?

helm/values.yaml
peterj: Added to #49

Dockerfile
peterj: yes

peterj: runs the UI and the autogen backend

PR #60, opened 2025-02-25
Single agent teams

ilackarms:


go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -8,50 +8,81 @@ import (
...
+			APIVersion: "agent.ai.solo.io/v1alpha1",
+		},
+		Spec: v1alpha1.AutogenTeamSpec{
+			Participants:         []string{agent.Name},
+			Description:          agent.Spec.Description,
+			RoundRobinTeamConfig: &v1alpha1.RoundRobinTeamConfig{},
+			TerminationCondition: v1alpha1.TerminationCondition{
+				MaxMessageTermination: &v1alpha1.MaxMessageTermination{MaxMessages: 1},
+			},
+			MaxTurns: 1,
```
EItanya: So we don't want any max turns here, or max messages. The termination will be controlled by the UI or the CLI. We should use this type for that: https://github.com/EItanya/autogen/blob/main/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py#L12-L46

helm/templates/modelconfig.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: agent.ai.solo.io/v1alpha1
```
EItanya, line 1: I will make an issue for this but we need to switch to `kagent.dev`

PR #65, opened 2025-02-25
Add helm tools

EItanya:


python/src/kagent/tools/argo/_argo_crds.py
```diff
@@ -106,42 +106,16 @@ async def _generate_crd(
         except Exception as e:
             return f"Error generating policy: {str(e)}"
 
-    async def _generate_crd(
```
peterj, line 109: what's the reason for removing this one?

EItanya: It was declared twice, I think a cursor issue

PR #83, opened 2025-03-02
Task agent

EItanya:
The `SocietyOfMindAgent` is mostly copied from upstream with a couple of small additions which I am intending to try and upstream.

The `TextMessageTermination` is due to merge upstream soon so we can delete it then.

### DIFF
--- a/python/src/kagent/agents/_society_of_mind_agent.py
+++ b/python/src/kagent/agents/_society_of_mind_agent.py
@@ -156,7 +156,6 @@ class SocietyOfMindAgent(BaseChatAgent, Component[SocietyOfMindAgentConfig]):
         # Add the messages to the context.
         for message in messages:
             self._context.append(message)
-        # Run the team of agents.
         task = self._context
 
         # Run the team of agents.
--- a/python/src/kagent/tools/helm/_helm.py
+++ b/python/src/kagent/tools/helm/_helm.py
@@ -63,12 +63,6 @@ helm_list_releases = FunctionTool(
     description="""
 This command lists all of the releases for a specified namespace (uses current namespace context if namespace not specified).
 
-By default, it lists only releases that are deployed or failed. Flags like
-'--uninstalled' and '--all' will alter this behavior. Such flags can be combined:
-'--uninstalled --failed'.
-
-By default, items are sorted alphabetically. Use the '-d' flag to sort by
-release date.
 
 If the --filter flag is provided, it will be treated as a filter. Filters are
 regular expressions (Perl compatible) that are applied to the list of releases.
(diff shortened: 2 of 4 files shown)


python/src/kagent/agents/_society_of_mind_agent.py
```diff
@@ -0,0 +1,283 @@
...
+        assert response is not None
+        return response
+
+    async def on_messages_stream(
+        self, messages: Sequence[ChatMessage], cancellation_token: CancellationToken
+    ) -> AsyncGenerator[AgentEvent | ChatMessage | Response, None]:
+        # Add the messages to the context.
+        for message in messages:
+            self._context.append(message)
+        # Run the team of agents.
```
peterj: misplaced comment

python/src/kagent/agents/_society_of_mind_agent.py
```diff
@@ -0,0 +1,283 @@
...
+                    continue
+
+                # TODO: Filter out the messages that are just context, they break the next step.
+                # We want the summary to be a pure function of the inner messages.
+
+                inner_messages.append(inner_msg)
+        assert result is not None
+
+        # Filter out the task messages from the inner messages.
+        # inner_messages = inner_messages[current_context_length:]
```
peterj, line 185: remove?

python/src/kagent/agents/_society_of_mind_agent.py
```diff
@@ -0,0 +1,283 @@
...
+                self._context.append(TextMessage(source=self.name, content=completion.content))
+                yield Response(
+                    chat_message=TextMessage(
+                        source=self.name, content=completion.content, models_usage=completion.usage
+                    ),
+                    inner_messages=inner_messages,
+                )
+
+        # Reset the team.
+        await self._team.reset()
```
peterj, line 243: should this be `self.on_reset`?

python/src/kagent/tools/helm/_helm.py
```diff
@@ -78,13 +78,54 @@ def _helm_list(
...
+    namespace: Annotated[str, "The namespace to get the release from"],
+    resource: Annotated[
+        str,
+        "The resource to get information about. If not provided, all resources will be returned, can be one of: all, hooks, manifest, notes, values",
+    ],
+) -> str:
+    return _run_helm_command(f"get {resource} {name} -n {namespace}")
+
 
+helm_get_release = FunctionTool(
```
peterj, line 94: update the `tool_gen.py` file and add "helm" to the list of dirs there

python/src/kagent/agents/_society_of_mind_agent.py
EItanya: It's in both places, we also want to run this at the end of a "task"

PR #87, opened 2025-03-03
package make targets

EItanya:


Dockerfile
```diff
@@ -94,4 +94,7 @@ RUN mkdir -p /app/python/.cache/uv && \
 
 EXPOSE 80
 
+LABEL org.opencontainers.image.source=https://github.com/kagent-dev/kagent
+LABEL org.opencontainers.image.description="Kagent is a tool for managing Kubernetes clusters."
```
peterj: is this correct? :)

EItanya: hahaha

helm/Chart-template.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v2
 name: kagent
 description: A Helm chart for KAgent, built with AutogenStudio
 type: application
-version: 0.1.0
\ No newline at end of file
+version: ${VERSION}
```
peterj, line 5: where will/is this coming from?

EItanya:
```
helm-publish:
	VERSION=$(VERSION) envsubst < helm/Chart-template.yaml > helm/Chart.yaml
	helm package helm/
	helm push kagent-$(VERSION).tgz oci://ghcr.io/kagent-dev/kagent/helm
```

peterj: wouldn't renaming the file to chart-template break the helm-install command? perhaps the first line here also needs to be in the helm-install target?

EItanya: good point, fixed

PR #88, opened 2025-03-04
adding tests fro generateresource tool (only authpolciy for now)

peterj:


python/src/kagent/tests/test_generate_resource.py
```diff
@@ -0,0 +1,141 @@
...
+import yaml
+from autogen_core import CancellationToken
+
+from ..tools.k8s import GenerateResourceTool, GenerateResourceToolConfig, GenerateResourceToolInput, ResourceTypes
+from ._test_result import GenerateResourceTestResult
+from ._yaml_comparer import YAMLComparer
+
+logger = logging.getLogger(__name__)
+
+TEST_CASES = str(Path(__file__).parent / "testcases")
```
EItanya, line 16: `python` 😆

python/src/kagent/tests/_yaml_comparer.py
EItanya, line 1: Not blocking, but I do think we should have unit tests for this, it's complex.

peterj: added an issue, will fix

PR #104, opened 2025-03-06
[WIP] Add tool for collecting L4 connection event from ambient-enrolled pods in the Solo mesh

bleggett:
This will only function with workload pods enrolled in ambient, under licensed Solo mesh builds 1.25 and newer.

It will subscribe to the in-pod L4 events stream and observe live connection events for a given duration.

This is WIP because I haven't tested or optimized this much (`kagent` backend isn't working in-cluster for me generally for some unrelated reason)


bleggett:
> This looks awesome, but I have 2 main comments:
> 
>     1. Where is the `proto` file for this located? I think there should be some sort of tooling to copy it and re-generate just in case that ever needs to happen.

Currently it resides in Solo's internal `ztunnel` repo. I'll add a make target/script to make the regen process more obvious.

>     2. We need to handle the case when this is running in cluster, when it's running in cluster we can just get endpoints and use those rather than port-forward which is pretty annoying.

Yeah the underlying API is primarily designed to be used from outside the cluster. It does not expose endpoints to the outside world, and only works with port-forwarding. AFAIK portforwarding used here should work just as well from inside the cluster as it does outside of it, and one nice thing about using portforwarding is that access to this endpoint can be controlled simply with k8s RBAC - if for instance you had specific workloads you wanted to make inaccessible to inspection by the agent (or anyone/anything else)

At least, since other tools here also use `kubectl` I assume we have client creds available already.

EItanya:
Closing for now

PR #109, opened 2025-03-07
Fixup arm64 container builds

bleggett:
1. Pinned `astral-uv` images by SHA are single-arch.
2. Use TARGETARCH for istioctl

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -20,7 +20,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN uv python install 3.12
 
 # Install kubectl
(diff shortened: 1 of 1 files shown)


python/Dockerfile
```diff
@@ -23,7 +25,7 @@ RUN curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/s
     && mv kubectl /usr/local/bin/
 
 # Install Istio
-RUN curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.25.0 TARGET_ARCH=x86_64 sh - \
+RUN curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.25.0 TARGET_ARCH=$TARGETARCH sh - \
```
EItanya, line 28: Won't kubectl have the same issue?

bleggett: duh, missed that

PR #117, opened 2025-03-10
statuses + Istio agent fixes

ilackarms:
adds status reporting +

### DIFF
--- a/python/src/kagent/tools/k8s/_kubectl.py
+++ b/python/src/kagent/tools/k8s/_kubectl.py
@@ -104,6 +104,7 @@ def _create_resource_from_url(
 ) -> str:
     return _run_kubectl_command(f"create -f {resource_yaml_url} {f'-n {namespace}' if namespace else ''}")
 
+
 def _get_events() -> str:
     return _run_kubectl_command("get events")
 
--- a/helm/templates/helm-agent.yaml
+++ b/helm/templates/helm-agent.yaml
@@ -1,0 +1,54 @@
+{{- if .Values.contrib.agents.helm.enabled }}
+---
+apiVersion: kagent.dev/v1alpha1
+kind: Agent
+metadata:
+  name: helm-agent
+spec:
+  description: The Helm Expert AI Agent specializing in using Helm for Kubernetes cluster management and operations. This agent is equipped with a range of tools to manage Helm releases and troubleshoot Helm-related issues.
+  systemMessage: |-
+    You are a Kubernetes expert agent specialized in cluster management and operations. Your primary responsibilities include:
+
+        1. Helm Operations:
+           - Managing Helm releases (list, get details, upgrade, and uninstall)
+           - Understanding Helm chart structures and dependencies
+           - Validating release states and configurations
+
+        2. Kubernetes Resource Management:
+           - Monitoring and managing various Kubernetes resources (Pods, Deployments, Services, etc.)
+           - Analyzing resource configurations and relationships
+           - Applying and validating Kubernetes manifests
+           - Understanding API resource availability and versioning
+
+        3. Documentation and Troubleshooting:
+           - Accessing and interpreting Kubernetes and related project documentation
+           - Providing context-aware recommendations
+           - Identifying potential issues in configurations
+
+        Always follow these principles:
+        - Validate operations before execution
+        - Provide clear explanations for your actions
+        - Consider security implications
+        - Follow Kubernetes best practices
+        - Use appropriate tools based on the context
+
+        When handling tasks:
+        1. First analyze the current state
+        2. Plan your actions
+        3. Execute with appropriate tools
+        4. Verify results
+        5. Report outcomes clearly
+  tools:
+    - provider: kagent.tools.helm.ListReleases
+    - provider: kagent.tools.helm.GetRelease
+    - provider: kagent.tools.helm.Upgrade
+    - provider: kagent.tools.helm.Uninstall
+    - provider: kagent.tools.helm.RepoAdd
+    - provider: kagent.tools.helm.RepoUpdate
+    - provider: kagent.tools.k8s.GetResources
+    - provider: kagent.tools.k8s.GetAvailableAPIResources
+    - provider: kagent.tools.k8s.ApplyManifest
+    - provider: kagent.tools.docs.QueryTool
+      config:
+        docs_download_url: https://doc-sqlite-db.s3.sa-east-1.amazonaws.com
+{{- end }}
--- a/helm/templates/istio-agent.yaml
+++ b/helm/templates/istio-agent.yaml
@@ -134,10 +134,10 @@ subjects:
 apiVersion: kagent.dev/v1alpha1
 kind: Agent
 metadata:
-  name: kube-expert
+  name: istio-agent
   namespace: team-ns
 spec:
-  description: The Kubernetes Expert AI Agent specializing in cluster operations,
+  description: The Istio Expert AI Agent specializing in Istio operations,
     troubleshooting, and maintenance.
   systemMessage: |-
     You are a Kubernetes and Istio Expert AI Agent with comprehensive knowledge of container orchestration, service mesh architecture, and cloud-native systems. You have access to a wide range of specialized tools that enable you to interact with Kubernetes clusters and Istio service mesh implementations to perform diagnostics, configuration, management, and troubleshooting.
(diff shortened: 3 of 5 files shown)


go/autogen/client/validate.go
```diff
@@ -18,6 +21,24 @@ type ValidationResponse struct {
 	Warnings []*ValidationError `json:"warnings"`
 }
 
+func (r ValidationResponse) ErrorMsg() string {
```
EItanya, line 24: `String()` so you can `%s` directly?

Makefile
```diff
@@ -14,6 +14,8 @@ APP_IMG ?= $(DOCKER_REGISTRY)/$(DOCKER_REPO)/$(APP_IMAGE_NAME):$(APP_IMAGE_TAG)
 DOCKER_BUILDER ?= docker
 DOCKER_BUILD_ARGS ?=
 
+ISTIO_INSTALLER_ENABLED ?= true
```
EItanya: This seems very very specific for the Makefile, can we allow passing in values instead?

go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -45,7 +49,17 @@ type AnyType struct {
 }
 
 // AgentStatus defines the observed state of Agent.
-type AgentStatus struct{}
+type AgentStatus struct {
+	ObservedGeneration int64            `json:"observedGeneration,omitempty"`
+	Conditions         []AgentCondition `json:"conditions,omitempty"`
+}
+
+type AgentCondition struct {
```
EItanya: Should we use the upstream condition type?

go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -28,7 +35,17 @@ type ModelConfigSpec struct {
 }
 
 // ModelConfigStatus defines the observed state of ModelConfig.
-type ModelConfigStatus struct{}
+type ModelConfigStatus struct {
+	Conditions         []ModelConfigCondition `json:"conditions"`
+	ObservedGeneration int64                  `json:"observedGeneration"`
+}
+
+type ModelConfigCondition struct {
```
EItanya: Same comment about using upstream types

go/controller/api/v1alpha1/autogenteam_types.go
```diff
@@ -87,7 +91,17 @@ type OrTerminationCondition struct {
 }
 
 // TeamStatus defines the observed state of Team.
-type TeamStatus struct{}
+type TeamStatus struct {
+	Conditions         []TeamCondition `json:"conditions"`
+	ObservedGeneration int64           `json:"observedGeneration"`
+}
+
+type TeamCondition struct {
```
EItanya: Same comment about using upstream types

go/controller/internal/autogen/autogen_reconciler.go
```diff
@@ -75,7 +117,38 @@ func (a *autogenReconciler) ReconcileAutogenModelConfig(ctx context.Context, req
 		return fmt.Errorf("failed to find teams for model %s: %v", req.Name, err)
 	}
 
-	return a.reconcileTeams(ctx, teams...)
+	return a.reconcileModelConfigStatus(
```
EItanya, line 123: Do we actually have validation or models? It seems like we should update `AgentStatuses` or something like that, as what we've actually broken is the agent

python/src/kagent/tools/k8s/_kubectl.py
```diff
@@ -89,14 +89,21 @@ def _label_resource(
...
+    namespace: Annotated[Optional[str], "The namespace of the resource to create"],
 ) -> str:
     with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=True) as tmp_file:
         tmp_file.write(resource_yaml)
         tmp_file.flush()  # Ensure the content is written to disk
-        return _run_kubectl_command(f"create -f {tmp_file.name}")
+        return _run_kubectl_command(f"create -f {tmp_file.name} {f'-n {namespace}' if namespace else ''}")
 
 
+def _create_resource_from_url(
```
EItanya, line 101: Why do we need a separate tool for this, can't we just add it to the above?

helm/templates/istio-installer-clusterrole.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if .Values.permissions.istioInstaller.enabled }}
```
EItanya: Our core RBAC already includes a lot of this stuff, shouldn't we just add the things that we don't already have. Also I honestly don't love the UX here, the user has to know to add this, I think a much better solution is to either do it dynamically, or instruct the user how to update it.

go/controller/internal/autogen/autogen_reconciler.go
ilackarms: invalid secret ref would be the main one; this is more of an indicator that the modelconfig is used by an invalid agent. we can omit the status here (or just check the secret ref)

helm/templates/istio-installer-clusterrole.yaml
ilackarms: dynamically how? i think we should be prioritizing ease of use for demo / new users

go/controller/internal/autogen/autogen_reconciler.go
EItanya: Gotcha, I think what I'm referencing is that this will also invalidate any team that references it, so it might be worth adding a condition to the team/agent which mentions that

helm/templates/istio-installer-clusterrole.yaml
EItanya: Agreed, but I think this is actually harder, if the agent outputs the RBAC that will fix the problem, that's probably easier than having separate helm values for each project we want to interact with, does that make sense?

Makefile
ilackarms: im not sure wym
this does allow passing in the value, it just sets a default

go/autogen/client/validate.go
ilackarms: personally i like ` ErrorMsg()` because it makes usage easier to track
also it is specifically for the error message, it doesn't produce anything if the validation result is successful

go/controller/api/v1alpha1/autogenagent_types.go
ilackarms: which upstream?

python/src/kagent/tools/k8s/_kubectl.py
ilackarms: they're different; one takes the raw yaml to apply, the other accepts a url to fetch from
how do you imagine the function signature looking?

EItanya: I'm just worried about tool sprawl, I'm imagining something like `kubectl apply` where it can really take any combination of things and then we do some logic to check that it's valid

go/controller/api/v1alpha1/autogenagent_types.go
EItanya: https://github.com/kubernetes-sigs/gateway-api/blob/c3ed0633feacd8532e9fb52ae13c2f68d3db986c/apis/v1/gateway_types.go#L802

EItanya: `metav1` has a condition type which is typically used for this so that you can use `k8s` native functions for acting on these lists

Makefile
EItanya: Mostly this just comes down to my comment about adding this to the chart rather than having the agent tell you how to update the RBAC, this approach is super limited and will break for any resources outside of a few blessed paths.

helm/templates/istio-installer-clusterrole.yaml
peterj: I agree with Eitan here; I don't think we should be hardcoding this for each project/tool we support in helm.

Makefile
ilackarms: so what are you proposing we do as the alternative

go/controller/internal/autogen/autogen_reconciler.go
ilackarms: if the team/agent fails to translate for whatever reason that will be reflected on the team / agent status directly
this just also makes sure it's reflected on the modelconfig
however it might be redundant to do in both places (especially if the error is not caused by something in the model config) so probably we could just do a simple validation function for MC

python/src/kagent/tools/k8s/_kubectl.py
ilackarms: the thing is that normally kubectl takes a path to a file which might be a url
our tool takes the raw yaml string which is not interchangeable with a URL / path

helm/templates/istio-installer-clusterrole.yaml
ilackarms: how do we plan to handle rbac for the built-in agents we are shipping with?

python/src/kagent/tools/k8s/_kubectl.py
EItanya: That's not true, kubectl apply can also take a manifest through `stdin`/`EOF`

helm/templates/istio-installer-clusterrole.yaml
EItanya: I think we should just add ALL OF IT honestly, we can worry about narrowing down later

Makefile
EItanya: https://github.com/kagent-dev/kagent/pull/117#discussion_r1991621068

python/src/kagent/tools/k8s/_kubectl.py
ilackarms: ok but how do you see the function signature looking?

PR #129, opened 2025-03-11
Cli auto port forward

### LINKED ISSUE #93: cli should automatically do a port forward

### LINKED ISSUE #120: Add a 20 tool limit to all agents
Agents perform best with smaller numbers of tools, 20 is the limit at which OpenAI says things start to go wonky. Therefore we should hard-cap at 20 for now.

EItanya:
1. Add auto port-forward from CLI when connection fails
2. Add maximum # of tools to Agent CRD
3. Temporarily removed the ability for the agent to update it's own RBAC 😆 

Closes #93 

Closes #120

### DIFF
--- a/go/config/crd/bases/kagent.dev_agents.yaml
+++ b/go/config/crd/bases/kagent.dev_agents.yaml
@@ -54,7 +54,7 @@ spec:
                     provider:
                       type: string
                   type: object
-                maxItems: 15
+                maxItems: 20
                 type: array
             type: object
           status:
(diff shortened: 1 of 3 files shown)


go/config/crd/bases/kagent.dev_agents.yaml
```diff
@@ -53,6 +54,7 @@ spec:
                     provider:
                       type: string
                   type: object
+                maxItems: 20
```
peterj, line 57: the UI limit is 10 :D

peterj: I can change that to 20

EItanya: Yea ik, but the UI is also meant to be the "simpler" walkthrough option, so I think that's fine for now

PR #140, opened 2025-03-12
Cli helper commands

EItanya:


go/cli/internal/config/config.go
```diff
@@ -35,6 +36,7 @@ func Init() error {
 	viper.SetDefault("ws_url", "ws://localhost:8081/api/ws")
```
peterj, line 36: should this be 8082?

EItanya: No, `8082` here is explicitly for the UI, not related to the websocket or standard API client

PR #142, opened 2025-03-12
Controller Agent fixes

ilackarms:


go/controller/internal/autogen/autogen_reconciler.go
```diff
@@ -257,6 +268,9 @@ func (a *autogenReconciler) reconcileAgents(ctx context.Context, agents ...*v1al
 }
 
 func (a *autogenReconciler) upsertTeam(team *autogen_client.Team) error {
+	// lock to prevent races
```
EItanya, line 271: In the future we can add an `upsert` call to the backend

PR #144, opened 2025-03-13
use model_config for the LLMTools

EItanya:


EItanya:
FYI @peterj, I swapped the config for the LLMTool so that we can use any model/provider we want, this helped @ilackarms with the translation as well

PR #160, opened 2025-03-16
Fix tool yamls + update system prompts

ilackarms:


helm/values.yaml
```diff
@@ -81,11 +81,22 @@ securityContext: {}
...
   agents:
+    k8s:
+      enabled: true
     istio:
       enabled: true
     helm:
       enabled: true
     observability:
+      prometheus:
+        url: ""
```
EItanya, line 92: I wonder if we should default this to `grafana.grafana:<port>` or something like that, so that if we have a doc it "just works"

Issue #162, opened 2025-03-17
add ambient db to the docs tool


peterj:

peterj:
fixed by #163

PR #163, opened 2025-03-17
add ambient mesh db to query_documentation.py

peterj:


EItanya:
closes #162

python/src/kagent/tools/docs/query_documentation.py
```diff
@@ -28,6 +28,7 @@
     "gloo-edge": "gloo-edge.db",
     "otel": "otel.db",
     "cilium": "cilium.db",
+    "istio ambient mesh": "ambient.db",
```
EItanya, line 31: This is interesting, we might need multiple keywords to get to this DB

PR #169, opened 2025-03-17
Fix crd agent creation

EItanya:
1. Fix `yaml` for observability agent to not have `null` for `str` fields
2. Fix improper `nil` check in translator for tools
3. Add smoke test for agent acceptance to CI

### DIFF
--- a/go/cli/internal/cli/install.go
+++ b/go/cli/internal/cli/install.go
@@ -49,7 +49,28 @@ func InstallCmd(ctx context.Context, c *ishell.Context) {
 		return
 	}
 	s.Stop()
-	c.Println("kagent installed successfully")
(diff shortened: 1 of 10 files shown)


go/cli/internal/cli/install.go
```diff
@@ -35,7 +35,6 @@ func InstallCmd(ctx context.Context, c *ishell.Context) {
 		"--namespace",
 		cfg.Namespace,
 		"--create-namespace",
-		"--wait",
```
peterj: should this be removed?

peterj (dismissed):

go/cli/internal/cli/install.go
EItanya: Yes, that's my bad

Issue #132, opened 2025-03-11
[EPIC] Runtime/Engine


EItanya:
## Description

Runtime/Engine improvements comprises all of the work that has to be done to the core agent framework.

`Autogen` laid the groundwork for an expressive, declarative API to configure agents/workflows. However, there are certain additions we will have to make in order for the system to solve all of the use-cases we're envisioning.

### Multi-Agent systems

The first release of `kagent` supports single agents in order, but more complex systems build up of many agents will be a fast follow.

### True Graph Execution

In order to unlock all use-cases, true declarative graph style execution needs to be added to `autogen`. There is already an issue [here](https://github.com/microsoft/autogen/issues/4623?notification_referrer_id=NT_kwDOAD_b_rMxMzc3MzY2MzE1NDo0MTg1MDg2#issuecomment-2712898472) upstream. 

### Workflows

A workflow is a pre-determined set of steps a set of agents to perform. This workflow can either be a simple linear progression, or a DAG (directed acrylic graph). The system we design should support both types.

### Support for multiple LLM providers

Currently we only support `OpenAI` as our LLM provider, this was an expedient decision when testing and releasing, but in future this will no longer be required, and testing different scenarios with different models will be prudent.

jrbe228:
> ### True Graph Execution
> In order to unlock all use-cases, true declarative graph style execution needs to be added to `autogen`. There is already an issue [here](https://github.com/microsoft/autogen/issues/4623?notification_referrer_id=NT_kwDOAD_b_rMxMzc3MzY2MzE1NDo0MTg1MDg2#issuecomment-2712898472) upstream.
> 
> ### Workflows
> A workflow is a pre-determined set of steps a set of agents to perform. This workflow can either be a simple linear progression, or a DAG (directed acrylic graph). The system we design should support both types.

1. Is it possible to merge the above categories into a single category "DAGs"? Saying "graphs" opens the door to cyclic graphs, which I believe is out-of-scope for the project. Also the term "workflows" is applied to so many situations... it might be confusing. 
2. Is it possible to use an existing DAG [engine](https://github.com/meirwah/awesome-workflow-engines)? Integration with existing engines may have already been evaluated. I'm new to the Autogen and Kagent roadmaps. In theory Kagent might integrate with a declarative, K8s-native DAG engine. 

> Support for multiple LLM providers

3. Looking forward to it! Especially this [one](https://github.com/kagent-dev/kagent/issues/174). 

[...]

PR #180, opened 2025-03-19
Use independent pods / deployments for each container service

jrbe228:
TODO 
- Edit default [port ](https://github.com/jrbe228/kagent/blob/0ab88aefb230abe110f5b3818ecea459ed179289/helm/values.yaml#L60)for controller service. 
- Containers should use the hostname: `<SERVICE_NAME>.<NAMESPACE>.svc.cluster.local
` instead of `localhost` for 
pod-->ClusterIP-->pod communication. [Here ](https://github.com/search?q=repo%3Akagent-dev%2Fkagent%20localhost&type=code)are all current usages of localhost.


EItanya:
Hi @jrbe228, thanks so much for the contribution. 

I left a comment on the issue with more information about my thoughts: https://github.com/kagent-dev/kagent/issues/132#issuecomment-2738345867.

sbx0r:
Hi @jrbe228 
Do you need any assistance with this?

marcinkubica:
> Containers should use the hostname: <SERVICE_NAME>.<NAMESPACE>.svc.cluster.local  instead of localhost 

not all deployments use .cluster.local - ie. GKE with CloudDNS

sbx0r:
> > Containers should use the hostname: <SERVICE_NAME>..svc.cluster.local  instead of localhost
> 
> not all deployments use .cluster.local - ie. GKE with CloudDNS

Right.
We should make it configurable or do you know some more mature approach?

EItanya:
This has been implemented by #599

Issue #165, opened 2025-03-17
switch the UI backend from autogen to controller


peterj:
The UI is currently using autogen as its backend. The UI should use the controller as its backend and the controller can either proxy or make requests to the autogen.

EItanya:
Finished except for the `wss` calls, which we may never transition.

PR #186, opened 2025-03-24
Implement REST API on the controller

peterj:
This implements the rest api on the controller and moves the UI to use that. Note that UI goes to the websocket server directly (on the autogen).


peterj:
Fixes #165

peterj:
I also added an endpoint for model configs that will return whatever model config resources are deployed in the cluster. However, in order to support different models we'll probably need to add the model config to the Agent resource.

ShalinChh:
Hello, I'm looking to access MCP-based agents programmatically via the API instead of relying on the UI. It would be great if this could be merged soon so we can proceed with testing. Thanks for the great work!

go/autogen/client/team.go
```diff
@@ -14,6 +14,21 @@ func (c *Client) CreateTeam(team *Team) error {
 	return c.doRequest("POST", "/teams/", team, team)
 }
 
+func (c *Client) GetTeamByID(teamID int, userID string) (*Team, error) {
```
EItanya, line 17: You can use this:
```
@router.get("/{team_id}")
async def get_team(team_id: int, user_id: str, db=Depends(get_db)) -> Dict:
```

go/autogen/client/toolservers.go
```diff
@@ -0,0 +1,46 @@
...
+	return server, err
+}
+
+func (c *Client) ListToolServers(userID string) ([]*ToolServer, error) {
+	var toolServers []*ToolServer
+	err := c.doRequest("GET", fmt.Sprintf("/toolservers/?user_id=%s", userID), nil, &toolServers)
+	return toolServers, err
+}
+
+func (c *Client) GetToolServer(serverID *int, userID string) (*ToolServer, error) {
```
EItanya, line 19: Why is the input here a pointer, it's always required

EItanya:
```
@router.get("/{server_id}")
async def get_server(server_id: int, user_id: str, db=Depends(get_db)) -> Dict:
```

go/controller/internal/httpserver/server.go
```diff
@@ -0,0 +1,144 @@
...
+	autogen_client "github.com/kagent-dev/kagent/go/autogen/client"
+	"github.com/kagent-dev/kagent/go/controller/internal/httpserver/handlers"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// Default namespace for resources
+	DefaultResourceNamespace = "kagent"
```
EItanya, line 18: If it isn't exposed already we should get this from the env from `k8s`

EItanya (commented):
Comments aren't blocking

EItanya (approved):

PR #194, opened 2025-03-25
Support for Anthropic/AzureAI models

peterj:
This PR updates the ModelConfig CRD and adds support for Anthropic and Azure OpenAI (still needs to be tested). I've also updated the Agent CRD to reference the ModelConfig + other changes that were necessary to dynamically create the model in autogen based on the ModelConfig resource. 

There are also bunch of changes in the UI to support the backend changes.

Here's how to test this out:

1. Create the secret wit the API key:
```
kubectl create secret generic kagent-anthropic -n kagent --from-literal ANTHROPIC_API_KEY=<key>
```

2. Apply the ModelConfig referencing the secret and setting the provider to Anthropic:

```
kind: ModelConfig
metadata:
  name: claude-model-config
  namespace: kagent
spec:
  apiKeySecretName: kagent-anthropic
  apiKeySecretKey: ANTHROPIC_API_KEY
  model: claude-3-sonnet-20240229
  provider: Anthropic
  ```

Similar for Azure:

1. Secret first:
```
kubectl create secret generic kagent-azureopenai -n kagent --from-literal AZURE_OPENAI_API_KEY=<key>
```

2. Then the modelconfig with azure-specific config:

```
kind: ModelConfig
metadata:
  name: azureopenai-model-config
  namespace: kagent
spec:
  apiKeySecretName: kagent-azureopenai
  apiKeySecretKey: AZURE_OPENAI_API_KEY
  model: gpt-4o-mini
  provider: AzureOpenAI
  azureOpenAI:
    azureEndpoint: "https://<endpointname>.openai.azure.com/"
    apiVersion: "2025-03-01-preview"
    azureDeployment: "gpt-4o-mini"
    azureAdToken: <token>
```

closes: #174

### CODE REVIEW

File: go/controller/api/v1alpha1/autogenagent_types.go [hunk 1]
```diff
@@ -31,6 +31,8 @@ type AgentSpec struct {
 	Description string `json:"description,omitempty"`
 	// +kubebuilder:validation:MinLength=1
 	SystemMessage string `json:"systemMessage,omitempty"`
+	// +kubebuilder:validation:Required
+	ModelConfigRef string `json:"modelConfigRef"`
```
EItanya: Why is this required now? Currently we support a default model config which feels like a nice default UX

go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -24,11 +24,131 @@ const (
...
+	// The provider of the model
+	// +kubebuilder:default=OpenAI
+	Provider ModelProvider `json:"provider"`
+
 	APIKeySecretName string `json:"apiKeySecretName"`
 	APIKeySecretKey  string `json:"apiKeySecretKey"`
+
+	// Provider-specific configuration
+	// +optional
+	ProviderConfig *ProviderConfig `json:"providerConfig,omitempty"`
```
EItanya: I think the fields in here should be flattened into the `ModelConfigSpec`. The Gateway API does validation via `cel` to ensure that the `ModelProvider` matches the field chosen. It's a nice UX.

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -562,3 +563,169 @@ func addModelClientToConfig(
 	(*toolConfig)["model_client"] = cfg
 	return nil
 }
+
+// createModelClientForProvider creates a model client component based on the model provider
+func createModelClientForProvider(modelConfig *v1alpha1.ModelConfig, apiKey string) (*api.Component, error) {
+	switch modelConfig.Spec.Provider {
+	case v1alpha1.ProviderAnthropic:
+		anthropicConfig := map[string]interface{}{
```
EItanya: Can you create a type like we do here? https://github.com/kagent-dev/kagent/blob/main/go/autogen/api/models.go. Generic `map[string]interface{}` are not super nice to use in `go`

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -562,3 +563,169 @@ func addModelClientToConfig(
...
+
+		return &api.Component{
+			Provider:      "autogen_ext.models.anthropic.AnthropicChatCompletionClient",
+			ComponentType: "model",
+			Version:       makePtr(1),
+			Config:        anthropicConfig,
+		}, nil
+
+	case v1alpha1.ProviderAzureOpenAI:
+		azureConfig := map[string]interface{}{
```
EItanya: Same comment as above

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -562,3 +563,169 @@ func addModelClientToConfig(
...
+
+		return &api.Component{
+			Provider:      "autogen_ext.models.azure_openai.AzureOpenAIChatCompletionClient",
+			ComponentType: "model",
+			Version:       makePtr(1),
+			Config:        azureConfig,
+		}, nil
+
+	case v1alpha1.ProviderOpenAI:
+		openAIConfig := map[string]interface{}{
```
EItanya: Same comment as above

EItanya (commented):
My main comment on this PR has to do with the use of `map[string]interface{}`. Totally makes sense why these would look like a good option coming from a JS/Python world. But typically in `go` you should create a type to represent it when using it within the code. `map[string]interface{]` is typically used to represent JSON you don't know the format of, but we know the format of all of the JSON in this case.

Jmadih:
Hello, 
Maybe in the future, you could check out LiteLLM (litellm.ai) for managing all LLMs in the OpenAI format.

go/autogen/api/models.go
```diff
@@ -72,3 +73,39 @@ func (c *AzureOpenAIClientConfig) ToConfig() (map[string]interface{}, error) {
...
 	return fromConfig(c, config)
 }
+
+type AnthropicCreateArguments struct {
+	MaxTokens      *int              `json:"max_tokens,omitempty"`
+	Temperature    *float64          `json:"temperature,omitempty"`
+	TopP           *float64          `json:"top_p,omitempty"`
+	TopK           *int              `json:"top_k,omitempty"`
+	StopSequences  []string          `json:"stop_sequences,omitempty"`
+	ResponseFormat interface{}       `json:"response_format,omitempty"`
```
EItanya: What's supposed to go here?

go/autogen/api/models.go
```diff
@@ -72,3 +73,39 @@ func (c *AzureOpenAIClientConfig) ToConfig() (map[string]interface{}, error) {
...
+	ModelInfo         *ModelInfo        `json:"model_info,omitempty"`
+	Timeout           *float64          `json:"timeout,omitempty"`
+	MaxRetries        *int              `json:"max_retries,omitempty"`
+	DefaultHeaders    map[string]string `json:"default_headers,omitempty"`
+	AnthropicCreateArguments
+}
+
+type AnthropicClientConfiguration struct {
+	Tools      []map[string]interface{} `json:"tools,omitempty"`
+	ToolChoice interface{}              `json:"tool_choice,omitempty"`
```
EItanya: What's supposed to go here?

go/config/crd/bases/kagent.dev_agents.yaml
```diff
@@ -64,6 +70,8 @@ spec:
                   type: object
                 maxItems: 20
                 type: array
+            required:
```
EItanya: This still says required

go/controller/api/v1alpha1/autogenmodelconfig_types.go
EItanya: I want to bring this comment back up, I think it's important

go/autogen/api/models.go
peterj: copied this from whatever was there on openai :) 

This would be either json or text output, but I don't think we should necessarily expose this

go/controller/api/v1alpha1/autogenmodelconfig_types.go
peterj: yep, still working on this

go/autogen/api/models.go
peterj: List[Dict[str, Any]] and Union[Literal["auto", "any", "none"], Dict[str, Any]]

peterj: I guess I can make a type for ToolChoice to be one of "auto", "any" or "none"

peterj: will remove these

go/autogen/api/models.go
peterj: removed

peterj:
> Looking better, but still have the same major comment about the API

I think I address all feedback with the latest commit.

EItanya (approved):

PR #198, opened 2025-03-26
Add toolserver crd

ilackarms:
configure mcp toolservers and enable agents to consume it via crds


go/autogen/client/types.go
```diff
@@ -6,22 +6,50 @@ import (
...
+type ToolServerConfig struct {
+	BaseObject
+	Component *api.Component `json:"component"`
+}
+
+type StdioMcpServerConfig struct {
+	Command string            `json:"command"`
+	Args    []string          `json:"args,omitempty"`
+	Env     map[string]string `json:"env,omitempty"`
+	Stderr  interface{}       `json:"stderr,omitempty"`
```
EItanya: What is this for?

go/controller/PROJECT
```diff
@@ -35,4 +35,13 @@ resources:
   kind: AutogenModelConfig
   path: github.com/kagent-dev/kagent/go/controller/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: ai.solo.io
```
EItanya: We should probably get `solo.io` out of here

go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -36,6 +36,12 @@ type AgentSpec struct {
 }
 
 type Tool struct {
+	// ONEOF: BuiltinTool, McpServerTool
+	BuiltinTool   `json:",inline"`
+	McpServerTool `json:",inline"`
```
EItanya, line 43: What does this actually look like from a rendering/writing YAML perspective.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -0,0 +1,107 @@
...
+	ComponentVersion int     `json:"component_version"`
+	Description      *string `json:"description"`
+	Label            *string `json:"label"`
+	// note: this implementation is due to the kubebuilder limitation https://github.com/kubernetes-sigs/controller-tools/issues/636
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	Config map[string]AnyType `json:"config,omitempty"`
+}
+
+type MCPToolServerParams struct {
```
EItanya, line 77: How is this different from `ToolServerConfig`

go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -45,6 +51,15 @@ type Tool struct {
 	Config map[string]AnyType `json:"config,omitempty"`
 }
 
+type McpServerTool struct {
+	// the name of the ToolServer that provides the tool. must exist in the same namespace as the Agent
+	ToolServer string `json:"toolServer,omitempty"`
+	// The name of the tool to be provided by the ToolServer
+	// For a list of all the tools provided by the server,
+	// the client can query the status of the ToolServer object after it has been created
+	ToolName string `json:"toolName,omitempty"`
```
EItanya: What if a user wants to use more than one, a totally valid answer is that we can add that later just curious.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ToolServerSpec defines the desired state of ToolServer.
+type ToolServerSpec struct {
+	Description string           `json:"description"`
+	Config      ToolServerConfig `json:"config"`
+}
+
+type ToolServerConfig struct {
+	Stdio *StdioMcpServerConfig `json:"stdio,omitempty"`
+	Sse   *SseMcpServerConfig   `json:"sse,omitempty"`
```
EItanya, line 31: Same question about `oneofs` here. I think we need a unified approach in our APIs to this sort of thing.

go/controller/internal/controller/toolserver_controller.go
```diff
@@ -0,0 +1,68 @@
...
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.0/pkg/reconcile
+func (r *ToolServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	return ctrl.Result{
+		// loop forever because we need to refresh tools server status
```
EItanya, line 56: How does this work if the object is updated? Will this event continue to requeue forever

go/autogen/client/types.go
ilackarms: simply based on the config in the UI, im not sure myself tbh.  i can remove 

```ts
export interface StdioMcpServerConfig {
  /**
   * The executable to run to start the server.
   */
  command: string;
  /**
   * Command line arguments to pass to the executable.
   */
  args?: string[];
  /**
   * The environment to use when spawning the process.
   */
  env?: Record<string, string>;
  /**
   * How to handle stderr of the child process.
   */
  stderr?: string | number;
  /**
   * The working directory to use when spawning the process.
   */
  cwd?: string;
}
```

go/controller/PROJECT
ilackarms: i'm not totally sure the impact of modifying the PROJECT file but i'm down to try. i don't see this value used anywhere else

go/controller/api/v1alpha1/autogenagent_types.go
ilackarms: it inlines the json inside the parent struct as though the fields of the child structs were embedded directly in the parent message.  https://go.dev/play/p/egH9MaiL6kv

go/controller/api/v1alpha1/toolserver_types.go
ilackarms: subtle difference but mainly that kubebuilder doesn't allow `interface{}` types so we have to use `Anytype`.

```go
// kube type
type SseMcpServerConfig struct {
	URL string `json:"url"`
	// +kubebuilder:pruning:PreserveUnknownFields
	// +kubebuilder:validation:Schemaless
	Headers        map[string]AnyType `json:"headers,omitempty"`
	Timeout        string             `json:"timeout,omitempty"`
	SseReadTimeout string             `json:"sse_read_timeout,omitempty"`
}
```
vs
```go
// autogen api type
type SseMcpServerConfig struct {
	URL            string                 `json:"url"`
	Headers        map[string]interface{} `json:"headers,omitempty"`
	Timeout        int                    `json:"timeout,omitempty"`
	SseReadTimeout int                    `json:"sse_read_timeout,omitempty"`
}
```

go/controller/api/v1alpha1/autogenagent_types.go
ilackarms: good point i can make this an array no prob

go/controller/api/v1alpha1/toolserver_types.go
ilackarms: i'm open to suggestions. i found tihs issue on kubebuilder: https://github.com/kubernetes-sigs/controller-tools/issues/461 seems like they are suggesting using cel rules to validate. i'm open to this or other suggestions

go/controller/internal/controller/toolserver_controller.go
ilackarms: yeah this just means that there will always be an event in the queue for every toolserevr, regardless of whether it was updated or not

go/autogen/client/types.go
peterj: looking at the MCP code -- stderr and the cwd can be removed as they aren't there anymore

EItanya: Gotcha, I think we can remove for now, let's keep it bare bones at first

go/controller/api/v1alpha1/toolserver_types.go
EItanya: This is what gateway-api does: https://github.com/kubernetes-sigs/gateway-api/blob/f1fd808386c4dd24e2f3edaff0dd5707038ba721/apis/v1/httproute_types.go#L783. It doesn't look pretty, but the actual UX is pretty good.

EItanya (approved):
I tested this locally and it works!! 

I'm going to approve and merge this for now, but I do think we need to do a couple of API related follow-ups, specifically for the inline JSON and `ONEOF` support.

Issue #200, opened 2025-03-26
tool calls don't work with Anthropic


peterj:
```
Stream error for run 67: Object of type TextBlock is not JSON serializable 
```

Waiting for #194  to merge and will investigate this one.

peterj:
The fix will be coming from autogen: https://github.com/microsoft/autogen/pull/6344

peterj:
This was fixed. Closing.

Issue #171, opened 2025-03-18
expose kagent functionality through the API for external consumption


peterj:
Feedback/request from the discord: https://discord.com/channels/1346225185166065826/1346225185841221644/1351530953939484705

peterj:
Another use case for invoking agents through the API: https://discord.com/channels/1346225185166065826/1346225185841221644/1354480403616108644 

Theoretically there's a way to do this through the API -- at least invoking the "chat" and sending the initial/filled out prompt; however, it's not a single step at the moment; it would require you to create a run and a sesssion, then send the initial message (i.e. multiple calls to the API vs. just a single one).

peterj:
We could use a better version of this: https://github.com/kagent-dev/autogen/blob/main/python/packages/autogen-studio/autogenstudio/web/serve.py

zanuka:
happy to take this one

zanuka:
updating my questions / plan as I review codebase...

zanuka:
I've started working on exposing kagent functionality through the API for external consumption. Since this feature this will enable use cases like incident response via Slackbot and MCP tool integration, I've included a bunch of questions to ensure things are on the right track. Might seem like overkill, but since there will likely be specific preferences / opinions on a lot of this stuff I thought I'd ask first. 

1) API Design & Functionality
- Which specific kagent functionality should be prioritized for API exposure? (agent invocation, chat, tool discovery, etc.)
- For incident response via Slackbot, should we expose dedicated endpoints for predefined scenarios (latency dashboards, manual mitigations) or a more generic agent invocation API?
- How should we handle the payload structure for different use cases (e.g., templated prompts with placeholders for client IDs, endpoints)?

2) Authentication & Security
- What authentication mechanism should these new endpoints use?
- Are there additional security requirements for external API access vs. UI access?
- Should we implement rate limiting to prevent abuse, especially for resource-intensive operations?

3) Response Handling
- For chat operations, should the API return immediately with session/run IDs or wait for initial agent responses?
- How should streaming responses be handled for long-running operations?
- What's the preferred approach for WebSocket connections with these simplified endpoints?

4) Integration Patterns
- Should we provide client libraries or example code for common integrations (Slack, etc.)?
- How should the API handle context from external systems (e.g., translating business domain knowledge to cluster context)?
- Are there specific formats or protocols that external systems would require for integration?

5) Implementation Approach
- Should we extend the existing API that the UI uses or create new dedicated endpoints?
[...]

PR #208, opened 2025-03-27
feat(controller): add comprehensive logging to HTTP handlers

### LINKED ISSUE #203: add/improve logging in the controller
We should add logging to the controller, specifically in the REST API handlers: https://github.com/kagent-dev/kagent/tree/main/go/controller/internal/httpserver/handlers 

Also, returning better/more specific error messages - especially when validation fails in the create team handler - would be useful: https://github.com/kagent-dev/kagent/blob/e8069000e777b3b2cb1c54239122faae71991ad4/go/controller/internal/httpserver/handlers/teams.go#L116

zanuka:
This commit adds detailed logging to all REST API handlers in the controller:

- Add structured logging with context information (operation type, user ID, resource IDs)
- Implement different log levels (Info for important events, V(1) for detailed operations)
- Enhance error logging with detailed context
- Improve validation error messages in the team creation handler
- Create consistent logging patterns across all handlers
- Add logging to helper functions for complete traceability

The changes follow best practices by using request-scoped loggers that carry request context, allowing for better debugging and monitoring in production.

Fixes #203

### DIFF
--- a/go/controller/internal/httpserver/handlers/health.go
+++ b/go/controller/internal/httpserver/handlers/health.go
@@ -2,6 +2,8 @@ package handlers
 
 import (
 	"net/http"
+
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // HealthHandler handles health check requests
@@ -14,6 +16,9 @@ func NewHealthHandler() *HealthHandler {
 
 // HandleHealth handles GET /health requests
 func (h *HealthHandler) HandleHealth(w http.ResponseWriter, r *http.Request) {
+	log := ctrllog.FromContext(r.Context()).WithName("health-handler")
+	log.V(1).Info("Handling health check request")
+
 	w.WriteHeader(http.StatusOK)
 	w.Write([]byte("OK"))
 }
--- a/go/controller/internal/httpserver/errors/errors.go
+++ b/go/controller/internal/httpserver/errors/errors.go
@@ -1,0 +1,62 @@
+package errors
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// APIError represents an API error with HTTP status code and message
+type APIError struct {
+	Code    int
+	Message string
+	Err     error
+}
+
+// Error implements the error interface
+func (e *APIError) Error() string {
+	if e.Err != nil {
+		return fmt.Sprintf("%s: %v", e.Message, e.Err)
+	}
+	return e.Message
+}
+
+// Unwrap returns the wrapped error
+func (e *APIError) Unwrap() error {
+	return e.Err
+}
+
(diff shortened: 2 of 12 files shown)


go/controller/internal/httpserver/handlers/health.go
```diff
@@ -14,6 +16,9 @@ func NewHealthHandler() *HealthHandler {
 
 // HandleHealth handles GET /health requests
 func (h *HealthHandler) HandleHealth(w http.ResponseWriter, r *http.Request) {
+	log := ctrllog.FromContext(r.Context()).WithName("health-handler")
```
EItanya, line 19: Why are you using `FromContext` in some places, and not in others?

EItanya (commented):
Hey @zanuka, thanks so much for the contribution :)

These changes look awesome! I have a couple of quick questions:
1. Have you looked into a way to do logging at a middleware level. Golang is notoriously verbose so this may be the best way, but I'm wondering if we can insert some sort of generic middleware function which always logs on entrance, and then logs any errors. I doubt we're the first person with that requirement. 
2. If the above is possible, I think it could be good to define error types which allow us to customize the logging from the middleware rather than having custom code for each new handler. 

If we accomplish the 2 above it will make this code more maintainable going forward. In addition it can lay the groundwork for further telemetry such as metrics and tracing.

What do you think?

zanuka:
@EItanya thanks for the feedback, totally agree and will have a revision pushed up soon!

zanuka:
@EItanya updates pushed. happy to make further mods to address any further issues. Mike

EItanya (approved):
So funnily enough this is not exactly the solution I had in mind, but it's pretty clever! Seems like a great start to me. I think you will have to rebase the latest changes.

zanuka:
Thanks for the feedback @EItanya, I've rebased and re-pushed 👍🏼

zanuka:
@EItanya I just fixed the merge conflict resolution error that was causing build to error

EItanya (approved):

EItanya:
Hey @zanuka, if you run `make build-controller` in your local env it should find the errors before you push.

zanuka:
@EItanya I think I've got it fixed up now 🤞🏼

EItanya (approved):

zanuka:
Thanks @EItanya, am happy to pick up more issues if you have any specific ones in mind.

EItanya:
That would be awesome. Here are a couple of potential options:
1. Exposing agent runs as an API: #171 
2. incorrect behavior with Anthropic tools #200 

Do either of those interest you?

zanuka:
@EItanya Cool, you can assign me #171 and I'll get started 🛠️

PR #212, opened 2025-03-29
Api enum

EItanya:
Use CEL based `oneof` for model config and tools.


EItanya:
@ilackarms and @peterj, this PR will make the API `oneofs` better validated, but it is a break from the current API, which is ok given that it's Alpha. However, before merging this we should definitely change the docs.

Also, I'm wondering if we should move to `v1alpha2` so that it's net-new for all users and the upgrade will be a bit more seamless

peterj:
Testing this!

peterj:
@EItanya added a commit that updates the UI to support the updated CRD.

peterj (approved):

EItanya:
I think my plan today is to switch the version to v1alpha2 so the break is more clear, but in order to do that we also need to update the documentation. In addition I am going to update the tag minor version.

peterj:
with the exception of https://kagent.dev/docs/getting-started/configuring-providers, there's nothing else in the docs referring to the CRDs/yaml.

peterj (approved):

PR #216, opened 2025-04-01
fix: cli improvements to handle port forward correctly

harmindersinghk:
The issue was that kagent cli was trying to port forward before the installation and there was no port forward after installation. The k8s resources needed for port forward are added by installation so any port forward before that is useless and we have to do a port forward after the installation.

Also improved cli by adding messages so that if someone tries the commands either as arguments or on cli before installation, they are told to perform installation first.


EItanya:
Hey, thanks so much for the contribution! I am pretty busy at KubeCon right now but will do my best to have a look tomorrow.

go/cli/internal/cli/install.go
```diff
@@ -49,8 +49,34 @@ func InstallCmd(ctx context.Context, c *ishell.Context) {
...
+        return
+    }
+
+    s.Stop()
+    c.Println("kagent installed successfully")
+
+    // Store the port-forward command and context in the shell's context
+    c.Set("portForwardCmd", portForwardCmd)
+    c.Set("portForwardCtx", pfCtx)
+    c.Set("portForwardCancel", pfCancel)
```
EItanya: So I totally understand why you did this, but I think this is a bit too much magic for my taste. Does the CLI library being used here have lifecycle hooks we can hook into to do this. For example `BeforeEach() {checkIfConnected()}`. 

Another option would be to port-forward only the first time it's needed, for example it won't be needed before the install command.

EItanya (commented):
Thanks so much again for the review, one major comment.

go/cli/internal/cli/install.go
harmindersinghk: We will need port forward in main.go anyway before install command. The reason for that is once you do install and exit the cli , port forward will be closed. So when we get into CLI again we need to start the port forward even without install command. 

Happy to remove those 3 lines as we don't use the saved values anywhere for now.

go/cli/cmd/kagent/main.go
```diff
@@ -29,8 +42,13 @@ func main() {
...
 
-	// Attempt to connect to the server
-	_, err = client.GetVersion()
+	// Check server connection for commands passed as arguments
+	if len(os.Args) > 1 && os.Args[1] != "install" && os.Args[1] != "version" {
+		if err := checkServerConnection(client); err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+	}
```
EItanya: How does this work? We don't currently support passing in any args from the command line afaik? This won't actually run the `install` command, right?

go/cli/internal/cli/install.go
EItanya: Yes, please remove, but more importantly how are you cancelling the `port-forward` process. The cancel gets lost here. I think this needs a bit of a refactor to take this into account

EItanya (changes requested):
I really appreciate this PR, but I think we need to take a small step-back and figure out if we can handle this at a higher level, either via lifecycle in the cli library, or lifecycle that we create. For example, if the connection check fails can we try and reconnect, rather than doing it from only the install command?

go/cli/internal/cli/install.go
harmindersinghk: This is where the port-forward will be cancelled when shell is closed:

[https://github.com/kagent-dev/kagent/pull/216/files#:~:text=defer%20func()%20%7B-,defer%20func()%20%7B,%7D(),-//%20create%20new%20shell](url)

harmindersinghk: This where port-forward is cancelled when shell is closed:

`// Ensure the context is cancelled when the shell is closed
	defer func() {
		cancel()
		// cmd.Wait()
		if err := cmd.Wait(); err != nil {
			// These 2 errors are expected
			if !strings.Contains(err.Error(), "signal: killed") && !strings.Contains(err.Error(), "exec: not started") {
				fmt.Fprintf(os.Stderr, "Error waiting for port-forward to exit: %v\n", err)
			}
		}
	}()`

harmindersinghk: port-forward is cancelled when the shell is closed.

go/cli/cmd/kagent/main.go
harmindersinghk: Sorry, thats my misunderstanding the docs in terms of arguments being supported. Will take this out.

harmindersinghk:
> I really appreciate this PR, but I think we need to take a small step-back and figure out if we can handle this at a higher level, either via lifecycle in the cli library, or lifecycle that we create. For example, if the connection check fails can we try and reconnect, rather than doing it from only the install command?

We are not doing it only from install command. Port-forward is set up in two places:

1. When you enter kagent shell. This is going to fail if install hasn't been done yet. No matter how many times you retry its going to fail as the resources to which we are port-forwarding are not there.Once install command is run, and we exit and enter the shell again this is going to work ok as we now have the resources.
2. When install command is run. This is done so that if someone doesn't exit and renters shell after install command the subsequent commands in the same shell still work.

harmindersinghk:
Sorry you will have to rebase with main again.

EItanya (approved):

PR #223, opened 2025-04-08
feat(api): implement external agent invocation endpoint

### LINKED ISSUE #171: expose kagent functionality through the API for external consumption

zanuka:
Add RESTful API endpoint for agent invocation with both sync and async modes. Includes comprehensive unit and e2e tests to validate functionality.

Closes #171


go/controller/internal/httpserver/handlers/invoke.go
```diff
@@ -0,0 +1,166 @@
...
+	Status      string `json:"status"`
+	CompletedAt string `json:"completedAt,omitempty"`
+}
+
+// HandleInvokeAgent handles POST /api/agents/{agentId}/invoke requests
+func (h *InvokeHandler) HandleInvokeAgent(w ErrorResponseWriter, r *http.Request) {
+	log := ctrllog.FromContext(r.Context()).WithName("invoke-handler").WithValues("operation", "invoke")
+
+	// Extract agent ID from path
+	vars := mux.Vars(r)
```
peterj: you could use the GetPathParam in `helpers.go` to extract the parameter from URL

go/controller/internal/httpserver/handlers/invoke.go
```diff
@@ -0,0 +1,166 @@
...
+// InvokeResponse represents a response from an agent invocation
+type InvokeResponse struct {
+	SessionID   string `json:"sessionId"`
+	Response    string `json:"response,omitempty"`
+	StatusURL   string `json:"statusUrl,omitempty"`
+	Status      string `json:"status"`
+	CompletedAt string `json:"completedAt,omitempty"`
+}
+
+// HandleInvokeAgent handles POST /api/agents/{agentId}/invoke requests
```
peterj: Using the ID here is probably fine for now, however I think we should use the agent's name. The ID from the autogen DB isn't exposed anywhere (i.e. the users can't see it); the only "id" they know about is the name of the agent resource.

go/controller/internal/httpserver/handlers/invoke.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Parse request body
+	var invokeRequest InvokeRequest
+	if err = DecodeJSONBody(r, &invokeRequest); err != nil {
+		w.RespondWithError(errors.NewBadRequestError("Invalid request body", err))
+		return
+	}
+
+	// Get user ID from request or header
+	userID := invokeRequest.UserID
```
peterj, line 130: check the GetUserID here https://github.com/kagent-dev/kagent/blob/main/go/controller/internal/httpserver/handlers/helpers.go

go/controller/internal/httpserver/handlers/invoke.go
```diff
@@ -0,0 +1,166 @@
...
+		return
+	}
+	
+	// Convert agent ID to int
+	agentID, err := strconv.Atoi(agentIDStr)
+	if err != nil {
+		if httpWriter, ok := w.(http.ResponseWriter); ok {
+			httpWriter.WriteHeader(http.StatusBadRequest)
+		}
+		w.RespondWithError(errors.NewBadRequestError("Invalid agent ID format, must be an integer", err))
```
peterj, line 119: check RespondWithError https://github.com/kagent-dev/kagent/blob/main/go/controller/internal/httpserver/handlers/helpers.go

go/controller/internal/httpserver/handlers/invoke.go
```diff
@@ -0,0 +1,166 @@
...
+	
+	// Handle client selection for testing vs. production
+	if h.testClient != nil {
+		if mockClient, ok := h.testClient.(mockAutogenClient); ok {
+			session, err = mockClient.CreateSession(sessionRequest)
+		}
+	} else if h.AutogenClient != nil {
+		session, err = h.AutogenClient.CreateSession(sessionRequest)
+	} else {
+		err = fmt.Errorf("no client available")
```
peterj: should we panic here?

go/controller/internal/httpserver/server.go
```diff
@@ -138,6 +139,9 @@ func (s *HTTPServer) setupRoutes() {
 	s.router.HandleFunc(APIPathTeams+"/{teamID}", adaptHandler(s.handlers.Teams.HandleGetTeam)).Methods(http.MethodGet)
 	s.router.HandleFunc(APIPathTeams+"/{teamLabel}", adaptHandler(s.handlers.Teams.HandleDeleteTeam)).Methods(http.MethodDelete)
 
+	// Agents
+	s.router.HandleFunc(APIPathAgents+"/{agentId}/invoke", adaptHandler(s.handlers.Invoke.HandleInvokeAgent)).Methods(http.MethodPost)
```
peterj, line 143: we should have two paths here, one for `invokeSync` and other one for `invokeAsync` (the backend HandleInvokeAgent can be a single func with an if statement as you have it right now).

peterj (changes requested):

peterj:
Thanks for the PR @zanuka! I added some comments.

zanuka:
No problem, thanks for the prompt review @peterj. I'll update today.

go/controller/internal/httpserver/server.go
EItanya: I wonder if there's prior art for APIs like this. I agree that having separate paths feels cleaner, but the more I think about this I wonder how others have done it, and if that's something we can learn from

go/controller/internal/httpserver/handlers/invoke.go
```diff
@@ -0,0 +1,166 @@
...
+		UserID: userID,
+		Name:   fmt.Sprintf("Invocation of agent %d", agentID),
+		TeamID: agentID,
+	}
+	
+	log.V(1).Info("Creating session for agent invocation")
+	var session *autogen_client.Session
+	
+	// Handle client selection for testing vs. production
+	if h.testClient != nil {
```
EItanya: Rather than using separate `test` and `prod` clients. Can we have `h.AutogenClient` be an interface which can be either the `real` or `mock` version. And that way we don't need multiple code paths in the server

go/controller/internal/httpserver/server.go
EItanya: Maybe something along the lines of `invoke` and `start`, or something like that.

EItanya (changes requested):

go/controller/internal/httpserver/handlers/invoke.go
zanuka: thanks for the suggestions, makes sense to me and will update

zanuka:
@peterj @EItanya thanks for the feedback, I think I've addressed everything with the exception of using agent name instead of ID which seems okay for now. 

### full details 💾  
- **updated API design**: added separate, descriptive endpoints (`/invoke` for synchronous and `/start` for asynchronous execution) to better reflect operation intent
- **added helper functions**: replaced direct parameter access with `GetPathParam` and `GetUserID` helpers
- **enhanced error handling**: standardized error responses using `RespondWithError` and proper error types
- **cleaner client architecture**: implemented proper interface for client operations, allowing easier testing without duplicate code paths
- **updated error handling**: added panic for critical configuration errors
- **updated comments**: added Go-idiomatic documentation to all key components

tests are passing, and the endpoints have been manually verified. The implementation maintains backward compatibility with consumers while providing a cleaner, more intuitive API interface.

nothing set in stone so let me know if you have any other suggested mods

peterj:
Looks good to me. Note that we'll have to do the actual implementation on the backend that will invoke the agent. @EItanya we could merge this knowing that it doesn't work yet, and then do the work on the backend and come back to update the code here to hook it up - what do you think?

EItanya:
That approach makes sense to me as well! @peterj and I discussed, and we were wondering if you would want to work on the autogen backend as well. We could point out some prior art and help you get started!

zanuka:
@EItanya thanks for the additional review, I've just rebased and pushed the branch. Working on the autogen backend sounds good to me, happy to help. I'll stay tuned for whatever you might want me to focus on 👍🏼

zanuka (commented):
let me know if there are any other changes to make

EItanya:
So I think there may have been a bit of a misunderstanding about removing the `testClient` from the code. What I mean is that the handler should only have 1 `client`. It should't care whether or not it's mocked or not, that's the problem of the component creating it, that way you can test the handler as a pure function. Does that make sense? I still see reference to `h.testClient`. I think that's the last comment from me.

In terms of the invoke endpoint in Autogen. Here is an existing endpoint which can be looked at for general guidance: https://github.com/kagent-dev/autogen/blob/faec0c007eba443a6a755fd363abbad82195bcb9/python/packages/autogen-studio/autogenstudio/web/serve.py#L4. There are a number of issues with this:
1. No JSON input
2. Requires loading the team from a FS
3. No Async support

We should fix all of those issues in the new API we create.

There are other API examples here: https://github.com/kagent-dev/autogen/blob/faec0c007eba443a6a755fd363abbad82195bcb9/python/packages/autogen-studio/autogenstudio/web/routes

Lmk if you have any other questions.

zanuka:
@EItanya Thanks for the clarification, I see what you mean about the handler and will fix.
Regarding the Autogen endpoint implementation, makes sense to me and I'll review the examples you shared. Will get updates in place soon.

zanuka:
@EItanya new version pushed up for re-review... I think it should address everything now

- removed the testClient field in favor of a single client dependency
- handler no longer cares about the client implementation type
- added a test case for the nil client panic path
- verified both endpoints work correctly against a running server

EItanya (approved):

peterj (approved):
Excellent - thanks for your contribution!

zanuka:
You're welcome! Feel free to assign me something else 🤓

christian-posta:
How should we be thinking about this in terms of A2A? Is this new "invoke" endpoint for custom integrations (ie, random agent or service wants to call the agent) without the need for task tracking and agent response which A2A does?

zanuka:
@christian-posta Thanks for your feedback! I’ll provide a more detailed explanation of the invoke endpoint and how it relates to A2A, with specific references to line numbers in the `invoke.go` file for clarity.

### invoke endpoint
The `/api/agents/{agentId}/invoke` (synchronous) and `/api/agents/{agentId}/start` (asynchronous) endpoints are designed to trigger agent executions a more lightweight way. These endpoints allow external services or agents to invoke an agent without requiring the full complexity of the A2A protocol. 

details:

- **sync invocation (`/invoke`)**: creates a session and a run for the specified agent, then immediately returns a placeholder response. This is handled in the `HandleInvokeAgent` function, where lines 47-50 in `invoke.go` set up the session and return a basic response like `{"status": "ok"}`.

- **async invocation (`/start`)**: creates a session and run but returns a `sessionId` and `statusUrl` so the caller can track progress later. Check out the `HandleStartAgent` function, lines 57-63, where it constructs this response with minimal overhead.

So these 2 endpoints prioritize simplicity over the richer feature set of A2A, making them quick entry points for agent interaction.

### ways that it differs from A2A
The way I understand it, the A2A protocol is all about structured, robust communication between agents... so more of a focus on features like detailed task tracking, state management, and artifact handling. 

Our invoke endpoint, however, takes a leaner approach:

- **no full task tracking**: The `/start` endpoint gives you a `sessionId` and `statusUrl` (lines 57-63), but that’s it—no deep task state tracking or artifact management like A2A provides. It’s a basic “fire and check later” mechanism.

- **simple response handling**: For the synchronous `/invoke`, you get a placeholder response right away (line 50), not the agent’s actual output. The asynchronous `/start` relies on the caller polling the `statusUrl`, which is less interactive than A2A’s real-time updates.

- **basic validation**: Both endpoints validate essentials like `agentId`, `userID`, and the request body—see the `extractAgentParams` function, lines 67-88 in `invoke.go`. But unlike A2A, we're not currently enforcing protocol-specific standards (e.g., Agent Card validation or context requirements).

so invoke skips the A2A bells and whistles for speed and simplicity.

### use case
This endpoint would ideally shine for more **custom integrations** where you need to trigger an agent without A2A’s overhead. Things like testing an MCP tool or firing off a quick agent action from an external script—it’s perfect for more stateless, lightweight scenarios.

### flexibility and future considerations
I think the current design is flexible enough to evolve:
- **A2A compliance**: We could extend the `InvokeRequest.context` struct (parsed around lines 67-88) to validate A2A-specific fields like `endpoint` or `auth`, bringing it closer to protocol standards.
- **richer responses**: The sync endpoint could be tweaked to wait for the agent’s real response instead of returning a placeholder (line 50). This would make it feel more A2A-like without much complexity.

it would be cool to get more feedback from team and capture these details / plans relating to A2A tweaks or other enhancements!

### related issue - [#230](https://github.com/kagent-dev/kagent/issues/230) ( add Test/invoke button )
our invoke endpoint is the key to making this new Test button work. 

Here’s the rundown:

- No A2A complexity: Unlike the full Agent-to-Agent (A2A) protocol, which handles detailed tracking and state management, invoke skips the heavy lifting. 
- Custom integration win: example of a custom integration—users get a simple button to test tools, powered by "invoke," without needing the deeper A2A structure.
- invoke endpoint (/invoke) is a lightweight way to trigger agent executions in the kagent project. 
- we send it an agentId, userID, and a request body, and it kicks off a session and run, returning a simple {"status": "ok"}.
- when a user clicks the "Test" button, we can send a request to the invoke endpoint with details specific to that MCP tool
- agent then runs the tool, and the user gets to test it—all without needing a heavy integration


## more notes on `invoke.go` 

### key functions and lines
- **`HandleInvokeAgent` (Lines 47-50)**  
  handles the synchronous `/invoke` endpoint. Creates a session and run, then returns a placeholder `{"status": "ok"}`.

- **`HandleStartAgent` (Lines 57-63)**  
  manages the asynchronous `/start` endpoint. Sets up a session and run, returning a `sessionId` and `statusUrl` for tracking.

- **`extractAgentParams` (Lines 67-88)**  
  validates request inputs like `agentId`, `userID`, and the request body. Ensures basic integrity but skips A2A-specific checks.

## potential enhancements
- **A2A Validation**: Add checks for `context` fields (e.g., `endpoint`, `auth`) in `extractAgentParams`.
- **Response Upgrade**: Modify line 50 to wait for agent output instead of a placeholder.

sorry for the large response... just decided to share what was in my dev notes

I likely have omitted some details and perhaps @peterj and @EItanya can weigh-in as well since I'm new to the project

PR #239, opened 2025-04-11
add preliminary E2e testing framework

ilackarms:
note that tests do not consistently pass as agent responses can vary. this set of tests lays the groundwork for iterating until we can mark them as "required"

for now, we should enable them to run in CI without requiring them, and manually check test run results on future PRs until we can rely on agent consistency


go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -413,15 +417,16 @@ func (a *apiTranslator) translateTaskAgent(
 		return nil, err
 	}
 
+	agentName := convertToPythonIdentifier(team.Name)
 	return &api.Component{
 		Provider:      "kagent.agents.TaskAgent",
 		ComponentType: "agent",
 		Version:       1,
-		Label:         "society_of_mind_agent",
+		Label:         agentName,
```
peterj, line 425: note that you'll have to update the UI as well as I am checking for this specific name in order to filter it out from the messages.

EItanya: I think I fixed this in: https://github.com/kagent-dev/kagent/pull/239/commits/77faf401f29b64ab0a5cc7eb4c151f1d21405003

EItanya (approved):

PR #249, opened 2025-04-17
feat(controller) read DefaultResourceNamespace from the env variable

### LINKED ISSUE #191: read the DefaultResourceNamespace from the environment/k8s
Check if the default resource NS (kagent) is set in the k8s somewhere and use it where needed (instead of hardcoding it in the code).

bhupixb:
Fixes #191
- Reads the resource namespace from environment variable and removes hardcoded value from code.
- Update the helm deployment controller container to inject namespace as environment variable

### DIFF
--- a/go/controller/internal/httpserver/handlers/teams.go
+++ b/go/controller/internal/httpserver/handlers/teams.go
@@ -9,14 +9,12 @@ import (
 	"github.com/kagent-dev/kagent/go/controller/internal/autogen"
 	"github.com/kagent-dev/kagent/go/controller/internal/client_wrapper"
 	"github.com/kagent-dev/kagent/go/controller/internal/httpserver/errors"
+	common "github.com/kagent-dev/kagent/go/controller/internal/utils"
 	"k8s.io/apimachinery/pkg/types"
 
 	autogen_client "github.com/kagent-dev/kagent/go/autogen/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
-
-// DefaultResourceNamespace is the default namespace for resources
-const DefaultResourceNamespace = "kagent"
 
 // TeamsHandler handles team-related requests
 type TeamsHandler struct {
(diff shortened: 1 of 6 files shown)


go/controller/internal/httpserver/server.go
```diff
@@ -13,10 +14,17 @@ import (
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-const (
-	// Default namespace for resources
-	DefaultResourceNamespace = "kagent"
+// DefaultResourceNamespace is the default namespace for resources
+var DefaultResourceNamespace = getResourceNamespace()
 
+func getResourceNamespace() string {
```
peterj: I'd suggest putting this function into /utils folder and re-use it, instead of implementing it in two places.

bhupixb: Done

go/controller/internal/httpserver/handlers/teams.go
```diff
@@ -9,14 +9,15 @@ import (
...
+	common "github.com/kagent-dev/kagent/go/controller/internal/utils"
 	"k8s.io/apimachinery/pkg/types"
 
 	autogen_client "github.com/kagent-dev/kagent/go/autogen/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // DefaultResourceNamespace is the default namespace for resources
-const DefaultResourceNamespace = "kagent"
+var DefaultResourceNamespace = common.GetResourceNamespace()
```
EItanya: Why do we need this var at all anymore? Can't we just call `common.GetResourceNamespace()`?

go/controller/internal/httpserver/server.go
```diff
@@ -8,15 +8,16 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-const (
-	// Default namespace for resources
-	DefaultResourceNamespace = "kagent"
+// DefaultResourceNamespace is the default namespace for resources
+var DefaultResourceNamespace = common.GetResourceNamespace()
```
EItanya: Same comment as above

EItanya (changes requested):

go/controller/internal/httpserver/handlers/teams.go
bhupixb: Addressed.

PR #252, opened 2025-04-18
Split CRDs into a different chart so they could be fully managed by helm

### LINKED ISSUE #234: old CRD versions are kept around when you run kagent uninstall
Helm has a limitation where CRDs in the CRD folder will never be updated after installation. To solve this most projects have a separate CRDs chart. We should do the same.

gossion:
This pull request introduces the changes to the Helm chart structure and CLI commands for managing the Kagent application. The primary goal is to separate the Kagent Custom Resource Definitions (CRDs) into their own Helm chart, improving modularity, version control, and adherence to Helm best practices. Additionally, the CLI commands for installation and uninstallation have been updated to align with this new structure.

original helm release: 
```
kagent-$version # including kagent and CRDs but CRDs are not fully managed by helm, so CRDs are not delete when uninstall helm
```

new helm releases:
```
kagent-crds-$version # fully managed by helm
kagetn-$version
```

It will also fix #234. Note: For the existing old CRDs deployed by old helm chart:
```
agents.kagent.dev
modelconfigs.kagent.dev
teams.kagent.dev
toolservers.kagent.dev
```
users need to delete those resources manually so the next `helm install` could use the separated CRDs chart.


EItanya:
Hi @gossion, thanks so so much for the contribution. I kicked off CI, and will hopefully have a chance to review later today :)

Makefile
```diff
@@ -94,6 +100,12 @@ helm-install: helm-version check-openai-key kind-load-docker-images
...
 		--set openai.apiKey=$(OPENAI_API_KEY)
 
+.PHONY: helm-uninstall
+helm-uninstall:
+	helm uninstall kagent --namespace kagent
+	helm uninstall kagent-crds --namespace kagent
+
 .PHONY: helm-publish
 helm-publish: helm-version
+	helm push kagent-crds-$(VERSION).tgz oci://ghcr.io/kagent-dev/kagent/helm
```
EItanya, line 110: Note to self: This may fail the release since the OCI repo won't exist beforehand, I can solve that if this happens.

go/cli/internal/cli/install.go
```diff
@@ -25,28 +53,28 @@ func InstallCmd(ctx context.Context, c *ishell.Context) {
...
+		// original kagent installation that had CRDs installed together with the kagent chart
+		if strings.Contains(output, "exists and cannot be imported into the current release") {
+			s.Stop()
+			c.Println("Warning: CRDs already exist but not managed by helm, you might need to delete them manually to make them fully managed by helm.")
+			s.Start()
+		} else {
+			c.Println("Error installing kagent-crds:", output)
+			return
+		}
+	}
```
EItanya, line 72: Not a requirement, but since we are in a `repl` anyway we could offer to do this for the user.

EItanya (commented):
Changes look great to me, looks like there's a merge conflict

Makefile
gossion: Thanks @EItanya ,  please help if that happen.

go/cli/internal/cli/install.go
gossion: That is a good point. 

Here is what I am thinking:
1. InstallCmd - still succeeds (to avoid breaking change) when CRDs are not managed by helm, but warns that users need to manually delete CRDs or run uninstall cmd to delete CRDs.
2. UninstallCmd - adding steps in uninstall to delete CRDs (by kubctl?) if helm release not found but CRDs are there.

how do you think?

EItanya (approved):

PR #259, opened 2025-04-19
Added a new `bug-report` command to the kagent CLI to help gather debug information.

Mukund-cpu:
This commit adds a new 'bug-report' command to the kagent CLI that:
- Gathers Agent, ModelConfig, and ToolServers YAMLs
- Lists secret names (without values)
- Collects pod logs
- Records versions and images used
- Creates a timestamped directory with all debug information

### DIFF
--- a/go/cli/internal/cli/bug_report.go
+++ b/go/cli/internal/cli/bug_report.go
@@ -1,0 +1,93 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/abiosoft/ishell/v2"
+)
+
+func BugReportCmd(c *ishell.Context) {
+	// Create a temporary directory for bug report
+	timestamp := time.Now().Format("20060102-150405")
+	reportDir := fmt.Sprintf("kagent-bug-report-%s", timestamp)
+	if err := os.MkdirAll(reportDir, 0755); err != nil {
+		c.Println("Error creating report directory:", err)
+		return
+	}
+
+	c.Println("Gathering bug report information...")
+
+	// Get Agent, ModelConfig, and ToolServers YAMLs
+	resources := []string{"agent", "modelconfig", "toolserver"}
+	for _, resource := range resources {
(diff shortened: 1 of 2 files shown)


go/cli/internal/cli/bug_report.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		c.Println("Error getting pod images:", err)
+	} else {
+		filename := filepath.Join(reportDir, "versions.txt")
+		if err := os.WriteFile(filename, output, 0644); err != nil {
+			c.Println("Error writing versions file:", err)
+		}
+	}
+
+	c.Printf("Bug report generated in directory: %s\n", reportDir)
```
EItanya, line 123: Can you add a warning here telling people to make sure to scrub the `agent.yaml` of any potential secrets. There are some known situations which we plan to fix, but for now worth telling people to double check.

go/cli/internal/cli/bug_report.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		c.Println("Error getting pod names:", err)
+	} else {
+		pods := strings.Split(string(output), "\n")
+		for _, pod := range pods {
+			if pod == "" {
+				continue
+			}
+			podName := strings.TrimPrefix(pod, "pod/")
+			cmd = exec.Command("kubectl", "logs", "-n", "kagent", podName)
```
EItanya: Right now this only gets the logs for the `controller`. It should get the logs for all of the containers.

EItanya (changes requested):
Just 2 small comments and then good to go!

go/cli/internal/cli/bug_report.go
mukundkumarjha: I have added the warning.

go/cli/internal/cli/bug_report.go
mukundkumarjha: Thanks @EItanya for pointing out i have updated my PR. Please check.

EItanya (approved):

PR #263, opened 2025-04-22
clean CRDs when uninstalling kagent

gossion:
This is implementation of https://github.com/kagent-dev/kagent/pull/252#discussion_r2051881145 to clean CRDs when uninstalling kagent.


gossion:
cc @EItanya

EItanya (approved):

PR #271, opened 2025-04-23
Tool api updates

ilackarms:
update the crds api for consistency & rename "inline" tool to "builtin" tool


go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -32,52 +32,59 @@ type AgentSpec struct {
 	// +kubebuilder:validation:MinLength=1
 	SystemMessage string `json:"systemMessage,omitempty"`
 	// +optional
-	ModelConfigRef string `json:"modelConfigRef"`
+	ModelConfig string `json:"modelConfig,omitempty"`
```
EItanya, line 35: Why change this? You changed `apiKeySecretName -> apiKeySecretRef`, for consistency shouldn't this also be a `ref`?

go/test/e2e/manifests/kube-expert-agent.yaml
```diff
@@ -159,41 +159,91 @@ spec:
...
+apiVersion: kagent.dev/v1alpha1
+kind: ToolServer
+metadata:
+  creationTimestamp: null
+  name: asdf
+  namespace: team-ns
+spec:
+  config:
+    sse:
+      url: https://www.mcp.run/api/mcp/sse?nonce=WrRYKc7jwXSnlwalvjHlzA&username=ilackarms&profile=ilackarms%2Fdefault&sig=GvCWTGTiNh0I_ZqOCx7CeID0KEIVZJnWGpP58eXNUuw
```
peterj: this url should probably be removed

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -42,6 +42,87 @@ var (
 		"kagent.tools.prometheus.GeneratePromQLTool",
 		"kagent.tools.k8s.GenerateResourceTool",
 	}
+
+	// TODO(ilackarms): must make sure that all new builtin tools are added here & their names are documented
+	BuiltinToolNameToProviders = map[string]string{
```
peterj: would this be done from the python side somehow, so the generated provider name is what we want?

peterj: looks like the name has to be that full provider name.

EItanya: But further, why change the name, I'm not sure this is any cleaner. I actually like the original more. I understand getting rid of inline, but the name change seems unnecessary. The more I think about it I actually think it might be a good idea to have the built in tools be strongly typed. If they're inherent to the system then we might as well get schemas as well.

go/controller/api/v1alpha1/autogenagent_types.go
ilackarms: i kept ApiKeySecretRef beacuse i don't want to confuse people that it's actually their api key (this was also Yuval's feedback originally on that PR)
everywhere else though we don't have `ref` in the name of a reference string `toolServer`, `participant`, etc)

go/test/e2e/manifests/kube-expert-agent.yaml
ilackarms: we are removing this whole section in #239 anyway. but sure i'll remove

go/controller/internal/autogen/autogen_api_translator.go
ilackarms: nothing needs to be done on the python side but we will need to update the documentation for builtin tools

EItanya: Bump on my comment

ilackarms: my bad i misunderstood. i thought we originally talked about changing the tool names to some kind of simplified mapping. i can just switch it back to provider if that's the consensus

EItanya (dismissed):

peterj:
This will also require changes in the UI types -- e.g.:

- https://github.com/kagent-dev/kagent/blob/a8b34af293a3e931780c23892e3604b098672f8d/ui/src/types/datamodel.ts#L373
- https://github.com/kagent-dev/kagent/blob/a8b34af293a3e931780c23892e3604b098672f8d/ui/src/lib/toolUtils.ts#L24
- https://github.com/kagent-dev/kagent/blob/a8b34af293a3e931780c23892e3604b098672f8d/ui/src/app/actions/teams.ts#L19 (and other instances the same file)

Just search for "Inline" in the UI and that needs to change.

EItanya (approved):

peterj (approved):

PR #273, opened 2025-04-24
Fix Go controller README and add missing kustomization.yaml

### LINKED ISSUE #250: README for go controller seems out of date
The doc https://github.com/kagent-dev/kagent/tree/main/go mentions commands like `make docker-build` but it doesn't work.
```shell
> make docker-build

make: *** No rule to make target `docker-build'.  Stop.
```

Other command such as `make install`, `make tests`, and `make tests-e2e` also fails for me e.g.:
```shell
/path/kagent/go/bin/kustomize build config/crd | kubectl apply -f -
Error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/Users/bhupendrayadav/Developer/bhupendray/oss/forks/kagent/go/config/crd'
error: no objects passed to apply
make: *** [install] Error 1
```

Mukund-cpu:
- Updated README to match actual Makefile targets
   - Added docker-build and docker-push targets to Makefile
   - Created missing kustomization.yaml in config/crd

   Fixes #250


go/Makefile
```diff
@@ -165,3 +165,18 @@ mv $(1) $(1)-$(3) ;\
...
+DOCKER_REPO ?= kagent-dev/kagent
+IMAGE_NAME ?= controller
+
+.PHONY: docker-build
+docker-build: ## Build the Docker image
+	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_REPO)/$(IMAGE_NAME):$(VERSION) -f Dockerfile .
+
+.PHONY: docker-push
+docker-push: ## Push the Docker image to registry
+	docker push $(DOCKER_REGISTRY)/$(DOCKER_REPO)/$(IMAGE_NAME):$(VERSION)
```
EItanya, line 182: These targets are already in the root `Makefile`. I put all of them there on purpose. Why do we need it here as well?

ui/src/app/layout.tsx
```diff
@@ -21,7 +21,10 @@ export default function RootLayout({ children }: { children: React.ReactNode })
   return (
     <TooltipProvider>
       <AgentsProvider>
-        <html lang="en" className="">
+        <html lang="en"
```
EItanya, line 24: What's the purpose of this change?

go/README.md
EItanya, line 1: In general I think rather than editing this file, we should redo it completely. It's mostly generated boilerplate from `kubebuilder`, but isn't super relevant anymore.

EItanya (changes requested):

PR #275, opened 2025-04-24
Fix tool server url protocol

### LINKED ISSUE #270: [BUG] Adding Tool without protocol causes tool to malfunction (no tools loaded)
**Describe the bug**
When adding a Tool in the UI without specifying the protocol (e.g., http://), the tool fails to function correctly. Specifically, no tools are loaded or displayed. The system does not handle missing protocols gracefully, leading to a silent failure.

**To Reproduce**
1. Go to the "Tools -> Manage tool servers -> Add Server" section in the UI.
2. Enter the Tool address without a protocol, e.g.: 10.0.10.25:8686/mcp
3. Set the name
4. Save the configuration.
5. Observe that no tools are loaded or displayed.
6. Edit the Tool and add the protocol explicitly, e.g.: http://10.0.10.25:8686/mcp
7. Save the configuration again.
8. Tools are now correctly fetched and displayed.

**Expected behavior**
The system should either:

Automatically prepend a default protocol (e.g., http://) if none is provided,
or
Validate the input and prompt the user to include the protocol before saving.

**System Info**
Browser: Chrome (Version 135.0.7049.95 (Official Build) (64-bit))
OS: Ubuntu 24.04
Kagent version: 0.1.16

Mukund-cpu:
When adding a tool server via the UI, URLs without a protocol now automatically get http:// prepended. This prevents the issue where tools weren't loading due to missing protocol in the URL.
   
Fixes #270


ui/src/components/AddServerDialog.tsx
```diff
@@ -309,6 +309,11 @@ export function AddServerDialog({ open, onOpenChange, onAddServer, onError }: Ad
         url: url.trim(),
       };
 
+      // Add http:// protocol if missing
+      if (params.url && !params.url.match(/^[a-z]+:\/\//i)) {
+        params.url = `http://${params.url}`;
+      }
+
```
EItanya: @peterj what do you think about this. It seems to me like we should potentially just validate rather than magically prepend.

EItanya (commented):
The scope of changes for this PR are actually just in `AddServerDialog.tsx`, but the changes from your other PR are present, can you remove them?

ui/src/components/AddServerDialog.tsx
peterj: I agree. I think we should just error out instead of prepending (e.g. call `setError("Please enter a valid URL")` or something likethat)

peterj (changes requested):
Sending this back to you, if you can address that one comment.

mukundkumarjha:
Dear @peterj I have updated my pr accordingly plz review.

EItanya (commented):
Looks good to me, will wait for @peterj for final review

peterj (approved):
looks good! thanks!

EItanya (approved):

PR #279, opened 2025-04-25
Better support additionnal env value

NargiT:
This feature allow to add any key/value map to the deployment from values.yaml

```yaml
app:
  extraEnvValue:
    UV_INDEX: https://example.com/repository/pypi/pypi
    UV_INDEX_URL: https://example.com/repository/pypi/simple

# also works for controller and ui if required
```

Any UV_* env are supported out of the box and allow any one to run the chart in a closed environment (no internet).


NargiT:
Done.

NargiT:
Sorry for the delay, now test are passing.

EItanya:
Hello @NargiT. 2 more quick nits:

1. Can you rename `extraEnvValue -> env`?
2. Can you add default empty maps to the `values.yaml` file as a form of documentation. The values file is typically where users go for documentation. For example:
```
app:
  # Additional environment variables for the app
  env: {}
```

NargiT:
Hey @EItanya, the reason I didn't use env is because the implementation I wrote doesn't support valueFrom values.

[Kubernetes docs on environment variables](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables)

Should I still rename for futur support ?

EItanya:
That's a good point, maybe instead we should just passthrough the value via `{{ toYaml }}` helm function, that was the user can add string values or env values?

NargiT:
sounds good for me.

EItanya (approved):

EItanya:
Hey there, I'm really sorry, I should've merged your PR before the other one, but the another one that modified the env values made it in first, would you mind resolving the conflicts? I don't have access to your fork

PR #282, opened 2025-04-25
Allow to disable completely openai model

NargiT:
This pr goal is to disable totaly openai model one or multiple instead that are local (like ollama)
This also unlock the ability to use different model depending of the agent and thus be able to compare them.

```yaml
# pseudo code for values.yaml

openai:
  enabled: false

contrib:
  agents:
    k8s:
      enabled: true
      modelConfigRef: ollama-model-config # ModelConfig for ollama needs to be define
    helm:
      enabled: true
      modelConfigRef: anthropic-model-config # ModelConfig for anthropic needs to be define
```


EItanya (approved):
LGTM, we were thinking of doing exactly this, thanks so much :)

PR #289, opened 2025-04-28
update the registry and download locations

peterj:


EItanya (dismissed):

Makefile
```diff
@@ -1,5 +1,5 @@
 # Image configuration
-DOCKER_REGISTRY ?= ghcr.io
+DOCKER_REGISTRY ?= cr.kagent.dev
```
EItanya: This will fail release because the build relies on this value being `ghcr.io` https://github.com/kagent-dev/kagent/blob/a8b34af293a3e931780c23892e3604b098672f8d/.github/workflows/tag.yaml#L54

You need to specifically retag and load for any targets that load into kind. That's the most important part in this PR

EItanya (changes requested):

EItanya (approved):

EItanya (approved):

PR #292, opened 2025-04-28
fix: deployment to custom namespace causes errors

### LINKED ISSUE #232: enable creating agents and deploying kagent to other namespaces
Currently, we default everything to the kagent namerspace, so if you install the chart into a different namespace, you'll see issues.

sbx0r:
resolves #232

### DIFF
--- a/helm/kagent/templates/deployment.yaml
+++ b/helm/kagent/templates/deployment.yaml
@@ -34,7 +34,7 @@ spec:
           resources:
             {{- toYaml .Values.controller.resources | nindent 12 }}
           env:
-            - name: POD_NAMESPACE
+            - name: KAGENT_NAMESPACE
               valueFrom:
                 fieldRef:
                   apiVersion: v1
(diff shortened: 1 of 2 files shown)


helm/kagent/templates/deployment.yaml
```diff
@@ -34,7 +34,7 @@ spec:
           resources:
             {{- toYaml .Values.controller.resources | nindent 12 }}
           env:
-            - name: POD_NAMESPACE
+            - name: KAGENT_NAMESPACE
```
EItanya, line 37: This is already added on line 61, can we delete this instance?

EItanya (commented):
Great catch!

helm/kagent/templates/deployment.yaml
sbx0r: Sure. But maybe from the other container to avoid these issues in the future?

sbx0r: Just a heads-up: if [this PR](https://github.com/kagent-dev/kagent/pull/180) gets merged, the KAGENT_NAMESPACE environment variable will no longer be available in the `controller` pod.

I double-checked the `app` container code, and it looks like there’s no dependency on that variable — so it should be safe. 👍

sbx0r: removed:

https://github.com/kagent-dev/kagent/commit/2105128db24d2bb2f6578c4d63cfd1dfcb223b7e#diff-d586510c2235608d20fae4dcc1af6e3288cca45fa738c3e4c3871d3e59c8fd56L61-L64

EItanya: Oh lol, I didn't even notice that was a different `container` 🙈. Well either way you're correct!

EItanya (approved):

PR #297, opened 2025-04-30
Support for Memory

peterj:
This PR implements the initial support for RAG/Memory concept in kagent (note the changes include the CRD and python implementation, but no UI implementation yet).

The initial memory provider is Pinecone. We only support the Pinecone indexes with [integrated embedding](https://docs.pinecone.io/guides/indexes/understanding-indexes#use-integrated-embedding), which means you don't have to provide an embedding model from the kagent side.

The PR adds a resource called Memory that looks like this: 

```yaml
kind: Memory
metadata:
  name: my-pinecone-memory
  namespace: kagent
spec:
  provider: Pinecone
  # Reference to pineone credentials, following the same pattern as with ModelConfig
  apiKeySecretRef: pinecone-credentials
  apiKeySecretKey: PINECONE_API_KEY
  pinecone:
    indexHost: https://mypineconehost.pinecone.io
    # The number of results to return from pinecone (defaults to 5 if not provided)
    topK: 10
    namespace: my-namerspace
    # the score threshold aka similarity - must be between 0.0 and 1.0. If not provided, 
    # all results will be returned (defaults to 0.0), otherwise, only results above the score threshold are returned.
    scoreThreshold: "0.5"
    # Fields to retrieve from the records and store in the memory. If not provided, all fields will be stored in memory. If specific fields are provided, those will be stored as memory, the rest will be stored as metadata.
    recordFields:
      - chunk_text
      - category
```

At the moment, the memory is read-only (i.e., the agent cannot add entries to the memory, but we should consider doing this in the future). The memory is consulted each time a user sends a query - we perform the similarity search and include memories in a new system message that gets sent to the LLM to create the response.  

The memories are added to the system message using this prefix "Your response should include the following memory content: [retrieved memories]". I didn't make this configurable, but we can later on if we want to.


go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -35,6 +35,8 @@ type AgentSpec struct {
 	ModelConfig string `json:"modelConfig,omitempty"`
 	// +kubebuilder:validation:MaxItems=20
 	Tools []*Tool `json:"tools,omitempty"`
+	// +optional
+	Memory []*Memory `json:"memory,omitempty"`
```
ilackarms: should this be a list of references to Memory objects? it looks like we are inlining the whole Memory crd inside of the agent CRD here

python/src/kagent/memory/_pinecone_memory.py
```diff
@@ -0,0 +1,163 @@
...
+    component_type = "memory"
+    component_provider_override = "kagent.memory.PineconeMemory"
+
+    def __init__(self, config: PineconeMemoryConfig):
+        self._config = config
+        self._pc: Pinecone | None = None
+        self._index: Index | None = None
+
+    async def _initialize(self):
+        """Initialize Pinecone and embedding clients if not already done."""
```
EItanya: I don't see the embedding client? Do we want to create a flexible embedding client so that users can bring their own models?

go/controller/api/v1alpha1/autogenagent_types.go
EItanya: Agreed, there is a new CRD referenced below, but then you inline it here. I'm cool with supporting inline as well as references, but that's not quite what's happening here.

python/src/kagent/memory/_pinecone_memory.py
```diff
@@ -0,0 +1,163 @@
...
+from loguru import logger
+from pinecone import Pinecone
+from pinecone.core.openapi.db_data.model.hit import Hit
+from pinecone.data import Index
+from pydantic import BaseModel, Field
+from typing_extensions import Self
+
+
+class PineconeMemoryConfig(BaseModel):
+    api_key: str = Field(..., description="The API key for the Pinecone API")
```
EItanya: Autogen has a way to store this data secretly, they do it for their API keys, can you take a look at that?

go/controller/api/v1alpha1/autogenmemory_types.go
```diff
@@ -0,0 +1,101 @@
...
+type MemoryProvider string
+
+const (
+	Pinecone MemoryProvider = "Pinecone"
+)
+
+// PineconeConfig contains Pinecone-specific configuration options
+type PineconeConfig struct {
+	// The index host to connect to
+	// +required
```
EItanya, line 38: Should we support a distance threshold to control how similar the results need to be?

python/src/kagent/memory/_pinecone_memory.py
```diff
@@ -0,0 +1,163 @@
...
+    component_config_schema = PineconeMemoryConfig
+    component_type = "memory"
+    component_provider_override = "kagent.memory.PineconeMemory"
+
+    def __init__(self, config: PineconeMemoryConfig):
+        self._config = config
+        self._pc: Pinecone | None = None
+        self._index: Index | None = None
+
+    async def _initialize(self):
```
EItanya: Why don't we just do this at `__init__` time?

go/controller/api/v1alpha1/autogenagent_types.go
peterj: yes, should be a list not sure why i inlined it

python/src/kagent/memory/_pinecone_memory.py
peterj: pinecone has built in embeddings - i purposefully didn’t add the embedding client for this pr. we can add it later as we add support for more memories

go/controller/internal/controller/autogenmemory_controller.go
```diff
@@ -0,0 +1,66 @@
...
+// the AutogenAgent object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.0/pkg/reconcile
+func (r *AutogenMemoryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	// TODO(user): your logic here
```
EItanya: Can we remove this and similar lines from the other reconcilers.

go/controller/internal/controller/autogenmemory_controller.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the AutogenAgent object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.0/pkg/reconcile
```
EItanya: Same comment about comments that don't really apply

python/src/kagent/memory/_pinecone_memory.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def _initialize(self):
+        """Initialize Pinecone if not already done."""
+        if self._pc is None:
+            try:
+                self._pc = Pinecone(api_key=self._config.api_key.get_secret_value(), host=self._config.index_host)
+                self._index = self._pc.Index(host=self._config.index_host)
+            except Exception as e:
+                logger.error(f"Failed to initialize Pinecone: {e}")
+                raise Exception(f"Failed to initialize Pinecone: {e}") from e
```
EItanya, line 47: Nit: `Pinecone` and `Index` can no longer be `None` since you are initializing or failing

EItanya (approved):

PR #299, opened 2025-04-30
Implementation of Namespace Filtering with Cache Optimization for KAgent Controller

### LINKED ISSUE #298: [FEATURE] Implementation of Namespace Filtering with Cache Optimization for KAgent Controller
### 📋 Prerequisites

### 📝 Feature Summary

This will add namespace filtering capability to the KAgent controller, allowing it to be configured to watch specified namespaces rather than the entire cluster.

Currently, the KAgent controller watches all namespaces in the Kubernetes cluster, which presents several challenges:

1. **Security Concerns:** The controller has visibility into all namespaces, potentially violating the principle of least privilege.

2. **Performance Impact:** Watching all resources across all namespaces creates unnecessary load on both the Kubernetes API server and the controller itself.

3. **Multi-tenancy Limitations:** In multi-tenant environments, there's no way to restrict the controller to only watch specific tenant namespaces.

4. **Resource Consumption:** Processing events from all namespaces consumes more memory and CPU, even when many of those namespaces aren't relevant to KAgent's operation.

By implementing namespace filtering, we address these issues and provide a more secure, efficient, and configurable controller.

### 💡 Proposed Solution

I propose implementing a namespace filtering mechanism that will:

1. **Enable Selective Namespace Watching:** Add the ability to specify which namespaces the controller should watch via an environment variable.

2. **Create a Custom Predicate Filter:** Implement a filtering predicate that will be applied to all controllers to filter events based on namespace.

3. **Optimize Cache Management:** Implement a namespace-aware cache that only stores resources from the specified namespaces, significantly reducing memory usage.

4. **Update Helm Configuration:** Enhance the Helm chart to allow users to configure the allowed namespaces during installation.

5. **Validate Namespace Names:** Implement validation to ensure that provided namespace names are valid according to Kubernetes naming rules.

6. **Default to Release Namespace:** If no additional namespaces are specified, default to watching all namespaces.

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

Benefits
- Improves security by restricting controller access to only specified namespaces
- Reduces resource usage by only watching relevant namespaces
- Significantly decreases memory footprint by only caching resources from watched namespaces
- Provides better multi-tenancy support in Kubernetes environments
- Gives administrators more control over the scope of the controller's operations
- Enables more efficient scaling in large Kubernetes clusters with many namespaces

sbx0r:
/resolves #298


helm/kagent/templates/deployment.yaml
```diff
@@ -61,10 +69,6 @@ spec:
               value: {{ .Values.otel.tracing.exporter.otlp.timeout | quote }}
             - name: OTEL_EXPORTER_OTLP_TRACES_INSECURE
               value: {{ .Values.otel.tracing.exporter.otlp.insecure | quote }}
-            - name: KAGENT_NAMESPACE
```
sbx0r, line 73: @EItanya 
I'm sure it was deleted. Maybe it was re-added due to wrong conflict resolution or it's mandatory for the app container (?)
Do you know something about it 😄 ?

EItanya: There have been A LOT of changes going into this area lately. I'll check it out

sbx0r: Alrighty.
Ping me whenever you want

EItanya: Ok, @dimetron just merged a pretty substantial (hopefully final) PR for these changes, can you update against that?

.github/workflows/test.yaml
```diff
@@ -2,36 +2,62 @@ name: Test
...
+          OPENAI_API_KEY: fake-key
+        run: |
+          make helm-install
+          kubectl wait --for=condition=Accepted  agents.kagent.dev -n kagent --all --timeout=60s
+
+      # - name: Run E2E tests
+      #   env:
+      #     OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+      #   run: |
+      #     make e2e-test
```
EItanya: Remove comments

go/controller/cmd/main.go
```diff
@@ -56,11 +60,17 @@ import (
...
+	setupLog        = ctrl.Log.WithName("setup")
+	kagentNamespace = utils_internal.GetResourceNamespace()
 )
 
+type ControllerConfig struct {
+	// Namespace configuration
+	AllowedNamespaces string `env:"ALLOWED_NAMESPACES" envDefault:""`
+	WatchNamespaces   string `env:"WATCH_NAMESPACES" envDefault:""`
+}
+
```
EItanya: Why are these env vars and not flags? I'm not saying I particularly like either one more, but all of the existing options are args so maybe we should stick with that for now?

.github/workflows/test.yaml
EItanya, line 1: Can you please move these into `ci.yaml`. This file is currently unused.

go/controller/cmd/main.go
```diff
@@ -225,7 +241,13 @@ func main() {
 		// if you are doing or is intended to do any operation such as perform cleanups
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
-	})
+	}
+
+	watchedNamespaces := ConfigureNamespaceWatching(&managerOptions, config)
+	namespaceFilter, allowedNamespaces := ConfigureNamespaceFiltering(config)
+	CheckNamespaceFilterConsistency(watchedNamespaces, allowedNamespaces)
```
EItanya: unused value?

helm/kagent/values.yaml
```diff
@@ -72,10 +86,12 @@ fullnameOverride: ""
...
 podAnnotations: {}
 
-podSecurityContext: {}
+podSecurityContext:
+  {}
   # fsGroup: 2000
 
-securityContext: {}
+securityContext:
+  {}
```
EItanya: These seem incorrect?

helm/kagent/values.yaml
```diff
@@ -11,10 +11,25 @@ openai:
   secretKey: "OPENAI_API_KEY"
 
 controller:
+  # -- Namespaces the controller is allowed to watch.
+  # The chart's installation namespace is always included automatically.
+  # Specify additional namespaces that should be watched by the controller.
+  # @default -- [] (watches all namespaces)
+  allowedNamespaces:
```
EItanya:
```suggestion
  allowedNamespaces: []
```

helm/kagent/values.yaml
```diff
@@ -11,10 +11,25 @@ openai:
...
+  #  - allowed-ns-1
+  #  - allowed-ns-2
+
+  # -- Namespaces the controller should watch.
+  # If empty, the controller will watch all namespaces (subject to RBAC permissions).
+  # @default -- [] (watches all namespaces)
+  watchNamespaces: []
```
EItanya, line 45: At a high level why would these ever need to be different?

EItanya (changes requested):
Looks good overall, just some questions

helm/kagent/values.yaml
sbx0r: it's valid.
it seems that i have to adjust settings of yaml plugin in my IDE 😄

go/controller/cmd/main.go
sbx0r: not really.
`CheckNamespaceFilterConsistency` function returns additional value for tests simplicity (no log validation and other crazy stuff).
this value isn't usefull for controller's logic.

go/controller/cmd/main.go
sbx0r: alrighty.
I can move it to flags.
just small suggestion for future consideration
https://www.12factor.net/config

helm/kagent/values.yaml
sbx0r: for example:
caching on [ns1,ns2,ns3]
filtering on:
- agents to the [ns1, ns2, ns3]
- tools to the [ns1, ns2]

just for future improvements.
one for cache optimization (for bigger clusters)
other one is recommended for security layer and more specific filtering
https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#watching-resources-in-a-set-of-namespaces

go/controller/cmd/main.go
EItanya: Why are we calling it in band if the value isn't used? Maybe I'm misunderstanding.

go/controller/cmd/main.go
EItanya: This is a great point, I'm definitely open to moving everything over in the future, just wanna be consistent for now.

sbx0r: Sure thing.
Will allign

go/controller/cmd/main.go
sbx0r: This function simply verifies if allowed namespaces are a subset of cached ones. It only displays a warning if they're not. It's purely a validation check with no other actions.

.github/workflows/test.yaml
sbx0r: done.
I've reverted the state of the file

go/controller/cmd/main.go
sbx0r: removed this validation

helm/kagent/values.yaml
sbx0r: After deeper research i've decided to delete predicates and all surrounding logic.
Only namespace watch will persist for now.
The additional layer of predicates may be helpful (but not necessarily) during the RBAC module implementation. For now it only makes a noise and complicates the codebase

helm/kagent/templates/_helpers.tpl
```diff
@@ -56,4 +56,19 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+{{- end }}
+
+{{/*
+Watch namespaces - transforms list of namespaces cached by the controller into comma-separated string
+Includes release namespace and removes duplicates
+*/}}
+{{- define "kagent.watchNamespaces" -}}
+{{- $nsSet := dict }}
+{{- /* Add release namespace */ -}}
+{{- $_ := set $nsSet .Release.Namespace "" }}
```
EItanya: So I think there's actually a bug here. In the controller, the only way to watch all namespaces is by making sure this list is empty, but this function ensures that there's always at least one value in the list, aka the release namespace.

sbx0r: You're right!
I've forgot to commit change for it 😫

EItanya: I agree that the better solution overall is to ONLY watch the release namespace by default, but I also don't want to break existing users. Ideally in the future we could offer an option where the user can use namespace scoped RBAC instead of cluster scoped, but that's definitely out of scope for this PR.

sbx0r: solved!
Thanks a lot for this finding!

EItanya (dismissed):

sbx0r:
@EItanya WAIT!

sbx0r:
Sorry for the alert.
Everything’s fine as far as you keep your CRs in the release namespace.
To support multiple namespaces, I’ll need to update the model — but that’s something I was already aware of (https://github.com/kagent-dev/kagent/issues/232 -- see here).

After merging this PR, I’ll start working on multi-namespace CR handling. It’ll require changes to the data models and updates across all layers of the system (app/controller/UI).

EItanya:
Would you mind creating a quick design in the issue so we're all on the same page?

go/controller/cmd/main.go
```diff
@@ -381,3 +390,45 @@ func waitForReady(f func() error, timeout, interval time.Duration) error {
...
+			setupLog.Info("Ignoring invalid namespace name",
+				"namespace", ns,
+				"validation_errors", strings.Join(errs, ", "))
+		} else {
+			validNamespaces = append(validNamespaces, ns)
+		}
+	}
+
+	return validNamespaces
+}
```
EItanya, line 434: I'm going to approve, but in a follow-up please move these functions and tests into a different package.

EItanya (approved):
LGTM

PR #301, opened 2025-04-30
A2a invocation

ilackarms:
adds support for A2A via the controller's HTTP Server + Skills which can now be defined on the Agent CRD. An Agent Card will automatically be generated from these skills + the agent description.

Implemented:
- Invoke (synchronous calls)
- A2A Server

Current todos for a followup:
- Streaming
- Testing
- A2A Client in CLI
- Deprecate websocket for chat
- Remove User Proxy & rely on A2A streaming for user interacton


go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -35,6 +37,8 @@ type AgentSpec struct {
 	ModelConfig string `json:"modelConfig,omitempty"`
 	// +kubebuilder:validation:MaxItems=20
 	Tools []*Tool `json:"tools,omitempty"`
+	// +optional
+	A2AConfig *A2AConfig `json:"a2aConfig,omitempty"`
```
EItanya, line 49: Can you add some docs here about when this is used etc.

go/controller/cmd/main.go
```diff
@@ -109,6 +111,7 @@ func main() {
 	flag.StringVar(&defaultModelConfig.Name, "default-model-config-name", "default-model-config", "The name of the default model config.")
 	flag.StringVar(&defaultModelConfig.Namespace, "default-model-config-namespace", podNamespace, "The namespace of the default model config.")
 	flag.StringVar(&httpServerAddr, "http-server-address", ":8083", "The address the HTTP server binds to.")
+	flag.StringVar(&a2aBaseUrl, "a2a-base-url", "http://127.0.0.1:8083", "The base URL of the A2A Server endpoint, as advertised to clients.")
```
EItanya, line 118: Can we serve this on the standard HTTP endpoint, why open up yet another port? Part of the value IMO is making it seamless with the rest of the system

go/controller/cmd/main.go
```diff
@@ -253,11 +256,20 @@ func main() {
 		defaultModelConfig,
 	)
 
+	a2aHandler := a2a.NewA2AHttpMux(httpserver.APIPathA2A)
+
+	a2aReconciler := a2a.NewAutogenReconciler(
+		autogenClient,
+		a2aHandler,
+		a2aBaseUrl+httpserver.APIPathA2A,
+	)
```
EItanya, line 274: I can dig in further, but at a high level why do we need a whole new reconciler for this? It seems like we should have it be a part of the same reconciler so that the translation is an atomic unit. Fails/Succeeds and then we can update the status.

go/controller/internal/a2a/a2a_task_processor.go
```diff
@@ -0,0 +1,110 @@
...
+	// Create response message.
+	responseMessage := protocol.NewMessage(
+		protocol.MessageRoleAgent,
+		[]protocol.Part{protocol.NewTextPart(fmt.Sprintf("Processed result: %s", result))},
+	)
+
+	// Update task status to completed.
+	if err := handle.UpdateStatus(protocol.TaskStateCompleted, &responseMessage); err != nil {
+		return fmt.Errorf("failed to update task status: %w", err)
+	}
```
EItanya, line 68: I just want to note this for later that this state mgmt will become a bit more awkward when the final state exists in the autogen DB and the client code is here.

EItanya (commented):
LGTM overall, just a few questions

go/controller/cmd/main.go
ilackarms: This is not opening another port. It's the same port we use for the HTTP server (check the line above). However we need to allow users to set this URL since it's the URL advertised to clients by the A2A protocol (when they first fetch the agent card). When they access via a proxy (or kube dns, etc) they'll want to change this to reflect the value expected by clients.

go/controller/cmd/main.go
ilackarms: this reconciler is called by the translation reconciler so it does get called in lockstep with the API translation. I simply wrapped the A2A logic in its own reconciler to wrap it all in a single function call from within the translation reconciler.

go/controller/internal/a2a/a2a_task_processor.go
ilackarms: i don't think we'll be storing the `task` state in autogen, just the session content if that makes sense

go/cli/cmd/kagent/main.go
```diff
@@ -43,23 +43,38 @@ func main() {
...
+			// These 2 errors are expected
+			if !strings.Contains(err.Error(), "signal: killed") && !strings.Contains(err.Error(), "exec: not started") {
+				fmt.Fprintf(os.Stderr, "Error waiting for port-forward to exit: %v\n", err)
 			}
-		}()
-	}
-
+		}
+	}()
+	a2aPortFwdCmd := exec.CommandContext(ctx, "kubectl", "-n", "kagent", "port-forward", "service/kagent", "8083:8083")
```
EItanya: Why do we need a separate port for this? I thought you said explicitly it's the same port

go/cli/internal/config/config.go
```diff
@@ -40,6 +41,7 @@ func Init() error {
 	viper.SetDefault("user_id", "admin@kagent.dev")
 	viper.SetDefault("output_format", "table")
 	viper.SetDefault("namespace", "kagent")
+	viper.SetDefault("a2a_url", "http://localhost:8083/api/a2a")
```
EItanya, line 44: Same question about a new port

go/controller/api/v1alpha1/autogenagent_types.go
```diff
@@ -37,6 +39,8 @@ type AgentSpec struct {
 	Tools []*Tool `json:"tools,omitempty"`
 	// +optional
 	Memory []string `json:"memory,omitempty"`
+	// +optional
+	A2AConfig *A2AConfig `json:"a2aConfig,omitempty"`
```
EItanya, line 49: Comments

go/cli/cmd/kagent/main.go
ilackarms: it's the 8083 http server port (not the 8081 autogen port)

go/cli/internal/config/config.go
EItanya: Ohhh, because the CLI uses autogen directly right now. Crap that's so confusing.

go/cli/cmd/kagent/main.go
EItanya: Since this is only used by the `a2a` functionality, can we just port-forward before making the call and then cancel the command. This code is super ugly right now and I don't think we should add to it.

ilackarms: sure i'll make that change

EItanya (approved):

PR #309, opened 2025-05-02
Add UI for memory

peterj:
+ required backend http handlers and a missing role.yaml that includees memories resource

### DIFF
--- a/go/controller/internal/httpserver/handlers/helpers.go
+++ b/go/controller/internal/httpserver/handlers/helpers.go
@@ -1,12 +1,18 @@
 package handlers
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/gorilla/mux"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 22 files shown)


peterj:
I am not too crazy about the term "Memory" -- it's too generic and doesn't really tell the user what it is. The backend objects/crds being called memory should be ok (as it's just a connection to the DB); however in our current case the memory is used more as a "knowledge" or knowledge base (a readonly database). 

I am sure we'll probably add other types of memory where either the user preferences are stored (e.g. per agent or per session, per deployment memory) or the user adds agent response to the "memory" by clicking 👍🏼 👎🏼  -- would that be called "memory" as well?

go/controller/internal/httpserver/handlers/helpers.go
```diff
@@ -99,3 +105,56 @@ func DecodeJSONBody(r *http.Request, target interface{}) error {
 	log.V(2).Info("Successfully decoded JSON request body")
 	return nil
 }
+
+// flattenStructToMap uses reflection to add fields of a struct to a map,
+// using json tags as keys.
+func FlattenStructToMap(data interface{}, targetMap map[string]interface{}) {
```
EItanya, line 111: I think a simpler way to do this would be to marshal into JSON, and the unmarshal into a map[string]interface{}

EItanya:
> I am not too crazy about the term "Memory" -- it's too generic and doesn't really tell the user what it is. The backend objects/crds being called memory should be ok (as it's just a connection to the DB); however in our current case the memory is used more as a "knowledge" or knowledge base (a readonly database).
> 
> I am sure we'll probably add other types of memory where either the user preferences are stored (e.g. per agent or per session, per deployment memory) or the user adds agent response to the "memory" by clicking 👍🏼 👎🏼 -- would that be called "memory" as well?

This is a great point. What about modes of memory? Short-term, read-only, write, etc.

EItanya (approved):

PR #310, opened 2025-05-02
UI: Prevent null value dereference

nimishamehta5:
While setting up the kagent Quickstart, https://kagent.dev/docs/getting-started/quickstart, creation of a simple Kubernetes agent chat does not work due to the error:
```
151-dadee709827cfc46.js:1 Uncaught TypeError: Cannot read properties of undefined (reading 'endsWith')
```
This error can be seen in the browser console. The stack trace points to the `source` being null and the program trying to access `endsWith` on a null source. This change adds a type check for the source so that it does not crash the UI.


peterj:
could you check the logs from the kagent pod? that will actually give us the root cause for the error (we will have to fix that as well at some point)

peterj (approved):

nimishamehta5:
> could you check the logs from the kagent pod? that will actually give us the root cause for the error (we will have to fix that as well at some point)

Makes sense, I will investigate further and report / fix any issues

Issue #288, opened 2025-04-28
[BUG] run chat panic & unknown flag -s


KIOSHIROI:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [ ] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

I tried to use CLI instead of UI and meet this.
```bash
kagent >> run chat k8s-agent
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x48 pc=0x104a6b590]

goroutine 1 [running]:
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd.func1(...)
	/home/runner/work/kagent/kagent/go/cli/internal/cli/chat.go:89
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd.Collect[...].func4-range3(...)
	/home/runner/work/kagent/kagent/go/cli/internal/cli/format.go:34
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd.Values[...].func3(...)
	/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.1.linux-amd64/src/slices/iter.go:40
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd.Filter[...].func4(...)
	/home/runner/work/kagent/kagent/go/cli/internal/cli/format.go:33
slices.AppendSeq[...](...)
	/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.1.linux-amd64/src/slices/iter.go:50
slices.Collect[...](...)
	/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.1.linux-amd64/src/slices/iter.go:58
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd(0x140001180d0)
	/home/runner/work/kagent/kagent/go/cli/internal/cli/chat.go:88 +0x530
main.main.func3(0x140001180d0)
[...]

PR #320, opened 2025-05-03
CLI: Minor fixes and improvements

nimishamehta5:
Fixes a couple of small issues.

* `run chat` with args used to result in a SIGSEGV because while getting the `team`, params were passed in the wrong order so we got a **nil** object for team. Also the `-s` flag for session was not respected. 

```
kagent >> run chat blah
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x48 pc=0x1026587c8]

goroutine 1 [running]:
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd.func1(...)
	/Users/nim/src/github.com/nimishamehta5/kagent/go/cli/internal/cli/chat.go:109
github.com/kagent-dev/kagent/go/cli/internal/cli.ChatCmd.Collect[...].func4-range3(...)
...
```

Now, with an incorrect chat name, it prompts the user to select from the available teams:
```
kagent >> run chat blah
Select an agent:
 ❯ my-first-k8s-agent
   promql-agent
   observability-agent
   k8s-agent
   istio-agent
   helm-agent
   argo-rollouts-conversion-agent
   RoundRobin Team
```

With a correct chat name, it now prompts for a session:
```
Select a session:
 ❯ [New Session]
   first
```

And with a session name passed in, the chat just works:
```
kagent >> run chat k8s-agent -s first
Enter a task: List nodes
Closing session, thank you :)
```

* `get run $id` did not work and gave the error:
```
kagent >> get run 20
Failed to get run 20: error unmarshaling into result: json: cannot unmarshal number into Go struct field Run.id of type string
```
This is because the autogen API actually returns an int, but we were trying to store it as a string, so the unmarshaling didn't work.
API response gives an int:
```
 ✗ curl -X GET "http://localhost:8081/api/runs/20" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  7722  100  7722    0     0   729k      0 --:--:-- --:--:-- --:--:--  754k
{
  "status": true,
  "data": {
    "updated_at": "2025-05-03T19:00:15.975556",
    "status": "error",
    "team_result": {
      "task_result": {
        "messages": [
          {
            "source": "system",
            "models_usage": null,
            "metadata": {}
          }
        ],
        "stop_reason": "An error occurred while processing this run"
      },
      "usage": "",
      "duration": 0.0
    },
    "version": "0.0.1",
    "user_id": "admin@kagent.dev",
    "created_at": "2025-05-03T19:00:13.953835",
    "id": 20,                <----- THIS PART
```

So now we store it as int and this works:
```
kagent >> get run 20
{
  "id": 20,
  "session_id": 13,
  "created_at": "2025-05-03T19:00:13.953835",
  "status": "error",
  "task": {
    "source": "user",
    "content": [
...
```

* In the print formatting, we now print the `#` column starting from human-friendly index 1 instead of 0.

### DIFF
--- a/go/autogen/client/run.go
+++ b/go/autogen/client/run.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/google/uuid"
 )
@@ -12,9 +13,15 @@ func (c *Client) CreateRun(req *CreateRunRequest) (*CreateRunResult, error) {
 	return &run, err
 }
 
-func (c *Client) GetRun(runID string) (*Run, error) {
+func (c *Client) GetRun(runIDStr string) (*Run, error) {
+	// Convert to integer runID
+	runID, err := strconv.Atoi(runIDStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid run ID: %s, must be a number: %w", runIDStr, err)
+	}
+
 	var run Run
-	err := c.doRequest("GET", fmt.Sprintf("/runs/%s", runID), nil, &run)
+	err = c.doRequest("GET", fmt.Sprintf("/runs/%d", runID), nil, &run)
 	return &run, err
 }
 
--- a/go/autogen/client/run.go
+++ b/go/autogen/client/run.go
@@ -2,7 +2,6 @@ package client
 
 import (
 	"fmt"
-	"strconv"
 
 	"github.com/google/uuid"
 )
@@ -13,15 +12,10 @@ func (c *Client) CreateRun(req *CreateRunRequest) (*CreateRunResult, error) {
 	return &run, err
 }
 
-func (c *Client) GetRun(runIDStr string) (*Run, error) {
-	// Convert to integer runID
-	runID, err := strconv.Atoi(runIDStr)
-	if err != nil {
-		return nil, fmt.Errorf("invalid run ID: %s, must be a number: %w", runIDStr, err)
-	}
+func (c *Client) GetRun(runID int) (*Run, error) {
 
 	var run Run
-	err = c.doRequest("GET", fmt.Sprintf("/runs/%d", runID), nil, &run)
+	err := c.doRequest("GET", fmt.Sprintf("/runs/%d", runID), nil, &run)
 	return &run, err
 }
 
--- a/go/cli/internal/cli/chat.go
+++ b/go/cli/internal/cli/chat.go
@@ -55,27 +55,10 @@ func ChatCmd(c *ishell.Context) {
 			c.Println(err)
 			return
 		}
-		if team == nil {
-			c.Printf("Error: Could not find team '%s'\n", teamName)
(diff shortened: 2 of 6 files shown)


nimishamehta5:
Just realized this also fixes https://github.com/kagent-dev/kagent/issues/288

go/autogen/client/run.go
```diff
@@ -12,9 +13,15 @@ func (c *Client) CreateRun(req *CreateRunRequest) (*CreateRunResult, error) {
 	return &run, err
 }
 
-func (c *Client) GetRun(runID string) (*Run, error) {
+func (c *Client) GetRun(runIDStr string) (*Run, error) {
+	// Convert to integer runID
+	runID, err := strconv.Atoi(runIDStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid run ID: %s, must be a number: %w", runIDStr, err)
+	}
```
EItanya: Should this argument be an int instead?

go/cli/internal/cli/chat.go
```diff
@@ -49,11 +50,31 @@ func ChatCmd(c *ishell.Context) {
...
+				return
+			}
+
+			c.Println("Available teams:")
+			if err := printTeams(teams); err != nil {
+				c.Printf("Failed to print teams: %v\n", err)
+				return
+			}
+			return
+		}
```
EItanya: Rather than printing teams, should this error move the user into the team selection logic? The same flow as the else below.

EItanya (changes requested):

EItanya:
Changes look awesome overall! Thanks so much for the contribution :)

go/autogen/client/run.go
nimishamehta5: I could move the conversion from string to int into the calling func of this func, but it has to be done somewhere, since the args are a list of strings by default.
```
// Context is an ishell context. It embeds ishell.Actions.
type Context struct {
...
	// Args is command arguments.
	Args []string

	// RawArgs is unprocessed command arguments.
	RawArgs []string
...
```

go/cli/internal/cli/chat.go
nimishamehta5: good suggestion, updated

go/autogen/client/run.go
EItanya: Makes sense. In this case I think it should be up to the caller, mostly because we know it has to be an int, why even allow a string. But that's pretty arbitrary so happy to go either way.

nimishamehta5: It's a better design to have the caller convert the string to an integer before passing it to GetRun, so I moved that logic into the caller now, thanks for the feedback!

EItanya (approved):

PR #322, opened 2025-05-04
fix(ui): add support for Ollama model tags in the UI

### LINKED ISSUE #307: [BUG] Allow providing model tags for Ollama models
### 📋 Prerequisites

### 🎯 Affected Service(s)

UI Service

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

We default to the :latest tag for Ollama models. In the UI we should add an input box that would allow users to explicitly set the model tag they want to use (e.g. qwen2.5:05.b). 

Note that it's not possible to pick specific tags in the UI as of right now (we always default to latest), but you can do it by updating the ModelConfig resource.

### 🔄 Steps To Reproduce

see above

### 🤔 Expected Behavior

_No response_

### 📱 Actual Behavior

_No response_

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

sbx0r:
This PR /resolves #307 issue. It adds ability to specify custom model tags when creating or editing Ollama model configurations in the UI

Changes:
- Added a "Model Tag" input field that appears only when Ollama provider is selected
- Default tag value is set to "latest" when not specified
- Automatically includes the tag in the model name when it's different from "latest"
- Sends the model name to the API in the format model:tag when a custom tag is specified
- Stores the tag value in the model parameters for reference
- Updates the auto-generated model resource name to include the tag for better identification
- The tag input helps users specify which version of an Ollama model they want to use, making it easier to work with different versions of the same model.

The tag input helps users specify which version of an Ollama model they want to use, making it easier to work with different versions of the same model.

> [!IMPORTANT]
> This change currently **prevents updating** the model tag on existing ModelProvider resources.
>
> **Here’s why:**
> When using the automatic name generator, the resulting ModelProvider name follows the convention
> {provider}-{model}-{tag} — for example, ollama-starcoder2-3b.
> If the tag is changed without updating the resource name, it may lead to confusion — e.g., a model named ollama-deepseek-r1-7b might actually point to 70b, or vice versa.
>
> To ensure stability, the controller must either:
>   - guarantee that all dependent Teams/Agents will use the updated model, or
>   - prohibit such updates altogether.
>
> This needs to be discussed further.

--- 

After this PR merges the 'Create New Model' card looks like this:
- [non-ollama]
image
- [ollama with latest tag]
image
- [ollama with non-latest tag]
image

Resulting model will look like this:

- [ui]
image
- [resulting k8s manifest]
  ```yaml
   kind: ModelConfig
   metadata:
     name: ollama-dolphin-mistral-70b
     namespace: kagent
   spec:
     apiKeySecretKey: ""
     apiKeySecretRef: ""
     model: dolphin-mistral:70b
     ollama: {}
     provider: Ollama
  ```


ui/src/app/models/new/page.tsx
```diff
@@ -366,13 +384,27 @@ function ModelPageContent() {
...
     setErrors({});
 
+    let finalModelName = name.trim();
+    let finalModelWithTag = modelName;
+    if (finalSelectedProvider.type === 'Ollama' && modelTag.trim() !== 'latest') {
+      if (!isEditingName && !isEditMode) {
+        const tagSuffix = `-${modelTag.trim()}`;
+        finalModelName = finalModelName + tagSuffix;
+
+        if (!isResourceNameValid(finalModelName)) {
```
peterj: recently I added `createRFC1123ValidName` -- perhaps use that to create the name? (so you don't have to check whether it's valid and use a regex).

ui/src/app/models/new/page.tsx
```diff
@@ -264,14 +278,18 @@ function ModelPageContent() {
       if (parts.length === 2) {
         const providerKey = parts[0];
         const modelName = parts[1];
-        const baseName = `${providerKey}-${modelName}`.toLowerCase();
+        let baseName = `${providerKey}-${modelName}`.toLowerCase();
+        const isOllama = selectedProvider?.type === "Ollama";
+        if (isOllama && modelTag && modelTag !== 'latest') {
```
peterj: nit: create a constant outside of the component for `latest` (E.g. `const LATEST_TAG="latest"`)

ui/src/app/models/new/page.tsx
```diff
@@ -114,6 +114,7 @@ function ModelPageContent() {
   const [providerModelsData, setProviderModelsData] = useState<ProviderModelsResponse | null>(null);
   const [selectedCombinedModel, setSelectedCombinedModel] = useState<string | undefined>(undefined);
   const [selectedModelSupportsFunctionCalling, setSelectedModelSupportsFunctionCalling] = useState<boolean | null>(null);
```
peterj, line 117: perhaps you can do the change in the wizard as well as a separate PR (doesn't have to be part of this one). Thanks a ton for doing this!

peterj (changes requested):

peterj:
another issue I noticed -- looks like the "0.5b" tag is applied twice to the name (I think this will be probably resolved once you switch to the method to create valid name).

Screenshot 2025-05-04 at 2 43 35 PM

peterj:
also, check that the editing works - it looks like the provider is not filled out and the tag is readonly
Screenshot 2025-05-04 at 2 50 06 PM
as well -- right now, it

ui/src/app/models/new/page.tsx
sbx0r: My pleasure!
I'm flexible. If you'd like this change quickly, I can open a follow-up issue an PR.
LMK what works best for you.

sbx0r:
> another issue I noticed -- looks like the "0.5b" tag is applied twice to the name (I think this will be probably resolved once you switch to the method to create valid name).
> 
> Screenshot 2025-05-04 at 2 43 35 PM

Thanks for checking it out!
I’ll use your validator and run it against the edge case you spotted.

sbx0r:
> also, check that the editing works - it looks like the provider is not filled out and the tag is readonly Screenshot 2025-05-04 at 2 50 06 PM as well -- right now, it

Yup. I found the issue with empty provider but didn't want to add noise to the PR. I can fix it here if that's okay with you.
For the model tag lock - I've some thoughts. Please check the PR description (see the 'Important' section)

ui/src/app/models/new/page.tsx
sbx0r: current implementation of createRFC1123ValidName have this logic

```ts
  //...

  // Final validation check - though sanitization should make it valid
  if (!isResourceNameValid(combined)) {
      console.warn(`Generated name '${combined}' is still not RFC 1123 valid after sanitization.`);
      // Returning potentially invalid name, caller should ideally re-validate
      // Or return "" to indicate failure? Returning the attempt might be more informative.
      return combined;
  }

  //...
```

Could you tell why is it there?

peterj: good point -- I think that was just a check in case it still couldn't create the valid name; I think we could keep this as is and see if anyone even runs into it.

ui/src/app/models/new/page.tsx
peterj: it's up to you -- if you don't mind adding it as part of this PR that would be good. but if not, a separate pr/issue is great as well.

peterj:
> > also, check that the editing works - it looks like the provider is not filled out and the tag is readonly Screenshot 2025-05-04 at 2 50 06 PM as well -- right now, it
> 
> Yup. I found the issue with empty provider but didn't want to add noise to the PR. I can fix it here if that's okay with you. For the model tag lock - I've some thoughts. Please check the PR description (see the 'Important' section)

It would be a pain to go and update all agents when renaming the model; but I don't want to prevent people from updating the models once they created them. Perhaps we can only allow editing the models parameters (e.g. API key, baseUrl, etc.) but not the actual model name/tag.

ui/src/app/models/new/page.tsx
sbx0r: alright. should I validate again after the function call or the console.warn will be enough?

peterj: I think warn should be enough -- and we'll see in which scenarios users run into this (if ever)

ui/src/app/models/new/page.tsx
sbx0r: I will add it to this PR then :)

ui/src/app/models/new/page.tsx
sbx0r: done

ui/src/app/models/new/page.tsx
sbx0r: done

sbx0r:
All is done.
I also added minor UX improvement to the wizard (set dynamic `model config name` if not entered manually by user) and fixed provider icons to make 'em visible in both - light and dark modes.

Ohh...
and I modified helm to expose the port of the engine (8083) which helps a lot in local development

helm/kagent/templates/deployment.yaml
```diff
@@ -50,6 +50,10 @@ spec:
           {{- with .Values.controller.env }}
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.service.ports.controller.targetPort }}
```
peterj, line 55: @EItanya you fine with exposing this (mostly for testing purposes)?

EItanya: Yea I don't see why not, seems good to have anyway

sbx0r: Just a heads-up - once the engine, controller, and UI are running in separate pods, we'll need to expose this port anyway. Right now, they're all sharing the same network interface.

peterj:
@sbx0r  Looks like there are some linting errors in the UI (you can run `npm run build` locally to see the issues/fix them:))

sbx0r:
> @sbx0r Looks like there are some linting errors in the UI (you can run `npm run build` locally to see the issues/fix them:))

Thanks!
I'm working on it.
Should be ready in a minute

Sorry for that. I was tesing it via `npm run dev` and forgot to run `make` before pinging you

peterj:
@sbx0r , looks like that one issue I mentioned earlier is still there -- if I use the tag "0.5b", in the creation page, the name of the resource looks right (e.g. `ollama-qwen2-5-0-5b`), however, when the resource is actually created, it creates "ollama-qwen2-5-0-5b-0.5b".

sbx0r:
> @sbx0r , looks like that one issue I mentioned earlier is still there -- if I use the tag "0.5b", in the creation page, the name of the resource looks right (e.g. `ollama-qwen2-5-0-5b`), however, when the resource is actually created, it creates "ollama-qwen2-5-0-5b-0.5b".

hmmm.....
Looks like an issue related to the OS/Browser/etc.

It works like a charm on my machine
image

image

sbx0r:
could you please tell me something more about your environment @peterj?

peterj:
you're completely right; this was me -- the GH pull request extension doesn't actually refresh the PR so I had old code -- anyway. it works -- let's wait for the CI and it's good to go!

sbx0r:
Great to hear (that it's not a platform issue LoL 😄)!

peterj (approved):

EItanya (approved):

PR #332, opened 2025-05-05
One click ENV setup with codespaces / devcontainer

dimetron:
One click ENV setup 
Support for codespaces / devcontainer

- [x] Updated readme with badge start in codespaces
- [x] Added required .devcontainer Dockerfile and settings


dimetron:
Build: https://github.com/dimetron/kagent/actions/runs/14845491446

go/Makefile
```diff
@@ -46,6 +46,8 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 	$(GOLANGCI_LINT) config verify
 
 ##@ Build
+
+.PHONY:bin/kagent-local
```
EItanya: This shouldn't be necessary because the target is an actual file.

go/Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION ?= $(shell git describe --tags --always --dirty)
+VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null | grep -i ^v.\*-dirty || echo "v0.0.0-local")
```
EItanya: Can we just strip dirty? This will make image loading into `kind` more brittle

Makefile
```diff
@@ -4,7 +4,7 @@ DOCKER_REPO ?= kagent-dev/kagent
 CONTROLLER_IMAGE_NAME ?= controller
 UI_IMAGE_NAME ?= ui
 APP_IMAGE_NAME ?= app
-VERSION ?= $(shell git describe --tags --always --dirty)
+VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null | grep -i ^v.\*-dirty || echo "v0.0.0-local")
```
EItanya: Same comment about image tagging

Makefile
```diff
@@ -177,4 +208,13 @@ helm-publish: helm-version
 .PHONY: kagent-cli-install
 kagent-cli-install: build-cli-local helm-version kind-load-docker-images
 kagent-cli-install:
-	KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local
\ No newline at end of file
+	KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local
+
+.PHONY: kagent-cli-port-forward
```
EItanya, line 213: Where is this used?

Makefile
```diff
@@ -177,4 +208,13 @@ helm-publish: helm-version
...
+kagent-cli-port-forward: use-kind-cluster
+	@echo "Port forwarding to KAgent CLI..."
+	kubectl port-forward -n kagent service/kagent 8081:8081 8082:80
+
+.PHONY: build-dev-container
+build-dev-container:
```
EItanya, line 219: Where is this used?

go/cli/internal/cli/install.go
```diff
@@ -115,15 +115,14 @@ func InstallCmd(ctx context.Context, c *ishell.Context) {
 	}
 
 	// Wait for port-forward to be ready
-	time.Sleep(2 * time.Second)
+	time.Sleep(3 * time.Second)
```
EItanya: Not necessary for this PR, but we need a better solution here. Maybe move away from `kind` and use `k3d` which ships with LB service support

go/cli/internal/config/utils.go
```diff
@@ -42,3 +46,29 @@ func BoldYellow(s string) string {
 func BoldRed(s string) string {
 	return color.New(color.FgRed, color.Bold).SprintFunc()(s)
 }
+
+func GetConfigDir(homeDir string) (string, error) {
```
EItanya, line 50: Why pass in `homeDir` and not call `os.GetHomeDir` inside of this function?

go/go.mod
```diff
@@ -133,3 +133,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace golang.org/x/net => golang.org/x/net v0.38.0
```
EItanya: Why is this necessary?

.vscode/settings.json
EItanya, line 1: What do you think about deleting this file? It should probably be gitignored 😆

go/go.mod
dimetron: fix [CVE-2025-22870](https://github.com/advisories/GHSA-qxp5-gwg8-xv66) and [CVE-2025-22872](https://groups.google.com/g/golang-codereviews/c/-Gvu3yp52js).

go/cli/internal/config/utils.go
dimetron: Because of tests

go/cli/internal/cli/install.go
dimetron: Maybe, it's not very stable - I have to run several times install

Makefile
dimetron: inside remote devcontainer for example

Makefile
dimetron: local testing of devcontainer Docker, since we do not publish it yet
later we can add it to release of tools devcontainer image

.vscode/settings.json
dimetron: I'd keep it for dev containers flow, when open project in GitHub Code Spaces browser 
this file disables some default annoying popups,

go/cli/internal/config/utils.go
EItanya: If it's a test issue, then perhaps the function should take a callback or an interface. e.g. `func () (*os.File, error)` or something like that?

go/go.mod
EItanya: Ok, but can you use `require` rather than `replace`?

go/cli/internal/config/utils.go
dimetron: It's simple string in both cases 

    in main - homeDir, err := os.UserHomeDir()
    in test - homeDir := t.TempDir()
    
https://github.com/kagent-dev/kagent/pull/332/files/a54912bfbf3f60a823d9d46b85e48b5275636405#diff-6b0807eb7c55414cab318e87270967bd29d054f6f84ea10bd86ee637fc552b54R20

```go
func TestHandlesErrorWhenCreatingConfigDir(t *testing.T) {
	homeDir := t.TempDir()
	nonExistentDir := path.Join(homeDir, "/invalid/path")
	result, err := GetConfigDir(nonExistentDir)
```

go/go.mod
dimetron: I've checked it's already resolved into that version anyway, removed

peterj (approved):

PR #337, opened 2025-05-06
CI: image cve scans

dimetron:


.github/workflows/ci.yaml
```diff
@@ -27,10 +26,28 @@ jobs:
...
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Image vulnerability scanner
+        if: steps.image-check.outputs.exists == 'true'
+        uses: aquasecurity/trivy-action@0.30.0
+        with:
+         input: ${{ runner.temp }}/${{ matrix.image }}.tar
+         severity: 'CRITICAL,HIGH'
```
sbx0r: what do you think about ignoring unfixed vulns?

dimetron:
I am facing issue with Docker builder tar format on CI is not recognized by scanner

.github/workflows/ci.yaml
dimetron: Yes, we should ignore those

docker-bake.hcl
```diff
@@ -0,0 +1,48 @@
+group "default" {
```
peterj: what is this file used for?

.github/workflows/ci.yaml
```diff
@@ -27,10 +26,30 @@ jobs:
         uses: docker/setup-buildx-action@v3
       - name: Run make build
         env:
-          DOCKER_BUILD_ARGS: "--platform linux/amd64,linux/arm64 --output type=tar,dest=${{ runner.temp }}/${{ matrix.image }}.tar"
           DOCKER_BUILDER: "docker buildx"
+          DOCKER_BUILD_ARGS: >-
+            --platform linux/amd64,linux/arm64
+            --cache-from=type=gha
+            --cache-to=type=gha,mode=max
+            --output type=tar,dest=${{ runner.temp }}/${{ matrix.image }}.tar
```
sbx0r: > I am facing issue with Docker builder tar format on CI is not recognized by scanner

the issue is with the `type=tar`.
From the [docs](https://docs.docker.com/build/exporters/local-tar/) you can read that this exports ONLY the rootfs.
Trivy expects OCI image (with manifest.json) so you have to change the `type` to `oci` or just `docker`.

sbx0r: the outcome from the issue below is that only the `docker` type of tarballs are supported currently
https://github.com/aquasecurity/trivy/issues/5775?plain=1

dimetron: right, and using tar make sense if we want to test 2 platforms, 
however need to find a better way to make it possible to scan

docker-bake.hcl
dimetron: [docker buildx bake](https://docs.docker.com/build/bake/ ) is a feature of Docker Buildx that lets you define your build configuration using a declarative file

dimetron: same way istio is building images, but they generating it from cli

dimetron: I will remove it - it was experiment

dimetron:
report example 
https://github.com/dimetron/kagent/actions/runs/14877471806/job/41777547716

dimetron:
+1 we need to run this on every commit. Perhaps we do it on tag or on merge to main

dimetron:
added if: github.event_name == 'pull_request'

reneleonhardt:
Is this to be expected?
https://github.com/dimetron/kagent/actions/runs/14877471806/job/41777547716#step:13:8
Cache hit occurred on the primary key cache-trivy-2025-05-07, not saving cache.
...
Failed to save: Unable to reserve cache with key trivy-binary-v0.60.0-Linux-X64, another job may be creating this cache.

.github/workflows/ci.yaml
```diff
@@ -127,4 +151,4 @@ jobs:
 
       - name: Run tests
         working-directory: ./ui
-        run: npm run test
+        run: npm run test
```
reneleonhardt: Whitespace error introduced (text files have to end with an empty line):
```suggestion
        run: npm run test

```

dimetron:
> Is this to be expected? 
There are a lot of improvements waiting to be implemented around cache layers and images :) 

1. we will move CVE scan as a separate job
2. docker images of Python + Node need refactor with 3 layers [base + build + app ]

I plan to look into this more later today and tomorrow

ui/Dockerfile
```diff
@@ -1,33 +1,66 @@
+ARG DOCKER_REGISTRY=ghcr.io
 ARG TOOLS_UV_VERSION=latest
-ARG TOOLS_NODE_VERSION=20
-FROM --platform=$BUILDPLATFORM node:${TOOLS_NODE_VERSION}-bookworm AS ui-builder
+ARG PROXY
+ARG NOPROXY
+
+### STAGE 1: base image
+FROM --platform=$BUILDPLATFORM $DOCKER_REGISTRY/astral-sh/uv:${TOOLS_UV_VERSION}-bookworm-slim AS base-os
```
peterj: what's the reason we're using the astral-sh/uv image for the UI instead of the node image?

dimetron: both ui and python using node and uv, 
by using the same stages we reduce time and increasing cached layers

peterj (dismissed):

ui/Dockerfile
EItanya: the UI should not be using python, this is a leftover from when the UI container also ran the backend

EItanya:
Hey @dimetron can we please separate the Dockerfile optimizations into a separate PR. I totally understand the desire to get things done as quick as possible, but I think it's important to keep the scope of PRs smaller so that we can evaluate them at their own merits rather than as a complex combination of things.

Issue #330, opened 2025-05-05
[BUG] Remove the old notebooks in the /python code


peterj:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

We should remove the notebooks in https://github.com/kagent-dev/kagent/tree/main/python/notebooks 

### 🔄 Steps To Reproduce

.

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

peterj:
Closing.

PR #343, opened 2025-05-07
Removing notebooks as per #330

Conor0Callaghan:
Removing the notebooks in the python folder as per #330


EItanya (approved):

EItanya:
Thanks so much for the contribution!

Issue #303, opened 2025-05-01
[BUG] Ability to cancel/stop the endless loop agent can get in


peterj:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [ ] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Multiple services / System-wide issue

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

An agent can get into an infinite loop as long as it thinks there are more tools to call (ie. it will keep calling the tools). We need a mechanism to stop an agent like that. 

### 🔄 Steps To Reproduce

/

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

PR #347, opened 2025-05-08
move to latest autogen fork

peterj:


peterj:
Partiall fixes #303 (we still want a way to completely stop/cancel any agent run)

PR #355, opened 2025-05-08
Add model info to model config API to support arbitrary models

EItanya:


go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -167,6 +167,12 @@ type ModelConfigSpec struct {
 	// +optional
 	APIKeySecretKey string `json:"apiKeySecretKey"`
 
+	// ModelInfo contains information about the model.
+	// This field is required if the model if the model is not one of the
```
peterj: typo "if the model"

helm/kagent/templates/deployment.yaml
```diff
@@ -60,14 +60,6 @@ spec:
           image: "{{ .Values.app.image.registry }}/{{ .Values.app.image.repository }}:{{ coalesce .Values.global.tag .Values.app.image.tag .Chart.Version }}"
           imagePullPolicy: {{ .Values.app.image.pullPolicy }}
           env:
-            {{- $model := index .Values.providers .Values.providers.default  }}
-            {{- if and (ne $model.apiKeySecretRef "") $model.enabled }}
-            - name: {{ $model.apiKeySecretKey }}
```
peterj, line 65: don't we still need to set the key here?

go/config/crd/bases/kagent.dev_modelconfigs.yaml
```diff
@@ -104,6 +104,25 @@ spec:
                 type: object
               model:
                 type: string
+              modelInfo:
+                description: |-
+                  ModelInfo contains information about the model.
+                  This field is required if the model if the model is not one of the
```
peterj: you need to re-create this file as well :)

EItanya: Copilot already caught these, check above 😆

helm/kagent/templates/deployment.yaml
EItanya: I don't think we do anymore? Right now it's NEVER being added anyway, I think we should wait for Scott's PR which changes how the docs tool adds the key.

go/config/crd/bases/kagent.dev_modelconfigs.yaml
peterj: yeah, I saw that once I commented :)

helm/kagent/templates/deployment.yaml
peterj: yes! I'd wait for his PR to go in first, that way we're 100% that the doc tool will continue working

go/config/crd/bases/kagent.dev_modelconfigs.yaml
peterj: this is still not fixed though - think you need to make generate again

peterj:
we'll also need docs for this https://github.com/kagent-dev/website/issues/41

peterj (dismissed):

peterj (approved):

nimishamehta5:
thanks for adding this, this is what I was missing trying to run a non-openAI Ollama model

Issue #350, opened 2025-05-08
[BUG] when open a link from agent reply, open a new window/tab


linsun:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

When clicking on the link from the agent reply, best to not use the existing chat window as it is hard to get back to the chat...

### 🔄 Steps To Reproduce

Interact with the mcp agent, if there is a link in the reply, click on the link

### 🤔 Expected Behavior

open the link in a new tab/window.

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

peterj:
This was fixed. closing.

PR #357, opened 2025-05-09
show the memory retrieval indicator in the ui

### LINKED ISSUE #315: [FEATURE] Show an indicator in the UI when memory is used
### 📋 Prerequisites

### 📝 Feature Summary

Show an indicator in the U when memory is used during a chat

Currently, there's no way of knowning whether the memory attached to the agent is used or not. We should that the memory is attached, however, we should also show the # of hits, similarities etc - just like we do with function calling.

### 💡 Proposed Solution

We'd probably need to emit a new event from the backend and then render that in the UI.

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

peterj:
Closes #315 

Screenshot 2025-05-08 at 5 10 23 PM


peterj:
Also added a fix for #350

EItanya (approved):

PR #365, opened 2025-05-11
GHA: Periodic image scan

dimetron:
Periodic image scan for CVE


.github/workflows/image-scan.yaml
```diff
@@ -0,0 +1,66 @@
...
+        if: steps.image-check.outputs.exists == 'true'
+        id: image-versions
+        run: make build-img-versions
+      - name: Image vulnerability scanner
+        if: steps.image-check.outputs.exists == 'true'
+        uses: aquasecurity/trivy-action@0.30.0
+        with:
+          image-ref: ghcr.io/kagent-dev/kagent/${{ matrix.image }}:${{ steps.vars.outputs.version }}
+          severity: 'CRITICAL,HIGH'
+          ignore-unfixed: true
```
EItanya: You can just remove `cli` as a matrix target on line `21` now that this is a separate job.

.github/workflows/image-scan.yaml
```diff
@@ -0,0 +1,66 @@
...
+        run: | 
+          echo "Version: ${{ steps.vars.outputs.version }}"
+      - name: Run make build
+        env:
+          DOCKER_BUILDER: "docker buildx"
+          DOCKER_BUILD_ARGS: >-
+            --load
+            --cache-from=type=gha
+            --cache-to=type=gha,mode=max
+            --build-arg=VERSION=${{ steps.vars.outputs.version }}
```
EItanya, line 44: I don't think this will build `multi-arch` anymore from what I can tell. Do we want to scan both architectures since they may have slightly different libraries?

.github/workflows/image-scan.yaml
dimetron: good idea :)

.github/workflows/image-scan.yaml
dimetron: you are correct --load will do it only on current arch 
and if we do not want to push - we will scan only one
there is a limitation of the scanner

.github/workflows/image-scan.yaml
```diff
@@ -0,0 +1,55 @@
...
+        run: make build-${{ matrix.image }}
+        working-directory: ./
+      - name: Load image versions
+        id: image-versions
+        run: make build-img-versions
```
EItanya, line 49: I don't see this target anywhere?

.github/workflows/image-scan.yaml
```diff
@@ -0,0 +1,55 @@
+name: Scan images
+
+on:
+  # Trigger once a week on the main branch
+  schedule:
+  - cron: '0 0 * * 0' # Every Monday at midnight UTC
+
+  # Trigger the workflow on push to the main branch
+  pull_request:
+    branches: [main]
```
EItanya: This will run the scan on every commit, we only want on merge to `main`

```
  push:
    branches: [ main ]
```

EItanya (approved):

PR #366, opened 2025-05-11
Build: docker images cleanup

dimetron:
docker images cleanup for faster builds and allow different from registry


ui/Dockerfile
```diff
@@ -1,66 +1,73 @@
...
 # Ensure correct permissions
-RUN chown -R nextjs:nodejs /app/ui && \
+RUN chown -R nextjs:nginx /app/ui && \
     chmod -R 755 /app
 
 EXPOSE 80
 
 LABEL org.opencontainers.image.source=https://github.com/kagent-dev/kagent
 LABEL org.opencontainers.image.description="Kagent app is the UI and apiserver for running agents."
+LABEL org.opencontainers.image.authors="Kagent Creators 🤖"
```
EItanya, line 71: 🤣

python/Dockerfile
```diff
@@ -1,101 +1,100 @@
...
-RUN curl -o /usr/local/bin/kubectl-argo-rollouts -L https://github.com/argoproj/argo-rollouts/releases/$TOOLS_ARGO_CD_VERSION/download/kubectl-argo-rollouts-linux-$TARGETARCH \
+RUN curl -Lo /usr/local/bin/kubectl-argo-rollouts -L https://github.com/argoproj/argo-rollouts/releases/$TOOLS_ARGO_CD_VERSION/download/kubectl-argo-rollouts-linux-$TARGETARCH \
     && chmod +x /usr/local/bin/kubectl-argo-rollouts
 
-# Install Istio
 RUN curl -L https://istio.io/downloadIstio | ISTIO_VERSION=$TOOLS_ISTIO_VERSION TARGET_ARCH=$TARGETARCH sh - \
-    && mv istio-$TOOLS_ISTIO_VERSION/bin/istioctl /usr/local/bin/istioctl \
-    && rm -rf istio-$TOOLS_ISTIO_VERSION
+    && mv istio-*/bin/istioctl /usr/local/bin/istioctl \
+    && rm -rf istio-*
```
EItanya, line 35: What was the purpose of this change? It should match now.

Makefile
```diff
@@ -1,13 +1,20 @@
 # Image configuration
 DOCKER_REGISTRY ?= ghcr.io
+IMAGE_FROM_REGISTRY ?= docker.io
```
EItanya: `BASE_IMAGE_REGISTRY`?

Makefile
```diff
@@ -21,6 +28,9 @@ DOCKER_BUILDER ?= docker
 DOCKER_BUILD_ARGS ?=
 KIND_CLUSTER_NAME ?= kagent
 
+PROXY ?= ""
+NOPROXY ?= ""
```
EItanya: What are these for?

Makefile
```diff
@@ -29,15 +39,16 @@ TOOLS_GO_VERSION ?= $(shell $(AWK) '/^go / { print $$2 }' go/go.mod)
...
+TOOLS_ISTIO_VERSION ?= 1.26.0
+TOOLS_ARGO_CD_VERSION ?= 3.0.0
 TOOLS_KUBECTL_VERSION ?= 1.33.4
 
 # build args
-GO_IMAGE_BUILD_ARGS = --build-arg TOOLS_GO_VERSION=$(TOOLS_GO_VERSION)
-
-TOOLS_IMAGE_BUILD_ARGS = $(GO_IMAGE_BUILD_ARGS)
+TOOLS_IMAGE_BUILD_ARGS =  --build-arg PROXY=$(PROXY)
+TOOLS_IMAGE_BUILD_ARGS =  --build-arg NOPROXY=$(NOPROXY)
```
EItanya: Same question

python/Dockerfile
```diff
@@ -1,101 +1,100 @@
-ARG TOOLS_UV_VERSION=latest
-FROM ghcr.io/astral-sh/uv:${TOOLS_UV_VERSION}-debian-slim AS builder
+### STAGE 1: base image
+ARG IMAGE_FROM_REGISTRY=docker.io
+FROM $IMAGE_FROM_REGISTRY/chainguard/wolfi-base:latest AS base-os
```
EItanya: Point to a version/sha, we don't want to be broken by an upgrade

python/Dockerfile
dimetron: we had a bug when ISTIO_VERSION env var did not work because of TOOLS_ ISTIO_VERSION
there was mismatch 1.25.2 and 1.26.0

Makefile
dimetron: leftovers - will remove

python/Dockerfile
dimetron: 🤨 I don't think we could use anything else except chainguard/wolfi-base:latest

EItanya:
Given the scope of the changes  I will do a more thorough testing tomorrow to ensure nothing has silently broken.

PR #369, opened 2025-05-12
Add shortNames to CRDs, remove Makefile refs to kustomize

nimishamehta5:
Small quality of life improvement to interact with common Kagent CRs.

Removed refs to Kustomize from the Makefile, as that flow is no longer used.

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -105,12 +105,12 @@ ifndef ignore-not-found
 endif
 
 .PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | $(KUBECTL) apply -f -
+install: manifests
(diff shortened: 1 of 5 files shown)


EItanya:
> Small quality of life improvement to interact with common Kagent CRs.
> 
> Added `kustomize.yaml` file, as `make install` from `go/` otherwise errors with
> 
> ```
> Error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/Users/nim/src/github.com/nimishamehta5/kagent/go/config/crd'
> ```

Agreed on short names. I don't think we need the `kustomization`. `make install` in `go/` is actually leftover from some bootstrapping done at the very beginning of the repo. Rather I think we should remove unused make targets.

nimishamehta5:
@EItanya - updated the make install / uninstall targets, let me know if that makes sense

EItanya (approved):

PR #376, opened 2025-05-13
Template namespace for resources produced by `kagent` Helm chart

onematchfox:
Ensures that rendered manifests have `namespace` set as per the value of the `namespace` flag used when running `helm template` (or optionally overriden via `values.yaml`). Enables deployment of manifests with commands like:

```
helm template --release-name kagent --namespace kagent . | kubectl apply -f -
```

FWIW, this is a fairly common pattern across a number of open source charts. Several examples can be seen within the [bitnami charts](https://github.com/search?q=repo%3Abitnami%2Fcharts%20namespaceOverride&type=code).


EItanya (approved):
Thanks so much for the contribution, this makes total sense to me :)

PR #389, opened 2025-05-15
add an option to not set the api key

peterj:
Fix on the UI side:
- I added a check box, so when check the api key isn't required (no errors/warnings)

Controller:
- when trying to retrieve the API key from the modelconfig we don't check for the Secret if the APIsecret ref is empty AND the BaseURL is set; otherwise we retrieve the secret (and fail if both apisecretref and baseurl are emptY).


sbx0r (approved):
Looks like a preparation for kgateway 😃 🚀

go/controller/internal/httpserver/handlers/modelconfig.go
```diff
@@ -164,6 +164,7 @@ type CreateModelConfigRequest struct {
 	AnthropicParams *v1alpha1.AnthropicConfig   `json:"anthropic,omitempty"`
 	AzureParams     *v1alpha1.AzureOpenAIConfig `json:"azureOpenAI,omitempty"`
 	OllamaParams    *v1alpha1.OllamaConfig      `json:"ollama,omitempty"`
+	IsUsingGateway  bool                        `json:"isUsingGateway"`
```
EItanya: Can this be implicit because the hostname isn't `openai.com`?

EItanya (commented):
I like these changes overall, I don't love a specific field in the REST API called `IsGateway`. Can we be more generic, or infer from the hostname?

go/controller/internal/httpserver/handlers/modelconfig.go
christian-posta: What about something like "DeferSecurity" or "AlternativeAuth"

peterj: The UI logic stays the same; the backend will check whether the API key was sent -- if it's not, we'll assume the user is using the custom host, otherwise, it will create a Secret.

ui/src/app/models/new/page.tsx
```diff
@@ -198,6 +199,12 @@ function ModelPageContent() {
             setSelectedCombinedModel(`${providerFormKey}::${modelName}`);
           }
 
+          if (!modelData.apiKeySecretRef) {
```
christian-posta, line 201: Wait so for this, it seems we are just deciding that if there is no key, then we must be using a gateway? That won't work tho. Because we may use a key with or without a gateway. I don't think we should just try to infer anything with this PR. We should be very explicit with an option that "ApiKeyRequired" which defaults to true, but can be overridden. Maybe for Ollama it can default to false.

peterj: I can change the UI option to say something like "I don't need an API key" or something like that, so it's not specific to the gateway.

christian-posta: sounds good

EItanya (approved):

PR #396, opened 2025-05-15
fix the typo in dockerfile

peterj:
Rebuilt this a couple of times locally without using cache and the build step took around ~100s


ui/Dockerfile
```diff
@@ -31,8 +31,9 @@ FROM node-os AS ui-builder
 
 WORKDIR /app/ui
 COPY package*.json ./
+COPY package-lock.json ./
```
EItanya: This should be handled by the line above :(

peterj:
```suggestion
```

EItanya:
Not relevant after #398

PR #398, opened 2025-05-15
Docker - improve slow CI multi arch build

dimetron:


go/Dockerfile
```diff
@@ -1,19 +1,21 @@
 ### STAGE 1: base image
+ARG LOCALARCH=amd64
```
EItanya: Why do we need this new env var, wasn't `$BUILDPLATFORM` already doing this?

ui/Dockerfile
```diff
@@ -1,6 +1,7 @@
 ### STAGE 1: base image
+ARG LOCALARCH=amd64
 ARG BASE_IMAGE_REGISTRY=cgr.dev
-FROM $BASE_IMAGE_REGISTRY/chainguard/wolfi-base:latest AS base-os
+FROM --platform=linux/${LOCALARCH} $BASE_IMAGE_REGISTRY/chainguard/wolfi-base:latest AS base-os
```
EItanya: use `$BUILDPLATFORM`?

python/Dockerfile
```diff
@@ -42,31 +61,12 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/s
...
 ### STAGE 4: final
 FROM python-os AS final
-ARG TOOLS_PYTHON_VERSION=3.12
 
 COPY --from=python-builder /usr/local/bin/kubectl /usr/local/bin/kubectl
 COPY --from=python-builder /usr/local/bin/istioctl /usr/local/bin/istioctl
 COPY --from=python-builder /usr/local/bin/helm /usr/local/bin/helm
 COPY --from=python-builder /usr/local/bin/kubectl-argo-rollouts /usr/local/bin/kubectl-argo-rollouts
 
-ENV PYTHONDONTWRITEBYTECODE=1
```
EItanya, line 52: How is this working since you're not copying over `.venv`

EItanya (commented):
These changes mostly make sense, I'm not sure why the move away from `$BUILDPLATFORM` and I don't understand the Python `Dockerfile`. How is Python available if it installs it in a different stage?

python/Dockerfile
dimetron: ENV UV_PROJECT_ENVIRONMENT="/app/python/.venv"

EItanya: Right, but you're not doing `COPY .venv .venv`

EItanya (approved):

peterj (approved):

PR #405, opened 2025-05-19
Multi namespace support

### LINKED ISSUE #339: [FEATURE] Multi namespace support
### 📋 Prerequisites

### 📝 Feature Summary

Add support for multi namespace resources

- Currently, if you create a CR in any namespace other than the release namespace, various parts of the system - UI, controller, and engine - struggle to handle it properly.
- I've looked into this and found that the issue is multi-layered. Supporting multi-namespace properly will likely require changes to both the API and data models.

### Team participants

```mermaid
classDiagram
    direction LR

    class Team_Old {
        +participants : []string
    }

    class Team_New {
        +participants : []ParticipantRef
    }

    class ParticipantRef {
        +name : string
        +namespace : string (omitempty)
    }

    Team_Old ..> Team_New : participants type changed
    Team_New "1" --> "*" ParticipantRef : contains

    style Team_New fill:#e6ffe6,stroke:#2ca02c,stroke-width:2px,color:#000
    style ParticipantRef fill:#e6ffe6,stroke:#2ca02c,stroke-width:2px,color:#000
```

### ToolAgent

```mermaid
classDiagram
    direction LR

    class Tool_Old {
        +agent : *AgentTool
    }

    class AgentTool {
        +ref : string
    }

    class Tool_New {
        +agent : *AgentToolRef
    }

    class AgentToolRef {
        +name : string
        +namespace : string (omitempty)
    }

    Tool_Old ..> Tool_New : agent type changed
    Tool_Old "0..1" --> "1" AgentTool     : refers
    Tool_New "0..1" --> "1" AgentToolRef : refers

    style Tool_New fill:#e6ffe6,stroke:#2ca02c,stroke-width:2px,color:#000
    style AgentToolRef fill:#e6ffe6,stroke:#2ca02c,stroke-width:2px,color:#000
```

## API Changes

```text
   ┌───────────────────────────────────────────┐
   │  OLD                                      │
   │  /api/<resources>/{id}                    │
   └──────────────┬────────────────────────────┘
                  │  change → use namespace/name
                  ▼
   ┌───────────────────────────────────────────┐
   │  NEW                                      │
   │  /api/<resources>/{namespace}/{name}      │
   └───────────────────────────────────────────┘


   ┌───────────────────────────────────────────────────────────┐
   │  OLD (action URI)                                         │
   │  /api/<resources>/{id}/<action>                           │
   └──────────────┬────────────────────────────────────────────┘
                  │  change → action after namespace/name
                  ▼
   ┌───────────────────────────────────────────────────────────┐
   │  NEW (action URI)                                         │
   │  /api/<resources>/{namespace}/{name}/<action>             │
   └───────────────────────────────────────────────────────────┘

```

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

### 📚 Additional Context

Let me understand why Team and Agent have ModelConfig? What is the reason or potential use-case of this?

sbx0r:
/resolves #339

**Summary:**
This pull request expands `kagent’s` namespace awareness. The controller can now watch and reconcile Custom Resources - `Agent`, `Memory`, `ModelConfig`, and `ToolServer` that live in any namespace granted by its RBAC rules (rather than just its own).

**Motivation:**
This change lays the groundwork for proper resource isolation and prepares `kagent` for future permission management in multi‑tenant clusters. Until now, `kagent` could interact only with Custom Resources located in its own namespace, forcing teams to keep every CR in a single space. Expanding the controller’s watch scope to all namespaces removes that limitation and provides a solid foundation for future tenant‑level access controls. Unifying how CRs are fetched across namespaces will also make it straightforward to introduce per‑tenant filtering and visibility limits later on.

**What’s changed:**
  - **Controller**
    - **CRD List/Watch scope** - switched from Namespaced to AllNamespaces informer for all CRDs.
    - **DTO updates** – added namespace field where missing and normalised existing ones.
    - **API Handlers** - added namespace to paths (where it was necessary)
    - **Reconciliation logic** – guarded against name collisions across namespaces.
    - **Unit tests** – new coverage for cross‑namespace scenarios; updated mocks.
  - **UI**
    - Adapted API calls to include namespace param.
    - Display component tweaks to show fully‑qualified resource names (<namespace>/<name>).
  - **Backward Compability**
    - Existing single‑namespace deployments continue to work without changes.
    - No CRD version bumps. Only DTO field additions (non‑breaking in JSON schema).

### DIFF
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -97,7 +97,7 @@ func (a *apiTranslator) getConfigMapValue(ctx context.Context, source *v1alpha1.
 	}
 
 	configMap := &corev1.ConfigMap{}
-	err := fetchObjKube(
+	err := common.GetObject(
 		ctx,
 		a.kube,
 		configMap,
@@ -122,7 +122,7 @@ func (a *apiTranslator) getSecretValue(ctx context.Context, source *v1alpha1.Val
 	}
 
 	secret := &corev1.Secret{}
-	err := fetchObjKube(
+	err := common.GetObject(
 		ctx,
 		a.kube,
 		secret,
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -445,7 +445,7 @@ func (a *apiTranslator) translateGroupChatForTeam(
 		return nil, fmt.Errorf("no team config specified")
 	}
 
-	teamConfig.Label = team.Namespace + "/" + team.Name
+	teamConfig.Label = common.GetObjectRef(team)
 
 	return &autogen_client.Team{
 		Component: teamConfig,
@@ -535,13 +535,12 @@ func (a *apiTranslator) translateAssistantAgent(
 				tools = append(tools, autogenTool)
 			}
 		case tool.Agent != nil:
-			toolRef := ""
-			if toolNamespacedName, err := common.ParseRefString(tool.Agent.Ref, agent.Namespace); err != nil {
+			toolNamespacedName, err := common.ParseRefString(tool.Agent.Ref, agent.Namespace)
+			if err != nil {
 				return nil, err
-			} else {
-				toolRef = toolNamespacedName.String()
 			}
 
+			toolRef := toolNamespacedName.String()
 			agentRef := common.GetObjectRef(agent)
 
 			if toolRef == agentRef {
@@ -559,7 +558,7 @@ func (a *apiTranslator) translateAssistantAgent(
 			// Translate a nested tool
 			toolAgent := &v1alpha1.Agent{}
 
-			err := common.GetObject(
+			err = common.GetObject(
 				ctx,
 				a.kube,
 				toolAgent,
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -549,19 +549,19 @@ func (a *apiTranslator) translateAssistantAgent(
 				tools = append(tools, autogenTool)
 			}
 		case tool.Agent != nil:
-			refFullName := common.GetRefFromString(tool.Agent.Ref, agent.Namespace).String()
+			toolRefFullName := common.GetRefFromString(tool.Agent.Ref, agent.Namespace).String()
 			agentFullName := fmt.Sprintf("%s/%s", agent.Namespace, agent.Name)
 
-			if refFullName == agentFullName {
+			if toolRefFullName == agentFullName {
 				return nil, fmt.Errorf("agent tool cannot be used to reference itself, %s", agentFullName)
 			}
 
-			if state.isVisited(tool.Agent.Ref) {
-				return nil, fmt.Errorf("cycle detected in agent tool chain: %s -> %s", agentFullName, refFullName)
+			if state.isVisited(toolRefFullName) {
+				return nil, fmt.Errorf("cycle detected in agent tool chain: %s -> %s", agentFullName, toolRefFullName)
 			}
 
 			if state.depth > MAX_DEPTH {
-				return nil, fmt.Errorf("recursion limit reached in agent tool chain: %s -> %s", agentFullName, refFullName)
+				return nil, fmt.Errorf("recursion limit reached in agent tool chain: %s -> %s", agentFullName, toolRefFullName)
 			}
 
 			// Translate a nested tool
@@ -571,7 +571,7 @@ func (a *apiTranslator) translateAssistantAgent(
 				ctx,
 				a.kube,
 				&toolAgent,
-				refFullName,
+				toolRefFullName,
 				agent.Namespace, // redundant
 			)
 			if err != nil {
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -313,33 +313,25 @@ func (a *apiTranslator) translateGroupChatForTeam(
 
 	modelConfigRef := a.defaultModelConfig
 	if team.Spec.ModelConfig != "" {
-		log.Printf("translateGroupChatForTeam: team %s/%s specifies ModelConfig '%s'",
-			team.Namespace, team.Name, team.Spec.ModelConfig)
 		// If ModelConfig contains only the name of the ModelConfig, use the namespace of the team.
 		modelConfigRef = common.GetRefFromString(team.Spec.ModelConfig, team.Namespace)
-		log.Printf("translateGroupChatForTeam: resolved ModelConfig ref: namespace='%s', name='%s'",
-			modelConfigRef.Namespace, modelConfigRef.Name)
 	} else {
 		log.Printf("translateGroupChatForTeam: team %s/%s using default ModelConfig: namespace='%s', name='%s'",
 			team.Namespace, team.Name, modelConfigRef.Namespace, modelConfigRef.Name)
 	}
 
 	modelConfigObj := &v1alpha1.ModelConfig{}
-	if err := common.FetchObjKube(
+	err := common.FetchObjKube(
 		ctx,
 		a.kube,
 		modelConfigObj,
 		modelConfigRef.Name,
 		modelConfigRef.Namespace,
-	); err != nil {
-		log.Printf("translateGroupChatForTeam: failed to fetch ModelConfig for team %s/%s: %v",
-			team.Namespace, team.Name, err)
+	)
+	if err != nil {
 		return nil, fmt.Errorf(
 			"failed to fetch ModelConfig %s: %w", modelConfigRef.String(), err)
 	}
-
-	log.Printf("translateGroupChatForTeam: successfully fetched ModelConfig %s/%s for team %s/%s",
-		modelConfigObj.Namespace, modelConfigObj.Name, team.Namespace, team.Name)
 
 	modelClientWithStreaming, err := a.createModelClientForProvider(ctx, modelConfigObj, true)
 	if err != nil {
@@ -480,7 +472,7 @@ func (a *apiTranslator) simpleRoundRobinTeam(ctx context.Context, agent *v1alpha
 		// If ModelConfig contains only the name of the ModelConfig, use the namespace of the agent.
 		modelConfig = common.GetRefFromString(agent.Spec.ModelConfig, agent.Namespace)
 	} else {
-		log.Printf("Agent %s/%s using default ModelConfig: namespace='%s', name='%s'",
+		log.Printf("ModelConfig for Agent %s/%s not defined. Using default ModelConfig: namespace='%s', name='%s'",
 			agent.Namespace, agent.Name, modelConfig.Namespace, modelConfig.Name)
 	}
 	if err := a.kube.Get(ctx, modelConfig, &v1alpha1.ModelConfig{}); err != nil {
@@ -575,13 +567,14 @@ func (a *apiTranslator) translateAssistantAgent(
 			// Translate a nested tool
 			toolAgent := v1alpha1.Agent{}
 
-			if err := common.FetchObjKube(
+			err := common.FetchObjKube(
 				ctx,
 				a.kube,
 				&toolAgent,
 				refFullName,
-				agent.Namespace,
-			); err != nil {
+				agent.Namespace, // redundant
+			)
+			if err != nil {
 				return nil, err
 			}
 
@@ -638,8 +631,8 @@ func (a *apiTranslator) translateAssistantAgent(
 	}
 
 	if agent.Spec.Memory != nil {
-		for _, memoryName := range agent.Spec.Memory {
-			autogenMemory, err := a.translateMemory(ctx, memoryName, agent.Namespace)
+		for _, memoryRef := range agent.Spec.Memory {
+			autogenMemory, err := a.translateMemory(ctx, memoryRef, agent.Namespace)
 			if err != nil {
 				return nil, err
 			}
@@ -657,9 +650,9 @@ func (a *apiTranslator) translateAssistantAgent(
 	}, nil
 }
 
-func (a *apiTranslator) translateMemory(ctx context.Context, memoryName string, memoryNamespace string) (*api.Component, error) {
+func (a *apiTranslator) translateMemory(ctx context.Context, memoryRef string, defaultNamespace string) (*api.Component, error) {
 	memoryObj := &v1alpha1.Memory{}
-	if err := common.FetchObjKube(ctx, a.kube, memoryObj, memoryName, memoryNamespace); err != nil {
+	if err := common.FetchObjKube(ctx, a.kube, memoryObj, memoryRef, defaultNamespace); err != nil {
 		return nil, err
 	}
 
@@ -739,18 +732,19 @@ func (a *apiTranslator) translateBuiltinTool(
 func translateToolServerTool(
 	ctx context.Context,
 	kube client.Client,
-	toolServerName string,
+	toolServerRef string,
 	toolName string,
-	agentNamespace string,
+	defaultNamespace string,
 ) (*api.Component, error) {
 	toolServerObj := &v1alpha1.ToolServer{}
-	if err := common.FetchObjKube(
+	err := common.FetchObjKube(
 		ctx,
 		kube,
 		toolServerObj,
-		toolServerName,
-		agentNamespace,
-	); err != nil {
+		toolServerRef,
+		defaultNamespace,
+	)
+	if err != nil {
 		return nil, err
 	}
 
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -14,7 +14,6 @@ import (
 	"github.com/kagent-dev/kagent/go/controller/api/v1alpha1"
 	common "github.com/kagent-dev/kagent/go/controller/internal/utils"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -1303,8 +1303,8 @@ func (a *apiTranslator) getMemoryApiKey(ctx context.Context, memory *v1alpha1.Me
 }
 
 func (a *apiTranslator) getModelConfigGoogleApplicationCredentials(ctx context.Context, modelConfig *v1alpha1.ModelConfig) (map[string]interface{}, error) {
-	googleApplicationCredentialsSecret := &v1.Secret{}
-	err := fetchObjKube(
+	googleApplicationCredentialsSecret := &corev1.Secret{}
+	err := common.GetObject(
 		ctx,
 		a.kube,
 		googleApplicationCredentialsSecret,
--- a/go/controller/internal/autogen/autogen_api_translator.go
+++ b/go/controller/internal/autogen/autogen_api_translator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"slices"
 	"strconv"
 	"strings"
@@ -65,7 +66,7 @@ func (a *apiTranslator) TranslateToolServer(ctx context.Context, toolServer *v1a
 			ComponentType: "tool_server",
 			Version:       1,
 			Description:   toolServer.Spec.Description,
-			Label:         toolServer.Name,
+			Label:         toolServer.Namespace + "/" + toolServer.Name,
 			Config:        api.MustToConfig(toolServerConfig),
 		},
 	}, nil
@@ -291,7 +292,7 @@ func (a *apiTranslator) translateGroupChatForAgent(
 	state *tState,
 ) (*autogen_client.Team, error) {
 
-	simpleTeam, err := a.simpleRoundRobinTeam(ctx, agent, agent.Name)
+	simpleTeam, err := a.simpleRoundRobinTeam(ctx, agent, agent.Name, agent.Namespace)
 	if err != nil {
 		return nil, err
 	}
@@ -312,29 +313,40 @@ func (a *apiTranslator) translateGroupChatForTeam(
 
 	modelConfigRef := a.defaultModelConfig
 	if team.Spec.ModelConfig != "" {
-		modelConfigRef = types.NamespacedName{
-			Name:      team.Spec.ModelConfig,
-			Namespace: team.Namespace,
-		}
+		log.Printf("translateGroupChatForTeam: team %s/%s specifies ModelConfig '%s'",
+			team.Namespace, team.Name, team.Spec.ModelConfig)
+		// If ModelConfig contains only the name of the ModelConfig, use the namespace of the team.
+		modelConfigRef = common.GetRefFromString(team.Spec.ModelConfig, team.Namespace)
+		log.Printf("translateGroupChatForTeam: resolved ModelConfig ref: namespace='%s', name='%s'",
+			modelConfigRef.Namespace, modelConfigRef.Name)
+	} else {
+		log.Printf("translateGroupChatForTeam: team %s/%s using default ModelConfig: namespace='%s', name='%s'",
+			team.Namespace, team.Name, modelConfigRef.Namespace, modelConfigRef.Name)
 	}
-	modelConfig := &v1alpha1.ModelConfig{}
-	err := fetchObjKube(
+
+	modelConfigObj := &v1alpha1.ModelConfig{}
+	if err := common.FetchObjKube(
 		ctx,
 		a.kube,
-		modelConfig,
+		modelConfigObj,
 		modelConfigRef.Name,
 		modelConfigRef.Namespace,
-	)
+	); err != nil {
+		log.Printf("translateGroupChatForTeam: failed to fetch ModelConfig for team %s/%s: %v",
+			team.Namespace, team.Name, err)
+		return nil, fmt.Errorf(
+			"failed to fetch ModelConfig %s: %w", modelConfigRef.String(), err)
+	}
+
+	log.Printf("translateGroupChatForTeam: successfully fetched ModelConfig %s/%s for team %s/%s",
+		modelConfigObj.Namespace, modelConfigObj.Name, team.Namespace, team.Name)
+
+	modelClientWithStreaming, err := a.createModelClientForProvider(ctx, modelConfigObj, true)
 	if err != nil {
 		return nil, err
 	}
 
-	modelClientWithStreaming, err := a.createModelClientForProvider(ctx, modelConfig, true)
-	if err != nil {
-		return nil, err
-	}
-
-	modelClientWithoutStreaming, err := a.createModelClientForProvider(ctx, modelConfig, false)
+	modelClientWithoutStreaming, err := a.createModelClientForProvider(ctx, modelConfigObj, false)
 	if err != nil {
 		return nil, err
 	}
@@ -350,23 +362,22 @@ func (a *apiTranslator) translateGroupChatForTeam(
 
 	var participants []*api.Component
 
-	for _, agentName := range team.Spec.Participants {
-		agent := &v1alpha1.Agent{}
-		err := fetchObjKube(
+	for _, agentRef := range team.Spec.Participants {
+		agentObj := &v1alpha1.Agent{}
+		if err := common.FetchObjKube(
 			ctx,
 			a.kube,
-			agent,
-			agentName,
+			agentObj,
+			agentRef,
 			team.Namespace,
-		)
-		if err != nil {
+		); err != nil {
 			return nil, err
 		}
 
 		participant, err := a.translateAssistantAgent(
 			ctx,
-			agent,
-			modelConfig,
+			agentObj,
+			modelConfigObj,
 			modelClientWithStreaming,
 			modelClientWithoutStreaming,
 			modelContext,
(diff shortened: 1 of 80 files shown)


sbx0r:
/resolves #339

Apologies for the oversized PR - it grew beyond the original scope.

Next steps
- Resolve merge conflicts
- Clean up the codebase
- Repair the CLI

I’ll document the rationale afterward to make the changes clear.

go/controller/internal/httpserver/server.go
sbx0r, line 1: Although the original goal was to keep the API and CRD unchanged, we had to break backward compatibility to properly support cases like:

```yaml
- name: model-a
  namespace: ns-1
- name: model-a
  namespace: ns-2
```

Previously, the API used the path:

`/<type>/{name}`

This made it impossible to distinguish between resources with the same name in different namespaces. To address this, I've updated relevant API paths to include the namespace:

`/<type>/{namespace}/{name}`

This change ensures unique identification of resources across namespaces and better supports future project needs.

go/controller/internal/utils/common.go
```diff
@@ -1,6 +1,13 @@
 package common
```
sbx0r, line 1: moved 
- `getRefFromString`
- `fetchObjKube`
- `convertToPythonIdentifier`
- `convertToKubernetesIdentifier`

functions to `commons`

go/controller/internal/utils/common.go
```diff
@@ -13,3 +20,44 @@ func GetResourceNamespace() string {
...
+
+	err := kube.Get(ctx, ref, obj)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func ConvertToPythonIdentifier(name string) string {
```
sbx0r, line 55: convertToPythonIdentifier was Moved to common.

go/controller/internal/utils/common.go
```diff
@@ -13,3 +20,44 @@ func GetResourceNamespace() string {
...
+	name = strings.ReplaceAll(name, "-", "_")
+	return strings.ReplaceAll(name, "/", "__NS__") // RFC 1123 will guarantee there will be no conflicts
```
sbx0r, line 220: These changes allow for lossless conversion between Kubernetes and Python identifiers, preserving the `namespace/name` structure.

go/controller/internal/utils/common.go
```diff
@@ -13,3 +20,44 @@ func GetResourceNamespace() string {
...
+}
+
+func ConvertToKubernetesIdentifier(name string) string {
```
sbx0r, line 60: convertToKubernetesIdentifier was Moved to common.

go/controller/internal/utils/common.go
```diff
@@ -13,3 +20,44 @@ func GetResourceNamespace() string {
...
+	name = strings.ReplaceAll(name, "__NS__", "/")
+	return strings.ReplaceAll(name, "_", "-")
```
sbx0r, line 227: were updated to ensure reversibility for resources following the `namespace/name` convention

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -795,35 +812,12 @@ func translateTerminationCondition(terminationCondition v1alpha1.TerminationCond
...
-}
-
 func toolNeedsModelClient(provider string) bool {
-	for _, p := range toolsProvidersRequiringModelClient {
-		if p == provider {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(toolsProvidersRequiringModelClient, provider)
```
sbx0r, line 893: simplified `toolNeedsModelClient`

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -795,35 +812,12 @@ func translateTerminationCondition(terminationCondition v1alpha1.TerminationCond
...
 }
 
 func toolNeedsOpenaiApiKey(provider string) bool {
-	for _, p := range toolsProvidersRequiringOpenaiApiKey {
-		if p == provider {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(toolsProvidersRequiringOpenaiApiKey, provider)
```
sbx0r, line 897: simplified `toolNeedsOpenaiApiKey`

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -1077,76 +1071,34 @@ func translateModelInfo(modelInfo *v1alpha1.ModelInfo) *api.ModelInfo {
 	}
 }
 
-func (a *apiTranslator) getMemoryApiKey(ctx context.Context, memory *v1alpha1.Memory) ([]byte, error) {
-	memoryApiKeySecret := &v1.Secret{}
-	err := fetchObjKube(
+func (a *apiTranslator) getSecretKey(ctx context.Context, secretRef string, secretKey string, namespace string) ([]byte, error) {
```
sbx0r, line 1276: added function which unifies the way of reading secrets

go/controller/internal/autogen/autogen_reconciler.go
```diff
@@ -388,6 +389,10 @@ func (a *autogenReconciler) reconcileToolServerStatus(
 func (a *autogenReconciler) ReconcileAutogenMemory(ctx context.Context, req ctrl.Request) error {
 	memory := &v1alpha1.Memory{}
 	if err := a.kube.Get(ctx, req.NamespacedName, memory); err != nil {
+		if errors.IsNotFound(err) {
```
sbx0r, line 396: prepared section with handles Memory deletion (due to exceptions)

ui/src/app/actions/teams.ts
```diff
@@ -20,17 +20,39 @@ function convertToolRepresentation(tool: unknown, allAgents: AgentResponse[]): T
   } else if (isMcpTool(typedTool)) {
     return tool as Tool;
   } else if (isAgentTool(typedTool)) {
-    const agentName = typedTool.agent.ref;
-    const foundAgent = allAgents.find(a => a.agent.metadata.name === agentName);
+    const agentRef = typedTool.agent.ref;
+    const foundAgent = allAgents.find((a) => {
+      // TODO: Resolve namespace ambiguity for agent references
```
sbx0r: I prefer to discuss this before actual implementation @EItanya

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -460,33 +471,36 @@ func (a *apiTranslator) translateAssistantAgent(
 				tools = append(tools, autogenTool)
 			}
 		case tool.Agent != nil:
-			if tool.Agent.Ref == agent.Name {
-				return nil, fmt.Errorf("agent tool cannot be used to reference itself, %s", agent.Name)
+			refFullName := common.GetRefFromString(tool.Agent.Ref, agent.Namespace).String()
+			agentFullName := fmt.Sprintf("%s/%s", agent.Namespace, agent.Name)
+
+			if refFullName == agentFullName {
```
sbx0r, line 477: Fixing the issue described below:

when
- referring agent is in namespace "ns-1"
- tool.Agent.Ref == "ns-1/agent-1", and agent.Name == "agent-1"

then
- cycle will be not detected

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -460,33 +471,36 @@ func (a *apiTranslator) translateAssistantAgent(
...
+				return nil, fmt.Errorf("agent tool cannot be used to reference itself, %s", agentFullName)
 			}
 
 			if state.isVisited(tool.Agent.Ref) {
-				return nil, fmt.Errorf("cycle detected in agent tool chain: %s -> %s", agent.Name, tool.Agent.Ref)
```
sbx0r, line 468: Fixing the issue described below:

- when
  - agent.Name == "agent-1"
  - agent.Namespace == "ns-1"
  - tool.Agent.Ref == "ns-1/agent-1"
- then
  - agent.Name != tool.Agent.Ref
  - cycle will be not detected

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -460,33 +471,36 @@ func (a *apiTranslator) translateAssistantAgent(
 				tools = append(tools, autogenTool)
 			}
 		case tool.Agent != nil:
-			if tool.Agent.Ref == agent.Name {
```
sbx0r, line 542: Fixing the issue described below:

- when
  - agent.Name == "agent-1"
  - agent.Namespace == "ns-1"
  - tool.Agent.Ref == "ns-1/agent-1"
- then
  - agent.Name != tool.Agent.Ref
  - agent reference to itself is not detected

go/controller/internal/autogen/autogen_reconciler.go
```diff
@@ -95,7 +96,7 @@ func (a *autogenReconciler) handleAgentDeletion(req ctrl.Request) error {
 
 	// TODO(sbx0r): temporary mock on GlobalUserID.
```
peterj, line 101: do you need to remove this comment? or is this still a todo here?

go/controller/internal/httpserver/handlers/memory.go
```diff
@@ -112,98 +127,127 @@ func (h *MemoryHandler) HandleCreateMemory(w ErrorResponseWriter, r *http.Reques
...
-	log.Info("Received request to delete memory")
+	log = log.WithValues(
+		"namespace", namespace,
+		"memoryName", memoryName,
+	)
 
-	log.V(1).Info("Checking if memory exists")
+	log.V(1).Info("Checking if Memory exists")
 	existingMemory := &v1alpha1.Memory{}
 	err = h.KubeClient.Get(r.Context(), types.NamespacedName{
```
peterj: I am assuming this will fail if the namespace and/or resource don't exist -- in other words, we don't have to check separately that the namespace exits, right?

ui/src/app/actions/memories.ts
```diff
@@ -1,41 +1,39 @@
...
-
-export async function deleteMemory(name: string): Promise<void> {
-  await fetchApi<void>(`/memories/${name}`, {
+export async function deleteMemory(namespace: string, name: string): Promise<void> {
+  await fetchApi<void>(`/memories/${namespace}/${name}`, {
     method: 'DELETE',
-  })
-} 
\ No newline at end of file
+  });
```
peterj: since there's a lot of reformatting -- can you also include your formatter settings (or were these fils just never formatted?)

sbx0r: Actually, I’d prefer to skip formatting for now so we don’t obscure the core changes in this PR. I would like to revert the Prettier edits - looks like I missed some 😄 - and then add a `.prettierrc` in a follow-up PR. Does that work for you?

go/controller/internal/httpserver/handlers/memory.go
sbx0r: Great question.
In this case, we don't differentiate whether the resource is missing due to a nonexistent namespace or because the memory itself doesn't exist. We assume the API caller knows exactly what they intend to delete. If the caller provides a nonexistent namespace, the function will simply return `{"error":"Memory not found"}` and nothing bad happens

go/controller/internal/autogen/autogen_reconciler.go
sbx0r: right! I can remove it

go/controller/internal/httpserver/handlers/memory.go
peterj: makes ssense.

EItanya:
Hey @sbx0r, I plan to do a deep review on this first thing tomorrow.

go/controller/internal/httpserver/server.go
EItanya: That makes sense, breaking in the REST API is ok, we don't have any promises there yet.

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -233,29 +235,32 @@ func (a *apiTranslator) translateGroupChatForTeam(
...
 	modelConfigRef := a.defaultModelConfig
 	if team.Spec.ModelConfig != "" {
-		modelConfigRef = types.NamespacedName{
-			Name:      team.Spec.ModelConfig,
-			Namespace: team.Namespace,
-		}
+		// If ModelConfig contains only the name of the ModelConfig, use the namespace of the team.
+		modelConfigRef = common.GetRefFromString(team.Spec.ModelConfig, team.Namespace)
+	} else {
+		log.Printf("translateGroupChatForTeam: team %s/%s using default ModelConfig: namespace='%s', name='%s'",
```
EItanya: Can we put this on a debug level?

go/controller/internal/httpserver/server.go
sbx0r: Happy to hear

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -382,37 +386,44 @@ func (a *apiTranslator) translateGroupChatForTeam(
 	}, nil
 }
 
-func (a *apiTranslator) simpleRoundRobinTeam(ctx context.Context, agent *v1alpha1.Agent, name string) (*v1alpha1.Team, error) {
+func (a *apiTranslator) simpleRoundRobinTeam(ctx context.Context, agent *v1alpha1.Agent, name, namespace string) (*v1alpha1.Team, error) {
```
EItanya: Every usage of this function just uses the `agent.Name` and `agent.Namespace`, so I think we can get rid of those arguments

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -372,7 +376,7 @@ func (a *apiTranslator) translateGroupChatForTeam(
 		return nil, fmt.Errorf("no team config specified")
 	}
 
-	teamConfig.Label = team.Name
+	teamConfig.Label = team.Namespace + "/" + team.Name
```
EItanya: Can we create a function which takes in a `client.object` and returns the concatenated name since we're doing this same thing in a lot of places?

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -382,37 +386,44 @@ func (a *apiTranslator) translateGroupChatForTeam(
...
-		modelConfig = types.NamespacedName{
-			Name:      agent.Spec.ModelConfig,
-			Namespace: agent.Namespace,
-		}
+		// If ModelConfig contains only the name of the ModelConfig, use the namespace of the agent.
+		modelConfig = common.GetRefFromString(agent.Spec.ModelConfig, agent.Namespace)
+	} else {
+		log.Printf("ModelConfig for Agent %s/%s not defined. Using default ModelConfig: namespace='%s', name='%s'",
+			agent.Namespace, agent.Name, modelConfig.Namespace, modelConfig.Name)
 	}
```
EItanya: This same logic about finding the modelConfig while falling back to the default is present in mutliple places, can we create a helper?

go/controller/internal/autogen/autogen_reconciler.go
EItanya: Without finalizers this will be best effort, not against it but just want to be clear

go/controller/internal/httpserver/handlers/memory.go
```diff
@@ -62,48 +62,63 @@ func (h *MemoryHandler) HandleListMemories(w ErrorResponseWriter, r *http.Reques
 		}
 	}
 
+	log.Info("Successfully listed Memories", "count", len(memoryResponses))
 	RespondWithJSON(w, http.StatusOK, memoryResponses)
 }
 
 type CreateMemoryRequest struct {
 	Name           string                   `json:"name"`
+	Namespace      string                   `json:"namespace"`
```
EItanya: Should these be `Refs` like the rest of the way we reference `name/namespace`

EItanya: This would apply to all handlers as well

go/controller/internal/utils/common.go
```diff
@@ -13,3 +20,44 @@ func GetResourceNamespace() string {
 func MakePtr[T any](v T) *T {
 	return &v
 }
+
+func GetRefFromString(ref string, parentNamespace string) types.NamespacedName {
```
EItanya: Can you add unit tests for this?

EItanya (commented):
CHanges make sense to me overall. But this is definitely a very big PR so gonna take it one step at a time.

sbx0r:
Hello @EItanya . 
Thanks a lot for a thorough review. I've just come back from a business trip so I'll need a bit of time to recover before continuing.

go/controller/internal/autogen/autogen_reconciler.go
sbx0r: Thanks. Totally agree. I'm aware of that limitation and actually added a TODO comment to address this in the Agent reconciler some time ago.
Appreciate the reminter!

go/controller/internal/autogen/autogen_api_translator.go
sbx0r: Great suggestion!
I will apply.
We could plan some bigger refactor in near future also. What do you think?

go/controller/internal/utils/common.go
sbx0r: sir yes sir! 😃 
I'll also add some comments to these functions.

go/controller/internal/httpserver/handlers/memory.go
sbx0r: Do you want me to update a2a handlers also?

```go
// go/controller/internal/a2a/a2a_handler_mux.go:20-24
	SetAgentHandler(
		agentNamespace string,
		agentName string,
		params *A2AHandlerParams,
	) error

// go/controller/internal/a2a/a2a_reconciler.go:55-58
	return a.a2aHandler.SetAgentHandler(
		agent.Namespace, agent.Name,
		params,
	)
```

replace with

```go
// go/controller/internal/a2a/a2a_handler_mux.go
	SetAgentHandler(
		agentRef string,
		params   *A2AHandlerParams,
	) error

// go/controller/internal/a2a/a2a_reconciler.go
        agentRef := common.GetObjectRef(agent) // this helper will be added with incoming commit

	return a.a2aHandler.SetAgentHandler(
		agentRef,
		params,
	)
```

sbx0r:
#448 was discovered

sbx0r:
@EItanya
I believe that the stuff is ready to be reviewed :)
I'm working on the documentation

go/controller/internal/utils/common.go
```diff
@@ -20,3 +46,183 @@ func GetGlobalUserID() string {
...
+}
+
+// GetObject retrieves a Kubernetes resource by name and namespace.
+// The found resource is stored in the provided obj parameter.
+func GetObject(ctx context.Context, kube client.Client, obj client.Object, objName, objNamespace string) error {
+	ref, err := ParseRefString(objName, objNamespace)
+	if err != nil {
+		return err
+	}
+
```
EItanya, line 207: This is a lot of params, maybe this should just take a `types.NamespacedName`?

go/controller/internal/utils/common.go
```diff
@@ -1,7 +1,33 @@
...
+)
+
+// ResourceInfo represents a Kubernetes resource that can be associated with a ModelConfig.
+// It extends client.Object to provide access to standard Kubernetes object metadata
+// while adding the ability to specify which ModelConfig should be used for the resource.
+// Implementers must provide a GetModelConfigName() method that returns either:
+// - An empty string: indicating the default ModelConfig should be used
+// - A name: indicating a ModelConfig in the same namespace as the resource
+// - A namespace/name reference: indicating a specific ModelConfig in a specific namespace
+type ResourceInfo interface {
```
EItanya: Can we call this something slightly more releavnt to the impl. For example: `ObjectWithModelConfig` or something like that. `ResourceInfo` is pretty generic.

go/controller/internal/autogen/autogen_api_translator.go
EItanya: this isn't using the function

go/controller/internal/httpserver/handlers/utils.go
```diff
@@ -0,0 +1,13 @@
...
+
+import common "github.com/kagent-dev/kagent/go/controller/internal/utils"
+
+// Helper function to update a reference string
+func updateRef(refPtr *string, namespace string) error {
+	ref, err := common.ParseRefString(*refPtr, namespace)
+	if err != nil {
+		return err
+	}
+	*refPtr = ref.String()
```
EItanya, line 19: Should we check if `nil` just in case?

go/controller/internal/httpserver/handlers/teams.go
```diff
@@ -212,82 +325,171 @@ func (h *TeamsHandler) HandleGetTeam(w ErrorResponseWriter, r *http.Request) {
...
+
+		default:
+			log.Info("Unknown tool type", "toolType", toolCopy.Type)
+		}
+	}
+
+	// Create a new object that contains the Team information from Team and the ID from the autogenTeam
+	teamWithID := &TeamResponse{
+		Id:             autogenTeam.Id,
+		Agent:          *team,
```
EItanya: Why are you derefing here, can't this just be a pointer?

go/controller/internal/httpserver/handlers/teams.go
```diff
@@ -212,82 +325,171 @@ func (h *TeamsHandler) HandleGetTeam(w ErrorResponseWriter, r *http.Request) {
...
+			continue
+		}
+		memoryRefs = append(memoryRefs, memoryRef.String())
 	}
 
-	log.Info("Successfully retrieved team")
+	// Get the tools for the team
+	tools := make([]*v1alpha1.Tool, 0, len(team.Spec.Tools))
+	for _, tool := range team.Spec.Tools {
+		toolCopy := tool.DeepCopy()
```
EItanya, line 379: Why are you deep copying here? Are you modifying the original?

go/controller/internal/autogen/autogen_api_translator.go
sbx0r: It looks like I totally meant to do that 😆
fixing

go/controller/internal/utils/common.go
sbx0r: Great point!
Actually, `objName` can be specified as "namespace/name" or simply as "name" - in the latter case, `objNamespace` will be utilized. So at least the param names and the docs should be updated. However, changing the number of parameters or the function’s responsibilities would require moving some logic outside, which might reduce its flexibility.
Maybe we should discuss what the function’s main purpose should be? Let me know what you think.

EItanya: Ahh I see, so maybe we should change the first arg to `objRef` so it's more clear?

go/controller/internal/httpserver/handlers/teams.go
sbx0r: The opposite. I'm using `DeepCopy()` because I **don't want** to modify the **original** k8s objects. I need to ensure that even if the agent's Spec doesn't use 'ref strings' format for tools (agent and MCP ones), the handler will transform these references and ensure that the response contains proper 'ref strings' format for all tools.

go/controller/internal/httpserver/handlers/teams.go
```diff
@@ -92,20 +164,47 @@ func (h *TeamsHandler) HandleListTeams(w ErrorResponseWriter, r *http.Request) {
...
+	log.Info("Received request to update Team")
 
 	var teamRequest *v1alpha1.Agent
 	if err := DecodeJSONBody(r, &teamRequest); err != nil {
 		w.RespondWithError(errors.NewBadRequestError("Invalid request body", err))
 		return
 	}
-	log = log.WithValues("teamName", teamRequest.Name)
 
+	if teamRequest.Namespace == "" {
```
EItanya, line 169: What do you think about using the namespace of the controller instead? `default` may get very confusing

go/controller/internal/httpserver/handlers/teams.go
EItanya: Ahh I see what you're doing, I think it may be slightly overkill though. We are already returning the Autogen Config which essentially represents the translated config. In other words the expanded references if one knows what they are looking for. This comment I less about this specific usage and more about the `TeamResponse` in general. It's now A LOT of repeated info. I'm ok with this as is, but maybe after it merges we should have a discussion about what this part of the REST API is really for, and what information is important for people to have.

go/controller/internal/httpserver/handlers/teams.go
sbx0r: I was also thinking about it, but my concern was if we want kagent to override default k8s behavior. Most controllers are defaulting to the `default` namespace rather than the controller's namespace, but it's up to you @EItanya

go/controller/internal/httpserver/handlers/teams.go
sbx0r: With pleasure. The API definitely needs a refactor. Handlers need commons, DTOs need alignments.

go/controller/internal/httpserver/handlers/teams.go
EItanya: I definitely think we should use the deployed namespace, that's been the most common in my experience. Otherwise RBAC gets confusing. I think we want to support a namespaced deployment of kagent eventually and this would become complicated.

sbx0r: Ahhh you're right! Okay. I'll adjust

go/controller/internal/httpserver/server.go
```diff
@@ -139,14 +139,14 @@ func (s *HTTPServer) setupRoutes() {
...
 	s.router.HandleFunc(APIPathToolServers, adaptHandler(s.handlers.ToolServers.HandleListToolServers)).Methods(http.MethodGet)
 	s.router.HandleFunc(APIPathToolServers, adaptHandler(s.handlers.ToolServers.HandleCreateToolServer)).Methods(http.MethodPost)
-	s.router.HandleFunc(APIPathToolServers+"/{toolServerName}", adaptHandler(s.handlers.ToolServers.HandleDeleteToolServer)).Methods(http.MethodDelete)
+	s.router.HandleFunc(APIPathToolServers+"/{namespace}/{toolServerName}", adaptHandler(s.handlers.ToolServers.HandleDeleteToolServer)).Methods(http.MethodDelete)
 
 	// Teams
 	s.router.HandleFunc(APIPathTeams, adaptHandler(s.handlers.Teams.HandleListTeams)).Methods(http.MethodGet)
 	s.router.HandleFunc(APIPathTeams, adaptHandler(s.handlers.Teams.HandleCreateTeam)).Methods(http.MethodPost)
 	s.router.HandleFunc(APIPathTeams, adaptHandler(s.handlers.Teams.HandleUpdateTeam)).Methods(http.MethodPut)
 	s.router.HandleFunc(APIPathTeams+"/{teamID}", adaptHandler(s.handlers.Teams.HandleGetTeam)).Methods(http.MethodGet)
```
EItanya, line 148: Why is this still exposed as `teamID` and not `{namespace}/{teamName}`?

go/controller/internal/httpserver/server.go
```diff
@@ -139,14 +139,14 @@ func (s *HTTPServer) setupRoutes() {
 	// Tool Servers
 	s.router.HandleFunc(APIPathToolServers, adaptHandler(s.handlers.ToolServers.HandleListToolServers)).Methods(http.MethodGet)
 	s.router.HandleFunc(APIPathToolServers, adaptHandler(s.handlers.ToolServers.HandleCreateToolServer)).Methods(http.MethodPost)
-	s.router.HandleFunc(APIPathToolServers+"/{toolServerName}", adaptHandler(s.handlers.ToolServers.HandleDeleteToolServer)).Methods(http.MethodDelete)
+	s.router.HandleFunc(APIPathToolServers+"/{namespace}/{toolServerName}", adaptHandler(s.handlers.ToolServers.HandleDeleteToolServer)).Methods(http.MethodDelete)
 
 	// Teams
```
EItanya, line 144: Why are the names of these `teams`? That seems like too much of an implementation detail. These are actually the `agents` from the user-facing API. I think we should change the names. Doesn't need to be this PR as I don't want to make the scope even bigger, just want to put it out there

EItanya (approved):
The scope here is large, but I've done a few thorough looks and the changes look good to me :) @peterj will rely on you for the UI side of things

go/controller/internal/httpserver/server.go
sbx0r: So the logic under the hood for the teams is based on the id from the beginning. I'm aware that bud didn't want to add more noise because it will work.
Changing this will require changes in:
- Reconciler
- Handler
- UI
- And DTOs ofc.

We should definitely normalize it - and align with other handlers etc, but let's make a follow-up story for that.

sbx0r: I've made an follow-up for that (not only for this specific endpoint)
https://github.com/kagent-dev/kagent/issues/518

ui/src/app/agents/new/page.tsx
```diff
@@ -260,7 +266,23 @@ function AgentPageContent({ isEditMode, agentId }: AgentPageContentProps) {
                 </div>
 
                 <div>
-                  <label className="text-base mb-2 block font-bold">Description</label>
+                  <label className="text-base mb-2 block font-bold">Agent Namespace</label>
+                  <p className="text-xs mb-2 block text-muted-foreground">
+                    This is the namespace of the agent that will be displayed in the UI and used to identify the agent.
+                  </p>
+                  <Input
+                    value={namespace}
```
peterj, line 274: add a default value of `kagent` here.

ui/src/components/chat/ChatMessage.tsx
```diff
@@ -66,13 +71,15 @@ export default function ChatMessage({ message, allMessages }: ChatMessageProps)
     setFeedbackDialogOpen(true);
   };
 
+  const formattedSource = convertToUserFriendlyName(source);
```
peterj, line 74: this issue came from me, but I figured since you're already making changes here ... :D 

can you add the `?` here (line 42/47):

```
content = message.data.task_result?.stop_reason || "An error occurred";
```

peterj (approved):
looks good!

ui/src/components/chat/ChatMessage.tsx
sbx0r: Done ✅

peterj (approved):

sbx0r:
e2e fails due to #523

PR #406, opened 2025-05-19
Feature/llm default headers

dimetron:
This allow to add custom headers for model config


go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -133,6 +133,10 @@ type AzureOpenAIConfig struct {
 	// Top-p sampling parameter
 	// +optional
 	TopP string `json:"topP,omitempty"`
+
+	// Default Headers
+	// +optional
+	DefaultHeaders map[string]string `json:"defaultHeaders,omitempty"`
```
EItanya: Can we add this to the top-level, this will be useful for all clients.

dimetron: New TopLevel structure for common config or ModelConfigSpec ?

helm/kagent/values.yaml
```diff
@@ -35,6 +35,8 @@ providers:
       azureAdToken: ""
       azureDeployment: ""
       azureEndpoint: ""
+      defaultHeaders:
+        application: "kagent"
```
peterj: should we be setting the value here or just leave it empty?

dimetron: since we moved all up - removing

EItanya (dismissed):

peterj (approved):

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -1038,6 +1039,14 @@ func (a *apiTranslator) createModelClientForProvider(ctx context.Context, modelC
 			}
 		}
 
+		if len(modelConfig.Spec.DefaultHeaders) > 0 {
```
peterj: this here is now only done for the Ollama case -- it should be in each of the case statements,  right?

peterj (dismissed):

go/autogen/api/models.go
```diff
@@ -117,6 +123,7 @@ type OllamaClientConfiguration struct {
 	ModelCapabilities interface{}       `json:"model_capabilities,omitempty"`
 	ModelInfo         *ModelInfo        `json:"model_info"`
 	Options           map[string]string `json:"options"`
+	BaseClientConfig
```
EItanya: This won't work for `Ollama`: see: https://github.com/microsoft/autogen/blob/f0b73441b6dde66616ce4400f11e994eaa7403f7/python/packages/autogen-ext/src/autogen_ext/models/ollama/config/__init__.py#L40

EItanya (dismissed):

dimetron:
there is small issue with modelConfig template, need to fix before

peterj (approved):

Makefile
```diff
@@ -197,7 +209,7 @@ helm-agents:
 	VERSION=$(VERSION) envsubst < helm/agents/cilium-crd/Chart-template.yaml > helm/agents/cilium-crd/Chart.yaml
 	helm package helm/agents/cilium-crd
 
-.PHONY: helm-version
+.PHONY: helm-cleanup helm-version
```
EItanya: Should `helm-cleanup` be on the line below so it's called as a dependency?

dimetron: yeah, my bad

PR #408, opened 2025-05-19
Non interactive cli [Dashboard, Get]

EItanya:
This PR furthers the work to create parity between the interactive and non-interactive versions of the CLI.

### DIFF
--- a/go/cli/cmd/kagent/main.go
+++ b/go/cli/cmd/kagent/main.go
@@ -124,7 +124,7 @@ func main() {
 			if len(args) > 0 {
 				resourceType = args[0]
 			} else {
-				fmt.Fprintf(os.Stderr, "Invalid resource type: %s\n", resourceType)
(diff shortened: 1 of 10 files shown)


go/cli/cmd/kagent/main.go
```diff
@@ -86,11 +91,62 @@ func main() {
...
+			if err := cli.CheckServerConnection(client); err != nil {
+				pf := cli.NewPortForward(ctx, cfg)
+				defer pf.Stop()
+			}
+			resourceType := ""
+			resourceName := ""
+			if len(args) > 0 {
+				resourceType = args[0]
+			} else {
+				fmt.Fprintf(os.Stderr, "Invalid resource type: %s\n", resourceType)
```
peterj: wouldn't resourceType be "" here?

EItanya: Good catch

PR #412, opened 2025-05-20
Initial feedback feature

peterj:
This PR adds the ability to 👍🏼 / 👎🏼  the agent responses. The collected feedback is then stored in the DB (where other chats, agents etc. are stored). 

For now, this is the extent of this feature, but at least we can enable users to start collecting feedback on the responses.


python/src/autogenstudio/datamodel/db.py
```diff
@@ -67,6 +67,30 @@ class Message(BaseDBModel, table=True):
...
+
+    is_positive: bool = Field(default=False, description="Whether the feedback is positive or negative")
+    feedback_text: str = Field(description="The feedback text provided by the user")
+    issue_type: Optional[str] = Field(default=None, description="Category of issue for negative feedback")
+
+    message_content: str = Field(description="Content of the message that received feedback")
+    message_source: str = Field(description="Source of the message (agent name)")
+
+    preceding_messages: Optional[List[str]] = Field(default=[], sa_column=Column(JSON),
+                                                  description="Contents of messages preceding the feedback")
```
EItanya, line 84: Messages have their own DB type, can't we just point to that. In fact `session_id` shouldn't be necessary as the messages have `session_id`

python/src/autogenstudio/datamodel/db.py
```diff
@@ -67,6 +67,30 @@ class Message(BaseDBModel, table=True):
     message_meta: Optional[Union[MessageMeta, dict]] = Field(default={}, sa_column=Column(JSON))
 
 
+class Feedback(BaseDBModel, table=True):
```
EItanya, line 70: Doesn't this model need a primary key?

EItanya (commented):
More generally, can't just store the feedback directly on the message itself rather than creating a new model?

python/src/autogenstudio/datamodel/db.py
peterj: I can; but the issue is that I don't get the message IDs (from the DB) on the frontend as the messages I am getting are the messages that are being streamed and they don't have the ID (at that point). It would require to do extra fetching for actual messages. 

I can do it, but i feels messy.

Issue #374, opened 2025-05-12
[FEATURE] Read tool server configuration from a secret/configmap


peterj:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Allow reading toolserver config from cm/secret

### ❓ Problem Statement / Motivation

At the moment, we can only configure toolservers directly in the CRD - i.e. we don't support providing configuration through configmaps or secrets. We should allow that as we don't want to store API keys directly in the resource itself.

### 💡 Proposed Solution

This will involve changes in the controller as well as the UI.

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

nimishamehta5:
How are tool server API keys currently configured? Is that done through [this Env map](https://github.com/kagent-dev/kagent/blob/8374aa4c694274039eb7e3cda882d4ac50cda3cd/go/autogen/api/tool_servers.go#L20) for the Stdio MCP servers, and some other method for SSE servers?

peterj:
Exactly, it's through the Env map for the stdio servers. There's no configuration per se for the SSE, however, I am thinking it might make sense to have an option of reading the header values from the secret/configmap (especially if auth header is provided).

nimishamehta5:
Makes sense, I can look into this issue then if that works? @peterj

peterj:
Of course - thank you!

nimishamehta5:
@peterj, I'd appreciate a review on this draft if it looks like the right direction: https://github.com/kagent-dev/kagent/pull/416
thanks!

PR #416, opened 2025-05-21
Read tool server configuration from a secret or configmap

nimishamehta5:
Fixes: #374

After this change, the `toolservers` CRD now includes:

- For Stdio MCP Servers:
`envFrom` field that allows sourcing environment variables from ConfigMaps or Secrets.

- For SSE MCP Servers:
`headersFrom` field that allows sourcing header values from ConfigMaps or Secrets.

This allows users to store sensitive information like API keys in Kubernetes Secrets and reference them in toolserver resources, rather than embedding them directly in the CRD. 

Example toolserver config tested in local kind cluster:
```
kind: ToolServer
metadata:
  name: mcp-toolserver
  namespace: kagent
spec:
  config:
    sse:
      url: http://mcp-website-fetcher.default.svc.cluster.local/sse
      headersFrom:
        - name: "Authorization"
          valueFrom:
            secretKeyRef:
              name: "kagent-openai"
              key: "OPENAI_API_KEY"
  description: Fetches a website and returns its content
```


go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -87,21 +88,131 @@ func (a *apiTranslator) TranslateToolServer(ctx context.Context, toolServer *v1a
...
+}
+
+// fetchValueFromSource fetches a value from either a ConfigMap or Secret
+func (a *apiTranslator) fetchValueFromSource(ctx context.Context, name types.NamespacedName, key string) (string, error) {
+	// Try to fetch from ConfigMap first
+	configMap := &corev1.ConfigMap{}
+	err := a.kube.Get(ctx, name, configMap)
+	if err == nil {
+		value, exists := configMap.Data[key]
+		if !exists {
```
peterj: The way this is written, it seems like it won't make it to line 120, where you check for Secret.

I think you should check whether the ConfigMap exists and if it does, then return. If it doesn't exist, ignore it and move to the Secret.

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -87,21 +88,131 @@ func (a *apiTranslator) TranslateToolServer(ctx context.Context, toolServer *v1a
...
+
+		// Add environment variables from EnvFrom
+		if len(config.Stdio.EnvFrom) > 0 {
+			for _, envVar := range config.Stdio.EnvFrom {
+				if envVar.ValueFrom != nil {
+					var value string
+					var err error
+
+					if envVar.ValueFrom.ConfigMapKeyRef != nil {
+						value, err = a.resolveValueSource(ctx, envVar.ValueFrom.ConfigMapKeyRef, namespace)
```
peterj: since you're explicitly checking ConfigMap and SecretKeyRef here, perhaps having a a.getConfigMapValue and getSecretValue (which I think we might already have), might work better.

peterj (changes requested):
left a couple of comments

go/controller/internal/autogen/autogen_api_translator.go
nimishamehta5: I see what you mean, I refactored this to be more modular and also implemented the fallback to secret value.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,54 @@ type ToolServerConfig struct {
...
+	ConfigMapKeyRef *ValueSource `json:"configMapKeyRef,omitempty"`
+	SecretKeyRef    *ValueSource `json:"secretKeyRef,omitempty"`
+}
+
+// Header represents an HTTP header
+type Header struct {
+	Name string `json:"name"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	Value     AnyType       `json:"value,omitempty"`
```
EItanya: Why is this `AnyType`? Headers need to be encoded as strings.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,54 @@ type ToolServerConfig struct {
 	Sse   *SseMcpServerConfig   `json:"sse,omitempty"`
 }
 
+// ValueSource defines a source for configuration values from a Secret or ConfigMap
+type ValueSource struct {
+	Name      string `json:"name"`
+	Key       string `json:"key"`
+	Namespace string `json:"namespace,omitempty"`
```
EItanya: Everywhere else in the codebase we're following an implicit `ref` practice rather than explicit `name` + `namespace`. Meaning that to indicate a namespace, it should be separated with a `/`. e.g. `name/namespace

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,54 @@ type ToolServerConfig struct {
...
+}
+
+// EnvVarSource represents a source for an environment variable value
+type EnvVarSource struct {
+	ConfigMapKeyRef *ValueSource `json:"configMapKeyRef,omitempty"`
+	SecretKeyRef    *ValueSource `json:"secretKeyRef,omitempty"`
```
EItanya: Rather than having separate `ConfigMapKey` and `SecretKey` fields, can we just have a `Type` field in `ValueSource` which is either `ConfigMap` or `Secret`

This would allow for removal of a level of nesting in the `_Source` types.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,54 @@ type ToolServerConfig struct {
...
+}
+
+// EnvVar represents an environment variable
+type EnvVar struct {
+	Name      string        `json:"name"`
+	Value     string        `json:"value,omitempty"`
+	ValueFrom *EnvVarSource `json:"valueFrom,omitempty"`
```
EItanya: Please add `CEL` rules to ensure that only one of these is set at a time.

go/controller/api/v1alpha1/toolserver_types.go
EItanya: Also, please add `CEL` rules to ensure only one of these is set at a time.

nimishamehta5: Do you mean one of either `Value` or `ValueFrom`?  I guess for a given header, there would only be one value so that makes sense to me

go/controller/api/v1alpha1/toolserver_types.go
nimishamehta5: @EItanya I've addressed some of the API comments here, PTAL

EItanya: This looks awesome 😄 

Is it ready for another review?

nimishamehta5: thanks! :) 
yeah, had to resolve some conflicts - can you re-review when you get a chance?

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,50 @@ type ToolServerConfig struct {
...
+
+const (
+	ConfigMapValueSource ValueSourceType = "ConfigMap"
+	SecretValueSource    ValueSourceType = "Secret"
+)
+
+// ValueSource defines a source for configuration values from a Secret or ConfigMap
+// The Name is in the format "name/namespace", if namespace is not provided, the default namespace is used
+type ValueSource struct {
+	Type ValueSourceType `json:"type"`
```
EItanya, line 44: Can you add the kubebuilder rule to make sure this is actually treated like an enum.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,50 @@ type ToolServerConfig struct {
...
+	Name string          `json:"name"`
+	Key  string          `json:"key"`
+}
+
+// EnvVar represents an environment variable
+// +kubebuilder:validation:XValidation:rule="(has(self.value) && !has(self.valueFrom)) || (!has(self.value) && has(self.valueFrom))",message="Exactly one of value or valueFrom must be specified"
+type EnvVar struct {
+	Name      string       `json:"name"`
+	Value     string       `json:"value,omitempty"`
+	ValueFrom *ValueSource `json:"valueFrom,omitempty"`
```
EItanya: I believe these both need to be `optional` or else creating the resource will fail.

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,50 @@ type ToolServerConfig struct {
...
 type StdioMcpServerConfig struct {
 	Command string            `json:"command"`
 	Args    []string          `json:"args,omitempty"`
 	Env     map[string]string `json:"env,omitempty"`
+	EnvFrom []EnvVar          `json:"envFrom,omitempty"`
+}
+
+// Header represents an HTTP header
+// +kubebuilder:validation:XValidation:rule="(has(self.value) && !has(self.valueFrom)) || (!has(self.value) && has(self.valueFrom))",message="Exactly one of value or valueFrom must be specified"
+type Header struct {
```
EItanya: This is identical to the EnvVar object, why do we need both?

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -70,21 +71,136 @@ func (a *apiTranslator) TranslateToolServer(ctx context.Context, toolServer *v1a
...
-func translateToolServerConfig(config v1alpha1.ToolServerConfig) (string, *api.ToolServerConfig, error) {
+// parseNamespacedName parses a string in the format "name/namespace" or just "name"
+// and returns the name and namespace separately.
+func parseNamespacedName(namespacedName string, defaultNamespace string) (name string, namespace string) {
+	parts := strings.Split(namespacedName, "/")
+	if len(parts) > 1 {
+		return parts[0], parts[1]
+	}
+	return namespacedName, defaultNamespace
+}
```
EItanya: We should already have a helper function for this, I can't recall off the top of my head

go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -31,17 +31,50 @@ type ToolServerConfig struct {
...
+const (
+	ConfigMapValueSource ValueSourceType = "ConfigMap"
+	SecretValueSource    ValueSourceType = "Secret"
+)
+
+// ValueSource defines a source for configuration values from a Secret or ConfigMap
+// The Name is in the format "name/namespace", if namespace is not provided, the default namespace is used
+type ValueSource struct {
+	Type ValueSourceType `json:"type"`
+	Name string          `json:"name"`
```
EItanya: If this can be cross-namespace can we call it `Ref`

go/controller/api/v1alpha1/toolserver_types.go
nimishamehta5: good point, I have consolidated this into one

go/controller/internal/autogen/autogen_api_translator.go
nimishamehta5: yeah, just found it and used it now, thanks for pointing this out!

EItanya (approved):
All of my comments have been addressed, super excited for this to merge :)

PR #419, opened 2025-05-22
feat(install): Adds a --home-bin option to the get-kagent script

ofrivera:
- Adds `--home-bin` option to the `get-kagent` script, allowing users to install kagent into `$HOME/bin` instead of `/usr/local/bin`. This is useful for environments where sudo access is unavailable. The script now creates `$HOME/bin` if it doesn't exist and warns the user if `$HOME/bin` is not in their `PATH`, providing instructions for adding it.


EItanya:
Rather than having a `--home-bin` option. Should we create/advertise an official way to run this in non-sudo mode?

PR #420, opened 2025-05-22
feat: surface tool server errors in API and UI

zanuka:
**Summary:**  
- enhances backend (`HandleListToolServers` in Go) to include a `status.error` field for each tool server, surfacing reconciliation and other errors from the controller
- updated the UI to display these errors prominently in the tool server card, so users are aware of issues (e.g., failed auth, misconfiguration) directly in the tool server list
- added status.error field to tool server API response for reconciliation and other errors
- improves visibility of tool server issues for users

Note: @peterj I could use help testing error display in the UI. I wasn't able to add a tool server in localhost

### DIFF
--- a/go/controller/internal/httpserver/handlers/toolservers.go
+++ b/go/controller/internal/httpserver/handlers/toolservers.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kagent-dev/kagent/go/controller/internal/httpserver/errors"
 	common "github.com/kagent-dev/kagent/go/controller/internal/utils"
 	"k8s.io/apimachinery/pkg/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -18,6 +19,16 @@ type ToolServersHandler struct {
 // NewToolServersHandler creates a new ToolServersHandler
 func NewToolServersHandler(base *Base) *ToolServersHandler {
 	return &ToolServersHandler{Base: base}
+}
+
+// getErrorFromConditions extracts error information from conditions
+func getErrorFromConditions(conditions []metav1.Condition) *string {
+	for _, condition := range conditions {
+		if condition.Status == metav1.ConditionFalse {
+			return &condition.Message
+		}
+	}
+	return nil
 }
 
 // HandleListToolServers handles GET /api/toolservers requests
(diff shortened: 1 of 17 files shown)


go/controller/internal/httpserver/handlers/toolservers.go
```diff
@@ -35,10 +46,23 @@ func (h *ToolServersHandler) HandleListToolServers(w ErrorResponseWriter, r *htt
 	for _, toolServer := range toolServerList.Items {
 		log.V(1).Info("Processing tool server", "toolServerName", toolServer.Name)
 
-		toolServerWithTools = append(toolServerWithTools, map[string]interface{}{
+		errorMsg := getErrorFromConditions(toolServer.Status.Conditions)
+		if errorMsg != nil {
+			log.Info("Tool server has error condition", "toolServerName", toolServer.Name, "error", *errorMsg)
```
EItanya: This should be a `debug` level log if anything. The information will already be returned by the handler.

go/controller/internal/httpserver/handlers/toolservers.go
```diff
@@ -35,10 +46,23 @@ func (h *ToolServersHandler) HandleListToolServers(w ErrorResponseWriter, r *htt
...
+			discoveredTools = []*v1alpha1.MCPTool{}
+		}
+		toolServerWithTools = append(toolServerWithTools, map[string]interface{}{ 
 			"name":            toolServer.Name,
 			"config":          toolServer.Spec.Config,
-			"discoveredTools": toolServer.Status.DiscoveredTools,
+			"discoveredTools": discoveredTools,
+			"status": map[string]interface{}{
+				"conditions": toolServer.Status.Conditions,
+				"error":     errorMsg,
```
EItanya: Why not just return the conditions and let the UI decide?

go/controller/internal/httpserver/handlers/toolservers.go
zanuka: agreed... since the error is now surfaced in the API response, the log line has been updated to use the debug level instead of info. have updated

go/controller/internal/httpserver/handlers/toolservers.go
zanuka: good point... I added the error field to simplify error display in the UI, but can remove it and let the UI parse the conditions array instead. Let me know which approach you prefer.

zanuka:
@EItanya The additional updates I've made include several improvements to the tool server functionality:

1. Aligns backend and frontend data structures by including the config field in responses
2. Implements consistent error handling across the tool server lifecycle (create, list, delete)
3. Maintains backward compatibility while improving type safety and error reporting
4. Updates tests to reflect the new data structure and error handling patterns
5. Removes outdated tests that were no longer relevant to the current handler logic

this should ensure better type safety, more reliable error handling, and a more consistent API contract between the frontend and backend.

peterj:
I was trying to test it and create an MCP server (using the github one), but without providing the token. I see this in the controller logs:

```
{"controller": "toolserver", "controllerGroup": "kagent.dev", "controllerKind": "ToolServer", "ToolServer": {"name":"server-github","namespace":"kagent"}, "namespace": "kagent", "name": "server-github", "reconcileID": "758ac5c6-df24-4c40-a208-1e5b7a72d2dc", "error": "failed to reconcile tool server server-github: ToolServer.kagent.dev \"server-github\" is invalid: status.conditions: Required value"}
```

Also, not sure how this was working previously, but if the mcp server doesn't exist/not found (e.g. "npx something-that-doesn'texist"), the status is never set on the resource.

zanuka:
@peterj I need to unassign myself from this task. Just not enough time to continue contributing with my existing commitments. Feel free to update this branch or close PR as needed.

PR #426, opened 2025-05-23
Adds Kubernetes Agent Benchmark

rinormaloku:
After running the benchmark 3 times (repeating only with the failed tests) we have these results:

1st pass, 3rd pass, 4th pass:
```
success: 16,19,20
failures: 8,5,4
```

**With 3 executions 20/24 passed ~83%**


.github/data/agent-framework/resources/agent.yaml
```diff
@@ -0,0 +1,184 @@
+apiVersion: kagent.dev/v1alpha1
```
peterj, line 1: a work item for later: we have these agents packaged up as Helm charts, so we might want to use that, so we don't have a drift between the agents here and in the yaml.

ui/src/app/agents/new/page.tsx
```diff
@@ -103,7 +103,7 @@ function AgentPageContent({ isEditMode, agentId }: AgentPageContentProps) {
                 model: agentResponse.model,
                 name: agent.spec.modelConfig,
               });
-
```
peterj: looks like there are only formatting changes here; can you revert this? (feel free to make the formatting change in a separate PR if needed).

peterj (dismissed):

peterj (approved):

Issue #101, opened 2025-03-06
[FEATURE] add a binary path configuration setting to kubectl, istioctl and helm


peterj:

marcinkubica:
should be fixed in #499 - re: https://github.com/kagent-dev/kagent/pull/434#issuecomment-2959035443

PR #434, opened 2025-05-24
feat: add binary path configuration for kubectl, istioctl, and helm

abhinav-1305:
fixes: #101, cc: @EItanya.


python/src/kagent/tools/common/_shell.py
```diff
@@ -1,13 +1,51 @@
...
+    elif command == "istioctl":
+        binary_path = _binary_paths.istioctl_path
+    elif command == "helm":
+        binary_path = _binary_paths.helm_path
+        
     try:
-        output = subprocess.check_output([command] + args, stderr=subprocess.STDOUT)
-        return output.decode("utf-8")
+        # Run the command with the specified binary path
+        result = subprocess.run(
```
peterj, line 34: what is the difference between the `subprocess.check_output` and the new `subprocess.run`?

python/src/kagent/tools/common/_binary_paths.py
```diff
@@ -0,0 +1,20 @@
+from typing import Optional
+from pydantic import BaseModel, Field
+
+class BinaryPathsConfig(BaseModel):
+    """Configuration for binary paths used by various tools."""
+    
+    kubectl_path: Optional[str] = Field(
+        default="kubectl",
```
peterj: I think the path here should default to `/usr/local/bin/[binary_name]`. 

Also, we have helm and kubectl-argo-rollouts binaries installed as well, so you might want to add those.

peterj (changes requested):
added a couple of comments/questions

python/src/kagent/tools/common/_shell.py
abhinav-1305: I used `subprocess.check_output()` because I only need the command’s output, and it automatically shows an error if the command fails. It's a simple fit here, but I’m happy to switch to `subprocess.run()` if you prefer.

python/src/kagent/tools/common/_shell.py
```diff
@@ -1,13 +1,51 @@
...
 
-# Function that runs the command in the shell
-def run_command(command: str, args: list[str]) -> str:
-    """Run the given command and return the output."""
+logger = logging.getLogger(__name__)
+
+# Global configuration instance
+_binary_paths = BinaryPathsConfig()
+
+def set_binary_paths(config: BinaryPathsConfig) -> None:
```
peterj: where is this used?

python/src/kagent/tools/common/_shell.py
peterj: I just wanted to make sure that the output of the both commands is the same

python/src/kagent/tools/common/_shell.py
abhinav-1305: removed it

abhinav-1305:
Requesting a review @peterj - let me know if any other changes are needed.

python/src/kagent/tools/common/_shell.py
```diff
@@ -1,13 +1,46 @@
...
 
+from ._binary_paths import BinaryPathsConfig
 
-# Function that runs the command in the shell
-def run_command(command: str, args: list[str]) -> str:
-    """Run the given command and return the output."""
+logger = logging.getLogger(__name__)
+
+# Global configuration instance
+_binary_paths = BinaryPathsConfig()
```
EItanya, line 10: How would a user modify the global config instance. I think the purpose is to allow users to modify this if necessary.

EItanya (changes requested):

python/src/kagent/tools/common/_shell.py
abhinav-1305: For example, by creating a new instance as shown below:

``` bash
from kagent.tools.common._shell import _binary_paths

# Change the path to kubectl
_binary_paths.kubectl_path = "/my/custom/path/to/kubectl"

# Change the path to istioctl
_binary_paths.istioctl_path = "/my/custom/path/to/istioctl"
```

EItanya: Right, but the way users modify the configuration in `kagent` is dynamically, NOT via the code. In other words this needs to be done in the `Tool` config, not for the function.

abhinav-1305: > Right, but the way users modify the configuration in `kagent` is dynamically, NOT via the code. In other words this needs to be done in the `Tool` config, not for the function.

I'm thinking of going in this direction:

First, I’ll turn BinaryPathsConfig into a proper tool config by:
- Making it inherit from Component
- Adding the required metadata and serialization support
- Registering it with kagent’s tool discovery system

Then in the shell module, I’ll:
- Load binary paths from the tool config
- Fall back to default paths if no config is set
- Handle dynamic config updates properly

I'll  also register the tool in `common/__init__.py` to complete the setup.

The idea is that users will be able to define custom binary paths through the standard config file at:
`$HOME/.autogenstudio/configs/tools.json`

``` json
{
  "config": {
    "kubectl_path": "/custom/path/to/kubectl",
    "istioctl_path": "/custom/path/to/istioctl",
    "helm_path": "/custom/path/to/helm",
    "kubectl_argo_rollouts_path": "/custom/path/to/kubectl-argo-rollouts"
  }
}
```
any thoughts on this @EItanya ?

abhinav-1305: Following up - cc: @EItanya

EItanya: I think in general we should close this PR for now in lieu of #499. That issue will supersede these tools in the long-run and obviate the need for these binaries at all most likely.

abhinav-1305: make sense

abhinav-1305:
As @EItanya said, this will be addressed in #499, so I'm closing it for now.

Issue #435, opened 2025-05-25
BUG: tool servers order in the UI is not consistent


peterj:
add a couple of tool servers, refresh the page. Notice the order in which the servers are displayed is non-deterministic. We should store the tool servers.

PR #440, opened 2025-05-27
Fix inconsistent servers order in the UI

abhinav-1305:
Fixed non-deterministic tool server ordering by implementing `sort()` with `localeCompare()` in frontend and `sort(key=lambda x: x.name or '')` in backend API. 

Fixes: #435, cc: @peterj.


ui/src/app/servers/page.tsx
```diff
@@ -36,10 +36,16 @@ export default function ServersPage() {
       // Fetch servers
       const serversResponse = await getServers();
       if (serversResponse.success && serversResponse.data) {
-        setServers(serversResponse.data);
+        // Sort servers by name
```
peterj: Do we need to resort on the frontend if we're already sorting on the backend?

peterj (changes requested):
added one question

ui/src/app/servers/page.tsx
abhinav-1305: No, sorting in frontend is redundant - thanks for pointing it out.

python/src/autogenstudio/web/routes/tool_servers.py
```diff
@@ -12,6 +12,9 @@
 @router.get("/")
 async def list_servers(user_id: str, db=Depends(get_db)) -> Dict:
     response = db.get(ToolServer, filters={"user_id": user_id})
+    if response.status and response.data:
+        # Sort servers by name
+        response.data.sort(key=lambda x: x.name or '')
```
peterj: This is incorrect and will result in an exception. There is no `name` in the ToolServer data model.

Issue #323, opened 2025-05-04
[BUG] Fix the navigation in header on smaller screens


peterj:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

UI Service

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

<img width="615" alt="Image" src="https://github.com/user-attachments/assets/341f72e8-b7bd-4cd5-b201-7c8b5629aab7" />

### 🔄 Steps To Reproduce

see above

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

PR #444, opened 2025-05-27
Fix: header navigation on smaller screens

abhinav-1305:
fixes: #323, cc: @peterj.

### DIFF
--- a/ui/src/app/globals.css
+++ b/ui/src/app/globals.css
@@ -6,7 +6,39 @@ body {
   font-family: Arial, Helvetica, sans-serif;
 }
 
-
+/* Responsive header navigation tweaks */
+@media (max-width: 768px) {
+  nav {
+    /* Ensure nav takes full width and stacks content */
+    width: 100%;
+  }
+  .mobile-menu {
+    width: 100%;
+    background: var(--background);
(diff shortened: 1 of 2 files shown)


ui/src/app/globals.css
```diff
@@ -6,7 +6,39 @@ body {
   font-family: Arial, Helvetica, sans-serif;
 }
 
-
+/* Responsive header navigation tweaks */
+@media (max-width: 768px) {
+  nav {
```
peterj: we're using tailwind, so it would be best if this is translated to tailwind attributes in the actual .tsx file.

peterj (changes requested):
we're using tailwind, so it would be best if this is translated to tailwind attributes in the actual .tsx file.

ui/src/app/globals.css
abhinav-1305: Ok

peterj:
The issue is still there. Please make sure you test the changes.

Screenshot 2025-05-27 at 1 33 05 PM

peterj (changes requested):
please see the comments

peterj:
@abhinav-1305  is this ready for review?

peterj:
closing this due to inactivity.

Issue #225, opened 2025-04-08
check whether mcp tools are used in any of the agents before deleting


peterj:
Should we warn the user and/or disable deleting the MCP server if any of the tools are being used in agents?

berkeyalciin:
I think it would be better warn the user, but also the user should be able to delete. It disapperas already from the right side of the screen when the tool removed. Users can see available tools there

PR #446, opened 2025-05-27
feat: Verify agent usage before removing MCP tools

abhinav-1305:
fixes: #225

### DIFF
--- a/ui/src/app/servers/page.tsx
+++ b/ui/src/app/servers/page.tsx
@@ -3,10 +3,11 @@
 import { useState, useEffect } from "react";
 import { Server, Globe, Trash2, ChevronDown, ChevronRight, MoreHorizontal, Plus, FunctionSquare } from "lucide-react";
 import { Button } from "@/components/ui/button";
-import { getToolDescription, getToolDisplayName, getToolIdentifier } from "@/lib/toolUtils";
-import {  ToolServer, ToolServerWithTools } from "@/types/datamodel";
+import { getToolDescription, getToolDisplayName, getToolIdentifier, isMcpServerUsedByAgents } from "@/lib/toolUtils";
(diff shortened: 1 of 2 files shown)


ui/src/app/servers/page.tsx
```diff
@@ -236,7 +257,11 @@ export default function ServersPage() {
           }
         }}
         title="Delete Server"
-        description="Are you sure you want to delete this server? This will also delete all associated tools and cannot be undone."
+        description={
+          showConfirmDelete && isMcpServerUsedByAgents(showConfirmDelete, agents)
```
peterj: we should show the deletion prompt regardless if the MCP server tools are used by agents or not. If they are used by agents, let's show the list of agents they are being used by as well, so it's easier for user to decide whether they want to delete the server or not.

abhinav-1305: The confirmation dialog now provides agent dependency context through `isMcpServerUsedByAgents` utility

ui/src/app/servers/page.tsx
```diff
@@ -230,14 +245,30 @@ export default function ServersPage() {
...
-          }
-        }}
+        onOpenChange={(open) => !open && setShowConfirmDelete(null)}
         title="Delete Server"
-        description="Are you sure you want to delete this server? This will also delete all associated tools and cannot be undone."
-        onConfirm={() => showConfirmDelete !== null && handleDeleteServer(showConfirmDelete)}
+        description={
+          showConfirmDelete ? (
+            <div className="space-y-2">
+              <p>Are you sure you want to delete the server &apos;{showConfirmDelete}&apos;? This action cannot be undone.</p>
```
peterj: this is incorrect here -- is it going to show "true"?

peterj:
@abhinav-1305 any updates on this PR?

EItanya:
I don't think we should be in the business of blocking deletions, closing for stale

Issue #450, opened 2025-05-27
update the CI so it only runs for code changes


peterj:
at the moment, it seems to run on any (even non) code changes.

PR #453, opened 2025-05-28
chore: run CI only for code changes

abhinav-1305:
Resolves: #450


.github/workflows/lint-python.yaml
```diff
@@ -6,6 +6,7 @@ on:
       - '**/*.py'  # Only trigger on changes to Python files
       - 'python/pyproject.toml'
       - 'python/uv.lock'
+  workflow_dispatch:
```
peterj: is this needed here?

abhinav-1305: No, I missed removing it.

abhinav-1305:
Hi @peterj  – could you please review this and let me know if other changes are needed?

abhinav-1305:
> Rather than deciding on paths on which the code should run, is there a way we can decide on which paths CI should NOT run. e.g. `README` etc. Deciding on which code paths CI should run I have found often ends up in CI accidentally not running when new code is added.

Ok cool - working on this.

abhinav-1305:
please review - @EItanya

EItanya:
These types of changes are rabbit holes, let's close for now

Issue #454, opened 2025-05-28
[BUG] `create_resource` and `apply_manifest` functions have a misleading description


EdmondDantes21:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Multiple services / System-wide issue

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

It seems that the descriptions of the `apply_manifest` and `create_resource` in `python/src/kagent/tools/k8s/_kubectl.py` unintentionally mislead the agent:

* `apply_manifest` description: _"Create a resource in Kubernetes. Must be a local file."_ Based on the implementation, it looks like the file doesn't need to be local. In my tests this consistently confused the agent when asked to apply a resource.

* `create_resource` description: _"Apply a YAML resource file to the Kubernetes cluster."_ This seems to have a similar issue as `apply_manifest`

Changing these to  _"Apply a YAML resource to the Kubernetes cluster."_ and _"Create a resource in Kubernetes."_ greatly improved the model's ability to use tools in my tests.

Please let me know if I am missing something.

### 🔄 Steps To Reproduce

1. Create an agent using qwen3:30b-a3b through ollama with the ApplyManifest and/or CreateResource k8s tools.
2. Ask the agent _"show me an example of a nginx pod"_
3. Then ask _"apply the nginx pod to my local kubernetes cluster"_

### 🤔 Expected Behavior

The agent calls either the create_manifest or the ApplyManifest or CreateResource tools with the manifest provided earlier as argument.

### 📱 Actual Behavior

The agent calls the ApplyManifest or CreateResource tools using a non existent 'nginx-pod.yaml' string as the argument, possibly being misled by the tool's description.

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR to fix this issue

EItanya:
Great catch, looking forward to the PR :)

PR #456, opened 2025-05-28
refactor: update descriptions for resource creation and manifest application in K8s tool

abhinav-1305:
Resolves: #454

### DIFF
--- a/python/src/kagent/tools/k8s/_kubectl.py
+++ b/python/src/kagent/tools/k8s/_kubectl.py
@@ -275,7 +275,7 @@ LabelResource, LabelResourceConfig = create_typed_fn_tool(
 
 create_resource = FunctionTool(
     _create_resource,
-    description="Create a resource in Kubernetes. Must be a local file.",
+    description="Create a resource in Kubernetes.",
     name="create_resource",
 )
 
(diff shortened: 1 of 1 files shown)


EdmondDantes21:
Two more lines should be changed in this PR at the functions' definitions, i.e. at `def _apply_manifest` and `def create_resource`. I may have made some mistakes, but this PR does not solve the issue for me without these two further adjustments.

EItanya:
> Two more lines should be changed in this PR at the functions' definitions, i.e. at `def _apply_manifest` and `def create_resource`. I may have made some mistakes, but this PR does not solve the issue for me without these two further adjustments.

Can you explain in more detail what you mean?

EdmondDantes21:
Sure, apologies for being so vague. 

Currently, the function's signature looks like this: 
```python
def _create_resource(
    resource_yaml: Annotated[str, "The YAML definition of the resource to create. Must be a local file"],
    namespace: Annotated[Optional[str], "The namespace of the resource to create"] = None,
) -> str:
```
The agent still refuses to comply due to that 'Must be a local file'. The same applies to the `apply_manifest` function. 
Btw, are you planning to get rid of built in tools?

EItanya:
That makes sense, I agree that this probably doesn't fully fix the issue, and yes we are planning to get rid of these built-ins so I don't think it's worth putting too much work into them right now.

PR #461, opened 2025-05-29
fix a2a handler removal + add a2a configs for other agents

ilackarms:


EItanya (approved):
LGTM. We may need to add finalizers at some point to ensure delete happens correctly.

PR #462, opened 2025-05-29
[feat] - added a2aConfig

dimetron:
- added more a2a config
- improve session name in create session logic due to  
  SQL insert  autogenstudio.database.db_manager:upsert:185 - 
  Error while updating/creating Session: (sqlite3.IntegrityError) 
  FOREIGN KEY constraint failed

### DIFF
--- a/go/controller/internal/a2a/a2a_reconciler.go
+++ b/go/controller/internal/a2a/a2a_reconciler.go
@@ -17,6 +17,11 @@ type A2AReconciler interface {
 		agent *v1alpha1.Agent,
 		autogenTeam *autogen_client.Team,
 	) error
+
+	ReconcileAutogenAgentDeletion(
+		agentNamespace string,
+		agentName string,
+	)
 }
 
 type a2aReconciler struct {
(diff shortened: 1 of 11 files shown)


go/controller/internal/a2a/autogen_a2a_translator.go
```diff
@@ -99,11 +99,15 @@ func (a *autogenA2ATranslator) makeHandlerForTeam(
 	return func(ctx context.Context, task string, sessionID *string) (string, error) {
 		var taskResult *autogen_client.TaskResult
 		if sessionID != nil && *sessionID != "" {
-			session, err := a.autogenClient.GetSession(*sessionID, common.GetGlobalUserID())
+			sessionName := fmt.Sprintf("%v-%s-%s", autogenTeam.CreatedAt, autogenTeam.UserID, *sessionID)
+			log.FromContext(ctx).Info("Using session for task", "sessionName", sessionName)
+			session, err := a.autogenClient.GetSession(sessionName, common.GetGlobalUserID())
```
EItanya: I'm not so sure about this change, why aren't we just using the `sessionID` directly?

dimetron: - a.autogenClient.GetSession - using session name not ID
- it was failing for me with SQL IntegrityError and I tried to make it more unique 
  SQL insert autogenstudio.database.db_manager:upsert:185 -
Error while updating/creating Session: (sqlite3.IntegrityError)

PR #463, opened 2025-05-29
fix/sessionName is not unique

dimetron:
Current implementation of create session not using proper name causing IntegrityError

kagent-fdc86b6b-2db9b:app 2025-05-29 20:28:45.519 | ERROR    | autogenstudio.database.db_manager:upsert:185 - Error while updating/creating Session: (sqlite3.IntegrityError) FOREIGN KEY constraint failed

Proposal to use: 

sessionName := fmt.Sprintf("%v-%s-%s", autogenTeam.CreatedAt, autogenTeam.UserID, *sessionID)

Which will make it sorted by date time as well user
"sessionName": "2025-05-29T23:32:34.858622-admin@kagent.dev-1209bbb5-eed8-4ebc-addb-130a8629d45d"}


EItanya:
I haven’t seen this error. What were you trying to use as the session_name. The session_name is shown in the UI as well so it should probably be user friendly.

dimetron:
> I haven’t seen this error. What were you trying to use as the session_name. The session_name is shown in the UI as well so it should probably be user friendly.

hmm, I was using external a2a voice client

dimetron:
I'll close it - will test more

PR #473, opened 2025-06-03
add Gemini support

### LINKED ISSUE #436: [FEATURE] Add Gemini as LLM provider
### 📋 Prerequisites

### 📝 Feature Summary

Add Gemini as LLM provider

The current implementation of `kagent` supports a limited set of LLM providers, which restricts flexibility and adaptability for users seeking to leverage state-of-the-art language models. As organizations adopt varied LLMs for performance, compliance, or cost reasons, the lack of support for Google Gemini becomes a notable limitation. This is especially relevant for teams already invested in Google Cloud ecosystems, where Gemini provides advantages in latency, integration, and potentially fine-tuned capabilities.

Adding Gemini support would enable more teams to adopt kagent without sacrificing their preferred LLM provider, improving usability and adoption in environments where Gemini is already being evaluated or used.

### 💡 Proposed Solution

Integrate Gemini as a supported LLM provider in the kagent framework.

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

peterj:
This adds Gemini support on the backend/controller.

I'll add the UI support in a separate PR.

Fixes #436


go/autogen/api/models.go
```diff
@@ -133,3 +133,47 @@ func (c *OllamaClientConfiguration) ToConfig() (map[string]interface{}, error) {
...
+	MaxOutputTokens  *int    `json:"max_output_tokens,omitempty"`
+	CandidateCount   *int    `json:"candidate_count,omitempty"`
+	ResponseMimeType *string `json:"response_mime_type,omitempty"`
+}
+
+type AnthropicVertexAIConfig struct {
+	BaseVertexAIConfig
+
+	MaxTokens *int `json:"max_tokens,omitempty"`
+	Stream    bool `json:"stream,omitempty"`
```
EItanya: This needs to be in the config, not the request?

python/src/kagent/models/vertexai/_anthropic_vertex_client.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+
+    def __getstate__(self) -> Dict[str, Any]:
+        state = self.__dict__.copy()
+        state["_client"] = None
+        return state
+
+    def __setstate__(self, state: Dict[str, Any]) -> None:
+        self.__dict__.update(state)
+        self._client = _anthropic_client_from_config(state["_raw_config"])
```
EItanya, line 119: What are these for?

python/src/kagent/models/vertexai/_anthropic_vertex_client.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class AnthropicVertexAIChatCompletionClient(
+    BaseAnthropicChatCompletionClient, Component[AnthropicVertexAIClientConfiguration]
+):
+    component_type = "model"
+    component_config_schema = AnthropicVertexAIClientConfiguration
+    component_provider_override = "kagent.models.vertexai.AnthropicVertexAIChatCompletionClient"
+
+    def __init__(self, **kwargs: Unpack[AnthropicVertexAIClientConfiguration]):
```
EItanya, line 58: Why not just pass in the whole config object, or have typed args?

python/src/kagent/models/vertexai/_anthropic_vertex_client.py
```diff
@@ -0,0 +1,121 @@
...
+        model_info: Optional[ModelInfo] = None
+        if "model_info" in kwargs:
+            model_info = kwargs["model_info"]
+            del copied_args["model_info"]
+
+        if "credentials" in kwargs:
+            credentials = kwargs["credentials"]
+            del copied_args["credentials"]
+        else:
+            raise ValueError("credentials is required for AnthropicVertexAIChatCompletionClient")
```
EItanya, line 80: What if the user wants to use the machine creds?

go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -25,16 +25,73 @@ const (
...
+	OpenAI            ModelProvider = "OpenAI"
+	Ollama            ModelProvider = "Ollama"
+	GeminiVertexAI    ModelProvider = "GeminiVertexAI"
+	AnthropicVertexAI ModelProvider = "AnthropicVertexAI"
 )
 
+type BaseVertexAIConfig struct {
+	// The project ID
+	// +optional
+	ProjectID string `json:"projectID,omitempty"`
```
EItanya: I think this should be required?

go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -25,16 +25,73 @@ const (
...
+type AnthropicVertexAIConfig struct {
+	BaseVertexAIConfig `json:",inline"`
+
+	// Maximum tokens to generate
+	// +optional
+	MaxTokens int `json:"maxTokens,omitempty"`
+
+	// Stream
+	// +optional
+	Stream bool `json:"stream,omitempty"`
```
EItanya: Same confusion about stream being in the config

go/autogen/api/models.go
```diff
@@ -133,3 +133,47 @@ func (c *OllamaClientConfiguration) ToConfig() (map[string]interface{}, error) {
 func (c *OllamaClientConfiguration) FromConfig(config map[string]interface{}) error {
 	return fromConfig(c, config)
 }
+
+type BaseVertexAIConfig struct {
```
EItanya, line 137: Doesn't need to be right this minute, but I do think we should support machine credential source for `google`

EItanya (commented):
Changes look good overall. I have some questions about machine identity, as well as the config code for `AnthropicVertex`. I didn't give the gemini client a super deep dive cuz it's A LOT of code 😆. Might be worth upstreaming it in the future.

python/src/kagent/models/vertexai/_anthropic_vertex_client.py
peterj: you mean if it runs on google cloud VM/infra? (I think there might be an env var I could use to check if the variable is set and use that instead of the credentials)

go/autogen/api/models.go
peterj: Let's open an issue and I can implement that as a followup

python/src/kagent/models/vertexai/_anthropic_vertex_client.py
peterj: it's used for serialization -- I follow the same pattern as the upstream implementation.

go/controller/api/v1alpha1/autogenmodelconfig_types.go
peterj: yep, forgot to change it once I removed the API key support

go/autogen/api/models.go
peterj: opened https://github.com/kagent-dev/kagent/issues/477

python/src/kagent/models/vertexai/_anthropic_vertex_client.py
peterj: with config object the caller has to pass in an object vs. if it's using kwargs you only pass in the args you need/want (you still get type hints and all that)

go/controller/api/v1alpha1/autogenmodelconfig_types.go
```diff
@@ -25,16 +25,69 @@ const (
...
-	Anthropic   ModelProvider = "Anthropic"
-	AzureOpenAI ModelProvider = "AzureOpenAI"
-	OpenAI      ModelProvider = "OpenAI"
-	Ollama      ModelProvider = "Ollama"
+	Anthropic         ModelProvider = "Anthropic"
+	AzureOpenAI       ModelProvider = "AzureOpenAI"
+	OpenAI            ModelProvider = "OpenAI"
+	Ollama            ModelProvider = "Ollama"
+	GeminiVertexAI    ModelProvider = "GeminiVertexAI"
+	AnthropicVertexAI ModelProvider = "AnthropicVertexAI"
```
marcinkubica, line 37: @peterj How does this work when user runs own model on VertexAI?

a-a-ron-stone:
would this PR enable us to use our Gemini API token within Kagent instead of the OpenAI token? just trying to understand

go/controller/api/v1alpha1/autogenmodelconfig_types.go
EItanya: Great question. This will specifically enable running Gemini and Anthropic models in Vertex. Adding custom models should be a pretty simple follow-up but isn't explicitly handled by this PR.

EItanya:
> would this PR enable us to use our Gemini API token within Kagent instead of the OpenAI token? just trying to understand

Great question. So this PR actually adds support for Gemini via Vertex AI, not via the original Gemini interface. The original Gemini interface is actually already supported via the OpenAI API. I have a PR up which exercises that functionality, we just need to document it.

EItanya (approved):

a-a-ron-stone:
> > would this PR enable us to use our Gemini API token within Kagent instead of the OpenAI token? just trying to understand
> 
> Great question. So this PR actually adds support for Gemini via Vertex AI, not via the original Gemini interface. The original Gemini interface is actually already supported via the OpenAI API. I have a PR up which exercises that functionality, we just need to document it.

apologies, I may not be explaining this very well - still new to this stuff. We have a paid enterprise license to use Gemini for business use etc. As we already pay for that, we were hoping to use that to use Kagent. So when we run

export OPENAI_API_KEY= I want to use our companies Gemini token to use those models instead of OpenAI if thats even possible?

EItanya:
Hey there. You should be able to, I was able to get it to work like this:
```
kind: ModelConfig
metadata:
  name: gemini-2.5-pro-model-config
  namespace: kagent
spec:
  apiKeySecretKey: GOOGLE_API_KEY
  apiKeySecretRef: kagent-google
  model: google/gemini-2.5-pro-preview-05-06
  provider: OpenAI
  openAI:
    baseUrl: https://generativelanguage.googleapis.com/v1beta/openai
  modelInfo:
    functionCalling: true
    jsonOutput: true
    family: gemini-2.5-pro
```

Let me know if that works for you.

a-a-ron-stone:
> Hey there. You should be able to, I was able to get it to work like this:
> 
> ```
> apiVersion: kagent.dev/v1alpha1
> kind: ModelConfig
> metadata:
>   name: gemini-2.5-pro-model-config
>   namespace: kagent
> spec:
>   apiKeySecretKey: GOOGLE_API_KEY
>   apiKeySecretRef: kagent-google
>   model: google/gemini-2.5-pro-preview-05-06
>   provider: OpenAI
>   openAI:
>     baseUrl: https://generativelanguage.googleapis.com/v1beta/openai
>   modelInfo:
>     functionCalling: true
>     jsonOutput: true
>     family: gemini-2.5-pro
> ```
> 
> Let me know if that works for you.

amazing! weve got it working based from this! we had to set the google api key and then store it as a secret in k8s, and the URL didnt work with openai on the end, but otherwise its all good! cheers for the help with this one 👍

Issue #468, opened 2025-06-02
[BUG] Missing line breaks in output (kagent install)


peterj:
Notice the missing line breaks in the output
```
kagent >> version
kagent version 0.3.11
git commit: ce4605f
build date: 2025-05-29
Warning: Could not fetch backend version
kagent >> install
███▒▒▒▒▒▒▒ Installing kagent-crds from oci://ghcr.io/kagent-dev/kagent/helm/Error installing kagent-crds: Error: Kubernetes cluster unreachable: Get "http://localhost:8080/version": dial tcp [::1]:8080: connect: connection refused

kagent >> install
█████▒▒▒▒▒ Installing kagent [OpenAI] Using oci://ghcr.io/kagent-dev/kagent/helm/:0.3.11kagent installed successfully
kagent >>  
```

marcinkubica:
Hi, seems stopping the spinner was the culprit
```
KAGENT_HELM_VERSION="0.3.12" ./go/bin/kagent-darwin-arm64 install
kagent installed successfully
```

PR #486, opened 2025-06-07
fix:(cli) line breaks in output (install)

marcinkubica:
fixes https://github.com/kagent-dev/kagent/issues/468


EItanya (approved):
Thanks so much for your first PR! Welcome to the community :)

Issue #489, opened 2025-06-07
[BUG] k8s secret for model is not deleted after model is deleted


marcinkubica:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

UI Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

k8s secret leftover after model is deleted

### 🔄 Steps To Reproduce

create `foobar` model from UI, check secrets
```
 k get -n kagent modelconfigs.kagent.dev
NAME                         PROVIDER   MODEL
default-model-config         OpenAI     gpt-4.1-mini
foobar-openai-gpt-4-1-mini   OpenAI     gpt-4.1-mini


❯ k get secrets -n kagent
NAME                                TYPE                 DATA   AGE
foobar-openai-gpt-4-1-mini          Opaque               1      42s
kagent-openai                       Opaque               1      22h
openai-claude-3-5-haiku-20241022    Opaque               1      13m
sh.helm.release.v1.kagent-crds.v8   helm.sh/release.v1   1      21h
sh.helm.release.v1.kagent-crds.v9   helm.sh/release.v1   1      21h
sh.helm.release.v1.kagent.v8        helm.sh/release.v1   1      21h
sh.helm.release.v1.kagent.v9        helm.sh/release.v1   1      21h
```

delete model from UI, check secrets

```
❯ k get -n kagent modelconfigs.kagent.dev
NAME                   PROVIDER   MODEL
default-model-config   OpenAI     gpt-4.1-mini

❯ k get secrets -n kagent
NAME                                TYPE                 DATA   AGE
foobar-openai-gpt-4-1-mini          Opaque               1      5m10s
kagent-openai                       Opaque               1      22h
openai-claude-3-5-haiku-20241022    Opaque               1      18m
sh.helm.release.v1.kagent-crds.v8   helm.sh/release.v1   1      21h
sh.helm.release.v1.kagent-crds.v9   helm.sh/release.v1   1      21h
sh.helm.release.v1.kagent.v8        helm.sh/release.v1   1      21h
[...]

PR #490, opened 2025-06-08
fix: delete associated k8s secret when model is deleted

abhinav-1305:
Resolves: #489 
- Added secret deletion logic in `HandleDeleteModelConfig`.
- Check for `non-Ollama` provider before attempting secret deletion (Since Ollama does not require secret deletion).

please review - @EItanya.


marcinkubica:
Shall we also start adding unit tests with the code? Or e2e, or both whichever is faster to add =)

abhinav-1305:
> Shall we also start adding unit tests with the code? Or e2e, or both whichever is faster to add =)

Yes, let’s start with unit tests first

EItanya:
I think there is a much easier way to do this. In the handler, simply create an `ownerReference` to the `ModelConfig`. That way we don't need any conditional logic, and the deletion will be handled by the ApiServer

go/controller/internal/httpserver/handlers/helpers.go
```diff
@@ -144,16 +146,40 @@ func FlattenStructToMap(data interface{}, targetMap map[string]interface{}) {
...
 	}
 
-	if err := kubeClient.Create(context.Background(), secret); err != nil {
+	if owner != nil {
+		secret.OwnerReferences = []metav1.OwnerReference{*owner}
+	}
+
+	if err := kubeClient.Create(ctx, secret); err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			// If the secret already exists, try to update it
```
EItanya, line 165: I'm actually not sure about this. In what case would the secret previously exist that we'd want to add the owner reference? If the secret previously exists, we by definition DO NOT own it.

go/controller/internal/httpserver/handlers/modelconfig.go
```diff
@@ -275,31 +254,43 @@ func (h *ModelConfigHandler) HandleCreateModelConfig(w ErrorResponseWriter, r *h
...
+		if k8serrors.IsAlreadyExists(err) {
+			w.RespondWithError(errors.NewConflictError("Model config API key secret already exists", err))
+			return
+		}
+		w.RespondWithError(errors.NewInternalServerError("Failed to create/update model config API key secret", err))
+		return
+	}
+	log.V(1).Info("Successfully created/updated model config API key secret")
+
+	log.Info("Model config created successfully")
```
EItanya, line 293: Why was this block moved out of the conditionals above? I believe they were there for a reason. @peterj may remember

EItanya (commented):
Overall I like the approach much better, just a couple of questions

go/controller/internal/httpserver/handlers/modelconfig.go
peterj: out of which conditionals?

sbx0r:
this one was moved to #594

Issue #488, opened 2025-06-07
[DOCS] kagent cli - get resource help not sufficient


marcinkubica:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Issue Description

when trying to use `kagent get` help doesn't provide much info of what can be done

```
kagent get
No resource type provided

❯ kagent get -h
Get a kagent resource

Usage:
  kagent get [flags]

Flags:
  -h, --help   help for get

Global Flags:
      --a2a-url string         A2A URL (default "http://localhost:8083/api/a2a")
      --api-url string         API URL (default "http://localhost:8081/api")
      --config string          config file (default is $HOME/.kagent/config.yaml) (default "/Users/marcin/.kagent/config.yaml")
  -n, --namespace string       Namespace (default "kagent")
  -o, --output-format string   Output format (default "table")
      --user-id string         User ID (default "admin@kagent.dev")
  -v, --verbose                Verbose output

❯ kagent get pods
Invalid resource type: pods

❯ kagent get crds
Invalid resource type: crds
```

### 🙋 Willing to Contribute?

- [x] I am willing to submit a PR for this documentation update

inFocus7:
I took a shot at this since I'm starting to delve into the project's codebase. I just saw you checked off the "willing to contribute" flag, so I can take mine down.

---

The issue is that the resource types for `kagent get` are defined and parsed in-line (within the run code). Because of this we'd either need to (1) create a custom help message (`cmd. SetHelpCommand`), or (2) define them as subcommands to the main `get` command. I believe the latter made more sense for this and would help with extensibility in the future if we ever add more resource types to `get`. 

I also added a personal note on how the cobra CLI portion of the code could possibly be improved, but that may work better as a separate issue / PR in case there are other commands that could benefit from a restructure.

marcinkubica:
@inFocus7 go take a shot! I'm marking most of my issues with willing to contribute anyway :D

I guess I was thinking to be waiting for project maintainers to maybe start moving issues into "todo" on the [board](https://github.com/orgs/kagent-dev/projects/3) as so not to jump into solving problems which "aren't there" ;)

PR #493, opened 2025-06-08
[CLI] Use subcommands for `kagent get`

inFocus7:
# Context

The `kagent get --help` message is not descriptive enough to understand what resource types can be gotten.

# Changes

Restructuring the `kagent get` cobra command to set up the 'resource type' as subcommands instead of internally-defined switch-case.

This:
1. Allows for a more descriptive `help` (automatically).
2. Makes it easier to extend this (ifever done), where added resource types would be added as subcommands and automatically be referenced in the `help`.
3. Allows the `get` and the subsequent resource type to have a single responsibility.

```
$ go run go/cli/cmd/kagent/main.go get --help

Get a kagent resource

Usage:
  kagent get [flags]
  kagent get [command]

Available Commands:
  agent       Get an agent or list all agents
  run         Get a run or list all runs
  session     Get a session or list all sessions
  tool        Get tools

Flags:
  -h, --help   help for get

Global Flags:
      --a2a-url string         A2A URL (default "http://localhost:8083/api/a2a")
      --api-url string         API URL (default "http://localhost:8081/api")
      --config string          config file (default is $HOME/.kagent/config.yaml) (default "/Users/fabiangonz98/.kagent/config.yaml")
  -n, --namespace string       Namespace (default "kagent")
  -o, --output-format string   Output format (default "table")
      --user-id string         User ID (default "admin@kagent.dev")
  -v, --verbose                Verbose output

Use "kagent get [command] --help" for more information about a command.
```

# Personal Thoughts

It _may_ be worth restructuring the cobra cli setup. For instance, setting up `get` as its own package with subcommands defined in it.

The below example structure, `get` and its subcommand would be defined outside of `main.go` and imported. That way the `main.go` file doesn't grow over time and the cli portion of this would hopefully become more easily understandable / maintainable.

```
/cli
  /cmd
    /kagent
      main.go
      /get
        get.go
        /session
        /tool
```

This structure also make it easier to write tests in the future (if wanted), so we can test commands and their expected outputs/actions.

Resolves: https://github.com/kagent-dev/kagent/issues/488

### DIFF
--- a/go/cli/cmd/kagent/main.go
+++ b/go/cli/cmd/kagent/main.go
@@ -142,7 +142,7 @@ func main() {
 
 	getSessionCmd := &cobra.Command{
 		Use:   "session [session_id]",
-		Short: "Get a session",
+		Short: "Get a session or list all sessions",
 		Long:  `Get a session by ID or list all sessions`,
 		Run: func(cmd *cobra.Command, args []string) {
 			client := autogen_client.New(cfg.APIURL)
@@ -160,7 +160,7 @@ func main() {
 
 	getRunCmd := &cobra.Command{
 		Use:   "run [run_id]",
-		Short: "Get a run",
+		Short: "Get a run or list all runs",
 		Long:  `Get a run by ID or list all runs`,
 		Run: func(cmd *cobra.Command, args []string) {
 			client := autogen_client.New(cfg.APIURL)
@@ -178,7 +178,7 @@ func main() {
 
 	getAgentCmd := &cobra.Command{
 		Use:   "agent [agent_name]",
-		Short: "Get an agent",
+		Short: "Get an agent or list all agents",
 		Long:  `Get an agent by name or list all agents`,
 		Run: func(cmd *cobra.Command, args []string) {
 			client := autogen_client.New(cfg.APIURL)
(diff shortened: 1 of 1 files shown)


EItanya (commented):
This is definitely a much better way to structure this code than what I did 😆. Thanks so much for the contribution!

inFocus7:
@EItanya I also noticed that in the non-interactive setup we define `a2aCmd`, but don't add it as a subcommand it to the root command. This means it's not accessible as `kagent a2a`. It's in the interactive mode though. 

Should/could I add it to the root command as part of this PR? Or is this expected behavior 👀 

---

```
➜  go git:(488-descriptive-help-for-get-cmd) ✗ go run cli/cmd/kagent/main.go a2a --help
Error: unknown command "a2a" for "kagent"
Run 'kagent --help' for usage.
exit status 1
```

EItanya:
> @EItanya I also noticed that in the non-interactive setup we define `a2aCmd`, but don't add it as a subcommand it to the root command. This means it's not accessible as `kagent a2a`. It's in the interactive mode though.
> 
> Should/could I add it to the root command as part of this PR? Or is this expected behavior 👀
> 
> ```
> ➜  go git:(488-descriptive-help-for-get-cmd) ✗ go run cli/cmd/kagent/main.go a2a --help
> Error: unknown command "a2a" for "kagent"
> Run 'kagent --help' for usage.
> exit status 1
> ```

I didn't port it over yet because A2A is more of an implementation detail in my opinion than an operation. The CLI has an `invoke` command which I think is enough for now. If we decide differently we can add it later.

PR #501, opened 2025-06-10
feat (all): add streamable HTTP MCP backend

### LINKED ISSUE #478: [FEATURE] Implement "Streamable HTTP" transport for MCP
### 📋 Prerequisites

### 📝 Feature Summary

Streamable HTTP is supposed to replace HTTP+SSE as transport.

The HTTP+SSE transport is deprecated[^1] and Streamable HTTP[^2] is its replacement.

[^1]: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#backwards-compatibility
[^2]: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

### 💡 Proposed Solution

Implement "Streamable HTTP" according to the specification[^2].

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

Not Sure

### 📚 Additional Context

_No response_

EItanya:
This PR adds support for streamable HTTP MCP backends. In addition it should fix timeout issues which users have been having with stdio tools. Currently there is a 5s timeout on the discovery read. When using `npx` or `uvx` tools this will often fail because it needs to download all of the dependencies in the new container. This will now be configurable, but it will default to `10s`.

resolves #478


go/controller/api/v1alpha1/toolserver_types.go
```diff
@@ -65,16 +66,30 @@ type StdioMcpServerConfig struct {
 	Args    []string          `json:"args,omitempty"`
 	Env     map[string]string `json:"env,omitempty"`
 	EnvFrom []ValueRef        `json:"envFrom,omitempty"`
+	// Default value is 10 seconds
+	// +kubebuilder:default:=10
+	ReadTimeoutSeconds uint8 `json:"readTimeoutSeconds,omitempty"`
```
abhinav-1305, line 71: Should this field be required rather than defaulted to make the config more explicit?

peterj (approved):

go/controller/api/v1alpha1/toolserver_types.go
EItanya: I don't think so, it's more of an implementation detail than anything. I just wanted to make it available just in case

peterj (approved):

abhinav-1305 (approved):
LGTM

PR #516, opened 2025-06-12
Remove stored team from session to allow for dynamic conversations

EItanya:


python/src/kagent/conditions/_termination.py
```diff
@@ -0,0 +1,62 @@
...
+from autogen_agentchat.messages import BaseAgentEvent, BaseChatMessage, StopMessage, TextMessage
+from autogen_core import Component
+from pydantic import BaseModel
+from typing_extensions import Self
+
+
+class FinalTextMessageTerminationConfig(BaseModel):
+    """Configuration for the FinalTextMessageTermination termination condition."""
+
+    source: str | None = None
```
sbx0r, line 13: why not using `Field()` from pydantic?

python/src/kagent/conditions/_termination.py
```diff
@@ -0,0 +1,62 @@
...
+    """The source of the text message to terminate the conversation."""
+
+
+class FinalTextMessageTermination(TerminationCondition, Component[FinalTextMessageTerminationConfig]):
```
sbx0r, line 17: is it used anywhere?

python/src/kagent/conditions/_termination.py
EItanya: No good reason, this is most copy-pasted from https://github.com/microsoft/autogen/blob/150ea0192d2c68d2fe08616804dabc58a6691074/python/packages/autogen-agentchat/src/autogen_agentchat/conditions/_terminations.py#L506-L510

python/src/kagent/conditions/_termination.py
EItanya: Yes, but not directly, it uses the component load method

https://github.com/kagent-dev/kagent/blob/eitanya/dynamic-session-teams/go/controller/internal/autogen/autogen_api_translator.go#L492-L496

https://github.com/kagent-dev/kagent/blob/eitanya/dynamic-session-teams/go/controller/internal/autogen/autogen_api_translator.go#L840-L849

peterj:
I am taking this on now to do the UI-side changes.

python/src/kagent/conditions/_termination.py
sbx0r: Sure thing. Your code is perfectly fine as-is. I only mentioned Field(...) because it makes Pydantic pick up the comment when it builds docs and schemas. If you don’t need that, please ignore my question :)

Makefile
```diff
@@ -234,6 +234,7 @@ helm-install-provider: helm-version check-openai-key
 		--history-max 2    \
 		--timeout 5m       \
 		--wait \
+		--set service.type=LoadBalancer \
```
onematchfox, line 237: @EItanya Was this change intentional? Not exactly sure how it relates to this PR, and it has the unfortunate side effect of breaking this target when using kind (as per the docs). Alternatively, perhaps some other doc updates were missed to [enable use of load balancers](https://kind.sigs.k8s.io/docs/user/loadbalancer/) services within kind?

EItanya: This was intentional, I use `metallb` locally and in CI, but definitely open to making it configurable

onematchfox: For posterity/reference - conversation continued over on #678.

Issue #524, opened 2025-06-13
[FEATURE] [UI] Improve namespace handling when creating kagent's resources


sbx0r:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### ❓ Problem Statement / Motivation

When creating a new `Agent`, `Memory`, etc., users must provide a namespace, but the system doesn't handle non-existent namespaces gracefully. If a user enters a namespace that doesn't exist (e.g., "blabbberjnbdf"), the system doesn't create it automatically, leading to potential errors or confusion.

**Current behavior:**
- User must manually enter a namespace name
- System doesn't validate if namespace exists
- No automatic namespace creation for new namespace names

### 💡 Proposed Solution

We need to decide on the approach for namespace handling and implement a consistent solution.

### 🔄 Alternatives Considered

- **Dropdown with existing namespaces only**
    Replace text input with dropdown showing all available namespaces
Users can only select from existing namespaces
Requires separate process for namespace creation
- **Auto-create namespaces**
    Keep text input but automatically create the namespace if it doesn't exist
Show confirmation/feedback when new namespace is created
Allows for dynamic namespace creation
- **Hybrid approach**
    Dropdown with existing namespaces + "Create new" option
When "Create new" is selected, show text input for new namespace name

### 🎯 Affected Service(s)

UI Service

### 📚 Acceptance criteria:

- [ ] Decision made on which approach to implement
- [ ] Users cannot create agents with invalid/non-existent namespaces
- [ ] Solution handles all kagent's resources (`Agent`, `ModelConfig`, `Memory`, `ToolServer`)
- [ ] Clear user feedback for namespace-related actions
- [ ] Input validation prevents invalid namespace names
- [ ] Error handling for namespace-related failures

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

inFocus7:
If sbx0r doesn't mind and this gets greenlit, I'd like to take a crack at this 👀 

---

I really like the **Hybrid Approach**. The resource creation process would heavily benefit from a dropdown-style choice for namespaces. If we can create new namespaces through the UI that would be cool too, although I'm not sure how production-infra/security folk would feel about the UI having the ability to create new namespace(s).

[...]

PR #540, opened 2025-06-18
(feat) implement namespace dropdown selection ux when creating resources

inFocus7:
# Changes

- Add a namespace handler to the server with ability to list watched namespaces.
    - Added tests.
- Add new component for namespace dropdown, which lists namespaces returned from the handler.

# Reproduction

Can reproduce by spinning up kagent and creating a new resource, seeing the new dropdown for namespaces.

```sh
export OPENAI_API_KEY="your-key"

make create-kind-cluster

# Try with some watchNamespaces set, or don't set to list all namespaces
HELM_EXTRA_ARGS="--set controller.watchNamespaces[0]=default --set controller.watchNamespaces[1]=test" make helm-install

kubectl -n kagent port-forward service/kagent 8001:80 &

# open below and any other creation page to try new dropdown.
# you'll only see the namespaces set in watchNamespaces that _also_ exist in the cluster.
# if no watchNamespaces set, then all namespaces will be returned.
open http://localhost:8001/models/new
```

To test the onboarding, on Chrome-based systems,
1. Open kagent page
1. \[Right Click\] > Developer Tools > Inspect
2. Application tab
3. Storage > Local Storage
4. Delete `kagent-onboarding` key
5. Reload kagent page

---

Resolves: https://github.com/kagent-dev/kagent/issues/524

### DIFF
--- a/ui/src/components/AddServerDialog.tsx
+++ b/ui/src/components/AddServerDialog.tsx
@@ -434,12 +434,12 @@ export function AddServerDialog({ open, onOpenChange, onAddServer, onError }: Ad
 
   return (
     <Dialog open={open} onOpenChange={handleClose}>
-      <DialogContent className="sm:max-w-2xl">
-        <DialogHeader>
+      <DialogContent className="sm:max-w-2xl flex flex-col max-h-[90vh]">
+        <DialogHeader className="px-6 pt-6 pb-2 border-b flex-shrink-0">
           <DialogTitle>Add Tool Server</DialogTitle>
         </DialogHeader>
 
-        <div className="py-4">
+        <div className="flex-1 overflow-y-auto px-6">
           {error && (
             <div className="mb-4 p-3 bg-red-50 border border-red-200 text-red-700 rounded-md text-sm flex items-start">
               <AlertCircle className="h-5 w-5 mr-2 mt-0.5 flex-shrink-0" />
@@ -664,7 +664,7 @@ export function AddServerDialog({ open, onOpenChange, onAddServer, onError }: Ad
           </div>
         </div>
 
-        <DialogFooter>
+        <DialogFooter className="px-6 py-4 border-t flex-shrink-0">
           <Button variant="outline" onClick={handleClose} disabled={isSubmitting}>
             Cancel
           </Button>


ui/src/components/onboarding/steps/AgentSetupStep.tsx
```diff
@@ -81,7 +82,11 @@ export function AgentSetupStep({ initialData, onNext, onBack }: AgentSetupStepPr
                             <FormItem>
                                 <FormLabel>Agent Namespace</FormLabel>
                                 <FormControl>
-                                    <Input {...field} />
+                                    <NamespaceCombobox
+                                        value={field.value || ""}
+                                        onValueChange={field.onChange}
+                                        placeholder="Select namespace..."
```
sbx0r: if this is component's default then why to set it here?

go/controller/cmd/main.go
```diff
@@ -215,6 +215,9 @@ func main() {
 		})
 	}
 
+	// filter out invalid namespaces from the watchNamespaces flag (comma separated list)
+	watchNamespacesList := filterValidNamespaces(strings.Split(watchNamespaces, ","))
```
inFocus7, line 219: separated this logic out since the validated namespaces would be used for both the cache option + passing to the namespace handler (where we list namespaces from)

ui/src/components/onboarding/steps/AgentSetupStep.tsx
inFocus7: Yeah, good call, I updated it. I was moving things around and missed some cleaning up 😅

sbx0r:
sorry @inFocus7. It looks like I forgot to respond to your last question:

> todo:
> ask if we want the dropdown to be the source-of-truth or if we want to allow for custom input
this would be for case there was a hiccup fetching namespaces, so users can can still input a namespace name and create a resource (assuming creation doesn't fail).
i.e. maybe get namespaces has issues but create resource in namespace doesn't).

I think the UI should rely on the controller's response, if fetching namespaces fails, show the error to the user.

EItanya (approved):
These go changes LGTM

peterj (approved):
LGTM!

inFocus7:
E2e failed when installing kagent. Looks like a change happened to the github action somewhere (haven't looked at it) and now the openai api key isn't set. I don't think my changes should have affected that. I can look into it later in the day 👀

`make: *** [Makefile:86: check-openai-key] Error 1`

inFocus7:
Maybe `secrets.OPENAI_API_KEY` is no longer set for the org/project?

https://github.com/kagent-dev/kagent/blob/e912c32ed9424f51aec33cc38d0d03524654cd27/.github/workflows/test.yaml#L27C29-L27C51

EItanya:
No, secrets don't get set for non org members, it's something we need to figure out :(

It will run on merge to main

sbx0r:
Great work @inFocus7 !

Issue #499, opened 2025-06-10
[FEATURE] Move built-in tools into an MCP server


EItanya:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Rather than having built-in tools be coded directly into the autogen framework, we should instead ship them via an MCP server.

### ❓ Problem Statement / Motivation

Currently `kagent` ships with a set of "builtin" tools which are coded directly into the python/autogen backend of the application. This was very nice when we first started the project as it made debugging very easy and allowed for very quick iteration. However, now its downsides are becoming more apparent.

1. Kuberentes Python tooling is not nearly as mature as `golang`
2. The tools are not easily sharable. These tools may be useful in contexts other than `kagent`, but because they're tied to tightly to `Autogen` we can't offer them elsewhere as MCP servers.

### 💡 Proposed Solution

I propose that we move the built-in tools into `golang` MCP servers which can be run in multiple ways depending on the use-case. When running within kagent, we can run them directly from the controller for the simplest possible UX. We can also ship binaries and docker containers for other projects which wish to consume them separately.

### 🎯 Affected Service(s)

Multiple services / System-wide

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

peterj:
Do they have to be `golang` servers? Should we consider doing "auto-discovery" of the mcp servers that get deployed or how do you see the "run them directly from the controller" would look like?

dimetron:
@peterj - for the start, Can we use existing ToolServerSpec CRD and Migrate all to a separate tools container ? I think later we will need some way to template those local tools server to support ArgoCD / FluxCD / Simple pod template

peterj:
Yes, I think we should use the existing CRD; but have the tools in a separate container as you say.

So we'd have a ToolServer for "built-in" tools that gets automatically deploed when one deploys kagent.

sbx0r:
I think we can close it :)

PR #552, opened 2025-06-23
Feature/move built in tools go

dimetron:
tools migration go https://github.com/kagent-dev/kagent/issues/499

## Tools Migrated to Go

- Go tools under go/tools

- Removed python tools

- contrib folder added with doc2vec Dockerfile and exposed as remote MCP

### Main challenges 

- naming convention of remote MCP is different

- DocQuery doc2vec MCP migrated to a separate container for now @djannot please have alook

- With toolList under mcpServer, there is no way to add config per tool like we had before for grafana prometheus


djannot:
@dimetron I'm not sure it's a good idea to copy the code from https://github.com/kagent-dev/doc2vec/tree/main/mcp into this repo. It would force us to maintain the code in 2 different locations.

helm/kagent/templates/toolserver-docs.yaml
```diff
@@ -0,0 +1,16 @@
+{{ if eq .Values.providers.default "openAI"}}
+apiVersion: kagent.dev/v1alpha1
+kind: ToolServer
+metadata:
+  name: {{ include "kagent.fullname" . }}-querydoc
+  namespace: {{ include "kagent.namespace" . }}
+spec:
+  config:
+    streamableHttp:
+{{/*      headers:*/}}
```
peterj: remove if it's commented out

dimetron:
> @dimetron I'm not sure it's a good idea to copy the code from https://github.com/kagent-dev/doc2vec/tree/main/mcp into this repo. It would force us to maintain the code in 2 different locations.

@djannot 
I agree, I've tried to move this part as well to Go, however there are some additional dependencies and I'd like to do migration of doc2vec to Go as part of separate task.
Also this would be good to sync with you, to have best solution and cover all use cases.

### Doc2Vec migration should be a separate task:

- 1. We separated image and reusing your code + some changes to make it run standalone
- 2. Move it to Go () with different backend plugins: sqlite, pgVec, milvus  etc. 
- 3. Make sure DB indexes are available and versioned with in sync to used versions of tools
- 4. Make sure we have fast cache of all HTML docs and re-index is fast
- 5. Add support of currently missing Ollama and Anthropic embeddings etc.

### Persistance in general

While selecting persistence backend should consider storage for KAgent app
- current runtime teams / agent configuration 
- chat sessions storage

Different backend providers have different HA and backup capabilities
- support vector storage
- HA multinode and replications

Autogen supports below DB options:

https://microsoft.github.io/autogen/stable//user-guide/autogenstudio-user-guide/faq.html#q-how-do-i-use-autogen-studio-with-a-different-database

- SQLite: sqlite:///database.sqlite
- PostgreSQL: postgresql+psycopg://user:password@localhost/dbname
- MySQL: mysql+pymysql://user:password@localhost/dbname
- AzureSQL: mssql+pyodbc:///?odbc_connect=DRIVER%3D%7BODBC+Driver+17+for+SQL+Server%7D%3BSERVER%3Dtcp%3Aservername.database.windows.net%2C1433%3BDATABASE%3Ddatabasename%3BUID%3Dusername%3BPWD%3Dpassword123%3BEncrypt%3Dyes%3BTrustServerCertificate%3Dno%3BConnection+Timeout%3D30%3B

pgVec has wide support and providers and is a good candidate for app DB and storage 
- https://neon.com/docs/extensions/pgvector
- https://supabase.com/docs/guides/ai
- https://cloudnative-pg.io

helm/kagent/templates/toolserver.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ToolServer
+metadata:
+  name: {{ include "kagent.fullname" . }}-tool-server
+  namespace: {{ include "kagent.namespace" . }}
+spec:
+  config:
+    sse:
+      sseReadTimeout: 5m0s
+      timeout: 30s
+      url: http://localhost:8084/sse
```
peterj, line 11: could we use the actual k8s service name here instead of localhost?

python/Dockerfile
```diff
@@ -101,4 +64,4 @@ LABEL org.opencontainers.image.description="Kagent app is the apiserver for runn
 LABEL org.opencontainers.image.authors="Kagent Creators 🤖"
 LABEL org.opencontainers.image.version="$VERSION"
 
-CMD ["uv", "run", "kagent-engine", "serve", "--host", "0.0.0.0", "--port", "8081"]
```
peterj, line 104: what was "serve" here removed?

python/src/kagent/cli.py
```diff
@@ -1,283 +1,9 @@
...
-def helm():
-    mcp.add_tool(helm_list_releases._func, helm_list_releases.name, helm_list_releases.description)
-    mcp.add_tool(helm_get_release._func, helm_get_release.name, helm_get_release.description)
-    mcp.add_tool(helm_uninstall._func, helm_uninstall.name, helm_uninstall.description)
-    mcp.add_tool(upgrade_release._func, upgrade_release.name, upgrade_release.description)
-    mcp.add_tool(helm_repo_add._func, helm_repo_add.name, helm_repo_add.description)
-    mcp.add_tool(helm_repo_update._func, helm_repo_update.name, helm_repo_update.description)
-    mcp.run()
-
+mcp = FastMCP("MCP Tools")
```
peterj: perhaps rename to "kagent" as it's not about MCP tools anymore.

python/Dockerfile
dimetron: since it's the only left command it was removed

helm/kagent/templates/toolserver.yaml
dimetron: for that we will need to move those containers in pods and make sure we do JWT tokens between etc

dimetron: For the moment localhost is also way to avoid security issues with unprotected MCP

djannot:
> > @dimetron I'm not sure it's a good idea to copy the code from https://github.com/kagent-dev/doc2vec/tree/main/mcp into this repo. It would force us to maintain the code in 2 different locations.
> 
> @djannot I agree, I've tried to move this part as well to Go, however there are some additional dependencies and I'd like to do migration of doc2vec to Go as part of separate task. Also this would be good to sync with you, to have best solution and cover all use cases.
> 
> ### Doc2Vec migration should be a separate task:
> * 1. We separated image and reusing your code + some changes to make it run standalone
> * 2. Move it to Go () with different backend plugins: sqlite, pgVec, milvus  etc.
> * 3. Make sure DB indexes are available and versioned with in sync to used versions of tools
> * 4. Make sure we have fast cache of all HTML docs and re-index is fast
> * 5. Add support of currently missing Ollama and Anthropic embeddings etc.
> 
> ### Persistance in general
> While selecting persistence backend should consider storage for KAgent app
> 
> * current runtime teams / agent configuration
> * chat sessions storage
> 
> Different backend providers have different HA and backup capabilities
> 
> * support vector storage
> * HA multinode and replications

Not sure there's real value in writing doc2vec in Go. If the goal is to support new backends, this is something which can be done in typescript.

dimetron:
> @dimetron I'm not sure it's a good idea to copy the code from https://github.com/kagent-dev/doc2vec/tree/main/mcp into this repo. It would force us to maintain the code in 2 different locations.

Let's discuss. 

- Option1 we keep https://github.com/kagent-dev/doc2vec/ as doc index pipeline and release. While we can have MCP image released in Kagent in one release cycle 

- Option2 we will release separate image in https://github.com/kagent-dev/doc2vec/ and only pin the version in Kagent. 


@EItanya - any suggestions ?

Goals: 
- I'd like to decouple DB index and MCP which is using DB
- doc2vec is good as preloaded Knowledge Base
- however it is very static and not allowing re-index or upset index
- Add support for more LLM embeddings types and storages
- Add dynamic update in vector DB

EItanya:
> > @dimetron I'm not sure it's a good idea to copy the code from https://github.com/kagent-dev/doc2vec/tree/main/mcp into this repo. It would force us to maintain the code in 2 different locations.
> 
> Let's discuss.
> 
> * Option1 we keep https://github.com/kagent-dev/doc2vec/ as doc index pipeline and release. While we can have MCP image released in Kagent in one release cycle
> * Option2 we will release separate image in https://github.com/kagent-dev/doc2vec/ and only pin the version in Kagent.
> 
> @EItanya - any suggestions ?
> 
> Goals:
> 
> * I'd like to decouple DB index and MCP which is using DB
> * doc2vec is good as preloaded Knowledge Base
> * however it is very static and not allowing re-index or upset index
> * Add support for more LLM embeddings types and storages
> * Add dynamic update in vector DB

Personally I prefer option 1, we can very easily add a released version from `doc2vec` to avoid duplicating code. We can also release to NPM in order to just pull the code released code instead.

@djannot, do you think you could handle releasing the code, if that doesn't already happen?

Those goals make sense to me, but I think we need a larger issue to discuss further features for the docs rather than try to squeeze all of that into this PR. For this PR specifically let's do what I mentioned above.

contrib/tools/githubcopilot.mcp.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: kagent.dev/v1alpha1
+kind: ToolServer
+metadata:
+  name: api.githubcopilot.com
+  namespace: kagent
+spec:
+  config:
+    streamableHttp:
+      headers:
+        Authorization: "Bearer TOKEN"
```
EItanya: Rather than requiring the user to change this, can we instead use `fromSecret` added by @nimishamehta5.

contrib/tools/server-sequential-thinking.mcp.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kagent.dev/v1alpha1
+kind: ToolServer
+metadata:
+  name: api.githubcopilot.com
```
EItanya: This is the same as the copilot toolserver?

contrib/tools/githubcopilot.mcp.yaml
```diff
@@ -0,0 +1,17 @@
+# Official MCP Server for GitHub Copilot
+# https://github.com/github/github-mcp-server
+apiVersion: kagent.dev/v1alpha1
+kind: ToolServer
+metadata:
+  name: api.githubcopilot.com
+  namespace: kagent
```
EItanya: Should we omit this so that it can be deployed anywhere?

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -545,6 +547,10 @@ func (a *apiTranslator) translateAssistantAgent(
 			}
 			tools = append(tools, autogenTool)
 		case tool.McpServer != nil:
+			if toolNeedsOpenaiApiKey(tool.McpServer.ToolServer) && modelConfig.Spec.Provider != v1alpha1.OpenAI {
+				log.V(1).Info("Skipping tool that requires OpenAI API key", "tool", tool.McpServer.ToolServer, "modelProvider", modelConfig.Spec.Provider)
+				continue
+			}
```
EItanya: Why are we skipping in this case? The user just needs to make sure to pre-configure the OPENAI key, but I don't think we should skip completely. I may also be misunderstanding.

go/controller/internal/controller/toolserver_controller.go
```diff
@@ -46,7 +46,6 @@ func (r *ToolServerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	return ctrl.Result{
 		// loop forever because we need to refresh tools server status
-		Requeue:      true,
 		RequeueAfter: 60 * time.Second,
```
EItanya, line 49: Why was this removed?

go/tools/cmd/tools-client/main.go
EItanya, line 1: Should this be deleted?

go/tools/cmd/tool-server/main.go
```diff
@@ -0,0 +1,97 @@
...
+var serverInstance *server.SSEServer
+var e error
+
+func main() {
+	// Initialize structured logging
+	logger.Init()
+	log := logger.Get()
+	defer logger.Sync()
+
+	port := os.Getenv("KAGENT_TOOLS_PORT")
```
EItanya: Can we use flags for this, we typically use flags for this sort of config time thing in go.

EItanya (commented):
So the actual implementation of these tools look mostly good to me, but I have a couple of important high level questions which I think we need to answer before I finish this review.

1. Do we want users to be able to run individual servers, for example just the k8s server, or the istio server. (I would argue yes because part of the goal here is to allow people to run these MCP servers outside of the context of `kagent`
2. Do we want to allow users to run these servers via `stdio` as well as over HTTP. (I would argue yes for the same reason as above. It expands the usefulness of these servers, and could allow them to also be used in local envs and expand the reach of these servers.)
3. Do we want to run an entirely separate container for the toolserver by default. (I would argue no. Currently the deployment method is as small and simple as possible, skin to the all-in-one jaeger deployment, and I think we should keep it that way. When we split up the pods for prod readiness this could get its own pod, but I don't think we should start there)

Thanks so much for all of the hard work, this is looking awesome :).

Most of my comments shouldn't require too many changes, just some overall directional things.

go/controller/internal/autogen/autogen_api_translator.go
dimetron: This condition is disabling QueryDoc tool if not openAI

1. Because doc2vec mcp fail to start without OPENAI key
2. With Ollama model -> tool is not discovered and Agents will never be visible

This is why we gracefully exclude tools in no openAI available.
Lets also think about embeddings as a separate model type.

go/controller/internal/controller/toolserver_controller.go
dimetron: Deprecated: Use `RequeueAfter` instead.

go/tools/cmd/tool-server/main.go
dimetron: tool server was initially MCP server with stdio 
and adding flags could break MCP protocol. 

since now it is http we can move it to args

dimetron:
> So the actual implementation of these tools look mostly good to me, but I have a couple of important high level questions which I think we need to answer before I finish this review.
> 
> 1. Do we want users to be able to run individual servers, for example just the k8s server, or the istio server. (I would argue yes because part of the goal here is to allow people to run these MCP servers outside of the context of `kagent`

- [x] this could be added via flags - one binary and several MCP

> 2. Do we want to allow users to run these servers via `stdio` as well as over HTTP. (I would argue yes for the same reason as above. It expands the usefulness of these servers, and could allow them to also be used in local envs and expand the reach of these servers.)

- [x] this could be added via flags - good for testing and integration into Cursor IDE

> 3. Do we want to run an entirely separate container/pod for the toolserver by default. (I would argue no. Currently the deployment method is as small and simple as possible, skin to the all-in-one jaeger deployment, and I think we should keep it that way. When we split up the pods for prod readiness this could get its own pod, but I don't think we should start there)

- [x] Agree, we are not ready to run MCP without JWT security as a separate POD
       after we will have mechanizm to to do it secure way (KGateway, MCP Gateway, Mesh),
       till then we should keep it local container with localhost networking. 

> 
> Thanks so much for all of the hard work, this is looking awesome :).
> 
> Most of my comments shouldn't require too many changes, just some overall directional things.

dimetron:
> > > @dimetron I'm not sure it's a good idea to copy the code from https://github.com/kagent-dev/doc2vec/tree/main/mcp into this repo. It would force us to maintain the code in 2 different locations.
> > 
> > 
> > @djannot I agree, I've tried to move this part as well to Go, however there are some additional dependencies and I'd like to do migration of doc2vec to Go as part of separate task. Also this would be good to sync with you, to have best solution and cover all use cases.
> > ### Doc2Vec migration should be a separate task:
> > 
> > * 1. We separated image and reusing your code + some changes to make it run standalone
> > * 2. Move it to Go () with different backend plugins: sqlite, pgVec, milvus  etc.
> > * 3. Make sure DB indexes are available and versioned with in sync to used versions of tools
> > * 4. Make sure we have fast cache of all HTML docs and re-index is fast
> > * 5. Add support of currently missing Ollama and Anthropic embeddings etc.
> > 
> > ### Persistance in general
> > While selecting persistence backend should consider storage for KAgent app
> > 
> > * current runtime teams / agent configuration
> > * chat sessions storage
> > 
> > Different backend providers have different HA and backup capabilities
> > 
> > * support vector storage
> > * HA multinode and replications
> 
> Not sure there's real value in writing doc2vec in Go. If the goal is to support new backends, this is something which can be done in typescript.

It's not typescript vs Go problem - is how to make images clean and maintainable overtime

grype docker:ghcr.io/kagent-dev/kagent/tools-querydoc:v0.3.17-13-gf4acb08       -o template -t reports/cve-report.tmpl --file reports//doc2vec-cve.csv
 ✔ Loaded image                                                                                                                                                                                                                                                            ghcr.io/kagent-dev/kagent/tools-querydoc:v0.3.17-13-gf4acb08 
 ✔ Parsed image                                                                                                                                                                                                                                                 sha256:7198b20da87dc9b455452ddcffab1a42b94a8ec617abee474d2b590d9721655d 
 ✔ Cataloged contents                                                                                                                                                                                                                                                  52657f03bd8c4e11f773bb8e8f36c208b0014e7f039b080f797cfc1da2688544 
   ├── ✔ Packages                        [458 packages]  
   ├── ✔ Executables                     [696 executables]  
   ├── ✔ File metadata                   [3,354 locations]  
   └── ✔ File digests                    [3,354 files]  
 ✔ Scanned for vulnerabilities     [74 vulnerability matches]  
   ├── by severity: 0 critical, 8 high, 12 medium, 8 low, 46 negligible
   └── by status:   2 fixed, 72 not-fixed, 0 ignored (1 dropped)

"perl-base","5.36.0-7+deb12u2","CVE-2023-31484","High","/var/lib/dpkg/status"
"cross-spawn","7.0.3","GHSA-3xgq-45jj-v275","High","/usr/local/lib/node_modules/npm/node_modules/cross-spawn/package.json"
"libpam-modules","1.5.2-6+deb12u1","CVE-2025-6020","High","/var/lib/dpkg/status"
"libpam-modules-bin","1.5.2-6+deb12u1","CVE-2025-6020","High","/var/lib/dpkg/status"
"libpam-runtime","1.5.2-6+deb12u1","CVE-2025-6020","High","/var/lib/dpkg/status"
"libpam0g","1.5.2-6+deb12u1","CVE-2025-6020","High","/var/lib/dpkg/status"
"libc-bin","2.36-9+deb12u10","CVE-2025-4802","High","/var/lib/dpkg/status"
"libc6","2.36-9+deb12u10","CVE-2025-4802","High","/var/lib/dpkg/status"

go/tools/internal/k8s/k8s.go
```diff
@@ -0,0 +1,890 @@
...
+	args := []string{"describe", resourceType, resourceName}
+	if namespace != "" {
+		args = append(args, "-n", namespace)
+	}
+
+	return k.runKubectlCommand(ctx, args)
+}
+
+// Fallback to kubectl command for logs operations
+func (k *K8sTool) handleKubectlLogsTool(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
```
EItanya: Unused

EItanya:
As mentioned on Discord, I think we should switch the names of the tools to be more consistent. 

For example, a current tool name is `k8s_GetPodLogs`. This reads a bit weird, can we instead change the names to: `k8s_get_pod_logs`.

Most future changes are easy to make in a non-user breaking way, but this one we should try hard not to change in the future.

dimetron:
> As mentioned on Discord, I think we should switch the names of the tools to be more consistent.
> 
> For example, a current tool name is `k8s_GetPodLogs`. This reads a bit weird, can we instead change the names to: `k8s_get_pod_logs`.
> 
> Most future changes are easy to make in a non-user breaking way, but this one we should try hard not to change in the future.

Done

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -27,7 +27,8 @@ var (
 		"kagent.tools.k8s.GenerateResourceTool",
 	}
 	toolsProvidersRequiringOpenaiApiKey = []string{
-		"kagent.tools.docs.QueryTool",
+		"query_Docs",
+		"kagent/kagent-querydoc",
```
sbx0r: is that a ref string?

go/tools/Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+COPY --from=tools --chown=65532:65532 /downloads/helm                  /bin/helm
+COPY --from=tools --chown=65532:65532 /downloads/kubectl-argo-rollouts /bin/kubectl-argo-rollouts
+# Copy the tool-server binary
+COPY --from=builder --chown=65532:65532 /workspace/tool-server           /tool-server
+
+ARG VERSION
+
+LABEL org.opencontainers.image.source=https://github.com/kagent-dev/kagent
+LABEL org.opencontainers.image.description="Kagent app is the UI and apiserver for running agents."
+LABEL org.opencontainers.image.authors="Kagent Creators 🤖"
```
sbx0r, line 95: are these LABELs valid for this image?

helm/agents/cilium-policy/templates/agent.yaml
```diff
@@ -449,14 +449,18 @@ spec:
...
-    - type: Builtin
-      builtin:
-        name: kagent.tools.cilium.DisplayPolicyNodeInformation
-    - type: Builtin
-      builtin:
-        name: kagent.tools.cilium.ValidateCiliumNetworkPolicies
-
+    - type: McpServer
+      mcpServer:
+        toolServer: kagent/kagent-querydoc
```
sbx0r: It shouldn't be hardcoded.
If someone will install this MCP Server to different namespace, it wouldn't work.

Following field should be configurable and set by the parent chart or the user during installation.

helm/agents/helm/templates/agent.yaml
```diff
@@ -77,7 +77,7 @@ spec:
     - `ApplyManifest`: Apply a YAML resource file to the cluster (useful for customizations).
 
     ### Documentation Tools
-    - `QueryTool`: Search documentation related to Helm, charts, and Kubernetes integration.
+    - `query_Docs`: Search documentation related to Helm, charts, and Kubernetes integration.
```
sbx0r: ooops :)
snake_Pascal

helm/agents/istio/templates/agent.yaml
```diff
@@ -326,5 +294,5 @@ spec:
           - "My requests to 'service-x' are failing with 503 errors, can you help troubleshoot?"
           - "How can I view the Istio telemetry for the 'payment-service'?"
           - "The sidecar injection is not working for pods in 'app-ns'. What should I check?"
-          - "Query documentation for best practices on Istio performance tuning."
+          - "query_Docs for best practices on Istio performance tuning."
```
sbx0r, line 297: Would it make sense to consider renaming the tool to something that clearly reflects its focus on Istio?

helm/agents/observability/templates/agent.yaml
```diff
@@ -76,30 +76,26 @@ spec:
...
+          - query_Docs
+    - type: McpServer
+      mcpServer:
+        toolServer: kagent/kagent-tool-server
+        toolNames:
+        - k8s_get_resources
+        - k8s_get_available_api_resources
     - type: Agent
       agent:
         ref: promql-agent
```
sbx0r, line 131: I know that your changes didn't change this line, but could you make it a ref string also? :)

helm/agents/observability/templates/agent.yaml
```diff
@@ -76,30 +76,26 @@ spec:
...
-          base_url: {{ .Values.grafana.url | default "grafana.kagent:3000" | quote }}
-          username: {{ .Values.grafana.username | default "" | quote }}
-          password: {{ .Values.grafana.password | default "" | quote }}
-          api_key: {{ .Values.grafana.apiKey | default "" | quote }}
+      ## TODO  - grafana.DashboardManagementTool
+      # config:
+      #   base_url: {{ .Values.grafana.url | default "grafana.kagent:3000" | quote }}
+      #   username: {{ .Values.grafana.username | default "" | quote }}
+      #   password: {{ .Values.grafana.password | default "" | quote }}
+      #   api_key: {{ .Values.grafana.apiKey | default "" | quote }}
```
sbx0r: Is it an intentional leftover (eg for next iteration)?

go/tools/cmd/tool-server/main.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func run(cmd *cobra.Command, args []string) {
+	logger.Init()
+	defer logger.Sync()
+
+	mcp := server.NewMCPServer(
+		"kagent-tools",
+		"1.0.0",
```
sbx0r: shouldn't it be the same version like the version of kagent?

go/tools/cmd/tool-server/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+var rootCmd = &cobra.Command{
+	Use:   "tool-server",
+	Short: "KAgent tool server",
+	Run:   run,
+}
+
+func init() {
+	rootCmd.Flags().IntVarP(&port, "port", "p", 8084, "Port to run the server on")
+	rootCmd.Flags().BoolVar(&stdio, "stdio", false, "Use stdio for communication instead of HTTP")
```
sbx0r, line 49: do we actually need it?

go/tools/README.md
```diff
@@ -0,0 +1,226 @@
...
+```bash
+go build -o kagent-tools .
+```
+
+### Running
+```bash
+./kagent-tools
+```
+
+The server runs using stdio transport for MCP communication.
```
sbx0r: the defaults says something different

go/controller/internal/autogen/autogen_api_translator.go
sbx0r: If QueryDoc needs the OpenAI API key here, why the actual image of QueryDoc also requires it?

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -545,6 +547,10 @@ func (a *apiTranslator) translateAssistantAgent(
 			}
 			tools = append(tools, autogenTool)
 		case tool.McpServer != nil:
+			if toolNeedsOpenaiApiKey(tool.McpServer.ToolServer) && modelConfig.Spec.Provider != v1alpha1.OpenAI {
```
sbx0r: Just wondering. Is it still needed if actual mcp server has been moved to the separate pod?

helm/agents/helm/templates/agent.yaml
dimetron: yeah, fixing this one as well

go/tools/cmd/tool-server/main.go
dimetron: yes, we need both web and stdio

go/controller/internal/autogen/autogen_api_translator.go
dimetron: @sbx0r if tool is OpenAI only and not found Agent is not reconciled. 
For example Ollama.

EItanya: But we don't have built in tools anymore, so why would we be translating the model config anyway?

go/tools/cmd/tool-server/main.go
EItanya: @sbx0r, for context, we are going to be releasing these tools for general consumption, as mentioned yesterday during the community meeting. In order to enable that use case having an stdio option will be important.

go/controller/internal/autogen/autogen_api_translator.go
EItanya: This also relates to the issue comment above. Why do we care about this case anymore if the key now has to be configured on the container.

go/controller/internal/autogen/autogen_api_translator.go
dimetron: Good point, I was not following that part in detail. We still need to provide tool list to autogen during agent reconcile, right ? And we excluding openAI specific tools if we use different provider. If Agent has doc-query tool in the list of tools we want to skip it. As I cannot do it in helm right now. for sub charts

dimetron: I think when we will add support of different providers [ollama,anthropic] to doc2vec MCP we can remove this condition

helm/agents/observability/templates/agent.yaml
dimetron: moving it to ENV var of the tool server

go/tools/cmd/tool-server/main.go
dimetron: definitely we can use app version - will need to add dynamic version during build

go/controller/internal/autogen/autogen_api_translator.go
dimetron: I will remove this hardcoded toolNeedsOpenaiApiKey and will make it work on tool side

EItanya: > Good point, I was not following that part in detail. We still need to provide tool list to autogen during agent reconcile, right ? And we excluding openAI specific tools if we use different provider. If Agent has doc-query tool in the list of tools we want to skip it. As I cannot do it in helm right now. for sub charts

After these changes I don't think we care anymore. The model is completely determined by how the MCP server itself is configured, so all of this logic is no longer relevant, we are no longer passing through any config.

go/controller/internal/autogen/autogen_api_translator.go
dimetron: fixed

go/tools/cmd/tool-server/main.go
EItanya: We already have one I believe, or at least we do it for the `cli` we can do a similar thing for the controller

Makefile
```diff
@@ -215,37 +234,39 @@ helm-test: helm-version
...
 	echo $$(helm template kagent ./helm/kagent/ --namespace kagent --set providers.default=anthropic    --set providers.anthropic.apiKey=your-anthropic-api-key 	| tee tmp/anthropic.yaml 	| grep ^kind: | wc -l)
 	echo $$(helm template kagent ./helm/kagent/ --namespace kagent --set providers.default=azureOpenAI  --set providers.azureOpenAI.apiKey=your-openai-api-key		| tee tmp/azureOpenAI.yaml	| grep ^kind: | wc -l)
+	helm plugin ls | grep unittest || helm plugin install https://github.com/helm-unittest/helm-unittest.git
+	helm unittest helm/kagent
 
 .PHONY: helm-agents
 helm-agents:
 	VERSION=$(VERSION) envsubst < helm/agents/k8s/Chart-template.yaml > helm/agents/k8s/Chart.yaml
-	helm package helm/agents/k8s
+	helm package -d dist helm/agents/k8s
```
EItanya: Can we make `dist` a variable: aka `HELM_DIST_FOLDER?=dist` or something.

Also this directory needs to be gitignored

Makefile
```diff
@@ -290,39 +311,47 @@ helm-uninstall:
...
+kagent-cli-install: clean build-cli-local kind-load-docker-images helm-version
 kagent-cli-install:
 	KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local install
 	KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local dashboard
 
 .PHONY: kagent-cli-port-forward
 kagent-cli-port-forward: use-kind-cluster
 	@echo "Port forwarding to kagent CLI..."
-	kubectl port-forward -n kagent service/kagent 8081:8081 8082:80
+	kubectl port-forward -n kagent service/kagent 8082:80 8084:8084
```
EItanya: Why did you get rid of `8081`

ui/conf/nginx.conf
```diff
@@ -22,6 +22,10 @@ http {
         server 127.0.0.1:8083;
     }
 
+    upstream kagent_tools {
+        server 127.0.0.1:8084;
```
EItanya: Does the UI need access to the tools? This is specifically for the UI. We can remove the ws_backend above as well.

python/.dockerignore
```diff
@@ -12,4 +12,7 @@ env/
 .pytest_cache/
 .coverage
 htmlcov/
-workspace/
\ No newline at end of file
+workspace/
+
+#exclude tools from the build
```
EItanya: Should we just delete the code? I would be in favor of that. It's all in git history anyway.

python/src/kagent/cli.py
```diff
@@ -1,282 +1,9 @@
 import typer
-from autogen_core import CancellationToken
 from mcp.server.fastmcp import FastMCP
```
EItanya, line 2: This can be removed

python/src/kagent/cli.py
```diff
@@ -1,282 +1,9 @@
...
-@app.command()
-def helm():
-    mcp.add_tool(helm_list_releases._func, helm_list_releases.name, helm_list_releases.description)
-    mcp.add_tool(helm_get_release._func, helm_get_release.name, helm_get_release.description)
-    mcp.add_tool(helm_uninstall._func, helm_uninstall.name, helm_uninstall.description)
-    mcp.add_tool(upgrade_release._func, upgrade_release.name, upgrade_release.description)
-    mcp.add_tool(helm_repo_add._func, helm_repo_add.name, helm_repo_add.description)
-    mcp.add_tool(helm_repo_update._func, helm_repo_update.name, helm_repo_update.description)
-    mcp.run()
+mcp = FastMCP("KAgent")
```
EItanya, line 6: Remove

helm/kagent/templates/toolserver.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kagent.dev/v1alpha1
+kind: ToolServer
+metadata:
+  name: {{ include "kagent.fullname" . }}-tool-server
+  namespace: {{ include "kagent.namespace" . }}
+spec:
+  config:
```
EItanya, line 7: Do we need to also optionally include the secret values for grafana/prometheus?

EItanya: Ignore me, I see now

go/controller/internal/autogen/autogen_api_translator.go
```diff
@@ -27,7 +27,8 @@ var (
 		"kagent.tools.k8s.GenerateResourceTool",
 	}
 	toolsProvidersRequiringOpenaiApiKey = []string{
-		"kagent.tools.docs.QueryTool",
+		"query_docs",
+		"kagent-querydoc",
 	}
```
EItanya: Can we remove these maps as we decided the config no longer lives at this level anyway? Now the config lives in the container itself iirc.

go/controller/internal/autogen/autogen_reconciler.go
```diff
@@ -619,8 +619,10 @@ func (a *autogenReconciler) findAgentsUsingApiKeySecret(ctx context.Context, req
 
 	var models []string
 	for _, model := range modelsList.Items {
+		if model.Spec.Provider == v1alpha1.Ollama {
+			continue
+		}
```
EItanya: Can we just check if `model.Spec.ApiKeySecretRef` is set?

go/tools/cmd/tools-client/main.go
EItanya, line 1: Can we delete this file?

go/tools/cmd/tool-server/main.go
EItanya: I will do this properly for both containers in a follow-up

go/tools/cmd/tool-server/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+	} else {
+		addr := fmt.Sprintf(":%d", port)
+		logger.Get().Info("Running KAgent Tools Server", "port", addr, "tools", strings.Join(tools, ","))
+		sseServer := server.NewSSEServer(mcp)
+
+		go func() {
+			if err := sseServer.Start(addr); err != nil {
+				if !errors.Is(err, http.ErrServerClosed) {
+					logger.Get().Error(err, "Failed to start SSE server")
```
EItanya, line 103: If this fails to start will this hang forever since the `signalChan` is waiting for an interrupt?

go/go.mod
```diff
@@ -7,131 +7,131 @@ require (
...
-	k8s.io/utils v0.0.0-20241210054802-24370beab758 // indirect
+	k8s.io/kube-openapi v0.0.0-20250610211856-8b98d1ed966a // indirect
+	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397 // indirect
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.32.0 // indirect
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 )
+
+replace github.com/google/cel-go => github.com/google/cel-go v0.25.0
```
EItanya: Why is this required, aren't these identical?

go/tools/internal/prometheus/prometheus_test.go
EItanya, line 1: Unused? Did you mean to add tests here?

go/tools/internal/argo/argo.go
```diff
@@ -0,0 +1,389 @@
+package argo
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
```
EItanya: `ioutil` is now deprecated, please remove references.

go/tools/internal/grafana/grafana.go
```diff
@@ -0,0 +1,362 @@
...
+// Grafana API client
+
+type GrafanaConfig struct {
+	BaseURL  string
+	Username string
+	Password string
+	APIKey   string
+}
+
+func newGrafanaClient(baseURL, username, password, apiKey string) *http.Client {
```
EItanya: Unused args?

go/tools/internal/grafana/grafana.go
```diff
@@ -0,0 +1,362 @@
...
+		return mcp.NewToolResultError("Grafana API request failed: " + err.Error()), nil
+	}
+
+	return mcp.NewToolResultText(string(respBody)), nil
+}
+
+func RegisterGrafanaTools(s *server.MCPServer) {
+	s.AddTool(mcp.NewTool("grafana_org_management",
+		mcp.WithDescription("Manage Grafana organizations"),
+		mcp.WithString("base_url", mcp.Description("The base URL of the Grafana API")),
```
EItanya: Why are `base_url, password, username, api_key` part of the args, those are secret data, shouldn't they be loaded at init/config time?

go/tools/internal/prometheus/prometheus.go
```diff
@@ -0,0 +1,230 @@
...
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// clientKey is the context key for the http client.
+type clientKey struct{}
+
+func getHTTPClient(ctx context.Context) *http.Client {
+	if client, ok := ctx.Value(clientKey{}).(*http.Client); ok && client != nil {
```
EItanya: I don't see this being set anywhere

EItanya: Additionally how are we doing auth?

Makefile
dimetron: HELM_DIST_FOLDER added

python/.dockerignore
dimetron: already deleted

python/src/kagent/tools/k8s/gateway_api/_reference_grant.py
```diff
@@ -1,200 +0,0 @@
-REFERENCE_GRANT_PROMPT = r"""
-# Role
```
peterj, line 2: I didn't see this or other tools that can generate resources to be moved -- are we not moving those?

go/tools/internal/grafana/grafana.go
```diff
@@ -0,0 +1,362 @@
...
+	"strings"
+	"time"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// Grafana API client
+
+type GrafanaConfig struct {
```
peterj: how can the users configure this? (and similarly, I think we have the configuration for the prometheus tools as well).

go/tools/README.md
```diff
@@ -0,0 +1,226 @@
...
+### 9. Documentation Tools (`docs.go`)
+Provides documentation query functionality:
+
+- **query_documentation**: Query documentation for supported products (simplified implementation)
+- **list_supported_products**: List supported products for documentation queries
+
+### 10. Common Tools (`common.go`)
+Provides general utility functions:
+
+- **shell**: Execute shell commands
```
peterj, line 119: suggestion: move the shell and the datetime tools into "Utilites" category. so we can name them like `utils.shell, utils.current_date_time` and `grafana.something...` -- this will help in the UI to determine the tool category and categorize the tools

helm/kagent/templates/deployment.yaml
```diff
@@ -103,3 +108,104 @@ spec:
...
+                  fieldPath: metadata.namespace
+            - name: OTEL_TRACING_ENABLED
+              value: {{ .Values.otel.tracing.enabled | quote }}
+            - name: OTEL_EXPORTER_OTLP_ENDPOINT
+              value: {{ .Values.otel.tracing.exporter.otlp.endpoint | quote }}
+            - name: OTEL_EXPORTER_OTLP_TRACES_TIMEOUT
+              value: {{ .Values.otel.tracing.exporter.otlp.timeout | quote }}
+            - name: OTEL_EXPORTER_OTLP_TRACES_INSECURE
+              value: {{ .Values.otel.tracing.exporter.otlp.insecure | quote }}
+            - name: PROMETHEUS_URL
```
peterj: ignore my previous comment regarding configuration. I think this is fine for now.

go/go.mod
dimetron: github.com/google/cel-go v0.24.1  bump v0.25.0

python/src/kagent/cli.py
dimetron: ??

go/tools/internal/grafana/grafana.go
dimetron: They were in Agent config - now they are in secret and ENV vars

go/tools/internal/grafana/grafana.go
dimetron: prometheus:
    url: ""
    username: ""
    password: ""
  grafana:
    url: ""
    username: ""
    password: ""
    apiKey: ""

python/src/kagent/tools/k8s/gateway_api/_reference_grant.py
dimetron: Need to see if it was lost

go/tools/internal/grafana/grafana.go
dimetron: tools:
  prometheus:
    url: ""
    username: ""
    password: ""
  grafana:
    url: ""
    username: ""
    password: ""
    apiKey: ""

go/tools/internal/grafana/grafana.go
dimetron: @EItanya maybe we can use https://github.com/grafana/mcp-grafana

helm/kagent/templates/deployment.yaml
```diff
@@ -55,6 +55,11 @@ spec:
             - name: http
```
peterj: I logged an issue for this, but you can fix it here: https://github.com/kagent-dev/kagent/issues/565 

Basically, we shouldn't name all ports "http" -- let's name them "http-app", "http-backend", "http-ui" or something like that.

EItanya (approved):

EItanya (approved):

PR #567, opened 2025-06-27
bump-a2a-0.2.0

### LINKED ISSUE #551: [FEATURE] Need to follow the upgrade of A2A specification?
### 📋 Prerequisites

### 📝 Feature Summary

_No response_

A2A protocol had a major incompatible upgrade recently.  Does the project plan to adopt this upgrade?

Previous Specification: [A2A v0.1.0 JSON Schema](https://github.com/google-a2a/A2A/blob/v0.1.0/specification/json/a2a.json)
Current Specification: [A2A v0.2.0 JSON Schema](https://github.com/google-a2a/A2A/blob/v0.2.0/specification/json/a2a.json)

And the [trpc-a2a-go](https://github.com/trpc-group/trpc-a2a-go/releases/tag/v0.2.0-beta) framework has completed adaptation to the new A2A specification.
Related changelog : https://github.com/trpc-group/trpc-a2a-go/blob/main/CHANGELOG.md

### 💡 Proposed Solution

Maybe we could consider trying to upgrade the trpc-a2a-go to its latest version?

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

EItanya:
fixes #551

### DIFF
--- a/go/controller/internal/a2a/a2a_handler_mux.go
+++ b/go/controller/internal/a2a/a2a_handler_mux.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"sync"
 
-	common "github.com/kagent-dev/kagent/go/controller/internal/utils"
 	"trpc.group/trpc-go/trpc-a2a-go/server"
 	"trpc.group/trpc-go/trpc-a2a-go/taskmanager"
 )
@@ -97,14 +96,12 @@ func (a *handlerMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	handlerName := common.ResourceRefString(agentNamespace, agentName)
-
 	// get the underlying handler
-	handlerHandler, ok := a.getHandler(handlerName)
+	handlerHandler, ok := a.getHandler(agentName)
 	if !ok {
 		http.Error(
 			w,
-			fmt.Sprintf("Agent %s not found", handlerName),
+			fmt.Sprintf("Agent %s not found", agentName),
 			http.StatusNotFound,
 		)
 		return
--- a/go/controller/internal/a2a/a2a_reconciler.go
+++ b/go/controller/internal/a2a/a2a_reconciler.go
@@ -60,7 +60,7 @@ func (a *a2aReconciler) ReconcileAutogenAgent(
 	}
 
 	return a.a2aHandler.SetAgentHandler(
-		agentRef,
+		agent.Name,
 		params,
 	)
 }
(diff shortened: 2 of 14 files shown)


go/controller/internal/a2a/a2a_reconciler.go
```diff
@@ -60,7 +60,7 @@ func (a *a2aReconciler) ReconcileAutogenAgent(
 	}
 
 	return a.a2aHandler.SetAgentHandler(
-		agentRef,
+		agent.Name,
```
sbx0r: are you sure about that? :)
what if we have 2 agents with the same name but different namespaces?

go/internal/utils.go
```diff
@@ -0,0 +1,6 @@
+package internal
+
+// MakePtr is a helper function to create a pointer to a value.
+func MakePtr[T any](v T) *T {
```
sbx0r: why not to use ptr.To(var) from [k8s.io/utils/ptr](https://pkg.go.dev/k8s.io/utils/ptr) lib?

EItanya: Good idea, I'll switch that!

go/controller/internal/a2a/a2a_reconciler.go
EItanya: That's a good point, 😭. I was hoping to not, but I did this to try and fix another issue I was seeing, so I'll probably need to make some other changes as well 😂

go/controller/internal/a2a/a2a_handler_mux.go
```diff
@@ -104,15 +104,12 @@ func (a *handlerMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if !ok {
 		http.Error(
 			w,
-			fmt.Sprintf("Agent %s not found", handlerName),
```
sbx0r: Why not using full refs in logs?

go/controller/internal/a2a/a2a_task_processor.go
```diff
@@ -24,38 +23,52 @@ type a2aTaskProcessor struct {
 	handleTask TaskHandler
 }
 
-var _ taskmanager.TaskProcessor = &a2aTaskProcessor{}
+var _ taskmanager.MessageProcessor = &a2aTaskProcessor{}
 
 // newA2ATaskProcessor creates a new A2A task processor.
-func newA2ATaskProcessor(handleTask TaskHandler) taskmanager.TaskProcessor {
+func newA2ATaskProcessor(handleTask TaskHandler) taskmanager.MessageProcessor {
```
sbx0r: Shouldn't this function be renamed?
Similar question about the parameter?

peterj (approved):

PR #572, opened 2025-06-28
fix(go): add owner reference to secrets created for modelConfigs

### LINKED ISSUE #489: [BUG] k8s secret for model is not deleted after model is deleted

sbx0r:
Resolves #489


go/controller/internal/httpserver/handlers/modelconfig.go
```diff
@@ -335,6 +336,26 @@ func (h *ModelConfigHandler) HandleCreateModelConfig(w ErrorResponseWriter, r *h
 		return
 	}
 
+	if providerTypeEnum != v1alpha1.Ollama && req.APIKey != "" {
```
EItanya, line 313: Rather than updating the secret, can we just create the secret AFTER the ModelConfig and add it at that point?

EItanya (commented):
I agree with your comment, this handler is getting pretty complex. I don't love mutating webhooks generally, but I think we can clean this code up to make things more clear as well. Smaller better tested functions.

Also, I think this logic needs to be added to the update function as well, in the case that the secret doesn't exist.

go/controller/internal/httpserver/handlers/modelconfig.go
sbx0r: I was thinking about that and i thought - if we create ModelConfig CR before the secret, then the reconciler could throw several errors. Please correct me if I'm wrong

sbx0r:
> I agree with your comment, this handler is getting pretty complex. I don't love mutating webhooks generally, but I think we can clean this code up to make things more clear as well. Smaller better tested functions.

Alrighty.

> Also, I think this logic needs to be added to the update function as well, in the case that the secret doesn't exist.

Oops. I missed that one 😨. Thanks!

go/controller/internal/httpserver/handlers/modelconfig.go
sbx0r: I read more about this problem I see that we have 2 ways:
- Go with *Graceful Degradation* - do not crash, just set status and RequeueAfter
- Use *Controller Pattern* - create missing resources at reconcile level

Which one do you prefer?

Also, if you know something better, please let me know :)

sbx0r: You're right!

I'll update the reconciler code to add proper exception handling for cases where the ModelConfig CR exists but the corresponding Secret doesn't

EItanya: I think it's probably ok for it to fail because it will be updated quickly afterward and then succeed?

sbx0r: Already did that lol 😆

sbx0r: But yes, it would be just a potential inconvenience

sbx0r: I also discovered a bug. We don't have watchers for dependent resources, so if I remove the secret used by the ModelConfig, it wouldn't be reconciled - The status will still show this CR is healthy

EItanya:
Waiting a bit on resolution of #578

PR #574, opened 2025-06-28
feat(tools): add resource generation tools to golang tool server

EItanya:


peterj:
Looking at this. Will push UI changes to the same branch (to remove the concept of builtin tools).

Issue #577, opened 2025-06-30
[BUG] External Secret Deletion Breaks ModelConfig Without Status Update


sbx0r:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

When users delete Kubernetes secrets that are referenced by ModelConfig resources using `kubectl` or `helm`, the ModelConfig becomes broken but the controller doesn't detect this change.
The ModelConfig Status remains unchanged, showing a healthy state while the resource is actually unusable.

### 🔄 Steps To Reproduce

1. Create a ModelConfig with an API key secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: openai-secret
  namespace: default
data:
  OPENAI_API_KEY: "abcabc"  # base64 encoded fake key
---
apiVersion: kagent.dev/v1alpha1
kind: ModelConfig
metadata:
  name: gpt4-config
  namespace: default
spec:
  provider: openai
  model: gpt-4
  apiKeySecretRef: "default/openai-secret"
```
2. Check ModelConfig status (should show as ready):
```bash
kubectl get modelconfig gpt4-config -o yaml
# Status shows healthy/ready condition
```
3. Delete the secret externally:
```bash
kubectl delete secret openai-secret -n default
```
4. Check ModelConfig status again:
```bash
kubectl get modelconfig gpt4-config -o yaml
# Status STILL shows healthy/ready - this is the bug!
```
5. Try to use an agent with this ModelConfig - it will fail with unclear errors

### 🤔 Expected Behavior

After secret deletion:
- ModelConfig controller should detect the missing secret
- ModelConfig status should be updated with condition like:
  ```yaml
  status:
    conditions:
    - type: Ready
      status: "False"
      reason: SecretNotFound
      message: "Referenced secret 'default/openai-secret' not found"
  ```
- Users should immediately see the broken state in `kubectl get modelconfig`

### 📱 Actual Behavior

- ModelConfig status shows as healthy/ready
[...]

Issue #575, opened 2025-06-29
[BUG] ModelConfig should gracefully handle missing API key secret errors


sbx0r:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

Currently, when a ModelConfig references a missing API key secret, the system doesn't handle this scenario gracefully. This can lead to unclear error states or unexpected behavior.

### 🔄 Steps To Reproduce

1. Create a ModelConfig that references a non-existent secret:
   ```yaml
   apiVersion: kagent.dev/v1alpha1
   kind: ModelConfig
   metadata:
     name: test-config
     namespace: default
   spec:
     provider: openai
     model: gpt-4
     apiKeySecretRef: default/missing-secret
     apiKeySecretKey: OPENAI_API_KEY
2. Create an Agent that uses this ModelConfig
3. Observe the ModelConfig status - it should show ModelConfigReconcileFailed with unclear error message
4. Check controller logs - should see generic "failed to fetch secret" errors
5. Even after creating the missing secret, reconciliation might not recover properly

### 🤔 Expected Behavior

- ModelConfig should detect when its referenced secret is missing
- Set appropriate status condition to indicate the issue
- Provide clear error message about the missing secret
- Allow for recovery when the secret becomes available

### 📱 Actual Behavior

- **Complete failure**: Missing secrets caused the entire reconciliation process to crash
- **Generic error messages**: Status only showed `ModelConfigReconcileFailed` without indicating the specific cause
- **No error distinction**: Missing secrets were indistinguishable from other types of reconciliation failures
- **No recovery mechanism**: System couldn't recover when the secret was later added
- **Cascade failures**: Secret errors blocked reconciliation of dependent agents and teams

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR to fix this issue

Issue #321, opened 2025-05-04
[FEATURE] Allow to pick existing secret for the model provider in 'create model' wizard


sbx0r:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add option to pick existing secret for the model

### ❓ Problem Statement / Motivation

- Everytime you define new model, you have to specify APIKey
- This leads to the situation where we have many secrets with exactly the same values

### 💡 Proposed Solution

- In `create model` wizard, add option which gives the user to pick an existing secret (from the same namespace [security]) for this provider

### 🎯 Affected Service(s)

UI Service

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

AnnaRevutsky:
I would like to work on this issue

peterj:
hey @AnnaRevutsky! Thanks for volunteering -- I am assigning this to you. Feel free to reach out here or on Discord in fou have any questions/need help!

sbx0r:
Hi @AnnaRevutsky !
How are you doing?

Do you need any assistance with this?

inFocus7:
^ Was going to ask the same lol.

I noticed there may be some UX/edge cases that could warrant a bit of discussion/aligning:
1. Currently model configs & secret naming is 1:1 with `<provider>-<model>`. So if Secret reuse is supported, it may feel a bit weird.
    - ex. `openai: gpt-4o` creates an `openai-gpt-4o` modelConfig + secret. later when creating an `openai: claude-3-7-sonnet-20250219` model config, it would reference a secret with the name `openai-gpt-4o` (or whatever else is chosen).
    - this isn't breaking or anything, just noting it may seem a bit weird at first glance.
2. When editing a model's api key, it seems to modify the secret's data based on the 1:1 modelConfig<->secret name `<provider>-<model>`, so we'd need to be wary about this gets handled when supporting secret re-use.

---

The following could be a bad idea and make the project more complex, but it could be easier to have a new `ApiKey` CRD that's user-definable:
```yaml
apiVersion: security.kagent/v1 # or something
kind: ApiKey 
metadata:
  name: my-shared-apikey
  namespace: default
  labels:
[...]

PR #578, opened 2025-06-30
EP-489: Secret Lifecycle Management for kagent resources

### LINKED ISSUE #489: [BUG] k8s secret for model is not deleted after model is deleted

sbx0r:
## Summary
  Add Enhancement Proposal EP-489 for Secret Lifecycle Management

  ## What this PR does
  - Adds design document for automatic cleanup of orphaned secrets
  - Proposes labels + watchers approach over OwnerReferences and finalizers
  - Covers ModelConfig and Memory resource secret management
  - Documents rejected alternatives (OwnerReferences, Finalizers)

  ## Key decisions
  - Uses label-based secret identification for explicit opt-in
  - Implements dedicated SecretCleanupReconciler watching both resources and secrets
  - Handles external secret modifications (kubectl/helm deletions)
  - Ensures consistency across all resource creation methods

  ## Next steps
  - [ ] Get maintainer approval on proposed solution
  - [ ] Get maintainer approval on rejected alternatives
  - [ ] Implement proof of concept
  - [ ] Address open questions in EP document

  Closes #489
  Related: #321, #575, #577

### DIFF
--- a/design/EP-489.yaml
+++ b/design/EP-489.yaml
@@ -1,0 +1,199 @@
+# EP-489: Secret Lifecycle Management for kagent resources
+
+* Issue: [#489](https://github.com/kagent-dev/kagent/issues/489)
+* Additional Issues:
+  - Issue: [#321](https://github.com/kagent-dev/kagent/issues/321)
+  - Issue: [#575](https://github.com/kagent-dev/kagent/issues/575)
+  - Issue: [#577](https://github.com/kagent-dev/kagent/issues/577)
+
+## Background 
+
+The kagent system creates and manages Kubernetes secrets containing API keys for ModelConfig resources and vector database credentials for Memory resources.
+Currently, when these resources are deleted or updated to no longer use a secret, orphaned secrets remain in the cluster, creating security risks and resource waste.
+
+This enhancement proposes a robust secret lifecycle management system that handles automatic cleanup of unused secrets while avoiding the complexity and race conditions inherent in OwnerReference-based approaches.
+The solution works consistently across all resource creation methods (API, kubectl, Helm).
+
+## Motivation
+
+### Current Problems
+
+1. **Orphaned Secrets**: When ModelConfigs or Memory resources are deleted, their associated secrets remain in the cluster indefinitely
+2. **Update Scenarios**: When resources change providers (e.g., OpenAI --> Ollama, Pinecone --> Mem0), old secrets become orphaned
+3. **Shared Secrets**: Multiple resources can reference the same secret, making ownership unclear
+4. **Manual Secret Creation**: Users can create secrets via kubectl/helm and reference them, bypassing our management
+5. **External Secret Modifications**: Users can delete or modify secrets externally (kubectl/helm), breaking kagent resources that depend on them
+6. **API-Only Solutions Create Inconsistency**: Implementing cleanup only in API handlers would create different behavior between resources created via API vs kubectl/Helm, leading to confusing user experience
+
+### Goals
+
+- Automatically clean up unused secrets when no kagent resources reference them
+- Handle secret cleanup during both deletion and update scenarios  
+- Support shared secrets across multiple ModelConfigs and Memory resources
+- Maintain compatibility with manually created secrets
+- Ensure consistent behavior regardless of resource creation method (API, kubectl, Helm)
+- Avoid race conditions and complex ownership management
+- Follow Kubernetes best practices and security principles
+- Provide extensible solution for future resource types that use secrets
+- Detect and handle external secret modifications (deletion/updates by kubectl/helm)
+
+### Non-Goals
+
+- Managing secrets not related to kagent resources (ModelConfig API keys, Memory credentials, etc.)
+- Cross-namespace secret references (secrets must be in same namespace as referencing resource)
+- Migration of existing secrets (will be handled separately)
+- UI changes for secret selection (future enhancement - issue [#321](https://github.com/kagent-dev/kagent/issues/321))
+
+## Implementation Details
+
+### Part 1: Label-Based Secret Identity
+
+Secrets managed by kagent will be explicitly labeled:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: openai-secret
+  labels:
+    kagent.dev/type: "modelconfig"  # or "memory" for vector DB credentials
+    kagent.dev/managed-by: "kagent-controller"
+data:
+  OPENAI_API_KEY: "c2stLi4u"  # base64 encoded
+```
+
+**Benefits:**
+- Explicit opt-in management
+- Safe by default (ignores unlabeled secrets)
+- Follows Kubernetes standard labeling conventions
+- Enables watching for external secret modifications
+
+### Part 2: Status Field Reference Tracking
+
+Extend ModelConfigStatus to track secret reference changes:
+
+```go
+type ModelConfigStatus struct {
+    Conditions         []metav1.Condition `json:"conditions,omitempty"`
+    ObservedGeneration int64              `json:"observedGeneration,omitempty"`
+    LastSecretRef      string             `json:"lastSecretRef,omitempty"`  // Detect changes
+    SecretStatus       SecretStatus       `json:"secretStatus,omitempty"`   // Current state
+}
+
+type SecretStatus struct {
+    InUse      bool   `json:"inUse"`
+    References int    `json:"references"`  // Reference count
+    Namespace  string `json:"namespace"`
+    Name       string `json:"name"`
+}
+```
+
+### Part 3: Controller Architecture
+
+#### Secret Cleanup Controller
+A dedicated controller following single responsibility principle that watches both resources and secrets:
+
+```go
+func (r *SecretCleanupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+    // 1. Get secret with kagent labels
+    // 2. Count references from ModelConfig and Memory resources using field indexing
+    // 3. If total reference count = 0, delete secret
+    // 4. If secret was deleted/modified externally, update resource status to reflect broken state
+    // 5. Update resource status fields
+}
+```
+
+**Key Advantage**: By watching secrets directly, this approach can detect when users delete or modify secrets externally (via kubectl/helm), allowing kagent to update resource status and alert users to broken configurations. Neither OwnerReferences nor Finalizers provide this capability.
+
+#### ModelConfig Controller Updates
+Enhanced to detect secret reference changes:
+
+```go
+func (r *ModelConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+    // Existing logic...
+    
+    // Detect secret reference changes
+    if mc.Status.LastSecretRef != currentSecretRef {
+        // Trigger cleanup check for old secret
+        // Update status with new reference
+    }
+}
+```
+
+### Part 4: Dynamic User Discovery
+
+Use controller-runtime field indexing for efficient secret usage lookups:
+
+```go
+// Setup field indexers for ModelConfig
+mgr.GetFieldIndexer().IndexField(ctx, &kagentv1alpha1.ModelConfig{}, 
+    "spec.apiKeySecretRef", func(rawObj client.Object) []string {
+        mc := rawObj.(*kagentv1alpha1.ModelConfig)
+        return []string{mc.Spec.ApiKeySecretRef}
+    })
+
+// Setup field indexers for Memory
+mgr.GetFieldIndexer().IndexField(ctx, &kagentv1alpha1.Memory{}, 
+    "spec.vectorDB.secretRef", func(rawObj client.Object) []string {
+        mem := rawObj.(*kagentv1alpha1.Memory)
+        return []string{mem.Spec.VectorDB.SecretRef}
+    })
+
+// Efficient lookup across both resource types
+var modelConfigs kagentv1alpha1.ModelConfigList
+err := r.List(ctx, &modelConfigs, 
+    client.MatchingFields{"spec.apiKeySecretRef": secretKey})
+
+var memories kagentv1alpha1.MemoryList
+err = r.List(ctx, &memories, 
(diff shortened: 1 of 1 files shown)


design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+This enhancement proposes a robust secret lifecycle management system that handles automatic cleanup of unused secrets while avoiding the complexity and race conditions inherent in OwnerReference-based approaches.
+The solution works consistently across all resource creation methods (API, kubectl, Helm).
+
+## Motivation
+
+### Current Problems
+
+1. **Orphaned Secrets**: When ModelConfigs or Memory resources are deleted, their associated secrets remain in the cluster indefinitely
+2. **Update Scenarios**: When resources change providers (e.g., OpenAI --> Ollama, Pinecone --> Mem0), old secrets become orphaned
```
EItanya, line 22: How is this true in our system. Specifically I'm thinking about the case when we are creating the `Secret`. If we are creating the secret for a given `ModelConfig` shouldn't we just use the same name for the `Secret`.

design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+3. **Shared Secrets**: Multiple resources can reference the same secret, making ownership unclear
+4. **Manual Secret Creation**: Users can create secrets via kubectl/helm and reference them, bypassing our management
+5. **External Secret Modifications**: Users can delete or modify secrets externally (kubectl/helm), breaking kagent resources that depend on them
+6. **API-Only Solutions Create Inconsistency**: Implementing cleanup only in API handlers would create different behavior between resources created via API vs kubectl/Helm, leading to confusing user experience
+
+### Goals
+
+- Automatically clean up unused secrets when no kagent resources reference them
```
EItanya, line 30: We need to be very careful to only do this in the case that we created the `Secret`. Which I think is probably just a demo use-case. I'm guessing after the first test most people will want to provide their own secrets.

design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+- Ensure consistent behavior regardless of resource creation method (API, kubectl, Helm)
+- Avoid race conditions and complex ownership management
+- Follow Kubernetes best practices and security principles
+- Provide extensible solution for future resource types that use secrets
+- Detect and handle external secret modifications (deletion/updates by kubectl/helm)
+
+### Non-Goals
+
+- Managing secrets not related to kagent resources (ModelConfig API keys, Memory credentials, etc.)
+- Cross-namespace secret references (secrets must be in same namespace as referencing resource)
```
EItanya, line 43: This unfortunately is not true in our API today, but I would like to change it so that this is true.

design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+Secrets managed by kagent will be explicitly labeled:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: openai-secret
+  labels:
+    kagent.dev/type: "modelconfig"  # or "memory" for vector DB credentials
+    kagent.dev/managed-by: "kagent-controller"
```
EItanya, line 60: Does `k8s` have a well known label for managed-by/owned-by? Also, why aren't we ALSO using `OwnerReference` here. In this case it makes the management easier I think.

design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+### Part 2: Status Field Reference Tracking
+
+Extend ModelConfigStatus to track secret reference changes:
+
+```go
+type ModelConfigStatus struct {
+    Conditions         []metav1.Condition `json:"conditions,omitempty"`
+    ObservedGeneration int64              `json:"observedGeneration,omitempty"`
+    LastSecretRef      string             `json:"lastSecretRef,omitempty"`  // Detect changes
+    SecretStatus       SecretStatus       `json:"secretStatus,omitempty"`   // Current state
```
EItanya, line 80: Can we make these conditions instead?

design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+}
+
+type SecretStatus struct {
+    InUse      bool   `json:"inUse"`
+    References int    `json:"references"`  // Reference count
```
EItanya, line 85: So I think this is a bit of an anti-pattern and we should probably avoid it. In general k8s reconciliation should focus on just the resource it cares about, whereas this field requires knowledge of the whole system, this can lead to very complex reconciliation.

design/EP-489.yaml
```diff
@@ -0,0 +1,199 @@
...
+    if mc.Status.LastSecretRef != currentSecretRef {
+        // Trigger cleanup check for old secret
+        // Update status with new reference
+    }
+}
+```
+
+### Part 4: Dynamic User Discovery
+
+Use controller-runtime field indexing for efficient secret usage lookups:
```
EItanya, line 125: I love this idea!

EItanya (commented):
Overall I think there are a lot of interesting ideas here, but I think we need to separate 2 things here very clearly. 

1. We only ever want to manage the lifecycle of secrets that we create, we absolutely never want to do anything with secrets we haven't created. In that case, the secret will only ever have one owner because the secret + ModelConfig are always 1:1. k8s supports multiple owner references if we ever needed that functionality, but I don't know that we ever want to add that feature to `kagent`.
2. Our reconcilers absolutely need to handle missing secrets/references better, and we can do a better job of indicating to the user what exactly is going wrong.

design/EP-489.yaml
sbx0r: That's actually still an open question (#321)
Currently we have 1:1 relation at the API and many:1 on Kubernetes level.
At the UI - whenever you want to use the same provider but with different parameters/options, you will have to create a new secret. 
At the k8s level - If you use Helm, Kustomize, Kubectl, etc. you can have many:1 relation and point any existing secret to use.
This shows that direct Kubernetes interactions currently have an advantage over the controller API, which introduces limitations for the UI.

design/EP-489.yaml
sbx0r: Fully agree!

This is why this proposal introduces `kagent.dev/managed-by` label.
If it's not set - user doesn't it want to be managed by kagent. If it's set to eg. controller name, then the controller will know that it can manage it.

design/EP-489.yaml
sbx0r: > Does k8s have a well known label for managed-by/owned-by?

Great question!
Yes, Kubernetes has the well-known label `app.kubernetes.io/managed-by` from the [official docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).
However, many tools like Helm, are using this label. I proposed kagent specific label to avoid conflicts.

> Also, why aren't we ALSO using OwnerReference here. In this case it makes the management easier I think.

This is also a very good question!
The later `Alternatives` section explains the limitations of OwnerReferences in detail, but I will add some more here :)
The key issue is that only one owner can have `controller: true` - when that main controller is deleted, Kubernetes garbage collector automatically removes the secret even if other ModelConfigs still reference it.

design/EP-489.yaml
sbx0r: Definitely!

Also I think now, that instead of setting LastSecretRef - which is poor, we could just use `kubectl.kubernetes.io/last-applied-configuration` annotation which is automatically  managed by kubernetes.

design/EP-489.yaml
sbx0r: That's true. I think this entire struct can be deleted.
`InUse` - has no purpose because if secret is no longer used, it shouldn't exist.
`References` - another terrible idea - this could be done by using field indexing to count total usage. I think that `MatchingFields` have complexity near O(1), so it's hard to beat 😄

design/EP-489.yaml
sbx0r: Glad you like it!

sbx0r:
> Overall I think there are a lot of interesting ideas here, but I think we need to separate 2 things here very clearly.
> 
> 1. We only ever want to manage the lifecycle of secrets that we create, we absolutely never want to do anything with secrets we haven't created. In that case, the secret will only ever have one owner because the secret + ModelConfig are always 1:1. k8s supports multiple owner references if we ever needed that functionality, but I don't know that we ever want to add that feature to `kagent`.
> 2. Our reconcilers absolutely need to handle missing secrets/references better, and we can do a better job of indicating to the user what exactly is going wrong.

Great feedback! Let me address both points:

1. Secret ownership & sharing
- This proposal doesn't automatically manage all secrets - it gives user the choice by opting-in via labels when creating resources.
   This ensures consistency whether users interact through the controller API or directly via Kubernetes API.
- The question of whether secrets can be reused across multiple ModelConfig/Memory resources is indeed one of the key design decisions we need to make here. **Whether it's worth building a Secret Lifecycle Manager largely depends on answering it.**
2. Better secret handling
- Absolutely agree! This approach could also solve (or at least provide solid foundations) related problems like detecting when a secret that ModelConfig/Memory depends on gets changed/deleted externally, and properly setting Conditions/Status accordingly. This can be easily leveraged in the UI to present clear resource state to users.

sbx0r:
Well.. I closed this EP by accident 🤣

design/EP-489.yaml
EItanya: I think this is somewhat by design though, at least until this point. If a user selected an existing secret then we wouldn't need to manage it at all, and if they choose to create one then it's directly tied to this model config/interaction. 

In general when designing kagent we imagined helm/argo would be the main way config would be managed, and that the UI would serve as an intro to building agents/tools, as well as a convenient easy way to execute/test them.

EItanya:
Here's my personal take on #1, but also open to other opinions.

In general secret management scares me, and I think the issue we originally were trying to solve feels like more of an implementation detail than a feature. We created those secrets as a result of the ModelConfig creation, but the user doesn't even necessarily know that, in that case they are basically one resource for all intents and purposes. If a user means to point to a secret that already exists then it's a more complex API interaction and I don't think we need to be involved, at least not right now.

What do you think?

sbx0r:
Yea. You're absolutely right. This might not be the right time to implement this solution. As long as our assumptions remain that:
- the API requires providing the API key during ModelConfig/Memory creation
- the API allows providing the new API key during ModelConfig/Memory update
- we block provider/model changes from the UI level (while the API allows it)
- we won't support re-using existing secrets from the API/UI level
the OwnerReferences should be an easy win and should be sufficient without any complications.

The problem of detecting changes to secrets associated with a given ModelConfig/Memory still remains to be solved, but that can be done with a simple watcher at the ModelConfig reconciler level.

We can shelve this EP for now and potentially return to it later if needed, and take the path of least resistance by using OwnerReferences instead.

EItanya:
> Yea. You're absolutely right. This might not be the right time to implement this solution. As long as our assumptions remain that:
> 
> * the API requires providing the API key during ModelConfig/Memory creation
> * the API allows providing the new API key during ModelConfig/Memory update
> * we block provider/model changes from the UI level (while the API allows it)
> * we won't support re-using existing secrets from the API/UI level
>   the OwnerReferences should be an easy win and should be sufficient without any complications.
> 
> The problem of detecting changes to secrets associated with a given ModelConfig/Memory still remains to be solved, but that can be done with a simple watcher at the ModelConfig reconciler level.
> 
> We can shelve this EP for now and potentially return to it later if needed, and take the path of least resistance by using OwnerReferences instead.

Sounds good, FWIW I still think a label is probably a good idea to limit the secrets that `kagent` has to watch. In my experience clusters often have a TON of secrets which cause a lot of unnecessary churn in the system.

sbx0r:
> Sounds good, FWIW I still think a label is probably a good idea to limit the secrets that kagent has to watch. In my experience clusters often have a TON of secrets which cause a lot of unnecessary churn in the system.

I also feel that it's a good approach, but if we go this way, the user will have to label their secrets anyway. So if we're requiring labels either way, maybe a secret lifecycle management controller could still be worth considering - just a thought 😄

PR #591, opened 2025-07-01
Pass description in agent handler

### LINKED ISSUE #544: [BUG] Agent description not visible when agent is a tool
### 📋 Prerequisites

### 🎯 Affected Service(s)

UI Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

If agent is added as a tool to another agent, the description of it is not visible on the right pane.

### 🔄 Steps To Reproduce

1. add agent as a tool for the other agent
2. open the agent
3. see the right pane (the agentTool doesn't have any description)

### 🤔 Expected Behavior

the description should be visible

### 📱 Actual Behavior

_No response_

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

inFocus7:
**wip**

- add missing (optional) description field for agent information
- fetch agent description alongside its ref when getting agent information
    - note: i didn't do this for the `list agentS` as it would be doing the kube call for every agent, and that method is only used for general -- at least in the sidebar, unsure if it could be helpful for others.

**todo**

- [ ] look into if we _shold_ add for `list` method as well.
- [ ] maybe some slightly diferent styling in the sidebar to differentiate agents from tools?

**question**
- [ ] why is this info gathered in the "teams" endpoint and not an "agents" one? (here: ui/src/app/agents/[agentId]/chat/layout.tsx)
    - when fetching the agent data for sidebar it calls teams, although there's no existing agents endpoint to get a specific agent's information.

todo:
- verify we're okay with this extra added kube getter (description field) in the existing `getTeam()` (for `api/teams/{id}/` path, i think) handler. we could also create a separate one.

Resolves #544


go/controller/internal/httpserver/handlers/teams.go
```diff
@@ -382,6 +407,25 @@ func (h *TeamsHandler) HandleGetTeam(w ErrorResponseWriter, r *http.Request) {
 				log.Error(err, "Failed to parse agent tool reference", "toolRef", toolCopy.Agent.Ref)
 				continue
 			}
+
+			// Fetch the referenced agent to get its description
+			agentRef, err := common.ParseRefString(toolCopy.Agent.Ref, team.Namespace)
```
sbx0r, line 412: at this stage, toolCopy.Agent.Ref is a full-ref guaranteed. So there is no need to to that again

go/controller/api/v1alpha1/agent_types.go
```diff
@@ -82,6 +82,9 @@ type AgentTool struct {
 	// Can either be a reference to the name of an Agent in the same namespace as the referencing Agent, or a reference to the name of an Agent in a different namespace in the form <namespace>/<name>
 	// +kubebuilder:validation:MinLength=1
 	Ref string `json:"ref,omitempty"`
+	// Description of the referenced agent.
```
sbx0r, line 85: I'm not sure if we want to add this field to CRD. Agents we reference as a tools has their own descriptions already, so it would be a duplicated information

sbx0r:
hi @inFocus7 !
Thanks for picking up that one

Lemme try to answer your question:

> **question**
why is this info gathered in the "teams" endpoint and not an "agents" one? (here: ui/src/app/agents/[agentId]/chat/layout.tsx)
when fetching the agent data for sidebar it calls teams, although there's no existing agents endpoint to get a specific agent's information.

From my personal point of view, there are 2 main reasons:
- the UI shouldn't do any additional logic when dealing with the API - presentation layer should present, rather than implement controller's business logic
- making it done on the controller side, will provide consistency between UI responses and responses from direct API calls

sbx0r:
> maybe some slightly diferent styling in the sidebar to differentiate agents from tools?

I think that's a really good idea!
We are already using icons when user is picking up tools to it's agent

what do you think @peterj ?

go/controller/api/v1alpha1/agent_types.go
inFocus7: Sorry for late follow-ups, have been busy!

---

I may be confused, although isn't this `AgentTool` CRD used specifically for the UI call? So even if the `Agent` spec defines a description, since this `AgentTool` is the spec used for the sidebar data it would also need a description. Currently the code assumes `AgentTool` had a `description` field so it tries accessing it, but the field doesn't exist and it fails to get information. [ref](https://github.com/kagent-dev/kagent/blob/fa636c730791aab84d4daa99b6510e40c70056bd/ui/src/components/sidebars/AgentDetailsSidebar.tsx#L120)

inFocus7: Another straightforward approach could be to simply update the `AgentDetailsSidebar` code ([ref](https://github.com/kagent-dev/kagent/blob/fa636c730791aab84d4daa99b6510e40c70056bd/ui/src/components/sidebars/AgentDetailsSidebar.tsx#L120)) to get the agent's description based on the existing `ref` by looking through `allAgents`. 

I attached an example below, **but** I believe the `find` would linearly scale with the number of agents an agent relies on -- unless a map-based data structure is used for the agents (currently unsure of the changes needed elsewhere to support this).

```go
if isAgentTool(tool) {
  agentRef := tool.spec.ref
  agent := allAgents.find(agent=>{match agentRef})
  description = agent?.description
}
```

inFocus7:
> We are already using icons when user is picking up tools to it's agent
nice, adding the icon for agents vs. tools was pretty simple to implement.

I'll push it in my next commit 👍🏼 I have local draft changes with [this](https://github.com/kagent-dev/kagent/pull/591#discussion_r2208767871) implemented, but waiting to see if we prefer that route or not before keeping/removing it. I'm too lazy to undo/stash them just to commit the icon changes, so will keep it all locally in the meantime 😅

EItanya:
Already done

Issue #525, opened 2025-06-13
[FEATURE] [UI] Improve agent card display format on front page


sbx0r:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### ❓ Problem Statement / Motivation

Currently, the front page displays agents in the format "namespace/name" on the cards. While this format is functional, it could be improved for better readability and visual hierarchy.

### 💡 Proposed Solution

Display the agent name as the primary/prominent text
Show the namespace as secondary text (smaller, below, or in a different style)

But if you have something better, feel free to leave a comment or edit this issue directly (if you have the power)

### 🎯 Affected Service(s)

UI Service

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

PR #595, opened 2025-07-02
refactor: display agent namespace in AgentCard layout

abhinav-1305:
resolves: #525


sbx0r:
I think that's already done #405, so we can close this one

abhinav-1305:
can be closed then

PR #599, opened 2025-07-02
Helm refactor

### LINKED ISSUE #184: [FEATURE] Use independent pods / deployments for each container service
Currently the runtime deploys as 3 containers within 1 pod -

Image

Eventually we would want 1 container per pod (plus any helpful sidecars) for independent scaling and resilience. Then pod-to-pod communication happens via ClusterIP services. I started a [PR](https://github.com/kagent-dev/kagent/pull/180) for that purpose.

This topic may be a longer-term goal, per the reasons explained [here](https://github.com/kagent-dev/kagent/issues/132#issuecomment-2740375801).

sbx0r:
Resolves #184


.github/workflows/ci.yaml
```diff
@@ -79,10 +79,14 @@ jobs:
           kubectl wait --for=condition=Accepted  agents.kagent.dev -n kagent --all --timeout=60s
 
       - name: Run e2e tests
+        working-directory: go
         run: |
-          export KAGENT_API_URL="http://$(kubectl get svc -n kagent kagent -o jsonpath='{.status.loadBalancer.ingress[0].ip}'):8083/api"
-          echo $KAGENT_API_URL
-          cd go
+          kubectl port-forward -n kagent services/kagent 8083:8083 &
+          timeout 30 bash -c 'until curl -s http://localhost:8083/health >/dev/null 2>&1; do sleep 1; done'
```
EItanya: Why do we need this? The statuses above should not be `Accepted` until everything is ready

sbx0r: Good question!

We're running port forwarding as a background process.

According to the POSIX documentation:
> "If a command is terminated by the control operator ampersand ( '&' ), the shell shall execute the command asynchronously in a subshell. This means that the shell shall not wait for the command to finish before executing the next command."
[Source](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_09_03_02)

And this results in a nondeterministic order of execution :)

You can see a simple example below:
https://unix.stackexchange.com/questions/195387/background-processes-mix-up-order-of-shell-script-execution

We could replace this line with a simple `sleep 2` if you prefer clarity over best practice.

sbx0r: I can run this step a dozen times on my fork without this command and we'll have empirically measured stats if you'd like :)

sbx0r: Okay. So I ran 25 tests and 2 of them failed. 
- https://github.com/sbx0r/kagent/actions/runs/16254137329
- https://github.com/sbx0r/kagent/actions/runs/16254143037

but it looks like these errors aren't strictly connected to the topic we're talking about. It's something different.

I'll run more with different config and let you know.

sbx0r: But for now, we can see that our background process with port-forwarding is faster than the `go run`. That's because `go run` is triggering multiple downloads before the actual tests.
If we add a cache for the go binaries it could punch us without saying hello. But I'll check that also and confirm

sbx0r: In parallel, I'm thinking about setting the controller's service type to LoadBalancer during those tests and use metallb 😅

sbx0r: Alright...
So I removed this spaghetti and turned back to metallb 😄
Now, if `CI` environment variable is set (which is set by default in github workers), `helm-install-provider` step in our Makefile will set controller's service type to LoadBalancer!

AD:
Letting alone misconfigured caches. I'll make a followups for them

python/src/autogenstudio/web/config.py
```diff
@@ -11,8 +11,19 @@ class Settings(BaseSettings):
     CONFIG_DIR: str = "configs"  # Default config directory relative to app_root
     DEFAULT_USER_ID: str = "admin@kagent.dev"
     UPGRADE_DATABASE: bool = False
+    CORS_ALLOW_ORIGINS: str = ""
```
sbx0r: what do we want to have here as a default?

python/src/autogenstudio/web/app.py
```diff
@@ -26,12 +26,7 @@
 # CORS middleware configuration
 app.add_middleware(
```
EItanya: You can remov `CORS` entirely, the browser is never reaching out to the engine anymore

python/src/autogenstudio/web/config.py
EItanya: As I commented in `app.py`, I don't think this is necessary anymore

python/src/autogenstudio/web/app.py
sbx0r: even if we want to have `all-in-one` deployment?

EItanya (commented):
Overall these changes make sense to me, just have a couple of questions:
1. Why did you decide to move the mcp-grafana server from an `stdio` server into a separate tool deployment?
2. Why did you decide to move the other 2 tool-servers into `contrib`. I think we should probably add a `helm/tools` folder for those

python/src/autogenstudio/web/app.py
EItanya: Yup, the UI never reaches out to the engine anymore, all communication happens through the golang httpserver.

sbx0r:
Thanks,
Happy to unpack the reasons behind those moves.

> 1. Why did you decide to move the mcp-grafana server from an `stdio` server into a separate tool deployment?

- The primary reason was to use the official Docker image instead of embedding a Python binary in the engine's container. This makes more sense from a security standpoint - there is no reason for the `mcp-grafana` server to inherit the engine's permissions. In the new setup, the Grafana server runs with its own service account with minimal privileges (no RBAC at all).
- ~~The second reason is scalability. The `stdio` type can only be configured through environment variables, so you can't spin up multiple tool servers that point to different Grafana instances. By switching to the `http` type, we can set the appropriate headers in ToolServer configuration (X-Grafana-URL, X-Grafana-API-Key [link to the code](https://github.com/grafana/mcp-grafana/blob/main/mcpgrafana.go#L27-L28)), and reuse the same deployment for many instances.~~

> 2. Why did you decide to move the other 2 tool-servers into `contrib`. I think we should probably add a `helm/tools` folder for those

Sure thing. My reasoning was as follows: each of these MCP servers lives in its own upstream repo. Since they're external tools, I treated them consistently and placed their chats in contrib/tools. Just to have them all in one place. So now, whenever a user wants to deploy a separate instance of the doc2vec, they can easily locate the helm-chart for that.
But if you'd rather keep them under a dedicated helm/tools dir, I can shift them there.

python/src/autogenstudio/web/app.py
sbx0r: Great!
So I'll remove this section entirely then.

.github/workflows/ci.yaml
```diff
@@ -79,10 +79,12 @@ jobs:
           kubectl wait --for=condition=Accepted  agents.kagent.dev -n kagent --all --timeout=60s
 
       - name: Run e2e tests
+        working-directory: go
         run: |
           export KAGENT_API_URL="http://$(kubectl get svc -n kagent kagent -o jsonpath='{.status.loadBalancer.ingress[0].ip}'):8083/api"
           echo $KAGENT_API_URL
-          cd go
+          export KAGENT_A2A_URL="http://${GW_IP}:8083/api/a2a"
```
EItanya: Is this needed?

sbx0r: Not something that we cannot live without, but definitely a good practice

helm/kagent/templates/engine-configmap.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
```
EItanya: Can we move these back into the deployment directly, there aren't any secrets in here or anything like that so it's just unecessary obfuscation

helm/kagent/templates/controller-configmap.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
EItanya: Same comments about moving these into the deployment

helm/kagent/templates/ui-configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
EItanya: Same comment about moving this into the deployment

helm/kagent-all-in-one/templates/deployment.yaml
```diff
@@ -165,9 +165,6 @@ spec:
               value: {{ .Values.otel.tracing.exporter.otlp.timeout | quote }}
             - name: OTEL_EXPORTER_OTLP_TRACES_INSECURE
               value: {{ .Values.otel.tracing.exporter.otlp.insecure | quote }}
```
EItanya: Why did you remove this?

helm/kagent-all-in-one/tests/modelconfig_test.yaml
EItanya, line 1: This test should also be in the standard chart tests?

Makefile
```diff
@@ -257,6 +268,45 @@ helm-install-provider: helm-version check-openai-key
 		--timeout 5m       \
 		--kube-context kind-$(KIND_CLUSTER_NAME) \
 		--wait \
+		--set ui.service.type=LoadBalancer \
+		--set ui.image.registry=$(RETAGGED_DOCKER_REGISTRY) \
+		--set ui.image.tag=$(UI_IMAGE_TAG) \
+		--set controller.image.registry=$(RETAGGED_DOCKER_REGISTRY) \
+		--set controller.image.tag=$(CONTROLLER_IMAGE_TAG) \
+		$(if $(CI),--set controller.service.type=LoadBalancer) \
```
EItanya: Can't this just always be LB

Makefile
```diff
@@ -257,6 +268,45 @@ helm-install-provider: helm-version check-openai-key
...
+		--set engine.image.registry=$(RETAGGED_DOCKER_REGISTRY) \
+		--set engine.image.tag=$(APP_IMAGE_TAG) \
+		--set providers.openAI.apiKey=$(OPENAI_API_KEY) \
+		--set providers.azureOpenAI.apiKey=$(AZUREOPENAI_API_KEY) \
+		--set providers.anthropic.apiKey=$(ANTHROPIC_API_KEY) \
+		--set providers.default=$(KAGENT_DEFAULT_MODEL_PROVIDER) \
+		$(if $(OPENAI_API_KEY),,--set querydoc.enabled=false) \
```
EItanya: I don't think we need to do these `ifs`. This is all for local anyway, plus the components "should" spin up regardless of whether or not there is a key

EItanya (commented):
I love the direction, but I have a couple issues/questions about the curent approach:

1. Currently upgrading from the existing chart to this fails. This isn't necessarily a blocker, but could be nice to make this seamless.
```Error: UPGRADE FAILED: cannot patch "kagent" with kind Deployment: Deployment.apps "kagent" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"controller", "app.kubernetes.io/instance":"kagent", "app.kubernetes.io/name":"kagent"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable```.
2. I'm not sure the complexity of creating separate helm charts for the querydoc and kagent tools is worth it just yet. It complicates this PR quite a bit, and if possible I think we should back this part out for now.

File: Makefile (hunk 9 above)
Comment (sbx0r) [reply Makefile]: could be :)
The question is - should we publish our API server to the outside world by default?

File: Makefile (hunk 10 above)
Comment (sbx0r) [reply Makefile]: I've seen that from version 1.1.10 querytool container won't fail without this key. When I started this refactor it just failed. same as the tool-server.

I'm removing this 'ifology'

File: helm/kagent-all-in-one/templates/deployment.yaml (hunk 8 above)
Comment (sbx0r) [reply helm/kagent-all-in-one/templates/deployment.yaml]: Good catch! Had to miss this one during the reabase

File: helm/kagent-all-in-one/tests/modelconfig_test.yaml
Comment (sbx0r) [reply helm/kagent-all-in-one/tests/modelconfig_test.yaml:1]: yea, I think we should test this logic in both.

File: helm/kagent/templates/engine-configmap.yaml (hunk 5 above)
Comment (sbx0r) [reply helm/kagent/templates/engine-configmap.yaml]: Keeping configuration in a ConfigMap is considered a best practice

Imagine deploying Kagent via Argo CD: instead of digging through Deployment manifests to understand environment variables, you simply open the ConfigMap.

Updating the config is also easier:

- **Embedded in the Pod template:** - the more env vars you have the uglier the manifest looks.
- **In a ConfigMap:** the Deployment only references it, and a Pod restart picks up the new values. _(We even have a reloader that automatically restarts Pods when their ConfigMap changes.)_

I know that currently we have not so many variables here. But my intention was to prepare a solid ground for more advanced configs.
Maybe we could strike a middle ground: keep the one‑off variables inline, **but create ConfigMaps for settings that can be reused across Deployments - for example, shared OpenTelemetry (OTel) configs**.
This gives us deduplication and clarity without over‑engineering every manifest.

File: .github/workflows/ci.yaml (hunk 4 above)
Comment (EItanya) [reply .github/workflows/ci.yaml]: What do you mean? What's this env var used for?

File: Makefile (hunk 9 above)
Comment (EItanya) [reply Makefile]: This is just for dev?

File: helm/kagent/templates/engine-configmap.yaml (hunk 5 above)
Comment (EItanya) [reply helm/kagent/templates/engine-configmap.yaml]: Ok interesting, I've never done it like this but I buy that. And just to make sure, the deployment will auto roll-out if these values change?

**Comment by sbx0r** [comment]
> I love the direction, but I have a couple issues/questions about the curent approach:

Thanks!

---

> 1. Currently upgrading from the existing chart to this fails. This isn't necessarily a blocker, but could be nice to make this seamless.
>    `Error: UPGRADE FAILED: cannot patch "kagent" with kind Deployment: Deployment.apps "kagent" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"controller", "app.kubernetes.io/instance":"kagent", "app.kubernetes.io/name":"kagent"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable`.

**Root cause**
The new controller Deployment keeps the old name **kagent**, but its selector now includes `app.kubernetes.io/component=controller`. Because `.spec.selector` is immutable, Helm cannot patch the existing Deployment and the upgrade fails.
```yaml
# old all‑in‑one Deployment
app.kubernetes.io/name: kagent
app.kubernetes.io/instance: kagent

# new controller Deployment (same name)
app.kubernetes.io/name: kagent
app.kubernetes.io/instance: kagent
app.kubernetes.io/component: controller   # must stay unique per component
```
**The** extra component **label is mandatory**: without it each Deployment would contend for the same Pods, Services would lose endpoints, and many other issues may happen.

There are few potential solutions to that:
N |Option | Description | Downtime | Pros | Cons
-- | -- | -- | -- | -- | --
1 | **Pre‑upgrade hook** | Job deletes the old Deployment, Helm then creates the new one. | Seconds | Fully automated, no user action. | Adds a hook + RBAC in this PR (more code).
2 | 🟢 **Rename Deployment** | e.g. `kagent‑controller` instead of `kagent`. | None (rolling) | No hook needed; old controller keeps running until new one is ready. | None.
3 | **Force replace** | helm upgrade --install --force --reuse-values | Full restart | One‐liner for operators. | Nuke 💣.
4 | **Manual delete** | Add a note to release notes | Short | Zero code change. | Requires operator action, not seamless.

I would recommend `2nd` option. WDYT?

---

> 2. I'm not sure the complexity of creating separate helm charts for the querydoc and kagent tools is worth it just yet. It complicates this PR quite a bit, and if possible I think we should back this part out for now.

Yeah. You're right that it might look a bit too complicated but:
- `querydoc`, `tool-server` and `kagent` are living in separate repositories, so it have independent lifecycles, so ultimately **each should live in its own chart alongside its source code**.  
- That way a change in, say, `querydoc` won't force a version bump of the entire umbrella chart, and we can release each component on its own cadence.  
- To keep this PR focused, I can put everything in one chart for now (`helm/kagent/charts`). 
- Move it to its upstream repositories in next iteration

helm/kagent/templates/engine-configmap.yaml
sbx0r: **If you modify the helm values - Yes**. Because I added this to the pod's template `metadata.annotations`:
```yaml
checksum/configmap: {{ include (print $.Template.BasePath "/controller-configmap.yaml") . | sha256sum }}
```
So whenever you change the config in the helm values, this value changes and this triggers the Deployment rollout.

---

**If you modify CM directly - No**. (via `kubectl edit cm ...`) But the question is how often you need to modify configuration manually 😛

Makefile
sbx0r: you're right!
Lemme remove that `if`

.github/workflows/ci.yaml
sbx0r: ahhh I thought you pointed the removed line :D

You're right. We don't need this env anymore. Removing it

EItanya:
> > I love the direction, but I have a couple issues/questions about the curent approach:
> 
> Thanks!
> 
> > 1. Currently upgrading from the existing chart to this fails. This isn't necessarily a blocker, but could be nice to make this seamless.
> >    `Error: UPGRADE FAILED: cannot patch "kagent" with kind Deployment: Deployment.apps "kagent" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"controller", "app.kubernetes.io/instance":"kagent", "app.kubernetes.io/name":"kagent"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable`.
> 
> **Root cause** The new controller Deployment keeps the old name **kagent**, but its selector now includes `app.kubernetes.io/component=controller`. Because `.spec.selector` is immutable, Helm cannot patch the existing Deployment and the upgrade fails.
> 
> ```yaml
> # old all‑in‑one Deployment
> app.kubernetes.io/name: kagent
> app.kubernetes.io/instance: kagent
> 
> # new controller Deployment (same name)
> app.kubernetes.io/name: kagent
> app.kubernetes.io/instance: kagent
> app.kubernetes.io/component: controller   # must stay unique per component
> ```
> 
> **The** extra component **label is mandatory**: without it each Deployment would contend for the same Pods, Services would lose endpoints, and many other issues may happen.
> 
> There are few potential solutions to that:
> 
> N	Option	Description	Downtime	Pros	Cons
> 1	**Pre‑upgrade hook**	Job deletes the old Deployment, Helm then creates the new one.	Seconds	Fully automated, no user action.	Adds a hook + RBAC in this PR (more code).
> 2	🟢 **Rename Deployment**	e.g. `kagent‑controller` instead of `kagent`.	None (rolling)	No hook needed; old controller keeps running until new one is ready.	None.
> 3	**Force replace**	helm upgrade --install --force --reuse-values	Full restart	One‐liner for operators.	Nuke 💣.
> 4	**Manual delete**	Add a note to release notes	Short	Zero code change.	Requires operator action, not seamless.
> I would recommend `2nd` option. WDYT?
> 
> > 2. I'm not sure the complexity of creating separate helm charts for the querydoc and kagent tools is worth it just yet. It complicates this PR quite a bit, and if possible I think we should back this part out for now.
> 
> Yeah. You're right that it might look a bit too complicated but:
> 
> * `querydoc`, `tool-server` and `kagent` are living in separate repositories, so it have independent lifecycles, so ultimately **each should live in its own chart alongside its source code**.
> * That way a change in, say, `querydoc` won't force a version bump of the entire umbrella chart, and we can release each component on its own cadence.
> * To keep this PR focused, I can put everything in one chart for now (`helm/kagent/charts`).
> * Move it to its upstream repositories in next iteration

I agree, approach 2 makes sense to me as well!

EItanya (dismissed):

PR #601, opened 2025-07-03
Add autogen loglevel to app container

antonioberben:
The standard Python logging mechanism only works if handlers are not set.

There are two ways:

Add the autogen way of handling logging level (in this PR)

Or, the alternative is to override all the handlers, forcing the same log_level to all handlers with (not in this PR):

```
logging.basicConfig(level=LOGLEVEL, force=True)
```

Which one do you think is a better approach? IMO this is better. But looking for second opinions.


antonioberben:
@peterj, or who might be concerned, are you ok merging this PR?

EItanya (approved):

PR #606, opened 2025-07-07
Add GitHub Action to run GitIngest on every release

### LINKED ISSUE #557: [FEATURE] gitingest on every release
### 📋 Prerequisites

### 📝 Feature Summary

Create LLM friendly digest of kagent 

Have LLM friendly version of kagent

### 💡 Proposed Solution

Add to github workflows
https://github.com/cyclotruc/gitingest

### 🎯 Affected Service(s)

### 📚 Additional Context

_No response_

cnu1812:
A GitHub Actions workflow that automatically runs [GitIngest](https://github.com/gitingest/gitingest) whenever a new release is published.

Fixes #557 

What’s included:
New workflow: `.github/workflows/gitingest.yml`

Triggers on `release.published` events

Uses Python 3.10 and installs `gitingest==0.4.2`

Logs GitIngest output to `ingest.log`

Uploads log as a downloadable artifact for debugging or auditing


cnu1812:
@marcinkubica Please review!

cnu1812:
Thanks for the suggestion @EItanya , I’ve now moved the GitIngest step into the existing tag.yml workflow under the release job. It runs after the CLI and Helm chart builds, and the ingest.log is now uploaded as part of the GitHub release assets via softprops/action-gh-release. Let me know if you'd like it placed earlier in the job or handled differently!

.github/workflows/tag.yaml
```diff
@@ -105,10 +105,27 @@ jobs:
...
+      with:
+        python-version: '3.10'
+
+    - name: Install GitIngest
+      run: pip install gitingest==0.4.2
+
+    - name: Run GitIngest
+      run: |
+        echo "⚙️ Running GitIngest..."
+        gitingest --repo-url https://github.com/kagent-dev/kagent > ingest.log
```
EItanya: Why are we using the repo URL and not just the code since we already have it here, and it's the correct tag.

EItanya (commented):
Can we do this in a parallel job and then copy the artifact over to the final release job.

.github/workflows/tag.yaml
cnu1812: Thanks for the feedback!

Regarding:

> "Why are we using the repo URL and not just the code since we already have it here, and it's the correct tag?"

Totally fair point. The current script was originally designed to work independently; it could ingest any public repo without needing to be inside a GitHub Action. But I agree: since this is running within the workflow and the repo is already checked out, we could simplify by using the local files directly. Happy to switch to that if preferred.

> "Can we do this in a parallel job and then copy the artifact over to the final release job?"

That makes sense for modularity and performance. In the current setup, I kept it in a single job for simplicity and traceability during early iterations. If you’d prefer, I can refactor this to split the ingest step into a separate job and pass the output via upload-artifact / download-artifact as suggested.

EItanya: That's a good point, let's do this for now: 
1. switch to ingesting the local files rather than the URL
2. Name the output file `gitingest.log`
3. Skip moving this into a separate job for now, as you mentioned simplicity is better for now

cnu1812: - Switched GitIngest to run on the local checkout instead of using --repo-url
- Output is now written to gitingest.log 
- Kept it within the same release job as discussed, to keep things simple for now

EItanya:
Hey there, DCO seems to be failing. Some of your commits must not be signed.

marcinkubica:
@EItanya thanks for leading on this PR, I'm beyond busy atmo :|

@cnu1812 you'd need to squash all your commits and put a single one with sign-off

marcinkubica:
@cnu1812 squash entire branch to a single, signed off commit ;) adding new won't help if the history has even one commit unsigned

cnu1812:
Opened new PR #648, can close this.

marcinkubica:
One thing though. With this we will be saving llm digest as a release artifact - which is fine, however the general idea is gitingest is added to the branch. I'm attaching the file taken from https://gitingest.com/

[kagent-dev-kagent.txt](https://github.com/user-attachments/files/21420056/kagent-dev-kagent.txt)

so we'd need run. a of gitingest, add digest file, commit it.

lastly - do we go with Gitingest default filename `digest.txt`, or want own `llmdigest.txt`, `gitdigest.txt` ?

PR #609, opened 2025-07-08
Migrate DB logic away from runtime and into go controller

EItanya:


dimetron:
Can we avoid CGO please https://github.com/cvilsmeier/go-sqlite-bench

dimetron:
We could check this as well https://gitlab.com/cznic/sqlite

dimetron:
Benchmarks https://pkg.go.dev/modernc.org/sqlite-bench

EItanya:
Closing in favor of #611 because of DCO

PR #644, opened 2025-07-24
feat(env): auto set GOMAXPROCS by go.uber.org/automaxprocs

dongjiang1989:
Auto set GOMAXPROCS.
add dependency on [go.uber.org/automaxprocs](https://github.com/uber-go/automaxprocs)


EItanya:
Thanks so much for the contribution, this definitely makes sense!

PR #648, opened 2025-07-24
add gitingestion with a signoff message

### LINKED ISSUE #557: [FEATURE] gitingest on every release

cnu1812:
Fixes #557, solves the #606 DCO issue

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -105,6 +105,22 @@ jobs:
         fi
         make build-cli
         make helm-version
+
+   # GitIngest setup and run
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+
+    - name: Install GitIngest
+      run: pip install gitingest==0.4.2
+
+    - name: Run GitIngest
+      run: |
+        echo "Running GitIngest using local checkout..."
+        gitingest . > gitingest.log
+        echo " GitIngest completed."
+
     - name: Release
       uses: softprops/action-gh-release@v2
       if: startsWith(github.ref, 'refs/tags/')
@@ -112,3 +128,4 @@ jobs:
         files: |
           go/bin/kagent-*
           kagent-*.tgz
+          gitingest.log


.github/workflows/tag.yaml
```diff
@@ -105,10 +105,27 @@ jobs:
...
         make helm-version
+
+   # GitIngest setup and run
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+
+    - name: Install GitIngest
+      run: pip install gitingest==0.4.2
```
EItanya: I just tried this locally and this release does not exist

```
➜  go git:(eitanya/adk) ✗ pip install gitingest==0.4.2
Defaulting to user installation because normal site-packages is not writeable
ERROR: Could not find a version that satisfies the requirement gitingest==0.4.2 (from versions: 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5)

ERROR: No matching distribution found for gitingest==0.4.2
```

.github/workflows/tag.yaml
```diff
@@ -105,10 +105,27 @@ jobs:
         fi
         make build-cli
         make helm-version
+
+   # GitIngest setup and run
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
```
EItanya, line 113: Why version 3.10?

EItanya (changes requested):
This will fail

.github/workflows/tag.yaml
cnu1812: oh, yeah, it is supported to be 0.1.2 , but will stick with the latest 0.1.5

.github/workflows/tag.yaml
cnu1812: went with Python 3.10 here cause it’s a very stable and well-supported version that works smoothly with gitingest. I tested it locally and in CI, and 3.10 avoids any potential issues that might pop up with 3.11+ . Also the workflow doesn’t need anything specific from newer versions, sticking to 3.10 felt like the safest and most predictable choice for now.

marcinkubica:
One thing though. With this we will be saving llm digest as a release artifact - which is fine, however the general idea is gitingest is added to the branch. I'm attaching the file taken from https://gitingest.com/

[kagent-dev-kagent.txt](https://github.com/user-attachments/files/21420056/kagent-dev-kagent.txt)

so we'd need a run of gitingest, add digest file, commit it. So it would need to run on pr close. Depending on decision if we would like to include it.

lastly - do we go with Gitingest default filename `digest.txt`, or want own `llmdigest.txt`, `gitdigest.txt` ?

Also, I like the idea of saving with pipe cause it gives token count for the entire file. However it doesn't work because all gitingest is outputting is the stats. @cnu1812 do you have file somewhere? Anyways we could combine those.


```
❯ gitingest . > myfile.log && cat myfile.log
Analyzing source, output will be written to 'digest.txt'...
Analysis complete! Output written to: digest.txt

Summary:
Directory: <dir>
Files analyzed: 299

Estimated tokens: 379.0k
```

cnu1812:
Totally agree that committing the digest file directly to the branch (rather than just attaching it to the release) is a more intentional and useful long-term approach. Running GitIngest on PR close and versioning the digest sounds great to me.
- Slight lean toward `llmdigest.txt` but happy to go with whatever the team prefers.

One quick question: where would we prefer to store the digest file in the repo?

marcinkubica:
I'm using defaults, saves to `digest.txt`, but that's my personal preference (I can't be bothered to enter custom filename each time :) ) `llmdigest.txt` also sounds fine.

one thing - digest has no option to add header, and number of files and estimated tokens are useful, that's when gitingest runs
```
❯ gitingest 
Analyzing source, output will be written to 'digest.txt'...
Analysis complete! Output written to: digest.txt

Summary:
Directory: cmd
Files analyzed: 14

Estimated tokens: 17.8k
```

also, for simplicity why not have all of that as .github/actions/gitingest/action.yaml and we can call it from whatever workflow we'd choose. Also needs to "sign off" the commit (which is as much as below in the commit message)
```
add git ingest 

Signed-off-by: Gitingest Github Actions <gitingest@kagent.dev>
```

EItanya:
closing for stale

PR #649, opened 2025-07-24
feat(adk): transition to ADK

EItanya:
This PR move the backend agent engine of kagent from autogen to ADK

TODO:
- Finish implementing model providers
- Decide what to do about `Memory`
- Fix UI to use A2A directly


dimetron:
Thanks @EItanya  - I've installed it and have few observations i'd like to discuss. 

- 1. it's kind of strange if I will now run helm deploy - there is a broken relation between helm deployment <--> agent containers 

- 2. resources - now we run a separate pod and all python apps over consume memory and cpu. I suggest we make resource limits to have a clear picture and expectations how each agent should consume
 
- 3. deployment spec in Agent Spec - looks like a duplicate or v1.DeploymentSpec 

- 4. I suggest we explore a possibility to have a separate python app SDK runtime container and Python generated app.

Lets discuss: 
- Can we keep using Helm for templates for all default agents   ?
- Do we need an option to consolidate default agents as one pod ?
- Can we also create deployments for BYOAgent from helm values or docker ?
  Lets say we can have a standard docker format to package agents
  and then run time on our side to plug in
  https://github.com/docker/compose-for-agents/blob/main/adk/agents/agent.py

go/controller/internal/a2a/a2a_reconciler.go
```diff
@@ -28,43 +29,39 @@ type A2AReconciler interface {
...
 	return &a2aReconciler{
-		a2aTranslator: NewAutogenA2ATranslator(a2aBaseUrl, autogenClient, dbService),
-		autogenClient: autogenClient,
-		a2aHandler:    a2aHandler,
+		a2aHandler: a2aHandler,
+		a2aBaseUrl: a2aBaseUrl,
 	}
 }
 
 func (a *a2aReconciler) ReconcileAutogenAgent(
```
yuval-k: nit: not autogen

go/controller/internal/reconciler/reconciler.go
```diff
@@ -0,0 +1,761 @@
...
+			req.NamespacedName.String(), err)
+	}
+
+	reconcileLog.Info("Agent was deleted", "namespace", req.Namespace, "name", req.Name)
+	return nil
+}
+
+func (a *kagentReconciler) handleExistingAgent(ctx context.Context, agent *v1alpha1.Agent, req ctrl.Request) error {
+	isNewAgent := agent.Status.ObservedGeneration == 0
+	isUpdatedAgent := agent.Generation > agent.Status.ObservedGeneration
```
yuval-k: you may also want to check the uid here

go/controller/translator/adk_api_translator.go
```diff
@@ -0,0 +1,730 @@
...
+		return nil, err
+	}
+
+	hash := sha256.Sum256(byt)
+	outputs.ConfigHash = binary.BigEndian.Uint64(hash[:8])
+
+	outputs.ConfigMap.Data["config.json"] = string(byt)
+	// Make sure the deployment is redeployed if the config changes
+	outputs.Deployment.Labels["config.kagent.dev/hash"] = fmt.Sprintf("%d", outputs.ConfigHash)
+	outputs.Deployment.Spec.Template.Labels["config.kagent.dev/hash"] = fmt.Sprintf("%d", outputs.ConfigHash)
```
yuval-k: why is label used here and not anotatoin? i think `k rollout` uses annotation

go/controller/translator/adk_api_translator.go
```diff
@@ -0,0 +1,730 @@
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Service",
+		},
+		ObjectMeta: objMeta,
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app":     "kagent",
+				"kagent":  agent.Name,
+				"version": "v1alpha1",
```
yuval-k: why do we need version as label?

go/controller/translator/adk_api_translator.go
```diff
@@ -0,0 +1,730 @@
...
+				FieldPath: "metadata.namespace",
+			},
+		},
+	})
+
+	return &v1alpha1.DeploymentSpec{
+		Replicas: ptr.To(int32(1)),
+		Container: corev1.Container{
+			Name:            "kagent",
+			Image:           fmt.Sprintf("cr.kagent.dev/kagent-dev/kagent/app:%s", version.Get().Version),
```
yuval-k: can the prefix be configurable, for people mirroring images?

go/controller/translator/adk_api_translator.go
```diff
@@ -0,0 +1,730 @@
...
+
+func (a *adkApiTranslator) translateDeclarativeAgent(ctx context.Context, agent *v1alpha1.Agent, state *tState) (*adk.AgentConfig, []corev1.EnvVar, error) {
+
+	model, envVars, err := a.translateModel(ctx, agent.Namespace, agent.Spec.ModelConfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cfg := &adk.AgentConfig{
+		KagentUrl:   fmt.Sprintf("http://kagent.%s.svc.cluster.local:8083", common.GetResourceNamespace()),
```
yuval-k: consider: remove `cluster.local` suffix for compatability with clusters not using this suffix? 
another option is to detect the suffix

go/internal/utils/common.go
```diff
@@ -193,6 +194,26 @@ func GetModelConfig(
 	return modelConfigObj, nil
 }
 
+func UpsertObject(ctx context.Context, kube client.Client, obj client.Object) error {
```
yuval-k: where is this used? if not used, remove?

python/src/kagent/a2a.py
```diff
@@ -0,0 +1,115 @@
...
+
+    async def build(
+        self,
+        params: MessageSendParams | None = None,
+        task_id: str | None = None,
+        context_id: str | None = None,
+        task: Task | None = None,
+        context: ServerCallContext | None = None,
+    ) -> RequestContext:
+        if not context:
```
yuval-k: i assume in the future we will want to grab the user from the session / request. is that going to be possible?

go/controller/internal/reconciler/reconciler.go
EItanya: We don't have the old one to check against, we just have the current one. This logic is only used for logs anyway so gonna just remove it.

go/controller/translator/adk_api_translator.go
EItanya: Those don't actually need to be there, they were just for me to debug

go/controller/translator/adk_api_translator.go
EItanya: We don't, I will remove

go/controller/translator/adk_api_translator.go
EItanya: Definitely, mind if I add in a follow-up?

python/src/kagent/a2a.py
EItanya: Yea it will definitely be possible, A2A has support for lots of auth schemes, and we can also add our own if we need

peterj:
I was trying to fetch all agents and I am getting this:

```
❯  curl localhost:8083/api/agents
{"error":"Agent not found: failed to get model: record not found"}
```

Let me know when this is ready, so I can do the UI portion.

go/controller/internal/controller/agent_controller.go
```diff
@@ -29,31 +29,35 @@ import (
...
+func (r *AgentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(controller.Options{
 			NeedLeaderElection: ptr.To(true),
 		}).
 		For(&agentv1alpha1.Agent{}).
-		Named("autogenagent").
+		Owns(&appsv1.Deployment{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&corev1.Service{}).
```
yuval-k, line 60:
```suggestion
		Owns(&corev1.Service{}).
		Owns(&corev1.ServiceAccount{}).
```

EItanya:
Closing in favor of #671 as getting DCO to pass on this PR will be nearly impossible at this point.

PR #650, opened 2025-07-25
feat(env): Add automatic memory limit handling

dongjiang1989:
add auto GOMEMLIMIT setting

### DIFF
--- a/go/controller/cmd/main.go
+++ b/go/controller/cmd/main.go
@@ -65,6 +65,8 @@ import (
 	// +kubebuilder:scaffold:imports
 )
 
+const defaultMemlimitRatio = 0.0
+
 var (
 	scheme          = runtime.NewScheme()
 	setupLog        = ctrl.Log.WithName("setup")
@@ -102,6 +104,7 @@ func main() {
 	var databasePath string
 	var databaseType string
 	var databaseURL string
+	var memlimitRatio float64
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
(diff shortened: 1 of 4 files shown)


dongjiang1989:
cc @EItanya PTAL

EItanya (changes requested):
So I've been thinking about this PR and your previous one a little more, and I think they may be overkill tbh. I've used the following approach before which has been super simple and worked well. 

```
         - name: GOMEMLIMIT
            valueFrom:
              resourceFieldRef:
                resource: limits.memory
                divisor: "1"
          - name: GOMAXPROCS
            valueFrom:
              resourceFieldRef:
                resource: limits.cpu
                divisor: "1"
```

Here it is in Istio: https://github.com/istio/istio/blob/b6df261b7c9f66c10b0d17b02f9c9b9ebb546033/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml#L209-L218

dongjiang1989:
thanks @EItanya 

This method is also effective. However, there is a drawback: under `VPA` adjustment, the Go runtime configuration cannot be dynamically adjusted. Just start it once for the first time.

dongjiang1989:
> thanks @EItanya
> 
> This method is also effective. However, there is a drawback: under `VPA` adjustment, the Go runtime configuration cannot be dynamically adjusted. Just start it once for the first time.

howardjohn:
> thanks @EItanya
> 
> This method is also effective. However, there is a drawback: under `VPA` adjustment, the Go runtime configuration cannot be dynamically adjusted. Just start it once for the first time.

I don't think thats true. a VPA change triggers a new pod, which will have a freshly computed value.

I agree  with @EItanya on the approach

dongjiang1989:
> > thanks @EItanya
> > This method is also effective. However, there is a drawback: under `VPA` adjustment, the Go runtime configuration cannot be dynamically adjusted. Just start it once for the first time.
> 
> I don't think thats true. a VPA change triggers a new pod, which will have a freshly computed value.
> 
> I agree with @EItanya on the approach

@howardjohn 

It might be that my description wasn't clear enough.  When a VPA change with in place pod resize.

howardjohn:
you are referring to https://kubernetes.io/blog/2025/05/16/kubernetes-v1-33-in-place-pod-resize-beta/ yeah?

Given we are only calling SetMemLimit, if the memory limit actually does change during the pod lifetime, this change won't handle it either...?

EItanya:
closing for stale

Issue #633, opened 2025-07-21
[FEATURE] Let us use existing secrets in helm chart


nodnarbnitram:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add support to not create the secret and allow us to use an existing secret

### ❓ Problem Statement / Motivation

- Current implementation doesn't allow for proper GitOps with secrets being handled by external-secrets.

### 💡 Proposed Solution

- Add `createSecret: false` or `useExistingSecret: true`

### 🔄 Alternatives Considered

- Current implementation prevents work around since it hard-codes it.

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

PR #652, opened 2025-07-26
introduces createApiKeySecret for #633

day0ops:
- adding flag `createApiKeySecret` to override the default secret injection. Users can opt out and inject their own secrets using external tools such as external-secret operator.
- also fixes a minor bug not allowing users to set a custom secret name.


EItanya:
Hey @day0ops, I believe there was an issue about exactly this created a few months ago, and I'm fairly certain the takeaway was that we could just use the missing `apikey` as a signal not to create the secret, rather than create a new helm variable. What do you think?

day0ops:
I'm not opposed to that. I also see a big refactor happening in another PR. Will close this

PR #658, opened 2025-07-29
fix for 657

day0ops:
adds the missing namespace to fix the `404`.


day0ops:
@peterj @EItanya any idea why the API key isnt getting picked up from the GA secrets ?

```
Error: OPENAI_API_KEY environment variable is not set
Please set it with: export OPENAI_API_KEY=your-api-key
make: *** [Makefile:84: check-openai-key] Error 1
```

PR #671, opened 2025-07-31
feat(all): add support for ADK

EItanya:
This is the official PR to add support for ADK to kagent :tada:

This PR modifies many core parts of the system:
1. `kagent-engine` has been removed
2.  Each agent now gets its own deployment which runs ADK
3. An SDK will be released to allow users to BYO their own ADK agents


yuval-k (approved):
approving based off of previous PR comments

nashtsai:
hi, I am curious that is there any reason for switching from Autogen (https://kagent.dev/docs/kagent/concepts/architecture has yet updated) to ADK?

PR #673, opened 2025-07-31
fix(controller): don't fail entire call to list agents based on results for one/more agents

onematchfox:
Currently, every agent must exist in both Kubernetes and the database for this call to succeed. Thus, if for whatever reason an agent fails to reconcile on creation, any calls to this endpoint will fail. This, in turn, breaks all agents in the UI. This solution is a bit of a hack/workaround, and I do think issues should be surfaced; however, it at least ensures that the UI stays functional in this scenario for the time being.

image

image

```
$ k get agents                
NAME           ACCEPTED   MODELCONFIG
k8s-agent      True       default-model-config
promql-agent   False      default-model-config
```


onematchfox:
Problematic code was commented out in #671

Issue #466, opened 2025-06-01
[BUG] Add buffer size limits to prevent memory leaks in streaming


abhinav-1305:
While working with the chat interface during long-running code analysis or security audit sessions, I noticed a critical `memory leak`. The problem showed up when handling large streaming responses - memory usage kept growing over time, eventually slowing down the browser.

Digging into it, the root cause was unbounded buffer growth in the message processing logic. Here's the original code that caused the issue:

https://github.com/kagent-dev/kagent/blob/33a48ede61be68c84f6adcfddde09db41aeb1ea7/ui/src/components/chat/ChatInterface.tsx#L245

There’s no `limit` on `buffer size` or total message size, so memory just keeps accumulating.

To fix this, I’m thinking of adding proper buffer size limits and chunking logic to manage memory more efficiently. Something along these lines:

```  typescript
const MAX_BUFFER_SIZE = 1024 * 1024;       
const CHUNK_SIZE = 16 * 1024;             
const MAX_MESSAGE_SIZE = 10 * 1024 * 1024; 

while (true) {
  const { value, done } = await reader.read();
  processedSize += value.length;

  if (processedSize > MAX_MESSAGE_SIZE) {
    throw new Error("Message size exceeds allowed limit");
  }

  if (buffer.length > MAX_BUFFER_SIZE) {
    const lines = buffer.split('\n');
    buffer = lines.pop() || '';

    if (buffer.length > MAX_BUFFER_SIZE) {
      buffer = buffer.slice(-CHUNK_SIZE);
    }
  }

  // Will also add cleanup logic when done
}
```

PR #676, opened 2025-08-01
fix: prevent memory leaks in streaming chat responses

Vijaykv5:
Fixes : #466 

### **What this PR does :** 

Added buffer size limits and memory management to prevent unbounded growth during long-running streaming sessions.
It Implements`1MB buffer limit`, `16KB chunk truncation`, and `10MB total message size` capacity to maintain browser performance during code analysis.


EItanya:
Closing for stale

PR #677, opened 2025-08-01
fix(ui): ensure markdown URLs render correctly

onematchfox:
Regression introduced in #357 - not sure what the intention there was, though, as from what I can tell it doesn't seem related to the rest of the changes in that PR. cc @peterj 

Currently:

image

vs with fix

image

(Note difference in links in bottom-left of screenshots above)

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -39,6 +39,7 @@
         "react-dom": "^18.3.1",
         "react-hook-form": "^7.58.1",
         "react-markdown": "^9.1.0",
+        "rehype-external-links": "^3.0.0",
         "remark-gfm": "^4.0.1",
         "sonner": "^2.0.5",
         "tailwind-merge": "^2.6.0",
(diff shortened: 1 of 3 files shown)


peterj:
we want a way to render links in the chat responses - that's what the `a: (props: any) => {` gives us. We want agents to be able to render links and make them clickable. Removing this isn't fixing it. One way to fix it would be to check whether the value being pass to the  `a: (props..`  actually includes a valid link or not. If it doesn't, then return a non-a element vs. an <a> if the link is valid.

onematchfox:
> we want a way to render links in the chat responses - that's what the `a: (props: any) => {` gives us. 

Apologies, but I'm not following you here, and, full disclaimer, I haven't worked with `react-markdown` before so I could be missing a piece of the puzzle. 

Isn't rendering links already part of `react-markdown`? From what I can tell, the `a` [component](https://github.com/remarkjs/react-markdown/tree/9.1.0?tab=readme-ov-file#appendix-b-components) is overwriting `react-markdown`'s default handling/rendering of links. I.e. the component wouldn't be triggered if `react-markdown` didn't _detect_ (want to render) a link in the first place. Which then begs the question as to what is wrong with `react-markdown`'s default handling of links? Did a couple more experiments and, even without the component, handling of responses with plain-text links (in case that was what you were referring to) seem to work fine.

image

Message from event stream that was used to render the above
```
{"source": "kagent__NS__argo_rollouts_conversion_agent", "models_usage": {"prompt_tokens": 1922, "completion_tokens": 23}, "metadata": {"duration": "2.4396862983703613", "created_at": "1754043755.4115524"}, "content": "The official documentation link for Argo Rollouts is: https://argo-rollouts.readthedocs.io/", "type": "TextMessage"}
```

peterj:
the component has attributes that allows us to open the link in a new tab.

onematchfox:
> the component has attributes that allows us to open the link in a new tab.

OK, missed that. Have pushed a commit to add back that functionality based on the [official advice](https://github.com/remarkjs/react-markdown/blob/main/changelog.md#remove-linktarget) post the removal of `linkTarget`. Any objections to the extra dependency?

image

PR #678, opened 2025-08-01
chore: don't use load balancer services for local development

onematchfox:
Causes this target to timeout when running using a local kind cluster as per the current [development docs](https://github.com/kagent-dev/kagent/blob/main/DEVELOPMENT.md). Change was introduced in #516 - not sure if this was intentional or whether docs need to be updated to allow for the use of load balancer services in kind clusters (see[ comment](https://github.com/kagent-dev/kagent/pull/516#discussion_r2247297097))? 

```
helm upgrade --install kagent helm/kagent \
	--namespace kagent \
	--create-namespace \
	--history-max 2    \
	--timeout 5m       \
	--kube-context kind-kagent \
	--wait \
	--set ui.service.type=LoadBalancer \
	--set ui.image.registry=cr.kagent.dev \
	--set ui.image.tag=v0.4.5-5-g4067b28 \
	--set controller.image.registry=cr.kagent.dev \
	--set controller.image.tag=v0.4.5-5-g4067b28 \
	--set controller.service.type=LoadBalancer \
	--set engine.image.registry=cr.kagent.dev \
	--set engine.image.tag=v0.4.5-5-g4067b28 \
	--set providers.openAI.apiKey=foo \
	--set providers.azureOpenAI.apiKey= \
	--set providers.anthropic.apiKey= \
	--set providers.default=openAI \
	--set querydoc.openai.apiKey=foo \
	
Error: UPGRADE FAILED: context deadline exceeded
make: *** [Makefile:252: helm-install-provider] Error 1
```


peterj:
Unfortunately, the development doc is outdated. There have been a lot of changes recently -- the biggest one that impacts the development setup is the fact that we run the controller, tools, and the UI in separate pods. 

I think the LoadBalancer type was added a while back and it's there, so one doesn't require port-forward. But I do agree it should be documented in the development.md. 

Would you want to update those docs perhaps?

EItanya:
As I commented on the original PR, I added this because in CI and locally I use `metallb` which simplifies connecting without needing to port-forward. Not sure this works on Mac, but it's worth discussing.

onematchfox:
Thanks for the response, both. I don't have a strong opinion on how to proceed here. I just wanted to be able to get a working local dev environment in the easiest way possible, and happened to notice that the process in the docs broke at some point. So, I thought it was just easiest to fix the local environment to match the docs. 

I completely understand that there are a lot of changes being made at the moment. So, if you have other recommendations, I'm all ears. However, I'm not sure if I'm the best person to be updating the docs; someone closer to the project will have far more insight into how to work locally than myself. :sweat_smile:

onematchfox:
Took a quick look at what's done in CI and replaced my original commit with one that ensures that MetalLB is setup on the kind cluster. WOMM, but yeah, as mentioned already 
> Not sure this works on Mac

EItanya (approved):

PR #682, opened 2025-08-02
fix: correct CLI API URL to match server endpoint

Lavelee954:
- Remove '/api' suffix from default API URL to properly connect to kagent server.
- The server's version endpoint is at '/version', not '/api/version', causing
- CLI connection failures with "Error connecting to server" message.


EItanya:
Fixed in another PR

PR #689, opened 2025-08-05
[FIX ] - fixes adk performance tuning

dimetron:
- dependency versions update
- fixes adk performance tuning


python/Dockerfile
```diff
@@ -49,10 +57,9 @@ ARG VERSION
 # Install dependencies
 RUN echo "Installing dependencies..."                   \
     && uv venv --python=python$TOOLS_PYTHON_VERSION     \
-    && uv sync --locked --refresh
-
-RUN uv version ${VERSION%%-*} --package kagent \
-    && uv build --package kagent \
+    && uv sync --locked --refresh                       \
+    && uv build --package kagent-adk                    \
```
EItanya, line 62: Why did you remove `uv version`. That allows for the proper version to be reported by the Python code.

helm/tools/querydoc/templates/deployment.yaml
```diff
@@ -44,10 +44,13 @@ spec:
             httpGet:
               path: /health
               port: http
-            periodSeconds: 1
-            initialDelaySeconds: 1
+            periodSeconds: 15
```
EItanya, line 47: 15 seems like a for a startup probe, the point of this is just to run at startup, I understand for a readiness probe that runs for the whole lifetime.

EItanya (approved):

PR #700, opened 2025-08-07
fix(controller): ensure live state for agents matches desired state

onematchfox:
At present, the controller will ignore any changes to the live state of resources it owns due to the check that's being removed in this PR. This means that, for example, if the deployment for an agent is deleted, it will not be recreated until:
- The agent specification changes
- The agent is recreated
- The controller is restarted (and this is only true because the database is not persistent). 

However, removing this optimisation will increase load on the Kube API and also surfaces other issues with the controller code (e.g. controllers are currently updating objects they do not own). Opening this PR in draft for now so the issue is visible and we can discuss how best to proceed. I have also started looking into and will try to open smaller, related PRs that aim to minimise unnecessary reconciliation, thereby diminishing the impact of removing this check.

### DIFF
--- a/go/controller/internal/reconciler/reconciler.go
+++ b/go/controller/internal/reconciler/reconciler.go
@@ -458,11 +458,6 @@ func (a *kagentReconciler) upsertAgent(ctx context.Context, agent *v1alpha1.Agen
 		return fmt.Errorf("failed to store agent %s: %v", agentOutputs.Config.Name, err)
 	}
 
-	// If the config hash has not changed, we can skip the patch
-	if bytes.Equal(agentOutputs.ConfigHash[:], agent.Status.ConfigHash) {
-		return nil
-	}
-
 	for _, obj := range agentOutputs.Manifest {
 		if err := a.kube.Patch(ctx, obj, client.Apply, &client.PatchOptions{
 			FieldManager: "kagent-controller",


onematchfox:
One question I do have is, if we don't have this check in place, then do we even need to [store the hash](https://github.com/kagent-dev/kagent/blob/2824a26851177c0ef6247e8878b7e91cfa68cb46/go/controller/api/v1alpha1/agent_types.go#L134-L136) anymore?

onematchfox:
On this, unfortunately, there is going to be no getting around having to _fully_ reconcile the live state of the owned objects if we want to avoid all unnecessary/redundant API calls. However, reducing unnecessary API calls could be seen as a performance improvement that can come after ensuring that the bug is fixed. Or not :smile:

onematchfox:
Have now updated this to fully reconcile the objects owned by the agents controller with much _inspiration_ from https://github.com/open-telemetry/opentelemetry-operator. FYI @EItanya Given the API version update, I figured I'd also take the opportunity to drop the `ConfigHash` from agent status; although, to be honest, I'm not an expert on versioning Kubernetes APIs, so I'm not 100% sure on what the correct course of action is here. Happy to revert that out if needed.

I've also left the cleanup of owned objects that are no longer needed as a TODO for the sake of expediency.

onematchfox:
> At a high level though I'm not 100% sure why this is better than the configHash approach.

The issue with the config hash approach is that the live state of the resources that were created is only ever reconciled (which, in the current code, is patching) if: 
- The agent specification changes (the ConfigHash changes)
- The agent is recreated (a new ConfigHash is created)
- The controller is restarted (essentially the same as a new Agent and only because the database is not persistent)

This means that drift can occur. For example, if I manually delete the `Deployment` for an `Agent`, it will not be re-created until one of the above conditions is met. The same holds for updates and across all owned resources. Hopefully that makes more sense?

go/controller/internal/reconciler/reconciler.go
```diff
@@ -476,16 +471,93 @@ func (a *kagentReconciler) reconcileAgent(ctx context.Context, agent *v1alpha2.A
...
+		// we obtain the existing object by deep copying the desired object because it's the most convenient way
+		existing := desired.DeepCopyObject().(client.Object)
+		mutateFn := translator.MutateFuncFor(existing, desired)
+
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			_, createOrUpdateErr := controllerutil.CreateOrUpdate(ctx, a.kube, existing, mutateFn)
+			return createOrUpdateErr
+		})
+
+		if err != nil && errors.As(err, &translator.ImmutableChangeErr) {
```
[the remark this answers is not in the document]
onematchfox: Not true. If the pointer were nil, then the code would panic.

go/internal/controller/reconciler/reconciler.go
```diff
@@ -103,44 +106,19 @@ func (a *kagentReconciler) handleAgentDeletion(req ctrl.Request) error {
...
 		reason  string
 	)
-	if inputErr != nil {
+	if err != nil {
 		status = metav1.ConditionFalse
-		message = inputErr.Error()
-		reason = "AgentReconcileFailed"
-		reconcileLog.Error(inputErr, "failed to reconcile agent", "agent", utils.GetObjectRef(agent))
+		message = err.Error()
+		reason = "ReconcileFailed"
```
EItanya, line 118: Not for this PR, but we should probably define these `Reasons` as `consts` in the API types.

onematchfox: Agreed :)

go/internal/controller/reconciler/reconciler.go
```diff
@@ -424,30 +396,99 @@ func (a *kagentReconciler) reconcileRemoteMCPServerStatus(
...
+		mutateFn := translator.MutateFuncFor(existing, desired)
+
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			_, createOrUpdateErr := controllerutil.CreateOrUpdate(ctx, a.kube, existing, mutateFn)
+			return createOrUpdateErr
+		})
+
+		if err != nil && errors.As(err, &translator.ImmutableChangeErr) {
+			l.Error(err, "detected immutable field change, trying to delete, new object will be created on next reconcile", "existing", existing.GetName())
+			delErr := a.kube.Delete(ctx, existing)
```
yuval-k: it's a bit scary to delete something without informing the user. ideally we shouldn't modify immutable fields, so we have a consistent behavior.

Users might be surprised if the result of them changing a field is that pods get deleted

onematchfox: Whilst I agree that the controller ideally shouldn't modify immutable fields (it doesn't at present), the time may come when this is needed. That said, I don't see deleting as scary or surprising; these resources are owned by the controller, and it should be free to do as needed to ensure that the live state matches the desired state.

EItanya: I agree with @yuval-k  here. If we’re going to support deletes I believe we’re going to need some sort of rollout mechanism. We do own these resources, but agent downtime for an update seems like a bad idea.

onematchfox: :+1: Will rip this bit of code out now - as mentioned, the controller doesn't try to update immutable fields at present so this is really just a safe-guard against a _potential_ future problem. I'm more than happy to kick this can down the road and make it someone else's problem :laughing:

onematchfox: Addressed in https://github.com/kagent-dev/kagent/pull/700/commits/87cb91a85aa9421d018646bad491e5c82683a4e9

Issue #701, opened 2025-08-07
[BUG] Getting failed to build: invalid tag "ghcr.io/kagent-dev/kagent/controller:v0.0.0+3022b24": invalid reference format


AayushSaini101:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

Failed to build: invalid tag "ghcr.io/kagent-dev/kagent/controller:v0.0.0+3022b24": invalid reference forma

### 🔄 Steps To Reproduce

Run: make helm-install

### 🤔 Expected Behavior

No error 

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

EItanya:
Can you try this again on latest `main`. I believe @dimetron fixed this

PR #702, opened 2025-08-07
fix: helm install failure while installing the kagent local

AayushSaini101:
Close: #701


peterj:
Looks like this was already fixed yesterday: https://github.com/kagent-dev/kagent/commit/4e010d4275b8ee954c16b550afcaf1954d4a6142

PR #710, opened 2025-08-07
feat: add VertexAI machine identity support (#477)

### LINKED ISSUE #477: Implement machine identity in VertexAI provider
Right now, we expect the JSON credentials to be provided, but we should support the machine identity as well as means to auth with vertext. https://cloud.google.com/docs/authentication/set-up-adc-containerized-environment

Transcendental-Programmer:
Right now, we support only explicit credentials (via a JSON key) for VertexAI providers. This PR adds support for machine identity (ADC) as well essentially letting the provider authenticate via default credentials when no secret ref is provided.

The core logic in the translator was already set up to conditionally inject the `GOOGLE_APPLICATION_CREDENTIALS` env var only when needed, so most of the work here was wiring up the provider config and ensuring the frontend could handle the new shape.

* Added `projectID` and `location` as required fields for the VertexAI providers (Gemini + Anthropic)
* Downgraded Go from 1.24.4 to 1.24.0, some stuff wasn’t working correctly with the newer patch
* Frontend changes mostly involved updating the types, form handling, and provider config mapping
* Golden tests were added for both ADC and JSON-based auth flows

Tested both flows locally and in golden tests, both providers seem to behave as expected now. ADC skips the env var injection entirely, while JSON uses it the same as before. No regressions in existing behavior.

Leaving out screenshots since this is mostly backend/auth-related.

Still ironing out some edge cases, but this should be safe to merge once reviewed.

Fixes #477.

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -1,6 +1,6 @@
 module github.com/kagent-dev/kagent/go
 
-go 1.24.5
+go 1.24.0
 
 require (
 	github.com/abiosoft/ishell/v2 v2.0.2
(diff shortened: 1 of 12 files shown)


go/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kagent-dev/kagent/go
 
-go 1.24.5
+go 1.24.0
```
EItanya, line 3: Why this change, we just upgraded to `1.24.6`, does that fix things? In general I'm not a fan of downgrading dependencies unless there's a really good reason

EItanya:
Closing for stale

PR #713, opened 2025-08-08
perf(controller): limit unnecessary reconcilation

onematchfox:
Introduces the use of 3 [predicates](https://book.kubebuilder.io/reference/watching-resources/predicates-with-watch) that aim to limit unnecessary reconciliation within the various controllers:

1. [`GenerationChangedPredicate`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#GenerationChangedPredicate), which has been applied to the main controller objects to limit unnecessary reconciliation based on status updates and/or changes to labels/annotations. See note in 53bde269846409b69759c4dedd2aa2b8237b8246.
1. A custom `ownedObjectPredicate` which prevents reconciliation due to the creation of owned objects under the assumption that the controller will be the one who created the object in the first place.
1. [`ResourceVersionChangedPredicate`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#ResourceVersionChangedPredicate), which has been applied to all owned/watched objects. This predicate ensures that we only reconcile the owning/watching object if there is a _material_ change to the secondary resource. Which, ultimately, means that it just ignores any default update events triggered by the configured resync period (every [10 hours by default](https://github.com/xrmzju/controller-runtime/blob/5b53c3940eeb8939bdc088882a9074342c65ef5a/pkg/cache/cache.go#L107)) for all objects in the underlying cache. The assumption/motivation here is that, unless something material changes, we don't need to reconcile, and that reconciliation of the owning/watching resource is sufficient to ensure that any owned/watched objects are reconciled.

More than happy to split these changes into separate PRs if you feel that any of them deserve more discussion/consideration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ RETAGGED_APP_IMG = $(RETAGGED_DOCKER_REGISTRY)/$(DOCKER_REPO)/$(APP_IMAGE_NAME):
 #take from go/go.mod
 AWK ?= $(shell command -v gawk || command -v awk)
 TOOLS_GO_VERSION ?= $(shell $(AWK) '/^go / { print $$2 }' go/go.mod)
+export GOTOOLCHAIN=go$(TOOLS_GO_VERSION)
 
 # Version information for the build
 LDFLAGS := "-X github.com/kagent-dev/kagent/go/internal/version.Version=$(VERSION)      \
@@ -103,7 +104,7 @@ build-all: buildx-create
 create-kind-cluster:
 	docker pull kindest/node:v$(TOOLS_KIND_IMAGE_VERSION) || true
 	kind create cluster --name $(KIND_CLUSTER_NAME) --image kindest/node:v$(TOOLS_KIND_IMAGE_VERSION) --config ./scripts/kind/kind-config.yaml
-	./scripts/kind/setup-metallb.sh
+	sh ./scripts/kind/setup-metallb.sh
 
 .PHONY: use-kind-cluster
 use-kind-cluster:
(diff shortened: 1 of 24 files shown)


EItanya:
> Introduces the use of 3 [predicates](https://book.kubebuilder.io/reference/watching-resources/predicates-with-watch) that aim to limit unnecessary reconciliation within the various controllers:
> 
> 1. [`GenerationChangedPredicate`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#GenerationChangedPredicate), which has been applied to the main controller objects to limit unnecessary reconciliation based on status updates and/or changes to labels/annotations. See note in [53bde26](https://github.com/kagent-dev/kagent/commit/53bde269846409b69759c4dedd2aa2b8237b8246).
> 2. A custom `ownedObjectPredicate` which prevents reconciliation due to the creation of owned objects under the assumption that the controller will be the one who created the object in the first place.
> 3. [`ResourceVersionChangedPredicate`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#ResourceVersionChangedPredicate), which has been applied to all owned/watched objects. This predicate ensures that we only reconcile the owning/watching object if there is a _material_ change to the secondary resource. Which, ultimately, means that it just ignores any default update events triggered by the configured resync period (every [10 hours by default](https://github.com/xrmzju/controller-runtime/blob/5b53c3940eeb8939bdc088882a9074342c65ef5a/pkg/cache/cache.go#L107)) for all objects in the underlying cache. The assumption/motivation here is that, unless something material changes, we don't need to reconcile, and that reconciliation of the owning/watching resource is sufficient to ensure that any owned/watched objects are reconciled.
> 
> More than happy to split these changes into separate PRs if you feel that any of them deserve more discussion/consideration.

I Agree with all of the changes in these comments. I would like to better understand the removal of the `configHash` and the check for `conditionChanged`. I understand that these changes reduce churn, but there are still reconciles which may occur in which we don't want to update the resources. Updated cause unnecessary churn on the apiserver

onematchfox:
> I would like to better understand the removal of the `configHash` and the check for `conditionChanged` ... but there are still reconciles which may occur in which we don't want to update the resources. Updated cause unnecessary churn on the apiserver

Yes, you are correct - my bad. I thought this was no longer necessary since the `GenerationChangedPredicate` guarantees that we only reconcile the main object if its `spec` has been updated, and so the `ObservedGeneration` would need to be updated regardless. However, if a secondary resource is the source of the reconciliation and that reconciliation does not change the `Status` of the object, then yes, the update would be redundant. I'll revert that bit out.

PR #720, opened 2025-08-08
perf(controller): limit secret reconciliation based on label

onematchfox:
Currently, the controller will check whether any/every `Secret` in the cluster is used by any `ModelConfig` whenever a change is made. We can eliminate a few CPU cycles and unnecessary lookups by requiring that secrets referenced by a ModelConfig. Change would, however, require some documentation updates over at https://github.com/kagent-dev/website/ as well, so I will leave this in draft for now.


onematchfox:
Stale.

Issue #637, opened 2025-07-22
[BUG] Move to per-message structured GenAI events from prompt/completion span events


krisztianfekete:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Multiple services / System-wide issue

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

The community is working towards specifying how one should model prompts and replies here https://github.com/open-telemetry/semantic-conventions/issues/2010

Until a convention is specified, we can leverage this https://github.com/open-telemetry/semantic-conventions/pull/980, instead of using the now deprecated `gen_ai.prompt` and `gen_ai.completion` attributes.

### 🔄 Steps To Reproduce

Enable tracing and take a look at them.

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

PR #725, opened 2025-08-08
Make it possible to emit input/output as log events

krisztianfekete:
This PR makes it configurable to emit input/output messages as events if OTel logging is configured to be compliant with https://github.com/open-telemetry/semantic-conventions/pull/2179. By default, input/output is being emitted as span attributes.

Fixes https://github.com/kagent-dev/kagent/issues/637

### DIFF
--- a/helm/kagent/values.yaml
+++ b/helm/kagent/values.yaml
@@ -225,3 +225,10 @@ otel:
         endpoint: http://host.docker.internal:4317
         timeout: 15
         insecure: true
+  logging:
+    enabled: true
+    exporter:
+      otlp:
+        endpoint: http://host.docker.internal:4317
+        timeout: 15
+        insecure: true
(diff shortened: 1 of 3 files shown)


helm/kagent/values.yaml
```diff
@@ -225,3 +225,10 @@ otel:
         endpoint: http://host.docker.internal:4317
         timeout: 15
         insecure: true
+  logging:
+    enabled: true
```
EItanya: Should this be `enabled: false` by default?

python/packages/kagent/src/kagent/cli.py
```diff
@@ -9,12 +9,19 @@
 import uvicorn
 from kagent_adk import AgentConfig, KAgentApp
 from opentelemetry import trace
+from opentelemetry import _logs
```
EItanya, line 12: Do you know why all of these are private. Python can't force you not to use `_` items, but they're typically meant to be private.

EItanya (approved):

PR #735, opened 2025-08-11
feat: Bring you own agent

EItanya:
Add the ability to bring your own agent - provided as a docker image.
This is an alternative to the built in declarative agents, for agents that need more logic than the declarative option can offer.


python/Dockerfile.app
```diff
@@ -0,0 +1,16 @@
+ARG KAGENT_ADK_VERSION=latest
+ARG DOCKER_REGISTRY=ghcr.io
+FROM $DOCKER_REGISTRY/kagent-dev/kagent/kagent-adk:$KAGENT_ADK_VERSION
```
jjamroga: `FROM $DOCKER_REGISTRY/$DOCKER_REPO/kagent-adk:$KAGENT_ADK_VERSION`?

EItanya: I think for now this should do the trick, most places in our code vase we just support registry, name, and version

go/controller/api/v1alpha2/agent_types.go
```diff
@@ -24,13 +24,39 @@ import (
...
+// +kubebuilder:validation:XValidation:message="inline must be specified if type is Inline, or byo must be specified if type is BYO",rule="(self.type == 'Inline' && has(self.inline)) || (self.type == 'BYO' && has(self.byo))"
 type AgentSpec struct {
+	// +kubebuilder:validation:Enum=Inline;BYO
+	// +kubebuilder:default=Inline
+	Type AgentType `json:"type"`
+
+	// +optional
+	BYO *BYOAgentSpec `json:"byo,omitempty"`
+	// +optional
+	Inline *InlineAgentSpec `json:"inline,omitempty"`
```
yuval-k: re - naming. maybe if we have `Inline`, instead of `BYO` we should have `External`?

go/controller/internal/a2a/a2a_reconciler.go
```diff
@@ -53,17 +53,17 @@ func NewReconciler(
 func (a *a2aReconciler) ReconcileAgent(
 	ctx context.Context,
 	agent *v1alpha2.Agent,
-	adkConfig *adk.AgentConfig,
+	card server.AgentCard,
 ) error {
-	cardCopy := adkConfig.AgentCard
 	// Modify card for kagent proxy
```
yuval-k: should this comment move couple of lines below?

go/controller/internal/reconciler/reconciler.go
```diff
@@ -471,37 +480,47 @@ func (a *kagentReconciler) reconcileRemoteMCPServerStatus(
 	return nil
 }
 
-func (a *kagentReconciler) reconcileAgent(ctx context.Context, agent *v1alpha2.Agent) (*[sha256.Size]byte, error) {
+func (a *kagentReconciler) reconcileAgent(ctx context.Context, agent *v1alpha2.Agent) (*[]byte, error) {
```
yuval-k: previously this was an array, so a pointer made sense. now that it is a slice, can we remove the pointer?

python/samples/adk/basic/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+COPY  basic/ basic/
+COPY  pyproject.toml pyproject.toml
+COPY  README.md README.md
+COPY  .python-version .python-version
+COPY  uv.lock uv.lock
+
+RUN uv sync --locked --refresh
+
+ENTRYPOINT ["kagent-adk", "run", "basic", "--host", "0.0.0.0", "--port", "8080"]
```
yuval-k: can we move the entrypoint ["kagent-adk", "run"] to the base image, and then the user just needs a `CMD ["basic"]` here (assuming we can defalt host and port and remove them)

yuval-k (commented):
looks good overall!
left some comments

python/samples/adk/basic/Dockerfile
EItanya: We can, but I'm a little worried about magic 😆

yuval-k (approved):

PR #753, opened 2025-08-14
feat: make agents to be able include all tools from mcp tool server

sarco3t:
Allow agents to automatically include all tools from an MCP Tool Server, simplifying configuration and ensuring new tools are available without manual updates.

### DIFF
--- a/go/controller/api/v1alpha1/agent_types.go
+++ b/go/controller/api/v1alpha1/agent_types.go
@@ -109,7 +109,7 @@ type McpServerTool struct {
 	// The names of the tools to be provided by the ToolServer
 	// For a list of all the tools provided by the server,
 	// the client can query the status of the ToolServer object after it has been created
-	ToolNames []string `json:"toolNames,omitempty"`
+	ToolNames []string `json:"toolNames"`
 }
 
 type AnyType struct {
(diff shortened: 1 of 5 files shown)


go/controller/api/v1alpha1/agent_types.go
```diff
@@ -109,7 +109,7 @@ type McpServerTool struct {
 	// The names of the tools to be provided by the ToolServer
 	// For a list of all the tools provided by the server,
 	// the client can query the status of the ToolServer object after it has been created
-	ToolNames []string `json:"toolNames,omitempty"`
+	ToolNames []string `json:"toolNames"`
```
EItanya, line 112: Why did you remove `omitEmpty`? This just makes the field required. Now you would have to put `toolNames: []`. Is that intentional?

sarco3t:
yes, it's as I made `[]` to be equal to include all tools from mcp server
to agent, i guess it might not be the optimal or line-up with yours
architecture, so I'm seeking for the advice how to line-up that feature
with it. As it might be pretty useful for small tool servers, and with
automating binding tools/toolservers to the agent.

чт, 14 серп. 2025 р., 18:20 користувач Eitan Yarmush <
***@***.***> пише:

go/controller/api/v1alpha1/agent_types.go
sarco3t: > Why did you remove `omitEmpty`? This just makes the field required. Now you would have to put `toolNames: []`. Is that intentional?

Yes, it's intentional as I made `[]` to be equal to include all tools from mcp server
to agent, i guess it might not be the optimal or line-up with yours
architecture, so I'm seeking for the advice how to line-up that feature
with it. As it might be pretty useful for small tool servers, and with
automating binding tools/toolservers to the agent.

PR #756, opened 2025-08-15
refactor(controller): fix logging during remote mcp server reconciliation

onematchfox:
OK, the PR does a little more than the title mentions - sorry :sweat_smile: Started out trying to fix a couple copy/pasta errors but then things evolved... Will comment in-line where I think things may need further justification.

### DIFF
--- a/go/internal/controller/reconciler/reconciler.go
+++ b/go/internal/controller/reconciler/reconciler.go
@@ -200,10 +200,11 @@ func (a *kagentReconciler) ReconcileKagentMCPService(ctx context.Context, req ct
 	if remoteService, err := translator.ConvertServiceToRemoteMCPServer(service); err != nil {
 		reconcileLog.Error(err, "failed to convert service to remote mcp service", "service", utils.GetObjectRef(service))
 	} else {
-		if err := a.upsertToolServerForRemoteMCPServer(ctx, dbService, remoteService, service.Namespace); err != nil {
+		if _, err := a.upsertToolServerForRemoteMCPServer(ctx, dbService, remoteService, service.Namespace); err != nil {
 			return fmt.Errorf("failed to upsert tool server for mcp service %s: %v", utils.GetObjectRef(service), err)
 		}
 	}
+
 	return nil
 }
 
@@ -292,11 +293,12 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
 		Description: "N/A",
 		GroupKind:   schema.GroupKind{Group: "kagent.dev", Kind: "MCPServer"}.String(),
 	}
+
 	if remoteSpec, err := translator.ConvertMCPServerToRemoteMCPServer(mcpServer); err != nil {
 		reconcileLog.Error(err, "failed to convert mcp server to remote mcp server", "mcpServer", utils.GetObjectRef(mcpServer))
 	} else {
-		if err := a.upsertToolServerForRemoteMCPServer(ctx, dbServer, remoteSpec, mcpServer.Namespace); err != nil {
-			reconcileLog.Error(err, "failed to upsert tool server for remote mcp server", "mcpServer", utils.GetObjectRef(mcpServer))
+		if _, err := a.upsertToolServerForRemoteMCPServer(ctx, dbServer, remoteSpec, mcpServer.Namespace); err != nil {
+			return fmt.Errorf("failed to upsert tool server for remote mcp server %s: %v", utils.GetObjectRef(mcpServer), err)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


go/controller/internal/reconciler/reconciler.go
```diff
@@ -394,13 +404,12 @@ func (a *kagentReconciler) reconcileRemoteMCPServerStatus(
 	if err != nil {
 		status = metav1.ConditionFalse
 		message = err.Error()
-		reason = "AgentReconcileFailed"
-		reconcileLog.Error(err, "failed to reconcile agent", "tool_server", utils.GetObjectRef(toolServer))
```
onematchfox: This all started with fixing the copy/pasta "agent" here. The log line has now been removed, as errors are logged before this function is called.

go/controller/internal/reconciler/reconciler.go
```diff
@@ -333,59 +333,69 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
...
 			// Delete from DB if the remote mcp server is deleted
 			dbServer := &database.ToolServer{
-				Name:      utils.GetObjectRef(toolServer),
-				GroupKind: schema.GroupKind{Group: "kagent.dev", Kind: "RemoteMCPServer"}.String(),
+				Name:      serverRef,
+				GroupKind: server.GroupVersionKind().GroupKind().String(),
 			}
+
 			if err := a.dbClient.DeleteToolServer(dbServer.Name, dbServer.GroupKind); err != nil {
-				reconcileLog.Error(err, "failed to delete tool server for remote mcp server", "remoteMCPServer", utils.GetObjectRef(toolServer))
```
onematchfox: `toolServer` will be empty at this point since we failed to get it - hence the broader refactoring to set `serverRef` at the start of the function and re-use that everywhere instead of calling `utils.GetObjectRef` multiple times.

go/controller/internal/reconciler/reconciler.go
```diff
@@ -333,59 +333,69 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
...
+	tools, err := a.upsertToolServerForRemoteMCPServer(ctx, dbServer, &server.Spec)
+	if err != nil {
+		l.Error(err, "failed to upsert tool server for remote mcp server")
+
+		// Fetch previously discovered tools from database if possible
+		var discoveryErr error
+		tools, discoveryErr = a.getDiscoveredMCPTools(ctx, serverRef)
+		if discoveryErr != nil {
+			err = multierror.Append(err, discoveryErr)
+		}
```
onematchfox: Lifted this out of the status reconciliation and used the opportunity to optimise things a bit - since we are already list all tools as part of the upsert, we only need the database call if the upsert fails.

go/controller/internal/reconciler/reconciler.go
```diff
@@ -394,13 +404,12 @@ func (a *kagentReconciler) reconcileRemoteMCPServerStatus(
 	if err != nil {
 		status = metav1.ConditionFalse
 		message = err.Error()
-		reason = "AgentReconcileFailed"
-		reconcileLog.Error(err, "failed to reconcile agent", "tool_server", utils.GetObjectRef(toolServer))
+		reason = "RemoteMCPServerReconcileFailed"
```
onematchfox: copy/pasta :smile:

go/controller/internal/reconciler/reconciler.go
```diff
@@ -333,59 +333,69 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
...
+	server := &v1alpha2.RemoteMCPServer{}
+	if err := a.kube.Get(ctx, req.NamespacedName, server); err != nil {
+		// if the remote MCP server is not found, we can ignore it
 		if k8s_errors.IsNotFound(err) {
 			// Delete from DB if the remote mcp server is deleted
 			dbServer := &database.ToolServer{
-				Name:      utils.GetObjectRef(toolServer),
-				GroupKind: schema.GroupKind{Group: "kagent.dev", Kind: "RemoteMCPServer"}.String(),
+				Name:      serverRef,
+				GroupKind: server.GroupVersionKind().GroupKind().String(),
```
onematchfox: Can use the object itself here (and below) instead of magic strings.

EItanya (commented):
These changes look good overall. Given how many important things they touch I will want to run/validate myself, I will do so when I am back Monday!

go/internal/controller/reconciler/reconciler.go
```diff
@@ -228,8 +228,8 @@ func (a *kagentReconciler) ReconcileKagentMCPService(ctx context.Context, req ct
 	if remoteService, err := translator.ConvertServiceToRemoteMCPServer(service); err != nil {
 		reconcileLog.Error(err, "failed to convert service to remote mcp service", "service", utils.GetObjectRef(service))
 	} else {
-		if err := a.upsertToolServerForRemoteMCPServer(ctx, dbService, remoteService, service.Namespace); err != nil {
-			return fmt.Errorf("failed to upsert tool server for mcp service %s: %v", utils.GetObjectRef(service), err)
+		if _, err := a.upsertToolServerForRemoteMCPServer(ctx, dbService, remoteService, service.Namespace); err != nil {
+			reconcileLog.Error(err, "failed to upsert tool server for mcp service", "service", utils.GetObjectRef(service))
```
EItanya: Hey, so I actually changed this from a log to an error on purpose. I was seeing behavior where it would fail to discover tools and never try again, this ensures that the reconciler requeues.

go/internal/controller/reconciler/reconciler.go
```diff
@@ -323,7 +323,7 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
 	if remoteSpec, err := translator.ConvertMCPServerToRemoteMCPServer(mcpServer); err != nil {
 		reconcileLog.Error(err, "failed to convert mcp server to remote mcp server", "mcpServer", utils.GetObjectRef(mcpServer))
 	} else {
-		if err := a.upsertToolServerForRemoteMCPServer(ctx, dbServer, remoteSpec, mcpServer.Namespace); err != nil {
+		if _, err := a.upsertToolServerForRemoteMCPServer(ctx, dbServer, remoteSpec, mcpServer.Namespace); err != nil {
```
EItanya, line 300: This should return an error like above most likely for the same reason

go/internal/controller/reconciler/reconciler.go
onematchfox: Ah, my bad - missed that when rebasing. Have fixed.

go/internal/controller/reconciler/reconciler.go
onematchfox: Changed to match the behaviour of `ReconcileKagentMCPService`.

EItanya (approved):

PR #683, opened 2025-08-04
Official KMCP Support

EItanya:
This PR bring official KMCP support to KAGENT 🎉 

1. Changes `Agent` API to allow pointing to `Service`, `MCPServer` or new `RemoteMCPServer`
2. Removed `ToolServer` in favor of new `RemoteMCPServer`

PR #757, opened 2025-08-15
fix(controller): ensure controller name matches managed object

onematchfox:
Missed in #683.


onematchfox:
Not so simple... `mcpserver` is already used by KMCP. Big code smell here :fish: Will try to revisit at a later stage

PR #758, opened 2025-08-15
fix(controller): ensure tool server headers are applied when reconciling tools

onematchfox:
The headers configured on a tool server will often be things like an `Authorization` header and should be applied when attempting to list tools as part of the tool server reconciliation process. Currently, if the remote MCP server requires authorization, reconciliation will fail with a status like:

```
Status:
  Conditions:
    Last Transition Time:  2025-08-15T10:54:11Z
    Message:               failed to upsert tool server kagent/foo: failed to fetch tools for toolServer kagent/foo: failed to initialize client for toolServer kagent/foo: Unauthorized: Missing or invalid Authorization header
    Reason:                AgentReconcileFailed
    Status:                False
    Type:                  Accepted
```

### DIFF
--- a/go/controller/internal/reconciler/reconciler.go
+++ b/go/controller/internal/reconciler/reconciler.go
@@ -490,7 +490,7 @@ func (a *kagentReconciler) upsertToolServerForRemoteMCPServer(ctx context.Contex
 		return fmt.Errorf("failed to store toolServer %s: %v", toolServer.Name, err)
 	}
 
-	tsp, err := a.createMcpTransport(ctx, remoteMcpServer.URL, remoteMcpServer.Protocol, remoteMcpServer.HeadersFrom, namespace)
+	tsp, err := a.createMcpTransport(ctx, remoteMcpServer, namespace)
 	if err != nil {
 		return fmt.Errorf("failed to create client for toolServer %s: %v", toolServer.Name, err)
 	}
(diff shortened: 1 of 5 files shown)


go/controller/internal/reconciler/reconciler.go
```diff
@@ -490,27 +490,28 @@ func (a *kagentReconciler) upsertToolServerForRemoteMCPServer(ctx context.Contex
...
-		if err != nil {
-			return fmt.Errorf("failed to create streamable http client for toolServer %s: %v", toolServer.Name, err)
-		}
-		tools, err = a.listTools(ctx, streamableHttpClient, toolServer)
-		if err != nil {
-			return fmt.Errorf("failed to fetch tools for toolServer %s: %v", toolServer.Name, err)
+		var headers map[string]string
+		if headers, err = a.adkTranslator.ResolveValueRefs(ctx, remoteMcpServer.HeadersFrom, namespace); err == nil {
+			tsp, err = transport.NewStreamableHTTP(remoteMcpServer.URL, transport.WithHTTPHeaders(headers))
 		}
```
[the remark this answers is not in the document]
onematchfox: Refactored to make the logic cleaner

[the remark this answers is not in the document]
onematchfox: Refactored to make the logic cleaner

go/controller/translator/adk_api_translator.go
```diff
@@ -715,6 +701,24 @@ func (a *adkApiTranslator) translateSseHttpTool(ctx context.Context, tool *v1alp
 	return params, nil
 }
 
+func (a *adkApiTranslator) ResolveValueRefs(ctx context.Context, vrs []v1alpha2.ValueRef, namespace string) (map[string]string, error) {
```
EItanya: I'm not sure this should be a method of `adkApiTranslator`. Rather it should be a helper function which takes in `kcli client.Client` or something like that. It's leaking impl detail for it to be on the translator. What do you think?

onematchfox: Yeah, I felt funny putting it here too, but wasn't sure where else to put it. :sweat_smile: I could easily put it somewhere in `go/internal/utils` if that works for you? Alternatively, we could do something like adding a func to `ValueRef` type - `func (r *ValueRef) Resolve(ctx, client, namespace)` - and perhaps even adding a type, `ValueRefList`, that iterates over the list? What feels more appropriate to you?

onematchfox: Refactored to expose helper functions on the underlying types. LMKWYT.

PR #761, opened 2025-08-15
feat: initial auth framework

yuval-k:
lay the ground work for auth & authz in kagent. Initial impl leaves current state - user is by default `admin@kagent.dev` and it's authorized to do anything. 

Added test agent - it replies with a constant text without using LLM. This approach has these advantages
- It's fast & reliable (no call to slow LLM on the ineternet)
- It's predictable (we can reliable make test asserts)
- Low cost (no need for API key)
Which makes it ideal for integration / e2e tests for most of kagent (with the exception of the bundled agents themselves).


helm/kagent-crds/templates/kagent.dev_modelconfigs.yaml
```diff
@@ -8,8 +8,6 @@ metadata:
 spec:
   group: kagent.dev
   names:
-    categories:
```
yuval-k: not sure how this happened. i'll check

peterj: I always get this as well when building locally

go/test/e2e/agents/kebab/README.md
```diff
@@ -0,0 +1,21 @@
...
+1. Build the agent image
+
+```bash
+docker build  . --push -t localhost:5001/kebab:latest
+```
+
+2. Create a secret with the google api key
+
+```bash
+kubectl create secret generic kagent-google -n kagent  --from-literal=GOOGLE_API_KEY=$GOOGLE_API_KEY   --dry-run=client -oyaml | k apply -f -
```
EItanya: This agent doesn't need a key

python/packages/kagent-adk/src/kagent_adk/_token.py
```diff
@@ -0,0 +1,80 @@
...
+        request.headers.update(headers)
+
+
+def read_token() -> str:
+    try:
+        with open(KAGENT_TOKEN_PATH, "r", encoding="utf-8") as f:
+            token = f.read()
+            return token.strip()
+    except OSError:
+        # TODO: figure out how to log - print(f"Error reading token from {KAGENT_TOKEN_PATH}: {e}")
```
EItanya: This method should change the signature or return "". Also, to add log you can just add a logger at the top of the file: `logger - logging.get(__name__)`

go/internal/httpserver/handlers/sessions.go
```diff
@@ -337,7 +344,12 @@ func (h *SessionsHandler) HandleAddEventToSession(w ErrorResponseWriter, r *http
 	}
 	log = log.WithValues("session_id", sessionID)
 
-	userID, err := GetUserID(r)
+	principle, err := GetPrincipal(r)
+	if err != nil {
+		w.RespondWithError(errors.NewBadRequestError("Failed to get user ID", err))
+		return
+	}
+	userID, err := getUserID(r)
```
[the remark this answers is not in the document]
yuval-k: it's not a duplicate. it can be removed in a later iteration where agents will use the user's token to auth to kagent

EItanya (approved):

PR #772, opened 2025-08-18
[BUG] Fix nil object reference in ToolServer reconciler deletion logic

### LINKED ISSUE #770: [BUG] deleting the MCPServer doesn't remove it from the DB
### 📋 Prerequisites

- [ ] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

MCPServer doesn't get deleted from the DB.

### 🔄 Steps To Reproduce

1. Create an MCPserver
2. Delete an mcpserver

Note the server is still displayed in the UI and available from the controller (rest api).

### 🤔 Expected Behavior

_No response_

### 📱 Actual Behavior

_No response_

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

ashleywang1:
## Problem

The original issue was that when Kubernetes objects are deleted, the reconciler receives deletion events but the objects themselves are no longer available in the API. This caused `utils.GetObjectRef()` to return empty strings when called on nil/empty objects, breaking database cleanup logic.

However, during the fix, we discovered a deeper database schema issue: the `DeleteToolsForServer` method was trying to filter by `group_kind` column on the Tool table, but Tools were missing the GroupKind field entirely, leading to SQL errors.

## Solution

This PR provides a comprehensive fix addressing both the original bug and the underlying database schema issues:

### 1. **Nil Object Reference Fix**
Fixed the nil object reference bug in three reconcile functions by using `req.NamespacedName.String()` instead of `utils.GetObjectRef()` when handling `IsNotFound` errors:

- **`ReconcileKagentMCPServer`**: Fixed deletion logic for MCP servers
- **`ReconcileKagentRemoteMCPServer`**: Fixed deletion logic for remote MCP servers  
- **`ReconcileKagentMCPService`**: Fixed deletion logic for MCP services

### 2. **Database Schema Enhancement (BREAKING CHANGE)**
Implemented proper GroupKind support throughout the Tool database operations:

- **Added GroupKind field** to Tool model as primary key
- **Updated all database operations** to use both serverName and groupKind
- **Enhanced tool isolation** between different server types
- **Fixed SQL logic errors** by ensuring proper schema consistency

### 3. **Complete GroupKind Integration**
- **DeleteToolsForServer**: Now filters by both `server_name` AND `group_kind`
- **ListToolsForServer**: Updated to require `groupKind` parameter
- **RefreshToolsForServer**: Enhanced to handle `groupKind` when creating/updating tools
- **All reconciler calls**: Updated to pass `toolServer.GroupKind` parameter
- **HTTP handlers**: Updated to use groupKind when listing tools
- **Service support**: Added complete Service GroupKind case to toolserver deletion handler

### Database Model Changes
- Added `GroupKind` field to `Tool` model as `gorm:"primaryKey;not null"`
- Tool table now has composite primary key: `(id, server_name, group_kind)`

### Client Interface Updates
- `ListToolsForServer(serverName string, groupKind string) ([]Tool, error)`
- `RefreshToolsForServer(serverName string, groupKind string, tools ...*v1alpha2.MCPTool) error`
- `DeleteToolsForServer(serverName string, groupKind string) error` - now properly filters by both parameters

### Reconciler Fixes
- All `IsNotFound` error handling now uses `req.NamespacedName.String()` for proper object identification
- All database operations now pass the correct `groupKind` parameter

### HTTP Handler Updates
- Toolserver deletion now supports all three GroupKind types:
  - `"Service"` - for Kubernetes Services
  - `"MCPServer.kagent.dev"` - for MCP Server CRDs
  - `"RemoteMCPServer.kagent.dev"` - for Remote MCP Server CRDs

## Breaking Change

⚠️ **BREAKING CHANGE**: This PR includes breaking database schema changes.

**Migration Required**: Users must delete their existing database and allow GORM to recreate the schema. GORM AutoMigrate cannot safely add NOT NULL primary key fields to existing data.

The Tool table primary key has changed from `(id, server_name)` to `(id, server_name, group_kind)`.

## Issues Resolved

- closes #770


go/internal/database/client.go
```diff
@@ -27,7 +27,7 @@ type Client interface {
 	DeleteToolServer(serverName string, groupKind string) error
 	DeleteTask(taskID string) error
 	DeletePushNotification(taskID string) error
-	DeleteToolsForServer(serverName string, groupKind string) error
+	DeleteToolsForServer(serverName string) error
```
EItanya: Why did you remove this parameter, I think it's still necessary since it's part of the primary key

ashleywang1: It's part of the primary key for the tool server table, but not for the tools table. I tried adding it to the tools table before realizing that it might be a breaking change, so I backed out and made the assumption that a tool is only associated with the server name and not both the server name and group kind.

go/internal/database/fake/client.go
```diff
@@ -225,7 +225,7 @@ func (c *InMemmoryFakeClient) DeleteToolServer(serverName string, groupKind stri
 
 // DeleteToolsForServer deletes tools for a tool server by name
 
-func (c *InMemmoryFakeClient) DeleteToolsForServer(serverName string, groupKind string) error {
+func (c *InMemoryFakeClient) DeleteToolsForServer(serverName string) error {
```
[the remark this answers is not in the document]
ashleywang1: We changed the implementation so that it also removes the groupKind param

go/internal/database/client.go
ashleywang1: Spoke offline: this breaking change is acceptable

PR #783, opened 2025-08-20
Remove deprecated toolserver CRD from kagent-crds helm chart

antweiss:
As the ToolServer api has been removed - the CRD is redundant and is just causing confusion


EItanya (dismissed):
Good catch!

EItanya:
Thanks so much for the PR, and a great catch. Unfortunately this will not work today. These CRDs are autogenerated from the go types, which still exist. `go/controller/api/v1alpha1`. You'll need to figure out how to get it to stop generating this CR

antweiss:
Thanks!  I believe I know how to do it. Will look into it tomorrow.

Anton Weiss
http://otomato.io

antweiss:
So, there's a way to avoid generating all CRs for v1alpha1 by adding the // +kubebuilder:skip directive on the v1alpha1 package. This will also remove the deprecated Memories CR. The question is - do we need older versions?

antweiss:
@EItanya I updated the PR so that all v1alpha1 CRs are skipped while generating. I believe this is ok at this stage of the project but I could be wrong. LMKWYT

EItanya:
It should be fine, however there is one situation I am worried about. Do upgrades from `0.5.x -> this commit` work properly after this change. I remember seeing issues upgrading CRDs when removing versions like this, but I may have been doing it wrong. I think if it fails we should wait until `0.7.x` to remove them completely.

What do you think?

antweiss:
I think I'll give it a try. Will install CRDs from the old chart and then upgrade to this.

antweiss:
Ugh, didn't actually find a way to create a clean upgrade. Looks like there should be a way to mark na API version as unserved and deprecated but it's not really working for me yet. Will look at it at a later time. This can be closed in the meantime.

antweiss:
Closing

PR #791, opened 2025-08-21
[BUG] add kmcp as a subchart to kagent-crds

chunter0:
Kagent requires `McpServers` resource which comes from kmcp project. This pull request adds those crds as a subchart of kagent-crds to make for easier installation UX.

### DIFF
--- a/helm/kagent-crds/Chart-template.yaml
+++ b/helm/kagent-crds/Chart-template.yaml
@@ -3,3 +3,7 @@ name: kagent-crds
 description: CRDs for kagent
 type: application
 version: ${VERSION}
+dependencies:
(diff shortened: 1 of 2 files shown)


helm/kagent-crds/Chart-template.yaml
```diff
@@ -2,4 +2,8 @@ apiVersion: v2
 name: kagent-crds
 description: CRDs for kagent
 type: application
-version: ${VERSION}
\ No newline at end of file
+version: ${VERSION}
+dependencies:
+  - name: kmcp-crds
+    version: 0.1.5
```
EItanya: Can this version come from the go.mod version?

Issue #766, opened 2025-08-17
[FEATURE] Add flag '--no-default-agents' on `kagent install` for a minimal kagent installation


antweiss:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add flag '--no-default-agents' on `kagent install` for a minimal kagent installation. This should set all default agents to enabled=false.

### ❓ Problem Statement / Motivation

- Current default kagent installation creates 10 agents. This makes it too big for training environments (that;s how I stumbled upon this). Moreover - istio, cilium and promql agents are installed without checking if the tools are available in cluster (i.e istio, cilium, prometheus). And in general - I think having a way of creating a minimal installation without editing the values file for the chart would be a great convenience and help with adoption.

### 💡 Proposed Solution

Add flag '--no-default-agents' on `kagent install`
Under the hood that would go over the list of agents in the values file and set them to enabled=false
Maybe we should exclude only the k8s-agent...

### 🔄 Alternatives Considered

- Installing with helm and passing mulitple `--set` flags
- Installing with helm and providing a custom override values file.

Both ways work but are cumbersome if one just wnats to have a clean install of kagent.

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR for this feature

PR #796, opened 2025-08-21
KAgent Install Profile Feature

inFocus7:
# Changes

Support "profiles" through values files with default settings. At the moment, these are just for setting up which agents to install into cluster.

`demo`: Installs all agents, as done now.
`minimal` (default): Installs no agent.

I created both explicitly in case we change defaults are anything happens, where these profiles should still act the same.

Profiles are not bundled with helm and are stored alongside the CLI code in go to be embedded. They can still be referenced through `helm` installations by providing the yaml filepath.

## Breaking/Changed Defaults

**IMPORTANT**: With these changes, the additional agents are _not_ installed by default. Useers would either need to `helm [...] -f <raw-gh-path-to-demo-file>` or `kagent install --profile demo` to install.

# Testing

## Helm Installation

```sh
# install kagent crds (note this also installs the kmcp ones
helm install kagent-crds ./helm/kagent-crds/  --namespace kagent

helm install kagent ./helm/kagent/ --namespace kagent --set providers.default=openAI --set providers.openAI.apiKey=$OPENAI_API_KEY -f ./go/cli/internal/profiles/demo.yaml # or minimal.yaml, or none for the default (no agents)
```

## CLI Installation

Build
```sh
make create-kind-cluster helm-install build-cli-local
```

Uninstall current kagent installation
```sh
./go/bin/kagent-local uninstall
```

Install, non-interactive
```sh
KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local install --profile <demo|minimal>
```

Uninstall current kagent installation
```sh
./go/bin/kagent-local uninstall
```

Install, interactive
```sh
KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local
```

_Then choose a profile from the profile selection and see installed pods._

# Notes

I was looking into [Istio's profile implementation](https://github.com/istio/istio/blob/b68cd04f9f132c1361d62eb14125e915e8011428/manifests/charts/base/templates/zzz_profile.yaml), but this may not be possible on our end because of us importing the agents as subchart dependencies in Chart.yaml. Afaik, Chart.yaml is processed before templates, so `--set profile=x` would not work. Doing a `-f <path>/<profile>.yaml` does work as it's not a template, and the values would process before the Chart.

---

Resolves: https://github.com/kagent-dev/kagent/issues/766

(not an implementation of the _exact_ request flag, but the concluded method to handle this issue)

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -64,11 +64,6 @@ make kagent-cli-install
 export KAGENT_DEFAULT_MODEL_PROVIDER=ollama
 export KAGENT_HELM_REPO=./helm/
 make kagent-cli-install
-
-# use local path to install kagent using locally-defined profiles
-export KAGENT_HELM_PROFILE_URL=./helm/kagent/files/profiles/
-export KAGENT_HELM_REPO=./helm/
-make kagent-cli-install
 ```
 
 ## Upgrading


inFocus7:
this (c/sh)ould have just been a comment on the relevant issue tbh 😅

go/cli/internal/cli/install.go
```diff
@@ -38,20 +54,122 @@ func installChart(ctx context.Context, chartName string, namespace string, regis
...
+
+	pf, err := NewPortForward(ctx, cfg)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error starting port-forward: %v\n", err)
+		return nil
+	}
+	return pf
+}
+
+func InteractiveInstallCmd(ctx context.Context, c *ishell.Context) *PortForward {
```
inFocus7, line 97: Need to d.r.y. this.

I was thinking the non-interactive mode could use `--profile` flag, and interactive mode would have an interactive selection. so _most_ of the `InstallCmd` + `InteractiveInstallCmd` functions are 1:1, aside from the signature and profile selection method.

inFocus7:
something's broken with my signing, i use `-S` + set the signing key yesterday, but still fails DCO :rip:

go/cli/internal/cli/install.go
```diff
@@ -38,20 +41,87 @@ func installChart(ctx context.Context, chartName string, namespace string, regis
 		args = append(args, "--set", setValue)
 	}
 
+	// If a profile is provided, write the embedded YAML to a temp file and pass its path
+	var tmpProfilePath string
+	if profile != "" {
+		profileYAML := profiles.GetProfile(profile)
```
inFocus7: converted back to a file, because the values are defined in Helm, and in order to properly unmarshal (to a non-generic interface{}) we'd need to redefine the values in a go struct, which could lead to drift + needing to keep both in sync.

helm/README.md
```diff
@@ -7,8 +7,6 @@ These Helm charts install kagent-crds,kagent, it is required that the Kagent CRD
 ### Using Helm
 
 ```bash
-helm install kmcp-crds oci://ghcr.io/kagent-dev/kmcp/helm/kmcp-crds --version 0.1.2 --namespace kagent
```
inFocus7, line 10: we no longer need to install kmcp-crds separately

go/cli/internal/cli/install.go
```diff
@@ -38,20 +41,87 @@ func installChart(ctx context.Context, chartName string, namespace string, regis
...
+				return "", fmt.Errorf("failed creating temp profile file: %w", err)
+			}
+			if _, err := tmpFile.WriteString(profileYAML); err != nil {
+				_ = tmpFile.Close()
+				_ = os.Remove(tmpFile.Name())
+				return "", fmt.Errorf("failed writing temp profile file: %w", err)
+			}
+			if err := tmpFile.Close(); err != nil {
+				_ = os.Remove(tmpFile.Name())
+				return "", fmt.Errorf("failed closing temp profile file: %w", err)
```
EItanya: Can you just do a defer close/remove?

EItanya: Also, I believe helm supports passing the values yaml as a string, so that may be easier than saving a file.

go/cli/internal/cli/install.go
```diff
@@ -38,20 +41,87 @@ func installChart(ctx context.Context, chartName string, namespace string, regis
...
 		return string(byt), err
 	}
+	if tmpProfilePath != "" {
+		_ = os.Remove(tmpProfilePath)
+	}
 	return "", nil
 }
 
-func InstallCmd(ctx context.Context, cfg *config.Config) *PortForward {
+func InstallCmd(ctx context.Context, cfg *config.Config, profile string) *PortForward {
```
EItanya, line 61: Why not just put `profile` into `config`, or maybe have an `InstallConfig`. For example see `InvokeConfig`

helm/kagent/values.yaml
EItanya, line 1: I don't think we can do this yet :(.

I don't want such a big change/break mid-release. I would like to potentially wait until `0.7.0`, or at least bring it up in the community meeting first

go/cli/internal/cli/install.go
```diff
@@ -65,6 +135,24 @@ func InstallCmd(ctx context.Context, cfg *config.Config) *PortForward {
...
+	selectedProfile := profiles.Profiles[profileIdx]
+
+	return install(ctx, cfg, helmConfig, selectedProfile, modelProvider)
+}
+
+// helmConfig is the config for the kagent chart
+type helmConfig struct {
+	registry string
+	version  string
+	values   []string
```
EItanya: Could put `profile` in here as well

helm/kagent/values.yaml
inFocus7: Darn 💔 

_Although_, if we're alright with this approach (after resolving review comments), we could keep this non-breaking by
1. Revert agent defaults back to `true`.
2. Remove `demo` profile.
3. Keep the `minimal` profile for users who want to easily opt-out of the additional agents.

Then in the future (i.e. `0.7.0`) we could update agent defaults to `false` and re-add `demo` profile. (or we keep both right now, and just revert agents back to `true`).

---

I'm not against _not_ merging this though 👍🏼

EItanya: I think keep them both for now and just revert the actual `values` file back to true. Having both profiles creates a good starting point for us to move forward with, even if the demo one is a tad redundant right this minute

go/cli/internal/cli/install.go
inFocus7: I couldn't find any docs on doing it inline through a parameter, _but_ since it's a shell command we can pipe/input it as below (tested locally):

```sh
<yaml content> | helm ... -f -
```

so the logic becomes:
```go
cmd := exec.CommandContext(ctx, "helm", args...)

// If a profile is provided, pass the embedded YAML to the stdin of the helm command.
// This must be the last set of arguments.
if profile != "" {
	profileYAML := profiles.GetProfile(profile)
	cmd.Stdin = strings.NewReader(profileYAML)
	cmd.Args = append(cmd.Args, "-f", "-")
}
// execute, etc.
```

go/cli/internal/cli/install.go
inFocus7: Did it. I named it `inlineValues` to keep the naming generic, even though at the moment it's for holding the profile values (if one is chosen).

inFocus7:
1. had merge conflicts.
2. fixed merge conflicts.
3. needed to rebase for signing.
4. rebased.
5. hit the exact same merge conflicts.
6. fixed merge conflicts.
7. Pushed rebase with signing.
8. loop back to 1
  - now with 500+ unrelated changes...

inFocus7:
merge conflicts, rebasing, and signing got this PR in a broken state.

Issue #797, opened 2025-08-21
Creating an invalid agent  (YAML)  causes all agents to not load in the UI


christian-posta:
try creating an invalid agent, the UI doesn't handle that gracefully

inFocus7:
No resolution branch on my end [yet], but anyone who wants to reproduce this issue:
1. install `kagent` with the default Agents installed
2. Edit an `Agent` CR to delete the `modelConfig` (or update string to non-existent `modelConfig`)
3. Failure:
  - UI Agents screen does not load.
  - The Agent itself does have an appropiate status: `Accepted -> False: AgentReconcileFailure`

UI logs OK, Controller logs:
```
2025-08-22T18:47:15Z    DEBUG    http    Request started    {"method": "GET", "path": "/api/agents", "remote_addr": "10.244.0.15:35114", "user_id": "admin@kagent.dev"}
2025-08-22T18:47:15Z    DEBUG    http.agents-handler    Processing Agent    {"method": "GET", "path": "/api/agents", "remote_addr": "10.244.0.15:35114", "user_id": "admin@kagent.dev", "operation": "list-db", "agentRef": "kagent/argo-rollouts-conversion-agent"}
2025-08-22T18:47:15Z    DEBUG    http.agents-handler    Processing Agent    {"method": "GET", "path": "/api/agents", "remote_addr": "10.244.0.15:35114", "user_id": "admin@kagent.dev", "operation": "list-db", "agentRef": "kagent/argo-rollouts-conversion-agent"}
2025-08-22T18:47:15Z    DEBUG    http.agents-handler    Processing Agent    {"method": "GET", "path": "/api/agents", "remote_addr": "10.244.0.15:35114", "user_id": "admin@kagent.dev", "operation": "list-db", "agentRef": "kagent/cilium-debug-agent"}
2025-08-22T18:47:15Z    DEBUG    http.agents-handler    Processing Agent    {"method": "GET", "path": "/api/agents", "remote_addr": "10.244.0.15:35114", "user_id": "admin@kagent.dev", "operation": "list-db", "agentRef": "kagent/cilium-debug-agent"}
2025-08-22T18:47:15Z    DEBUG    http.agents-handler    Processing Agent    {"method": "GET", "path": "/api/agents", "remote_addr": "10.244.0.15:35114", "user_id": "admin@kagent.dev", "operation": "list-db", "agentRef": "kagent/cilium-policy-agent"}
[...]

PR #808, opened 2025-08-25
KAgent Install Profile Feature

inFocus7:
original pr (which got in a terrible state due to merge conflicts + rebasing, so better to start fresh): https://github.com/kagent-dev/kagent/pull/796

---

Originally, when listing the agents, if a single agent failed (i.e. `modelConfig` is invalid), no Agents were shown.

# Changes

With these changes, we:
1. Add a new AgentResponse field `Accepted` (i.e. `Reconciled`, but front-end friendly name)
    - This is based on the `Accepted` state of the Agent.
2. **Don't** return an error response _to the frontend_ when an Agent (or multiple) had failures.
    - We now continue listing agents, making the UI usable for valid agents, and adding a `Not Accepted` (could also be `Not Reconciled`, or `Error` for simplest terms) status on broken agents -- similar to how we do `Not Ready` while they are being deployed.
3. No longer display unready/un-accepted Agents in the agent sidebar from chats.

## Alternative Approaches

Another approach would be to avoid returning the not-reconciled agent altogether. This would avoid needing to set up the new response field and UI changes. I was originally leaning towards that one, but figured it may be useful/quicker for users to notice any broken agents if they appear in the UI with their failed status.

Another idea I had, but would probably be overkill and could constitute as a new feature, is to have a "notification system", where we would _not_ display failed agents in the agents list, but instead have the reconciliation as a notification.

# Testing

## Dashboard

Build & run
```sh
make create-kind-cluster helm-install build-cli-local

KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local dashboard
```

Update an Agent to have an invalid `modelConfig`
```sh
kubectl patch agents.kagent.dev/k8s-agent -n kagent --type='json' -p='[{"op": "replace", "path": "/spec/declarative/modelConfig", "value": "default-model-config-2"}]'
```

Reload page.
View listing UI to see the failed state (Not Accepted).

Fix the agent.
```sh
kubectl patch agents.kagent.dev/k8s-agent -n kagent --type='json' -p='[{"op": "replace", "path": "/spec/declarative/modelConfig", "value": "default-model-config"}]'
```

Reload page.
It’s all good again!

## CLI

This would ideally also get tested through the CLI to test the behavior of `KAGENT_HELM_REPO=./helm/ ./go/bin/kagent-local get agent`, _but_ I’m hitting “userID is required” issues. My `$HOME/.kagent/config.yaml` file has it defined, so I'm not sure of the exact issue yet.

---

Resolves: https://github.com/kagent-dev/kagent/issues/797


go/cli/cmd/kagent/main.go
```diff
@@ -32,14 +33,19 @@ func main() {
 	rootCmd.PersistentFlags().StringVarP(&cfg.Namespace, "namespace", "n", "kagent", "Namespace")
 	rootCmd.PersistentFlags().StringVarP(&cfg.OutputFormat, "output-format", "o", "table", "Output format")
 	rootCmd.PersistentFlags().BoolVarP(&cfg.Verbose, "verbose", "v", false, "Verbose output")
+	var profile string
```
EItanya: Same comment about `profile` not being a config option in something like `InstallConfig`

inFocus7: :donezo:

PR #810, opened 2025-08-26
fix(install): cancel root context when receiving an OS signal

FalcoSuessgott:
This PR introduces actually cancelling the root context when receiving an OS signal such as CTRL+C so that the context is cancelled throughout the `kagent` CLI

### DIFF
--- a/go/cli/cmd/kagent/main.go
+++ b/go/cli/cmd/kagent/main.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"syscall"
 	"strings"
+	"os/signal"
 
 	"github.com/abiosoft/ishell/v2"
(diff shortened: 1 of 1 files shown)


FalcoSuessgott:
I believe [`test-e2e`](https://github.com/kagent-dev/kagent/actions/runs/17225819183/job/48911075247?pr=810) fails because the job does not have access to secrets when triggered from a fork. Adding something like 

```yaml
on:
  pull_request_target:
```

should fix it

EItanya:
> I believe [`test-e2e`](https://github.com/kagent-dev/kagent/actions/runs/17225819183/job/48911075247?pr=810) fails because the job does not have access to secrets when triggered from a fork. Adding something like
> 
> ```yaml
> on:
>   pull_request_target:
> ```
> 
> should fix it

We originally thought that, but that doesn't work as expected. We are going to mock those calls in the future.

PR #812, opened 2025-08-26
feat(helm): extract Helm chart for Grafana MCP

onematchfox:
Enables this to be deployed as a standalone component (i.e. without the `observability-agent`) and also upgrades to make use of [Streamable HTTP](https://github.com/grafana/mcp-grafana?tab=readme-ov-file#usage) as this is now supported.

### DIFF
--- a/helm/kagent/templates/NOTES.txt
+++ b/helm/kagent/templates/NOTES.txt
@@ -20,11 +20,15 @@ DEPLOYED COMPONENTS:
   - Controller: {{ include "kagent.fullname" . }}-controller (manages Kubernetes resources)
   - Engine: {{ include "kagent.fullname" . }}-engine (AI agent runtime)
   - UI: {{ include "kagent.fullname" . }}-ui (web interface)
+
+ENABLED TOOLS:
 {{- if index .Values "kagent-tools" "enabled" }}
   - Tool Server: Kagent's Built-in tools
 {{- end }}
-{{- if index .Values "querydoc" "enabled" }}
-  - QueryDoc: Document search and analysis
+{{- range $tool, $config := .Values.tools }}
+{{- if $config.enabled }}
+  - {{ $tool }}
+{{- end }}
 {{- end }}
 
 ENABLED AGENTS:


helm/agents/observability/values.yaml
```diff
@@ -1,10 +1 @@
 modelConfigRef: ""
-prometheus:
-  url: "prometheus.kagent:9090/api/v1"
-  username: ""
-  password: ""
```
onematchfox, line 5: All of this was unused so cleaned up while I was at it. Same goes for `grafana.username` and `grafana.password` just below.

helm/tools/grafana-mcp/templates/remotemcpserver.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kagent.dev/v1alpha2
+kind: RemoteMCPServer
```
onematchfox, line 2: I know I could have just used the `Service`, but I prefer this because, since the removal of the `ToolServer`, one has to look at controller logs to see if there are any issues with `Service`/`McpServer`-_based_ tooling.

helm/kagent/Chart-template.yaml
```diff
@@ -8,10 +8,14 @@ dependencies:
...
     repository: oci://ghcr.io/kagent-dev/tools/helm
     condition: kagent-tools.enabled
+  - name: grafana-mcp
+    version: ${VERSION}
+    repository: file://../tools/grafana-mcp
+    condition: tools.grafana-mcp.enabled, agents.observability-agent.enabled
   - name: querydoc
     version: ${VERSION}
     repository: file://../tools/querydoc
-    condition: external-mcp.querydoc.enabled
```
onematchfox, line 14: Fixed this condition while I was here

helm/tools/grafana-mcp/templates/remotemcpserver.yaml
onematchfox: I am cognisant that the original `stdio` implementation may have been used to dog-food the KMCP integration, but hopefully a different solution can be found for that.

helm/kagent/Chart-template.yaml
```diff
@@ -8,10 +8,14 @@ dependencies:
     version: 0.0.11
     repository: oci://ghcr.io/kagent-dev/tools/helm
     condition: kagent-tools.enabled
+  - name: grafana-mcp
+    version: ${VERSION}
+    repository: file://../tools/grafana-mcp
+    condition: tools.grafana-mcp.enabled, agents.observability-agent.enabled
```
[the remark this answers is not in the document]
onematchfox: `observability-agent` is correct - see [here](https://github.com/kagent-dev/kagent/blob/ef0985d4837c3dddcd4d4f579a0e20b0789032fe/helm/kagent/values.yaml#L194-L195).

helm/tools/grafana-mcp/templates/_helpers.tpl
```diff
@@ -0,0 +1,69 @@
...
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create the grafana server URL
+*/}}
+{{- define "grafana-mcp.serverUrl" -}}
+{{- printf "http://%s.%s:%d/mcp" (include "grafana-mcp.fullname" .) .Release.Namespace (.Values.service.port | int) }}
+{{- end }}
```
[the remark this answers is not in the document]
onematchfox: Addressed

helm/tools/grafana-mcp/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+nodeSelector: {}
+
+service:
+  type: ClusterIP
+  port: 8000
+
+resources: {}
+
+# Deployment configuration
```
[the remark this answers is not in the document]
onematchfox: Addressed

helm/tools/grafana-mcp/templates/deployment.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "grafana-mcp.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "grafana-mcp.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicas }}
```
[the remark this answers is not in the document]
onematchfox: Addressed

EItanya:
Just out of curiosity why did you move from the `MCPServer` towards a full `deployment/service/configmap` approach. We originally had this, but then we added `kmcp` support into `kagent` for exactly this purpose.

onematchfox:
> Just out of curiosity why did you move from the `MCPServer` towards a full `deployment/service/configmap` approach. We originally had this, but then we added `kmcp` support into `kagent` for exactly this purpose.

The main reason at this point is that I think that using a native CRD, `RemoteMCPServer`, provides a better, more idiomatic UX than is possible using either `Service` or `MCPServer`. Neither of the other objects provides any form of feedback regarding reconciliation, which means that, in the event of issues, a user is left to trawl through the controller logs to find out why things aren't working. Whereas when using a `RemoteMCPServer`, one can simply look at the `Status` of the object to see what went wrong. 

Also, whilst I think the KMCP integration is great, I do, however, also disagree with the current implementation, which couples the two projects - but that's a bigger discussion that I was hoping to save for another time :sweat_smile:.

EItanya:
> > Just out of curiosity why did you move from the `MCPServer` towards a full `deployment/service/configmap` approach. We originally had this, but then we added `kmcp` support into `kagent` for exactly this purpose.
> 
> The main reason at this point is that I think that using a native CRD, `RemoteMCPServer`, provides a better, more idiomatic UX than is possible using either `Service` or `MCPServer`. Neither of the other objects provides any form of feedback regarding reconciliation, which means that, in the event of issues, a user is left to trawl through the controller logs to find out why things aren't working. Whereas when using a `RemoteMCPServer`, one can simply look at the `Status` of the object to see what went wrong.
> 
> Also, whilst I think the KMCP integration is great, I do, however, also disagree with the current implementation, which couples the two projects - but that's a bigger discussion that I was hoping to save for another time 😅.

Leaving aside the larger architecture question for one moment. I think the fact that the status is not useful is a bug rather than an inherent feature. If the MCPServer status is not properly indicative of what’s happening that’s something we can absolutely fix. In terms of the service, I think we can pretty easily expose a status via the kagent CLI.

onematchfox:
> Leaving aside the larger architecture question for one moment. I think the fact that the status is not useful is a bug rather than an inherent feature. If the MCPServer status is not properly indicative of what’s happening that’s something we can absolutely fix. In terms of the service, I think we can pretty easily expose a status via the kagent CLI.

Absolutely. This should be addressed at some point, and was something else I was going to bring up at another point. IMO the _correct_ way to do this is to ensure that all tool servers are reflected in the cluster and not only stored in the database :sweat_smile: 

ToolServer

onematchfox:
FWIW - I have tried using the `grafana-mcp` in its current form (using `MCPServer` with `transportType: "stdio"`) but have been unable to get it to work. I suspect that it's related to the way env vars are populated when running via `agentgateway`.  Needless to say, this isn't an issue when using a `Deployment` and hosting via streamable HTTP :shrug: 

image

helm/kagent/values.yaml
```diff
@@ -210,11 +210,21 @@ agents:
...
 # ==============================================================================
-# EXTERNAL MCP SERVERS
+# MCP TOOLS
 # ==============================================================================
 
+tools:
+  grafana-mcp:
+    enabled: true
+  querydoc:
+    enabled: true
```
EItanya, line 220: Why do we have a separate block for this, and not just nest it in the individual blocks below?

helm/tools/grafana-mcp/templates/configmap.yaml
EItanya, line 1: Why not just define these in the deployment. Personally I much prefer an `env` block in the values file, and then fill them directly into the deployment.

helm/kagent/values.yaml
onematchfox: The `tools.x` block has a slightly different semantic meaning from the individual blocks below. The `tools.x` block is used to [enable the subcharts](https://github.com/kagent-dev/kagent/blob/36e979c245c5009cd404ee1eeaf6ef98d3d0a593/helm/kagent/Chart-template.yaml#L11-L18) for the various tools, whereas the individual `querydoc` and `grafana-mcp` blocks are used to [configure the subcharts](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart) themselves. I.e., any values/directives within the individual blocks themselves will be passed through/sent to the subchart. Were we to specify `enabled: true` within the individual blocks, this would also be passed through to the subcharts' values. Now, neither of the charts actually defines/makes use of `enabled: true` at present, so there is no real harm in doing that, but this sort of thing can lead to weird/unexpected behaviour, so generally I prefer maintaining a distinction between the two use-cases. I also noticed that the agents themselves are [enabled/disabled/configured](https://github.com/kagent-dev/kagent/blob/36e979c245c5009cd404ee1eeaf6ef98d3d0a593/helm/kagent/values.yaml#L190-L210) in a similar manner, and so I thought I'd be consistent with that for `tools`. That said, I don't have a hard opinion here, so if you'd prefer `enabled` to be part of the individual blocks, then I'll happily make it so.

helm/tools/grafana-mcp/templates/configmap.yaml
onematchfox: Don't have a strong opinion here either - both work. I did it this way to be consistent with how things are done in [querydoc](https://github.com/kagent-dev/kagent/blob/36e979c245c5009cd404ee1eeaf6ef98d3d0a593/helm/tools/querydoc/templates/configmap.yaml). Happy to refactor as per your desire.

EItanya: I'm fine with keeping it here, but can we change the `values.yaml` field to `env` to better reflect where these values end up, `config` is a little too generic

onematchfox: Are you sure about just changing the name? IMO it would be confusing to specify `env` as key/values. `config` is currently a key/value map. `env` on the other hand implies a list of objects like

``` yaml
 - name: FOO
   value: bar
```

... or at least that is how `env` is used on a `Deployment`. I honestly don't mind the use of either, but would recommend keeping to commonly expected formats.

EItanya: ok fine, we can always change later if we need

PR #813, opened 2025-08-26
feat: Add BYO langgraph agents

EItanya:
This PR adds the ability to run langgraph agents as BYO agents in `kagent`.

1. Add `kagent` `checkpointer` implementation to the controller
2. Create `kagent-langgraph` python library to spin up agent

### DIFF
--- a/go/internal/controller/a2a/a2a_reconciler.go
+++ b/go/internal/controller/a2a/a2a_reconciler.go
@@ -13,6 +13,7 @@ import (
 	authimpl "github.com/kagent-dev/kagent/go/internal/httpserver/auth"
 	common "github.com/kagent-dev/kagent/go/internal/utils"
 	"github.com/kagent-dev/kagent/go/pkg/auth"
+	"k8s.io/apimachinery/pkg/types"
 	a2aclient "trpc.group/trpc-go/trpc-a2a-go/client"
 	"trpc.group/trpc-go/trpc-a2a-go/server"
 )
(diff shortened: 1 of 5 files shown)


go/internal/httpserver/handlers/checkpoints.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	threadID := r.URL.Query().Get("thread_id")
+	if threadID == "" {
+		w.RespondWithError(errors.NewBadRequestError("thread_id is required", nil))
+		return
+	}
+
+	checkpointNS := r.URL.Query().Get("checkpoint_ns")
+	if checkpointNS == "" {
```
peterj: Is this check necessary?

go/internal/httpserver/handlers/checkpoints.go
```diff
@@ -0,0 +1,290 @@
...
+
+// KAgentCheckpointTupleResponse represents the response containing checkpoint tuples
+type KAgentCheckpointTupleResponse struct {
+	Data []KAgentCheckpointTuple `json:"data"`
+}
+
+// GetUserIDFromHeaderOrQuery gets user ID from X-User-ID header or user_id query param
+func GetUserIDFromHeaderOrQuery(r *http.Request) (string, error) {
+	// Prefer X-User-ID header
+	if userID := r.Header.Get("X-User-ID"); userID != "" {
```
yuval-k: avoid getting user id directly, prefer to use auth framework

go/internal/httpserver/handlers/checkpoints.go
EItanya: Definitely not :laughing:

python/packages/kagent-langgraph/src/kagent/langgraph/_converters.py
```diff
@@ -0,0 +1,165 @@
...
+                        metadata={
+                            "app_name": app_name,
+                            "session_id": context_id,
+                        },
+                    )
+                )
+
+            elif isinstance(message, ToolMessage):
+                # Handle tool responses
+                if message.content and isinstance(message.content, str):
```
peterj: we also need to handle dictionaries here -- in the example the content from the tool call is a dictionary:

```
ToolMessage: content='{"amount": 1.0, "base": "EUR", "date": "2025-09-01", "rates": {"USD": 1.1715}}' name='get_exchange_rate' id='51dad06a-3c83-4271-90cc-648de7211e5f' tool_call_id='0e356936-9fe5-4609-be30-50658942eba9'
```

(this will result in an empty/missing tool result in the UI)

EItanya: Fixed in https://github.com/kagent-dev/kagent/pull/813/commits/9f070baa3ac14b087b934e7a45e9b978174e2e51

yuval-k (approved):

PR #816, opened 2025-08-27
build(deps): bump versions of `google-adk` and `a2a-sdk`

onematchfox:
Not exactly sure what the process is here - so feel free to let me know if I've missed anything. Mostly interested in the `a2a-sdk` update, as [`v0.3.1`](https://github.com/a2aproject/a2a-python/releases/tag/v0.3.1) contains [a fix](https://github.com/a2aproject/a2a-python/pull/383) I put through to resolve context cancellation issues that have been plaguing me when testing multi-agent workflows. I did scan the release notes of both projects for breaking changes, and I don't think any of them affected this project's usage of the libraries. Ran the default set of agents locally, and everything seems to work fine.

### DIFF
--- a/python/packages/kagent-adk/pyproject.toml
+++ b/python/packages/kagent-adk/pyproject.toml
@@ -25,14 +25,14 @@ dependencies = [
   "anthropic[vertex]>=0.49.0",
   "fastapi>=0.115.1",
   "litellm>=1.74.3",
-  "google-adk>=1.8.0",
+  "google-adk>=1.11.0",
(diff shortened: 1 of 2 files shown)


EItanya:
I have been holding off doing/testing this because of this issue: https://github.com/trpc-group/trpc-a2a-go/issues/78. But if it works it works 🤷 😆 . In general we need to move away from that library because we really don't need 99% of what it's doing in `kagent`

I'm gonna do some testing on this myself to see what happens.

EItanya (approved):

Issue #820, opened 2025-08-27
[FEATURE] Remote Agent Type


inFocus7:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add a new Agent type (`Remote`) to communicate with agents outside of kagent

### ❓ Problem Statement / Motivation

- We can currently only communicate with agents that are either declared inline or byo (container).
- There is no way to set up an agent that exists outside the cluster/kagent system.

### 💡 Proposed Solution

Below is the proposed _partial_ design (sequence graph). There is still work/thought to be done for updating the agent card from remote url. Likely through polling.

<img width="2182" height="1434" alt="Image" src="https://github.com/user-attachments/assets/8a7bda44-8ccd-4167-bb6c-dc557aeb8e79" />

### 🎯 Affected Service(s)

Multiple services / System-wide

### 📚 Additional Context

I've begun work on this since yesterday, although can't self-assign here. It's a WIP, but I'll put a draft PR up by EOD 👍🏼 

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR for this feature

dhruvbehl:
Hi all, I've been running kagent across multiple clusters and kept running into the same wall: there was no clean way to have agents in one cluster delegate to agents in another. I loved the idea of remote agents and felt the need for something analogous to how RemoteMCP works, so I went looking and landed on this issue.

Started digging into the nuances and ended up with a proposal I wanted to share before opening a PR.

### Approach: `RemoteAgent` CRD

Rather than adding a new type value to the existing `Agent` CRD (as explored in #821), I introduced a dedicated `RemoteAgent` resource. A few things pushed me in this direction:

- **Reusability**: in a multi-cluster setup, multiple agents often need to call the same remote endpoint. A standalone CRD lets you define the connection and auth once and reference it from many agents
- **K8s-native observability**: the controller fetches `/.well-known/agent.json` proactively and surfaces `Accepted`/`Reachable` conditions on the resource, so `kubectl get remoteagents` gives you live health across all your remote peers
- **Auth from day one**: `headersFrom` lets you pull bearer tokens or API keys from Secrets, following the same pattern as Gateway API's `SecretObjectReference`
[...]

PR #821, opened 2025-08-27
(Feat) Implement `remote` Agent type

inFocus7:
**wip**

Adds a `Remote` type for Agents.

resolves: https://github.com/kagent-dev/kagent/issues/820


inFocus7:
holy moly, there's a lot of merge conflicts. ~~not sure when/if picking this up again, but note to self: it may be easier to begin a new branch and manually pick over changes.~~

**update**: resolved merge conflicts locally. unsure if it's all still working. need to test it again + do a self code review to make sure I didn't remove any work from main during merging.

need to find my reproduction setup again :rip:

ui/src/lib/messageHandlers.ts
```diff
@@ -26,6 +26,30 @@ export function extractMessagesFromTasks(tasks: Task[]): Message[] {
...
+function getUsageFromWellknownFields(metadata: ADKMetadata | undefined): TokenStats | undefined {
+  if (!metadata) {
+    return undefined;
+  }
+
+  const usage = metadata?.kagent_usage_metadata ||
+    metadata?.usage_metadata as ADKMetadata["kagent_usage_metadata"] ||
+    metadata?.token_usage as ADKMetadata["kagent_usage_metadata"] ||
+    metadata?.usage as ADKMetadata["kagent_usage_metadata"] ||
+    undefined;
```
inFocus7, line 40: this is based on our non-remote stats under `kagent_usage_metadata` and other possible fields that usage would be stored in for remote agents.

we could:
1. Add fields for `remote` type so a user can configure the metadata field(s) for this data (if exposed).
2. Keep as-is
3. [?] Come back to this after figuring out the streaming + Task/history setup.

go/internal/httpserver/server.go
```diff
@@ -181,6 +181,7 @@ func (s *HTTPServer) setupRoutes() {
 	s.router.HandleFunc(APIPathAgents, adaptHandler(s.handlers.Agents.HandleUpdateAgent)).Methods(http.MethodPut)
 	s.router.HandleFunc(APIPathAgents+"/{namespace}/{name}", adaptHandler(s.handlers.Agents.HandleGetAgent)).Methods(http.MethodGet)
 	s.router.HandleFunc(APIPathAgents+"/{namespace}/{name}", adaptHandler(s.handlers.Agents.HandleDeleteAgent)).Methods(http.MethodDelete)
+	s.router.HandleFunc(APIPathAgents+"/{namespace}/{name}/card", adaptHandler(s.handlers.Agents.HandleGetAgentCard)).Methods(http.MethodGet)
```
EItanya: Why do we need this? It should be exposed by the A2A handler already?

go/internal/a2a/recording_manager.go
```diff
@@ -0,0 +1,192 @@
...
+		request.Message.Kind = protocol.KindMessage
+	}
+
+	upstream, err := m.client.StreamMessage(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: If DB is not available, passthrough? or should we return an error since we won't be able to store chat history information?
+	if m.dbClient == nil {
```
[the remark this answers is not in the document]
EItanya: Why would `dbClient` ever not be available?

go/internal/a2a/recording_manager.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			session := &database.Session{
+				ID:      task.ContextID,
+				UserID:  userID,
+				AgentID: agentID,
+			}
+
+			// Store a session (fail silently if duplicate)
+			if err := m.dbClient.StoreSession(session); err != nil {
```
EItanya, line 86: Maybe do a `Get` and do branching logic that way?

go/internal/a2a/recording_manager.go
```diff
@@ -0,0 +1,200 @@
...
+
+	result, err := m.client.SendMessage(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	if task, ok := result.Result.(*protocol.Task); ok {
+		logger := ctrllog.FromContext(ctx).WithName("recording-manager")
+
+		if m.dbClient != nil {
```
EItanya: Same comment, I don't think we need to check this

go/internal/a2a/recording_manager.go
```diff
@@ -0,0 +1,200 @@
...
+
+			// Store a session (fail silently if duplicate)
+			if err := m.dbClient.StoreSession(session); err != nil {
+				logger.Error(err, "Failed to create session", "contextID", task.ContextID, "sessionID", session.ID, "agentID", session.AgentID)
+			}
+
+			// Store Task - always store whatever the remote agent provides
+			if err := m.dbClient.StoreTask(task); err != nil {
+				logger.Error(err, "Failed to store sync task", "taskID", task.ID, "contextID", task.ContextID)
+			}
```
EItanya: Why are we failing silently here? The whole point of this functionality is to record this

go/internal/a2a/recording_manager.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// TODO: If DB is not available, passthrough? or should we return an error since we won't be able to store chat history information?
+	if m.dbClient == nil {
+		return upstream, nil
+	}
+
+	out := make(chan protocol.StreamingMessageEvent)
+
+	// TODO: We shouldn't _expect_ a Task to be sent during streaming.
+	// If not, this means we'd need to have an internal Task. This way we can store it in the database for future reference.
```
EItanya, line 126: Rather than waiting for a `task` to be sent, can we instead create a task on update if one doesn't already exist. Similarly we may need to create a session, which we should do at very beginning.

go/internal/controller/reconciler/reconciler.go
```diff
@@ -563,10 +567,21 @@ func (a *kagentReconciler) upsertAgent(ctx context.Context, agent *v1alpha2.Agen
 	defer a.upsertLock.Unlock()
 
 	id := utils.ConvertToPythonIdentifier(utils.GetObjectRef(agent))
+
+	// Marshal remote agent's AgentCard to store in DB
+	var serializedCard string
+	if agent.Spec.Type == v1alpha2.AgentType_Remote && agentOutputs != nil {
```
EItanya, line 627: How would a remote agent have a card stored from the translation? If we're going to store the agent card we should do it in a separate controller from this one since there's this required async logic of the card.

go/internal/controller/translator/adk_api_translator.go
```diff
@@ -511,6 +558,15 @@ func (a *adkApiTranslator) translateInlineAgent(ctx context.Context, agent *v1al
 					Url:         url,
 					Description: toolAgent.Spec.Description,
 				})
+			case v1alpha2.AgentType_Remote:
+				/* TODO: Add support for remote agents.
```
EItanya: The translator is a pure function, if we are going to make a network call we should do it in the reconciler

go/test/e2e/agents/remote-kebab/app.py
EItanya, line 1: Why do we need a whole separate agent for this? Can't we just treat it as remote in the tests?

EItanya (commented):
The overall design makes sense, but there are some details that need to be worked out before we can merge this.

go/internal/a2a/recording_manager.go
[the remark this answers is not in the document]
inFocus7: There shouldn't be a 'normal' case, assuming _we_ don't explicitly do something nefarious like set it to `nil`. I was planning on adding this check just in case we made some weird mistake on our end that would lead to that, since it would then lead to NPEs.

I removed it now, but provided the context on the reasoning 👍🏼 I'm personally still leaning towards having this nil check just to avoid any potential issue in the future (someone maybe changes it to `nil` and we merge it), but it's not something that would normally/randomly occur on a user environment.

go/internal/a2a/recording_manager.go
inFocus7: Done, I did a more explicit `get -> if not exist, create` for the Session.

If a session did not exist (or get created successfully), then we do not store a task. This is assuming a session -> tasks relation. If we allow for Tasks to exists without a Session, I could change this to always create a Task.

go/test/e2e/agents/remote-kebab/app.py
inFocus7: I'm going to play around with this idea a bit more.

Initially I was against it because it would mean we _need_ the `BYO` kebab Agent to be created first, so that its deployment + service exists and the `Remote` Agent can reference the service URL. But should be simple to deal with.

**Now**, after trying it out: the remote agent test passes, **but** when I try to chat with the remote agent...

Response in chat:
```
Client error '403 Forbidden' for url 'http://kagent-controller.kagent:8083/api/sessions/ctx-a6f000bf-7e85-4465-ab06-f08dd9be04d7/events?user_id=admin@kagent.dev' For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
```

Kebab agent's deployment:
```
INFO:     10.244.0.51:43074 - "POST / HTTP/1.1" 200 OK
ERROR:google_adk.kagent.adk._agent_executor:Error handling A2A request: Client error '403 Forbidden' for url 'http://kagent-controller.kagent:8083/api/sessions/ctx-a6f000bf-7e85-4465-ab06-f08dd9be04d7/events?user_id=admin@kagent.dev'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
Traceback (most recent call last):
  File "/.kagent/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py", line 125, in execute
    await self._handle_request(context, event_queue, runner)
  File "/.kagent/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py", line 206, in _handle_request
    async for adk_event in runner.run_async(**run_args):
    ...<4 lines>...
            await event_queue.enqueue_event(a2a_event)
  File "/.kagent/python/.venv/lib/python3.13/site-packages/google/adk/runners.py", line 250, in run_async
    async for event in agen:
      yield event
  File "/.kagent/python/.venv/lib/python3.13/site-packages/google/adk/runners.py", line 228, in _run_with_trace
    await self._append_new_message_to_session(
    ...<5 lines>...
    )
  File "/.kagent/python/.venv/lib/python3.13/site-packages/google/adk/runners.py", line 357, in _append_new_message_to_session
    await self.session_service.append_event(session=session, event=event)
  File "/.kagent/python/packages/kagent-adk/src/kagent/adk/_session_service.py", line 168, in append_event
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/.kagent/python/.venv/lib/python3.13/site-packages/httpx/_models.py", line 829, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '403 Forbidden' for url 'http://kagent-controller.kagent:8083/api/sessions/ctx-a6f000bf-7e85-4465-ab06-f08dd9be04d7/events?user_id=admin@kagent.dev'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
INFO:     10.244.0.1:51740 - "GET /health HTTP/1.1" 200 OK
INFO:     10.244.0.1:46628 - "GET /health HTTP/1.1" 200 OK
```

---

I'll need to dig into this, but it's probable that it's due to the remote agent a2a storing information in the DB **and** the other agent doing so as well (since byo + declarative agents also store in the db themselves). I'm not even sure if this would be a "real world" issue, since a user likely wouldn't be defining an agent created by kagent as a remote agent. (unless there's valid cross-cluster use cases for this, but in this case, it would be good to config `remote` agent with a bool field to _not_ store in the db)

inFocus7: Yeah, so while the test (somehow) passes, the session storage is a big issue if the remote agent writes to the database as well.

```
kagent >> get session
+---+------------------------------------------+--------------------+--------------------------------+----------------------+
| # | ID                                       | NAME               | AGENT                          | CREATED              |
+---+------------------------------------------+--------------------+--------------------------------+----------------------+
| 1 | ctx-5b33e4b2-d2d6-4854-955c-0f6dea1a282f | remote-kebab-agent | kagent__NS__remote_kebab_agent | 2025-09-17T22:54:21Z |
+---+------------------------------------------+--------------------+--------------------------------+----------------------+
```
_remote agent "owns" this session id, so when the agent it communicates to tries accessing/writing it, it fails. which makes sense._

So what happens is that
1. The remote Agent handler creates a Session (as expected)
2. It communicates with the other Agent
3. The other agent was created by us, meaning it has DB access
4. The other agent tries creating/getting the Session
5. The other agent fails and returns an error, because it already exists and is owned by the Remote Agent.

If we want to allow the case where a remote agent is one which already writes its sessions to the DB, then we'll want to add a spec field to disable the database writing -- or do something else where it passes a different context/id to the remote agent. The main issue here would be figuring out how to handle their own sessions/display.
```yaml
[...]
    type: Remote
    remote:
      [urls...]
      persistData: false # or persistTask/Session, persist, store, etc..
```

go/internal/httpserver/server.go
inFocus7: Which a2a handler? Like during communication?

To answer this + [this comment](https://github.com/kagent-dev/kagent/pull/821#discussion_r2356339490), the idea I had for remote agents was:
1. We will poll for the agent card using the `agentCardUrl`.
2. We would first get the agent card during reconciliation in order to validate the endpoint and handle the status update.
    - this would allow us to update the CRD with a status error if it wasn't able to fetch the card endpoint.
3. It would be nice if users can view the remote agent card's details.
4. Because we're fetching it on a poll-basis, we should display the _latest polled_ agent card details - as opposed to dynamically calling the card endpoint and showing the latest.
5. Therefore, we would need to store the polled agent card somewhere.

inFocus7: Actually, I get you about the pure function for translation reworking it now to move the network call to the reconciliation.

Although my logic for the above still is the same (although agent card fetch is in reconciliation)

go/internal/controller/translator/adk_api_translator.go
inFocus7: Makes sense. Just need to figure out how the polling of agent card ties into this a bit better. I was/am planning on handling polling logic in a follow-up to avoid blowing up this PR with diffs.

Self-thought/note:
I _think_ we do something along these lines for MCP tools, where there's a watcher that updates agents when mcp tools update. So something like that would be done for remote agent-as-a-tool, where updates would occur when they get re-polled (or when the remote agent reconciles).

Additionally, doing something like storing the agent card _hashed_ in the remote agent annotation _could_ kick off its reconciliation, which would should then update the usages of it as a tool from agents watching it.

---

Would probably be worth huddling (unless this is an easy yes/no). For using it as a tool we can:
1. Add a watcher, so when remote agents reconcile, agents using them as a tool do so as well
2. During reconciliation, remote agent urls/configs are updated based on latest data (from the database)
This way it stays up to date. 

When we implement polling, we'll need to see how to handle it (in my opinion adding an annotation on remote agents that is a hash of the config should be good. this way it forces a reconciliation + caller agents update). I'm trying this locally. It works in theory, but getting a 503 adk error when the caller agent tries calling the remote agent 🤔

go/internal/httpserver/server.go
peterj: removed it as it wasn't needed/used anywhere.

go/internal/controller/translator/adk_api_translator.go
```diff
@@ -510,20 +505,28 @@ func (a *adkApiTranslator) translateInlineAgent(ctx context.Context, agent *v1al
...
 					Headers:     headers,
 					Description: toolAgent.Spec.Description,
 				})
+			case v1alpha2.AgentType_Remote:
+				cfg.RemoteAgents = append(cfg.RemoteAgents, adk.RemoteAgentConfig{
+					Name:        utils.ConvertToPythonIdentifier(utils.GetObjectRef(toolAgent)),
+					Url:         agent.Spec.Remote.DiscoveryURL,
+					Headers:     headers,
+					Description: toolAgent.Spec.Description,
+				})
```
inFocus7, line 529: Q: Since we're not using Watches, does this mean that if a remote agent's discovery URL is updated, any Agent referencing it as a tool won't get it until they manually reconcile?

(iiirc, the deployment for declarative agents have their agent-as-tools urls "baked in" their secrets, which update on reconciliation.)

ae12345678910:
When will this be merged into kagent? This is a feature that I would find very valuable

Huimintai:
When this feature will be merged, it is very important for us to use remote agents.

PR #822, opened 2025-08-28
Decouple KMCP integration

onematchfox:
This PR decouples the kmcp integration, ensuring that kmcp is a separate, optional component that can be deployed alongside, and independently of, kagent. 

The kmcp integration remains fully in place and, by default, is enabled as is the case in the current version. The main difference being that, instead of the hard-coupling that is in place today, where the `kagent-controller` [runs the kmcp controller](https://github.com/kagent-dev/kagent/blob/bdbde1109e7fd214a2536940043a6afef8e07c92/go/pkg/app/app.go#L323-L329), the kmcp controller is deployed as a separate, standalone component/pod. This is important for numerous reasons... not least of which is the fact that these are separate projects with separate lifecycles and one should be able to manage them independently if desired.

Notes:
- The `KMCP_ENABLED` flag (and associated Helm setting) merely determines whether the kmcp controller and associated CRDs are installed as part of the `kagent` and `kagent-crds`. Users can set this to false and install kmcp independently of the `kagent` Helm chart, and the controller will still work with any additional configuration. The controller simply enables/disables reconciliation of `MCPServer` resources based on whether or not the CRD is present in the cluster. If desired, in the future, a flag could be added to the controller to disable/enable the integration manually, but for now, I think that's overkill - the (automatic) integration is great :love_letter: 
- The default setup will currently break if `KMCP_ENABLED` is set to `false`. This is due to the use of the `MCPServer` for the [`grafana` MCP](https://github.com/kagent-dev/kagent/blob/bdbde1109e7fd214a2536940043a6afef8e07c92/helm/agents/observability/templates/mcp-server.yaml#L12-L31) as part of the `observability` agent. This is addressed as part of #812, which switches to using `RemoteMCPServer`, ensuring that all built-in components/agents/tools work regardless of whether KMCP is enabled.

### DIFF
--- a/go/cli/internal/cli/agent/deploy.go
+++ b/go/cli/internal/cli/agent/deploy.go
@@ -8,11 +8,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/kagent-dev/kagent/go/api/v1alpha1"
 	"github.com/kagent-dev/kagent/go/api/v1alpha2"
 	"github.com/kagent-dev/kagent/go/cli/internal/agent/frameworks/common"
 	commonimage "github.com/kagent-dev/kagent/go/cli/internal/common/image"
 	"github.com/kagent-dev/kagent/go/cli/internal/config"
+	"github.com/kagent-dev/kmcp/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"


Makefile
```diff
@@ -272,6 +274,8 @@ helm-install-provider: helm-version check-openai-key
 		--set providers.azureOpenAI.apiKey=$(AZUREOPENAI_API_KEY) \
 		--set providers.anthropic.apiKey=$(ANTHROPIC_API_KEY) \
 		--set providers.default=$(KAGENT_DEFAULT_MODEL_PROVIDER) \
+		--set kmcp.enabled=$(KMCP_ENABLED) \
+		--set kmcp.image.tag=$(KMCP_VERSION) \
```
onematchfox, line 323: Only required due to https://github.com/kagent-dev/kmcp/issues/74

helm/kagent/values.yaml
```diff
@@ -172,6 +172,17 @@ providers:
...
 
+# ==============================================================================
+# KMCP
+# ==============================================================================
+
+kmcp:
+  enabled: true
+  nameOverride: "kmcp"
+  fullnameOverride: "" # Override if desired
+  namespaceOverride: "" # Override if desired (in conjunction with fullnameOverride to drop the `kagent-` prefix if you want)
```
onematchfox, line 189: These don't actually do anything - yet. So for now, the KMCP controller will be deployed as a `Deployment` name `kagent-controller-manager` within the `kagent` namespace. Support for these values was added in https://github.com/kagent-dev/kmcp/pull/72 and should be released in `0.1.7`. Once released, the `Deployment` will be renamed to `kagent-kmcp-controller-manager` by default with user's being able to override the full name using `fullnameOverride` and the namespace via `namespaceOverride`.

helm/kagent/values.yaml
```diff
@@ -172,6 +172,17 @@ providers:
       azureDeployment: ""
       azureEndpoint: ""
 
+# ==============================================================================
+# KMCP
+# ==============================================================================
+
+kmcp:
+  enabled: true
```
onematchfox, line 186: Related discussion [here](https://github.com/kagent-dev/kagent/pull/812#discussion_r2304884876). Have used `kmcp.enabled` here instead of something like `integrations.kmcp.enabled`, but happy to rework things if desired.

go/internal/controller/mcp_server_controller.go
```diff
@@ -51,11 +54,16 @@ func (r *MCPServerController) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return nil
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(controller.Options{
 			NeedLeaderElection: ptr.To(true),
 		}).
 		For(&v1alpha1.MCPServer{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Named("toolserver").
+		Named("mcpserver").
```
onematchfox: This change wasn't possible previously (see https://github.com/kagent-dev/kagent/pull/757) because the code was already running [a controller named `mcpserver`](https://github.com/kagent-dev/kmcp/blob/41071d12350f22dd20492f5848e0629ba5ee7c0c/pkg/controller/mcpserver_controller.go#L108). I.e. this code was running 2 separate controllers reconciling the same resources. :anguished:

onematchfox:
Just realised that I haven't even considered the UI implications of this change, and I'm going to go ahead and guess that there's stuff that needs to be handled there as well. Will try to take a look ASA,P but feel free to point me in the right direction if you like :pray:

go/internal/httpserver/handlers/toolservertypes.go
```diff
@@ -0,0 +1,71 @@
...
+	ToolServerTypeRemoteMCPServer ToolServerType = "RemoteMCPServer"
+	ToolServerTypeMCPServer       ToolServerType = "MCPServer"
+)
+
+var toolServerTypes = ToolServerTypes{
+	ToolServerTypeRemoteMCPServer,
+}
+
+// HandleListToolServerTypes handles GET /api/toolservertypes requests
+func (h *ToolServerTypesHandler) HandleListToolServerTypes(w ErrorResponseWriter, r *http.Request) {
```
onematchfox, line 69: Not 100% sure about this approach. I.e., adding an API endpoint and then calling that via the UI. But, it seemed to fit best with the current usage patterns between UI<->API vs. e.g., having the UI query the cluster directly. It may also prove a useful extension point if other "tool server" types/integrations are added later on. :shrug: 

More than happy to switch approach if anyone has any thoughts.

ui/src/components/AddServerDialog.tsx
```diff
@@ -526,16 +564,19 @@ export function AddServerDialog({ open, onOpenChange, onAddServer, onError }: Ad
...
-                <TabsTrigger value="command" className="flex items-center gap-2">
-                  <Terminal className="h-4 w-4" />
-                  Command
-                </TabsTrigger>
                 <TabsTrigger value="url" className="flex items-center gap-2">
                   <Globe className="h-4 w-4" />
                   URL
                 </TabsTrigger>
+                {renderTabTrigger(
+                  "command",
```
onematchfox, line 513: Left the tabs as "url" and "command" for now although I can't help but feel like this dialog/the tabs need a bit of an overhaul given that ultimately, [`url` = `RemoteMCPServer`](https://github.com/kagent-dev/kagent/blob/bdbde1109e7fd214a2536940043a6afef8e07c92/ui/src/components/AddServerDialog.tsx#L347-L393) and [`command` = `MCPServer`](https://github.com/kagent-dev/kagent/blob/bdbde1109e7fd214a2536940043a6afef8e07c92/ui/src/components/AddServerDialog.tsx#L283-L345) (which incidentally now needs to support a far broader range of `cmd` than just `npx` and `uvx`).

peterj:
> Just realised that I haven't even considered the UI implications of this change, and I'm going to go ahead and guess that there's stuff that needs to be handled there as well. Will try to take a look ASA,P but feel free to point me in the right direction if you like 🙏

In the UI, we'll have to block the creation of tools that use MCPServer resource (the stdio ones, the sse use the RemoteMCPServer crds).

onematchfox:
> In the UI, we'll have to block the creation of tools that use MCPServer resource (the stdio ones, the sse use the RemoteMCPServer crds).

Done :tada: Screenshot of what it looks like when `MCPServer` CRDs are not found. Moved URL to be the default/first tab so user's won't ever start on a disabled tab.

image

onematchfox:
Had a bit of a discussion with @EItanya on Discord regarding this PR, and I was made aware that there is some history as to why things were implemented the way they have been. We will discuss further at the next contributors meeting. In the meantime, I will open for review and try to address any comments/feedback that are received between now and then. That should also hopefully provide time for the comments [here](https://github.com/kagent-dev/kagent/pull/822#discussion_r2306673031) and [here](https://github.com/kagent-dev/kagent/pull/822#discussion_r2306682356) to be addressed.

ui/src/components/AddServerDialog.tsx
peterj: I agree. As a start, it might make sense to rename them `RemoteMCPServer` and `MCPServer`.

onematchfox: Do you want to do that as part of this PR or as a separate one?

ui/src/components/AddServerDialog.tsx
```diff
@@ -28,8 +29,8 @@ interface EnvPair {
   value: string;
 }
 
-export function AddServerDialog({ open, onOpenChange, onAddServer, onError }: AddServerDialogProps) {
-  const [activeTab, setActiveTab] = useState<"command" | "url">("command");
+export function AddServerDialog({ open, supportedToolServerTypes, onOpenChange, onAddServer, onError }: AddServerDialogProps) {
+  const [activeTab, setActiveTab] = useState<"command" | "url">("url");
```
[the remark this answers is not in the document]
onematchfox: This was intentionally changed to ensure that the default tab is always enabled.

ui/src/components/AddServerDialog.tsx
peterj: up to you

onematchfox: Let's deal with it separately since it will be needed regardless of what happens with this PR.

onematchfox:
Superseded by #939

PR #823, opened 2025-08-28
support litellm extra_headers

yanivmn:
support sending custom headers from model to litellm
@EItanya - should extra_headers be set to all model types?

### DIFF
--- a/go/internal/controller/translator/adk_api_translator.go
+++ b/go/internal/controller/translator/adk_api_translator.go
@@ -541,6 +541,7 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 	}
 
 	modelDeploymentData := &modelDeploymentData{}
+
 	switch model.Spec.Provider {
 	case v1alpha2.ModelProviderOpenAI:
 		if model.Spec.APIKeySecret != "" {
@@ -558,7 +559,8 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 		}
 		openai := &adk.OpenAI{
 			BaseModel: adk.BaseModel{
-				Model: model.Spec.Model,
+				Model:   model.Spec.Model,
+				Headers: model.Spec.DefaultHeaders,
 			},
 		}
 		if model.Spec.OpenAI != nil {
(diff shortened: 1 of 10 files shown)


EItanya:
> support sending custom headers from model to litellm @EItanya - should extra_headers be set to all model types?

If it's easy than yes

yanivmn:
@EItanya - applied to all LiteLLM usages

yanivmn:
@yuval-k & @EItanya - can someone please approve this PR?

peterj:
@yanivmn can you sign the commits? https://github.com/kagent-dev/kagent/pull/823/checks?check_run_id=49272505056

yanivmn:
@peterj - done please review again

go/internal/controller/translator/adk_api_translator.go
```diff
@@ -470,6 +470,17 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 	}
 
 	modelDeploymentData := &modelDeploymentData{}
+
+	// Propagate default model headers to the agent runtime so it can forward them to the provider.
+	if len(model.Spec.DefaultHeaders) > 0 {
+		if b, err := json.Marshal(model.Spec.DefaultHeaders); err == nil {
+			modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
+				Name:  "KAGENT_MODEL_DEFAULT_HEADERS",
```
EItanya: Why is this in an env value instead of the API models?

EItanya (changes requested):

go/internal/controller/translator/adk_api_translator.go
yanivmn: how can we share this to agent deployment? model config is not shared or mounted.

EItanya: Let me explain better. The declarative ADK agents have python types which define how they are configured, see [here](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent_adk/models.py#L71). Those types are mirrored in go and then configured in the translator, see [here](https://github.com/kagent-dev/kagent/blob/main/go/internal/adk/types.go#L224). I am saying that rather than defining these headers in an env var in Python, they should be passed via this well known config.

yanivmn: Both of your links are missing to its hard to follow

EItanya: Oh crap that PR merged, gimme a sec, just edited the message with the correct links

yanivmn: @EItanya - Thanks for your comments, PR updated to use model over ENV.
Please review again.

yanivmn:
@EItanya - PR modified are rebased to support latest OpenAI changes, please review

EItanya (approved):

PR #829, opened 2025-08-29
feat: add additional span info

EItanya:


python/packages/kagent-adk/src/kagent_adk/_agent_executor.py
```diff
@@ -166,6 +167,14 @@ async def _handle_request(
         # ensure the session exists
         session = await self._prepare_session(context, run_args, runner)
 
+        current_span = trace.get_current_span()
+        if context.task_id:
+            current_span.set_attribute("kagent.task_id", context.task_id)
+        if run_args["user_id"]:
+            current_span.set_attribute("kagent.user_id", run_args["user_id"])
+        if run_args["session_id"]:
+            current_span.set_attribute("kagent.session_id", run_args["session_id"])
```
krisztianfekete: Do we want to namespace some of these to `kagent.`? It makes sense if something is unique to kagent, but aren't all these covered by `gen_ai` semconv? E.g. session_id is essentially [gen_ai.converstation.id](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-conversation-id)?

EItanya: I'm open to anything, which names do you prefer? Maybe
`kagent.session_id -> gen_ai.converstation.id`, `kagent.task_id -> a2a.task_id`, and leave `kagent.user_id` as is.

go/internal/controller/translator/adk_api_translator.go
```diff
@@ -146,6 +152,57 @@ func (a *adkApiTranslator) TranslateAgent(
...
+
+	if state.isVisited(agentRef) {
+		return fmt.Errorf("cycle detected in agent tool chain: %s -> %s", agentRef, agentRef)
+	}
+
+	if state.depth > MAX_DEPTH {
+		return fmt.Errorf("recursion limit reached in agent tool chain: %s -> %s", agentRef, agentRef)
+	}
+
+	if agent.Spec.Type != v1alpha2.AgentType_Declarative {
```
peterj, line 167: should you check this before the previous two checks or does the max depth and cycle checks apply to declarative agents asw ell?

python/packages/kagent-adk/src/kagent_adk/_agent_executor.py
krisztianfekete: Currently it looks like `gen_ai.task.id` will be a thing as per https://github.com/open-telemetry/semantic-conventions/issues/2665, so I'd go with the gen_ai ones for the ones that match/will match, and keep user_id for kagent.

PR #831, opened 2025-08-29
add downstream extension points

ilackarms:
adds extension points for downstream forks / extensions of the core kagent code base

depends on https://github.com/kagent-dev/kmcp/pull/78

### DIFF
--- a/go/internal/controller/translator/adk_api_translator.go
+++ b/go/internal/controller/translator/adk_api_translator.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"maps"
 	"os"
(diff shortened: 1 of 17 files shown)


go/internal/controller/translator/adk_api_translator.go
```diff
@@ -6,12 +6,15 @@ import (
...
 	"encoding/json"
 	"fmt"
+	"github.com/kagent-dev/kagent/go/pkg/translator"
 	"maps"
 	"os"
 	"slices"
 	"strconv"
 	"strings"
 
+	"go.uber.org/multierr"
```
yuval-k: we can use stdlib error.Join instead

yuval-k (commented):
overall lgtm

go/internal/controller/translator/adk_api_translator.go
```diff
@@ -317,25 +334,21 @@ func (a *adkApiTranslator) buildManifest(
...
 	if len(dep.Cmd) != 0 {
 		cmd = []string{dep.Cmd}
 	}
 
+	podTemplateAnnotations := dep.Annotations
+	if podTemplateAnnotations == nil {
+		podTemplateAnnotations = map[string]string{}
+	}
+	// Add config hash annotation to pod template to force rollout on config changes
+	podTemplateAnnotations["kagent.dev/config-hash"] = fmt.Sprintf("%d", configHash)
```
EItanya, line 347: We only want to add this if config hash is set

ilackarms: apparently we always set it now

EItanya (dismissed):
Changes look good to me, but commits need to be signed for DCO before we can merge

EItanya (approved):

PR #838, opened 2025-09-02
(Feat) Implement `init`, `build` and `deploy cmd for kagent cli

jmhbh:


go/cli/internal/frameworks/adk/python/templates/Dockerfile.tmpl
```diff
@@ -0,0 +1,15 @@
+ARG DOCKER_REGISTRY=ghcr.io
+# TODO: use latest tag once it is available
+ARG VERSION=0.6.7
```
jmhbh: This is currently hardcoded to the latest version as we do not publish the `kagent-adk` image with the tag `latest`. This should be updated once we do publish an image with the tag `latest`

go/cli/internal/frameworks/adk/python/generator.go
```diff
@@ -0,0 +1,124 @@
...
+	fmt.Printf("   └── README.md\n")
+	fmt.Printf("\n🚀 Next steps:\n")
+	fmt.Printf("   1. cd %s\n", agentConfig.Name)
+	fmt.Printf("   2. Customize the agent in %s/agent.py\n", agentConfig.Name)
+	fmt.Printf("   3. Build the agent image and push it to the local registry\n")
+	fmt.Printf("      kagent build %s --push\n", agentConfig.Name)
+	fmt.Printf("   4. Deploy the agent to your local cluster\n")
+	fmt.Printf("      kagent deploy %s --api-key-secret <secret-name>\n", agentConfig.Name)
+	fmt.Printf("      Or use --api-key for convenience: kagent deploy %s --api-key <api-key>\n", agentConfig.Name)
+	fmt.Printf("      Support for using a credential file is coming soon\n")
```
jmhbh, line 123: Currently blocked on procuring the appropriate credentials to test `AnthropicVertexAI` and `VertexAI` which is why I have temporarily removed support for these model providers for now and will add them in a follow up PR.

go/cli/cmd/kagent/main.go
```diff
@@ -209,7 +210,166 @@ func main() {
...
+			// Get project directory from positional argument
+			buildCfg.ProjectDir = args[0]
+
+			if image, _ := cmd.Flags().GetString("image"); image != "" {
+				buildCfg.Image = image
+			}
+
+			if push, _ := cmd.Flags().GetBool("push"); push {
+				buildCfg.Push = true
+			}
```
EItanya: Why are you doing `gets` directly rather than setting these values on a config struct?

go/cli/cmd/kagent/main.go
```diff
@@ -494,6 +654,139 @@ Example:
 		},
 	})
 
+	shell.AddCmd(&ishell.Cmd{
```
EItanya: You can remove the interactive commands, we don't plan to fully support that UX going forward

go/cli/internal/cli/deploy.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	var config *rest.Config
+	var err error
+
+	if kubeconfig != "" && fileExists(kubeconfig) {
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
+	} else {
+		// Try in-cluster config
+		config, err = rest.InClusterConfig()
```
EItanya: Why are you trying this? this should never be running in cluster

go/cli/cmd/kagent/main.go
```diff
@@ -209,7 +210,166 @@ func main() {
...
+  kagent init adk python dice
+  kagent init adk python dice --instruction-file instructions.md
+  kagent init adk python dice --model-provider Gemini --model-name gemini-2.0-flash`,
+		Args: cobra.ExactArgs(3),
+		Run: func(cmd *cobra.Command, args []string) {
+			initCfg := &cli.InitCfg{
+				Config:    cfg,
+				Framework: args[0],
+				Language:  args[1],
+				AgentName: args[2],
```
EItanya: This can panic, you need to make sure args is long enough, or use cobra built in validators

go/cli/cmd/kagent/main.go
```diff
@@ -209,7 +210,166 @@ func main() {
...
+				initCfg.ModelProvider = modelProvider
+			}
+
+			if modelName, _ := cmd.Flags().GetString("model-name"); modelName != "" {
+				initCfg.ModelName = modelName
+			}
+
+			if description, _ := cmd.Flags().GetString("description"); description != "" {
+				initCfg.Description = description
+			}
```
EItanya: Bind these to config vars

go/cli/internal/cli/deploy.go
```diff
@@ -0,0 +1,296 @@
...
+			},
+		},
+	}
+
+	// Check if agent already exists
+	existingAgent := &v1alpha2.Agent{}
+	err := k8sClient.Get(context.Background(), client.ObjectKey{Namespace: cfg.Config.Namespace, Name: manifest.Name}, existingAgent)
+
+	if err == nil {
+		// Agent exists, update it
```
EItanya: This isn't necessarily true, you ned to check if `errors.IsNotFound`

go/cli/internal/cli/init.go
```diff
@@ -0,0 +1,97 @@
...
+	Config          *config.Config
+}
+
+func InitCmd(cfg *InitCfg) error {
+	// Validate framework and language
+	if cfg.Framework != "adk" {
+		return fmt.Errorf("unsupported framework: %s. Only 'adk' is supported", cfg.Framework)
+	}
+
+	if cfg.Framework == "adk" && cfg.Language != "python" {
```
EItanya: You've already checked for `adk`

go/cli/internal/cli/init.go
```diff
@@ -0,0 +1,97 @@
...
+	if cfg.ModelProvider != "" {
+		if err := validateModelProvider(cfg.ModelProvider); err != nil {
+			return err
+		}
+	}
+
+	if cfg.ModelProvider == "" && cfg.ModelName == "" {
+		cfg.ModelProvider = string(v1alpha2.ModelProviderGemini)
+		cfg.ModelName = "gemini-2.0-flash"
+	}
```
EItanya: I don't think we should default here, it's a little too user specific.

go/cli/internal/frameworks/adk/python/templates/Dockerfile.tmpl
EItanya: Can you create an issue to tag this with latest.

go/cli/cmd/kagent/main.go
jmhbh: using `cobra.ExactArgs(3)` to ensure that the required number of args are passed to the cmd here and for the other cmds as well.

go/cli/internal/cli/deploy.go
jmhbh: this was a mistake I updated this to use `clientcmd.NewNonInteractiveDeferredLoadingClientConfig` instead which is a cleaner way to load the kubeConfig.

go/cli/cmd/kagent/main.go
jmhbh: updated these flags to bind to the config struct instead.

go/cli/internal/cli/init.go
jmhbh: Removed the default here but added defaults for the `--model-name` and `--model-provider` flags instead.

go/cli/internal/cli/deploy.go
jmhbh: refactored this to check for `errors.IsNotFound` and create the agent CRD if the agent isn't found. If there aren't any errors then we will update the agent CRD.

go/cli/internal/frameworks/adk/python/templates/Dockerfile.tmpl
jmhbh: created an issue [here](https://github.com/kagent-dev/kagent/issues/845)

go/cli/internal/frameworks/adk/python/templates/agent/agent.py.tmpl
```diff
@@ -0,0 +1,79 @@
...
+    if provider.lower() == "gemini":
+        # For Gemini, use the model name directly
+        return model_name
+    elif provider.lower() in provider_mapping:
+        # Use LiteLLM with the appropriate prefix
+        prefix = provider_mapping[provider.lower()]
+        return LiteLlm(model=f"{prefix}/{model_name}")
+    else:
+        # For unknown providers, try using the model name directly
+        return model_name
```
EItanya: Rather than generating this function every time, can't we generate a different function depending on the input vars: `("{{.ModelProvider}}", "{{.ModelName}}")`

go/cli/internal/cli/build.go
```diff
@@ -0,0 +1,138 @@
...
+	// Check if project directory exists
+	if _, err := os.Stat(cfg.ProjectDir); os.IsNotExist(err) {
+		return fmt.Errorf("project directory does not exist: %s", cfg.ProjectDir)
+	}
+
+	// Check if Dockerfile exists in project directory
+	dockerfilePath := filepath.Join(cfg.ProjectDir, "Dockerfile")
+	if _, err := os.Stat(dockerfilePath); os.IsNotExist(err) {
+		return fmt.Errorf("Dockerfile not found in project directory: %s", dockerfilePath)
+	}
```
EItanya, line 37: For build and push, should we make sure `docker` is installed/available first?

EItanya (dismissed):
Going to approve as is, but I would like to see some of these fixes in a follow-up

go/cli/internal/frameworks/adk/python/templates/agent/agent.py.tmpl
```diff
@@ -42,31 +42,32 @@ async def check_prime(nums: list[int]) -> str:
...
-        return model_name
+# Create model based on provider and model name
+{{if eq .ModelProvider "gemini"}}
+def create_model():
+    """Create a Gemini model instance."""
+    return "{{.ModelName}}"
+{{else if eq .ModelProvider "openai"}}
+def create_model():
+    """Create an OpenAI model instance using LiteLLM."""
+    return LiteLlm(model="openai/{{.ModelName}}")
```
EItanya, line 53: We no longer use `LiteLlm` for `openai`, please see the latest code.

go/cli/internal/frameworks/adk/python/templates/agent/agent.py.tmpl
```diff
@@ -42,31 +42,32 @@ async def check_prime(nums: list[int]) -> str:
...
+{{else if eq .ModelProvider "anthropic"}}
+def create_model():
+    """Create an Anthropic model instance using LiteLLM."""
+    return LiteLlm(model="anthropic/{{.ModelName}}")
+{{else if eq .ModelProvider "azureopenai"}}
+def create_model():
+    """Create an Azure OpenAI model instance using LiteLLM."""
+    return LiteLlm(model="azure/{{.ModelName}}")
```
EItanya, line 61: Same for azure

Issue #779, opened 2025-08-20
[FEATURE] Add customizable timeout for LLM calls


EdmondDantes21:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Allow for the creation of user defined timeout for LLM calls.

### ❓ Problem Statement / Motivation

Currently, when users run large prompts on thinking models such as Gemini 2.5 Pro, the response will take longer than 60 seconds to arrive, which triggers a timeout.

### 💡 Proposed Solution

Allow the definition of a user defined timeout for LLM calls.

### 🔄 Alternatives Considered

- Adding longer `proxy_read_timeout` and `proxy_send_timeout` for all routes in the `nginx.conf` file does not seem to work.

### 🎯 Affected Service(s)

App Service

### 📚 Additional Context

Below are the logs when a timeout occurs

app logs:

```py
WARNING:google_adk.kagent_adk._agent_executor:Runner.close() was cancelled; suppressing during cleanup

ERROR:asyncio:Exception in callback EventConsumer.agent_task_callback() at /.kagent/python/.venv/lib/python3.13/site-packages/a2a/server/events/event_consumer.py:153

handle: <Handle EventConsumer.agent_task_callback() at /.kagent/python/.venv/lib/python3.13/site-packages/a2a/server/events/event_consumer.py:153>

Traceback (most recent call last):

File "/.kagent/python/.venv/lib/python3.13/site-packages/google/genai/_api_client.py", line 1089, in _async_request_once

response = await session.request(

^^^^^^^^^^^^^^^^^^^^^^

...<6 lines>...

)

^

File "/.kagent/python/.venv/lib/python3.13/site-packages/aiohttp/client.py", line 770, in _request

resp = await handler(req)

^^^^^^^^^^^^^^^^^^

File "/.kagent/python/.venv/lib/python3.13/site-packages/aiohttp/client.py", line 748, in _connect_and_send_request

await resp.start(conn)

File "/.kagent/python/.venv/lib/python3.13/site-packages/aiohttp/client_reqrep.py", line 532, in start

message, payload = await protocol.read() # type: ignore[union-attr]

^^^^^^^^^^^^^^^^^^^^^

[...]

PR #839, opened 2025-09-02
feat: Implement user-configurable timeout for LLM API calls

xrwang8:
Summary

  This pull request addresses the issue of LLM API calls having a hardcoded timeout, which was often insufficient for models processing large or complex prompts, leading to
  premature timeout errors. This change introduces a user-configurable timeout mechanism to provide greater flexibility and robustness.

  Solution Implemented

  The solution is implemented across the full stack, from the Kubernetes API to the Python agent runtime:

   1. `ModelConfig` CRD Update (Go):
       * An optional timeout field (in seconds) has been added to the spec for each model provider within the ModelConfig Custom Resource Definition.
       * This allows platform operators to define a specific timeout for each LLM configuration directly in their Kubernetes manifests.
       * It defaults to 300 seconds and is validated to be within a range of 1 to 3600 seconds.

   2. Controller Logic (Go):
       * The controller has been updated to read the timeout value from the ModelConfig resource.
       * It then translates and propagates this configuration to the agent's runtime environment.

   3. Agent Development Kit (Python):
       * The Python ADK now correctly receives the timeout setting.
       * This value is used to configure the timeout on the underlying LLM client making the API request.
       * An overarching asyncio.timeout() is also applied to the agent execution task to ensure the entire operation respects the configured timeout.

  This comprehensive approach ensures that users have full, granular control over LLM timeouts, making the system more reliable for long-running tasks.


xrwang8:
https://github.com/kagent-dev/kagent/issues/779

xrwang8:
@EItanya  PTAL

python/packages/kagent-adk/src/kagent_adk/models.py
```diff
@@ -125,4 +129,5 @@ def to_agent(self, name: str) -> Agent:
             instruction=self.instruction,
             tools=mcp_toolsets,
             sub_agents=remote_agents,
+            timeout=timeout,
```
EItanya, line 132: This is actually slightly different than a model timeout? What does the `Agent` timeout mean in this case?

python/packages/kagent-adk/src/kagent_adk/_agent_executor.py
```diff
@@ -191,13 +191,21 @@ async def _handle_request(
             )
         )
 
+        # Extract timeout configuration from runner
+        timeout = getattr(runner, 'timeout', 300)  # Default 5 minutes
```
EItanya, line 195: Rather than getting this from the runner, can we use `A2aAgentExecutorConfig` above?

go/api/v1alpha2/modelconfig_types.go
EItanya, line 1: Since we are applying this to all Model Configs, can we create a `BaseModelConfig` struct which we embed in all other types?

Issue #853, opened 2025-09-03
[FEATURE] Need to disable auto discover from MCP servers deployed fromkmcp


linsun:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### ❓ Problem Statement / Motivation

I'm deploying my mcp servers using kmcp, I want to place agentgateway in front of my mcp servers and manually register the agentgw endpoint instead of using the auto discovered mcp servers.  

### 💡 Proposed Solution

Allow users to opt out the auto discover as needed when deploying the mcp server.

### 🔄 Alternatives Considered

The alternative is to not use kmcp and deploy my mcp servers manually.

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

linsun:
cc @EItanya @jmhbh when you have some time

cc @kcbabo

PR #856, opened 2025-09-03
(Feat) Disable mcpServer reconciliation via `kagent.dev/discovery-disabled` annotation

jmhbh:
This PR enables disabling reconciliation of `mcpServer` CRDs that explicitly set the  `kagent.dev/discovery-disabled=true` annotation.

Resolves https://github.com/kagent-dev/kagent/issues/853

### DIFF
--- a/go/internal/controller/predicates/discovery_disabled.go
+++ b/go/internal/controller/predicates/discovery_disabled.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package predicates
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// DiscoveryDisabledPredicate filters out resources with the discovery disabled annotation set to true
+type DiscoveryDisabledPredicate struct {
+	predicate.Funcs
+}
+
+func (DiscoveryDisabledPredicate) Create(e event.CreateEvent) bool {
+	return !isDiscoveryDisabled(e.Object)
+}
+
+func (DiscoveryDisabledPredicate) Update(e event.UpdateEvent) bool {
+	return !isDiscoveryDisabled(e.ObjectNew)
+}
+
+func (DiscoveryDisabledPredicate) Delete(e event.DeleteEvent) bool {
+	return !isDiscoveryDisabled(e.Object)
+}
+
+func (DiscoveryDisabledPredicate) Generic(e event.GenericEvent) bool {
+	return !isDiscoveryDisabled(e.Object)
+}
+
+func isDiscoveryDisabled(obj client.Object) bool {
+	if obj == nil {
+		return false
+	}
+
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


go/internal/controller/predicates/discovery_disabled.go
```diff
@@ -0,0 +1,58 @@
...
+	if obj == nil {
+		return false
+	}
+
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		return false
+	}
+
+	discoveryDisabled, exists := annotations["kagent.dev/discovery-disabled"]
```
EItanya: Can you make the annotation value a const?

jmhbh: done

EItanya (dismissed):

onematchfox:
Just a comment here. I'd love to see consistency in the way that secondary resources are discovered. It feels weird to me that `Service`s are **opt-in** via the **label** `kagent.dev/mcp-service=true` whereas `MCPServer` resources are **opt-out** via the `kagent.dev/discovery-disabled` **annotation**. IMO, all secondary resources should be opt-in using the same mechanism.

EItanya:
> Just a comment here. I'd love to see consistency in the way that secondary resources are discovered. It feels weird to me that `Service`s are **opt-in** via the **label** `kagent.dev/mcp-service=true` whereas `MCPServer` resources are **opt-out** via the `kagent.dev/discovery-disabled` **annotation**. IMO, all secondary resources should be opt-in using the same mechanism.

Interesting. I think the concept here is that this is in fact a primary resource, but I do see what you’re saying. Why would you consider this a secondary resource?

onematchfox:
> Why would you consider this a secondary resource?

Hopefully, the contributor's meeting cleared this up. But for the sake of posterity and a wider audience, I'll comment here as well...

I see this as a secondary resource because the CRD is not owned by this project. It belongs to [KMCP](https://github.com/kagent-dev/kmcp/blob/main/api/v1alpha1/mcpserver_types.go) as does [the controller](https://github.com/kagent-dev/kmcp/blob/main/pkg/controller/mcpserver_controller.go) that reconciles it (related to #822).

linsun:
> > Why would you consider this a secondary resource?
> 
> Hopefully, the contributor's meeting cleared this up. But for the sake of posterity and a wider audience, I'll comment here as well...
> 
> I see this as a secondary resource because the CRD is not owned by this project. It belongs to [KMCP](https://github.com/kagent-dev/kmcp/blob/main/api/v1alpha1/mcpserver_types.go) as does [the controller](https://github.com/kagent-dev/kmcp/blob/main/pkg/controller/mcpserver_controller.go) that reconciles it (related to #822).

Great thoughts @onematchfox!  We are also interested in looking at merging kmcp as an optional component in kagent which would make MCPServer as first citizen resource in kagent.

EItanya (approved):

go/internal/controller/predicates/discovery_disabled_test.go
```diff
@@ -0,0 +1,69 @@
...
+		labels   map[string]string
+		expected bool
+	}{
+		{
+			name:     "no label - should process",
+			labels:   nil,
+			expected: true,
+		},
+		{
+			name:     "discovery label set to disabled - should not process",
```
onematchfox, line 41: What happens in the case of a resource that has already been discovered? Will this prevent it from being "undiscovered"?

EItanya: That is a great question. There's really no surefire way to catch this without checking every single time which feels like overkill. What do you think about a doc explaining how to remove the old entries for now?

onematchfox: > There's really no surefire way to catch this without checking every single time which feels like overkill.

Actually, I think there is a better way of handling this. The update handler on the predicate has access to both the old (`e.ObjectOld`) and new (`e.ObjectNew`) versions of the object. So, rather than only checking the discovery configuration on the new object ([`!isDiscoveryDisabled(e.ObjectNew)`](https://github.com/jmhbh/kagent/blob/9a54b2e81b523906cd75487805c2448c64521ec1/go/internal/controller/predicates/discovery_disabled.go#L37)) the controller should check **both** the old and the new versions and process any object where either of them do not have discovery disabled. That way, the controller is able to handle objects where discovery is disabled on an existing object and ensure that the tool server is removed.

PR #878, opened 2025-09-10
fix(helm): regenerate CRDs

onematchfox:
Diff results from updates to `k8s.io` modules in #864 - although I'm a little puzzled as to what that upgrade had to do with "add[ing] native openai impl to adk" :detective:


EItanya:
> although I'm a little puzzled as to what that upgrade had to do with "add[ing] native openai impl to adk" 🕵️

lol, called out 😆

EItanya (approved):

PR #884, opened 2025-09-10
feat: allow per-agent header configuration for tools

onematchfox:
I want to be able to configure agent-specific headers on calls to any tools. The idea behind this is to facilitate agents being able to provide their own configuration (aka credentials :credit_card:) to the tools they use. One such example is a shared Slack MCP, which would currently be configured with a shared bot token. It would be beneficial for agents to be configured with and supply their own bot tokens, allowing the resulting Slack interaction to use agent-specific credentials rather than the default shared credentials.

This PR attempts to enable that kind of interaction by adding `HeadersFrom` to the tools section on the `Agent` configuration and then using that configuration when building the agent card, whereby any headers supplied via the agent configuration will take precedence over any configured directly on the `RemoteMCPServer`. 

Happy to hear any thoughts/suggestions on the approach - particularly if others are facing similar issues or have ideas around how to solve this in other ways. :smile: 

Notes: 

- I'm not directly trying to use this pattern on agent-to-agent communication at present, and I did at first add `HeadersFrom` to `McpServerTool`. However, I then realised that this requirement may also arise from agent-to-agent communication, so I decided to make the configuration common to cater to both use cases. I am more than happy for someone to call YAGNI on me regarding the agent-to-agent comms. But FWIW, the implementation does work and would facilitate, for example, being able to call remote agents that have some form of static bearer token requirement (or something similar). 
- I know that I've left the UI side of things out of this PR for now. Given that A2A config hasn't yet made it into the UI, I'm hoping this isn't a blocker for this PR :sweat_smile: I can look into it if needed - but also figured I get some feedback on the general approach before doing a bunch of unnecessary work.
- I have been looking into the [MCP Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization) spec of late, but
  - This is optional 
  - Not all upstream services/APIs support OAuth - some like API keys
  - The MCP auth spec doesn't seem to support headless flows like Client Credentials - then again, as long as a valid token is presented, auth will pass :shrug:  (so agents "MCP clients" could just obtain a valid token however they want)

### DIFF
--- a/go/internal/controller/translator/testdata/inputs/agent_with_http_toolserver.yaml
+++ b/go/internal/controller/translator/testdata/inputs/agent_with_http_toolserver.yaml
@@ -44,6 +44,12 @@ objects:
               kind: RemoteMCPServer
               toolNames:
                 - k8s_get_resources
+            headersFrom:
+              - name: MATH
+                valueFrom:
+                  type: Secret
+                  name: math-secret
+                  key: math
   - apiVersion: kagent.dev/v1alpha2
     kind: RemoteMCPServer
     metadata:
--- a/go/internal/controller/translator/testdata/inputs/agent_with_nested_agent.yaml
+++ b/go/internal/controller/translator/testdata/inputs/agent_with_nested_agent.yaml
@@ -9,6 +9,13 @@ objects:
       namespace: test
     data:
       api-key: c2stdGVzdC1hcGkta2V5  # base64 encoded "sk-test-api-key"
+  - apiVersion: v1
+    kind: Secret
+    metadata:
+      name: agent-secret
+      namespace: test
+    data:
+      foo: c3VwM3JzM2NyM3Q=  # base64 encoded "sup3rs3cr3t"
   - apiVersion: kagent.dev/v1alpha2
     kind: ModelConfig
     metadata:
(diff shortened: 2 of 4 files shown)


go/internal/controller/translator/adk_api_translator.go
```diff
@@ -463,12 +463,14 @@ func (a *adkApiTranslator) translateInlineAgent(ctx context.Context, agent *v1al
 		}))
 	}
 
-	toolsByServer := make(map[v1alpha2.TypedLocalReference][]string)
 	for _, tool := range agent.Spec.Declarative.Tools {
 		// Skip tools that are not applicable to the model provider
 		switch {
 		case tool.McpServer != nil:
-			toolsByServer[tool.McpServer.TypedLocalReference] = append(toolsByServer[tool.McpServer.TypedLocalReference], tool.McpServer.ToolNames...)
+			err := a.translateMCPServerTarget(ctx, cfg, agent.Namespace, tool.McpServer, tool.HeadersFrom)
```
onematchfox, line 486: Dropping the `toolsByServer` map does mean that some tool configuration may end up being duplicated in the agent config if folks (and the UI) don't end up grouping tools per server nicely. Things do still seem to work though, and I don't see any way around this. Thoughts?

EItanya: I think you're right, the purpose originally was to try and avoid duplicated resolution of headers, but now that it can come from multiple sources that's no longer possible.

EItanya (approved):
Changes make sense to me, just need to fix nits and lints

EItanya (approved):

Issue #625, opened 2025-07-21
[FEATURE] Add systemMessageFrom field in Agent CRD


vicentinileonardo:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add a systemMessageFrom field to Agent CRD in order to get prompts from ConfigMaps and Secrets

### ❓ Problem Statement / Motivation

Pros:
- **Decoupling & Lifecycle Management**: prompts and agent configurations could have different lifecycles. 
- **Security & RBAC**: RBAC rules can be established. For example, the platform team can manage Agent CRs, while a dedicated "AI/Prompt Engineering" team only has permissions to create/update ConfigMaps in specific namespaces. This enforces separation of duties.
- **Reusability & Consistency**: a single, well-crafted prompt stored in a ConfigMap / Secret could be reused by multiple agents across the same or different namespaces. This ensures consistency and reduces potential prompt duplication and drift. 
- **Auditability**: it could be easier to track changes to prompts when they are in their own version-controlled ConfigMap files. The commit history for prompts would be separate from the history of the agent's structural definition.

Cons:
- **Increased Indirection**: debugging would require an extra step as an engineer must check the Agent manifest and then ConfigMap / Secret. This probably introduces a small amount of cognitive overhead.

By keeping the `systemMessage` field for simpler use cases, a low barrier to entry is maintained. 
By adding `systemMessageFrom` field, flexibility required for complex, multi-team environments found in enterprises is provided

### 💡 Proposed Solution

A PR is almost ready (currently in the testing phase), primarily inspired by: https://github.com/kagent-dev/kagent/pull/416.

The solution involves:

- CRD Schema Update
The `agent_types.go` file has been updated to add the `SystemMessageFrom` field. A CEL validation rule has been introduced to enforce mutual exclusivity between `SystemMessage` and `SystemMessageFrom`.

- Type Refactoring
The `ValueSource` and `ValueSourceType` types have been moved from `toolserver_types.go` to a new file `common_types.go` to allow shared usage between the toolserver and agent components.

- System Message Resolution Logic
[...]

PR #894, opened 2025-09-13
feat: Set system message from ConfigMap or Secrets

supreme-gg-gg:
Implements #625 

This PR allows user to set system message for declarative agents by providing a `systemMessageFrom` field that specify either a `ConfigMap` or `Secret` where the system message is stored, as an alternative to specifying a string as `systemMessage` originally.

A minor refactoring was done on the types and the controller's RBAC config is extended to read from config maps and secrets.

The usage is shown as below. I will update the documentation on `kagent/website` once this is approved and merged.

1. Create configmap or secret:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-agent-config
  namespace: kagent
data:
  system-message: "You will speak in the style of Shakespeare"
```

2. Create agent

```yaml
kind: Agent
metadata:
  name: agent-from-configmap
  namespace: kagent
spec:
  description: This agent speaks in Shakespearean dialect
  type: Declarative

  declarative:
    systemMessageFrom:
      type: ConfigMap
      name: my-agent-config
      key: system-message
```

### DIFF
--- a/go/internal/controller/translator/testdata/outputs/agent_with_system_message_from_configmap.json
+++ b/go/internal/controller/translator/testdata/outputs/agent_with_system_message_from_configmap.json
@@ -23,6 +23,7 @@
     "instruction": "Speak in the style of Shakespeare.",
     "model": {
       "base_url": "",
+      "headers": null,
       "model": "gpt-4o",
       "type": "openai"
     },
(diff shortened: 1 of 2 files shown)


supreme-gg-gg:
> These changes look great to me. Can you please add a unit test as defined in `translator_golden_tests`

I've added two test yaml files for config map and secrets and the output json as well, pls have a look

EItanya:
Hey @supreme-gg-gg, tests seem to be failing in CI

supreme-gg-gg:
> Hey @supreme-gg-gg, tests seem to be failing in CI

this seems to come from a missing headers field because #823 changed the translator, I have updated the expected outputs for my test cases

also sorry for requesting review from @peterj it was a mistake 🙏

supreme-gg-gg:
@EItanya should I update the docs to include example for this?

EItanya:
That would be awesome

Issue #883, opened 2025-09-10
[BUG] Security Context and Tolerations are not propagated properly into agent deployments


AyoubBarich:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

Summary In the current version of kagent, agent pods created by the controller no longer inherit global securityContext and tolerations defined in Helm values. Prior to v0.5.0, these settings were correctly applied to all deployed pods (controller and agents). This regression causes agent pods to be rejected by admission policies (e.g., Kyverno) or remain unschedulable on tainted nodes, breaking deployments in constrained environments. 

### 🔄 Steps To Reproduce

Deploying kagent in a cluster with strict security policies .

### 🤔 Expected Behavior

The global securityContext and tolerations—or at minimum the controller-level ones (since the controller and agents share the same base image)—should be correctly propagated to all agent pods.

### 📱 Actual Behavior

The controller applies these settings to itself but does not merge/propagate them into the agent manifests during reconciliation. Agents are created without the required securityContext/tolerations.

### 💻 Environment

 -  Kubernetes: 
  -  Cluster: Rancher-managed K8s cluster with node taints
  -  Policy: Kyverno enforcing securityContext
  -  Install method: Helm
  -  Kagent version: v0.6.7

### 🔍 Additional Context

I have made a post in [Discord](https://discord.com/channels/1346225185166065826/1415342096520577164) about this issue that goes in more details on how i tried to solve this issue .

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR to fix this issue

PR #895, opened 2025-09-15
Fix(controller) : securitycontext and toleration propagation

AyoubBarich:
This PR adds support for specifying Kubernetes tolerations, podSecurityContext, and containerSecurityContext in agent deployment specs. These fields can now be set via command-line flags in the controller, and are correctly propagated to the generated Deployment manifests.

### Changes :
- Parse tolerations and security context from flags.
- Pass these fields through the deployment resolution logic.
- Set tolerations, podSecurityContext , and containerSecurityContext on the PodSpec and container spec in generated manifests.
- Update golden tests and add new test cases to verify correct propagation.
### Motivation:
This improves the configurability and security of agent pods, allowing users to control scheduling and isolation according to their cluster policies.

✅ Code changes are covered by tests.
✅All tests pass locally.
Closes: #883


go/pkg/app/app.go
EItanya, line 1: I think we should leave this part out for now, stuffing JSON into arguments seems like a bad idea, and prone to error. Rather I think in a follow-up we should think through a more robust way to configure "defaults" for the agents. What do you think?

AyoubBarich: I completely agree. The only approach I can think of is mounting a ConfigMap in the controller and using it when resolving the deployment. The implementation details are still a bit fuzzy, but I would love to hear if you or the KAgent team have any other suggestions or alternative methods.

EItanya: Lemme think about it a bit and get back to you, not 100% sure yet, but yes I think a CM will probably be the best option

EItanya: Ok, what do you think about having a configmap in the deployment NS called `deployment-overrides` or something like that. And it will have the overrides which will be used?

PR #899, opened 2025-09-16
Eitanya/mock llm tests

EItanya:


go/test/e2e/invoke_api_test.go
```diff
@@ -23,11 +37,122 @@ func a2aUrl(namespace, name string) string {
...
+		},
+	}
+}
+
+func buildK8sURL(baseURL string) string {
+	// Get the port from the listener address
+	splitted := strings.Split(baseURL, ":")
+	port := splitted[len(splitted)-1]
+
+	localHost := "172.17.0.1"
```
yuval-k: do you want to do if linux/mac here? i think this only works on linux

go/test/e2e/mocks/invoke_inline_agent.json
```diff
@@ -0,0 +1,65 @@
...
+              ]
+            },
+            "finish_reason": "tool_calls"
+          }
+        ]
+      }
+    },
+    {
+      "name": "k8s_get_resources_response",
+      "match": {
```
yuval-k, line 40: is this a type of assertion? or scenario matching logic? maybe explain in readme?

go/test/mockllm/DESIGN.md
```diff
@@ -0,0 +1,110 @@
...
+  - Headers: `anthropic-version` required
+  - Streaming: SSE events with `event: ...` and JSON payload lines per SDK docs
+  - Tool use: `content` blocks with `type: "tool_use"` and `tool_result` handling
+
+References: OpenAI Go SDK `openai-go` and Anthropic Go SDK `anthropic-sdk-go` for shapes and stream semantics.
+
+### Configuration
+Two ways to configure scenarios:
+1) In-code builder (for brevity in tests):
+   - Fluent API: `NewScenario("name").Expect(OpenAI()).WithModel("gpt-4o").WithMessages(...).Then().RespondWithText("hello").AsStream(...).Build()`
```
yuval-k: ??

go/test/e2e/invoke_api_test.go
```diff
@@ -23,11 +37,122 @@ func a2aUrl(namespace, name string) string {
...
+	err = cli.Create(t.Context(), modelCfg)
+	require.NoError(t, err)
+	agent := generateAgent()
+	err = cli.Create(t.Context(), agent)
+	require.NoError(t, err)
+
+	defer func() {
+		cli.Delete(t.Context(), modelCfg) //nolint:errcheck
+		cli.Delete(t.Context(), agent)    //nolint:errcheck
+	}()
```
inFocus7, line 142: nit: having the `defer` down here would mean if the agent failed to be created, we'll fail and end the test before deleting the model config.

super nit: I believe go tests are safer using `t.Cleanup()` for post-test cleanups, at least true for [parallel testing](https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests). We're not doing parallel tests here so nbd, but may be worth using it to err on the safer side of things.

go/test/e2e/invoke_api_test.go
inFocus7: ^ +1. we probably should
i tried `127.0.0.1` on mac and hit an issue where i got a timeout in the test. `host.docker.internal` work, since this url used so the agent within the cluster communicates with the locally-running mock llm.

PR #901, opened 2025-09-16
fix: get userid from parameters when agent calls us

yuval-k:
get userid from query param if agent creates a session.

note - in the future we may want to create the session for the agent internally instead of having the agent call create_session.


go/internal/httpserver/handlers/sessions.go
```diff
@@ -398,9 +398,32 @@ func getUserID(r *http.Request) (string, error) {
 	userID := r.URL.Query().Get("user_id")
 	if userID == "" {
 		log.Info("Missing user_id parameter in request")
+	}
+
```
EItanya, line 402: Can we just remove this function in favor of the official one?

EItanya (approved):

PR #908, opened 2025-09-18
add the simple tui for chat (e.g. kagent chat)

peterj:
Screenshot 2025-09-17 at 5 44 40 PM

Screenshot 2025-09-17 at 5 45 16 PM

Screenshot 2025-09-17 at 5 45 39 PM

Screenshot 2025-09-17 at 5 46 05 PM

### DIFF
--- a/go/cli/internal/tui/chat.go
+++ b/go/cli/internal/tui/chat.go
@@ -13,7 +13,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/kagent-dev/kagent/go/cli/internal/tui/theme"
-	jsonutil "github.com/kagent-dev/kagent/go/cli/internal/tui/util"
+	"github.com/kagent-dev/kagent/go/internal/a2a"
 	"trpc.group/trpc-go/trpc-a2a-go/protocol"
 )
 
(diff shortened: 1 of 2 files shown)


EItanya:
What do you think about completely killing the current TUI as a result of adding this?

peterj:
so kagent should just launch into this new UI? (note that we'll lose the ability to run other commands within the tui; only chat will be in tui, everything else would stay as is).

EItanya:
> so kagent should just launch into this new UI? (note that we'll lose the ability to run other commands within the tui; only chat will be in tui, everything else would stay as is).

So yes and no. To me it feels like running just `kagent` should launch into this, but if you specify a subcommand then we'll use the current non interactive impl. What do you think. That way it's kind of like `Claude` or other agentic CLIs in that the default experience is the TUI.

peterj:
yeah, that’s what i thought - running kagent drops you into tui. but
running “kagent install” or “kagent —help” invokes other cli commands

go/cli/internal/tui/chat.go
```diff
@@ -0,0 +1,442 @@
...
+		}
+		return a2aEventMsg{Event: ev}
+	}
+}
+
+func (m *chatModel) appendUser(text string) {
+	m.appendLine(theme.UserStyle().Render("You:") + " " + text)
+}
+
+func (m *chatModel) appendEvent(ev protocol.StreamingMessageEvent) {
```
EItanya, line 246: This function uses json utilities to parse the data, but most of this data is available properly via the go types, so please update to use those.

go/cli/internal/tui/chat.go
```diff
@@ -0,0 +1,442 @@
...
+	return b
+}
+
+// JSON helpers now provided by util package
+
+type tickMsg time.Time
+
+// historyEvent represents the stored session event payload.
+// Only fields we need are modeled; unknown fields are ignored by json.Unmarshal.
+type historyEvent struct {
```
EItanya: Also, is there a struct for this already in the `A2A` library that we can use?

EItanya (commented):
Small nits about the go types, but overall looking awesome

PR #909, opened 2025-09-18
Fix : Update azure api version variable

lannoc:
This PR updates then azure api version variable to the correct variable name


EItanya (changes requested):
Rather than this, can we change our env var to match the underlying value:
https://github.com/openai/openai-python/blob/0d85ca08c83a408abf3f03b46189e6bf39f68ac6/src/openai/lib/azure.py#L174-L180

EItanya (approved):

PR #920, opened 2025-09-20
[Feature] Support CrewAI for BYO agents

supreme-gg-gg:
This PR extends BYO agents support for agents created with the CrewAI framework. The design follows `kagent-langgraph`.

## Changes

- [x] Create new `kagent-crewai` package for creating A2A servers using CrewAI
- [x] Provide `samples/crewai` for Crew and Flow mode to show developers how to integrate (<10 lines of change from CrewAi code to setup the integration)

## Next steps

- Implement memory and persistence for CrewAI agents
- Support OpenTelemetry CrewAI Instrumentation

## Testing

To get started, follow instructions in `samples/crewai/research-crew`. Alternatively, follow `samples/crewai/poem_flow` for a Flow mode example. Both examples are taken from the CrewAI QuickStart tutorial, showcasing how to create agent, tasks, provider, crew, flow, etc. in CrewAI and integrate it seamlessly by just modifying the code entry point.

How it looks like from the Kagent UI:

Screenshot 2025-09-20 at 6 24 03 pm


python/samples/crewai/poem_flow/src/poem_flow/main.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
+import json
+import logging
+import os
+from random import randint
+
+import uvicorn
+from crewai.flow import Flow, listen, start
+from kagent.core import configure_tracing
```
krisztianfekete: What do you think setting this up in as part of the integration, so end users can have it automatically without specifying in each of their crewai agents?

Would you be also interested (either here, or as a follow-up) seeing how leveraging [this SDK](https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-crewai) would improve instrumentation?

supreme-gg-gg: I think it's better if it user manually enable tracing, I can put this as an argument when creating the app so the user doesn't need this import: `app = KAgentApp(..., tracing=True)`

For the `opentelemetry` CrewAI SDK would you mind elaborating on what features are you looking for? I'm down to implement this as well, but I saw that there are already tracing in kagent core

python/packages/kagent-crewai/src/kagent/crewai/__init__.py
```diff
@@ -0,0 +1,5 @@
+from ._a2a import KAgentApp
+from ._executor import CrewAIAgentExecutor
+
+__all__ = ["KAgentApp", "CrewAIAgentExecutor", "CrewConfig", "AgentConfig", "TaskConfig"]
+__version__ = "0.1.0"
```
EItanya: No need for this, it will be programmatically added by CI

python/packages/kagent-crewai/src/kagent/crewai/_executor.py
```diff
@@ -0,0 +1,390 @@
...
+                    "app_name": self.app_name,
+                    "session_id": getattr(context, "session_id", context.context_id),
+                },
+            )
+        )
+
+        loop = asyncio.get_running_loop()
+
+        def _enqueue_event(event: Any):
+            asyncio.run_coroutine_threadsafe(event_queue.enqueue_event(event), loop)
```
EItanya: Why do we need this helper func, why can't we just enqueue the envoy from the scoped handlers?

python/samples/crewai/poem_flow/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+WORKDIR /app
+
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install uv for fast Python package management
+RUN pip install uv
```
EItanya: We can use uv image to avoid installing uv separately

python/packages/kagent-crewai/src/kagent/crewai/_executor.py
```diff
@@ -0,0 +1,390 @@
...
+        self._crew = crew
+        self.app_name = app_name
+        self._config = config or CrewAIAgentExecutorConfig()
+
+    @override
+    async def cancel(self, context: RequestContext, event_queue: EventQueue):
+        raise NotImplementedError("Cancellation is not implemented")
+
+    @override
+    async def execute(
```
EItanya, line 52: This is a very very long function, Is there anyway we can clean this up so it's not so long? Maybe split up the handlers into a different class or something like that? I'm not a python expert but I bet Claude can help

EItanya (changes requested):
This is looking great so far! Mostly my comments are about the executor and some `Docker` config.

python/samples/crewai/poem_flow/src/poem_flow/main.py
EItanya: Specifically this would be enabling tracing for `crewai`. We bake in tracing for various LLM providers, `http` calls, and ADK, but we don't want to ALWAYS load tracing libraries for agent libraries we aren't actively using. I think I agree with @supreme-gg-gg about `app = KAgentApp(..., tracing=True)`, but maybe it should be `True` by default.

python/packages/kagent-crewai/src/kagent/crewai/_executor.py
supreme-gg-gg: The handlers must be synchronous because they're invoked by the crewAI event bus. However the `enqueue_event` function from A2A is async, so we need this helper to execute the enqueue instead of doing it directly in the handler

python/samples/crewai/poem_flow/Dockerfile
supreme-gg-gg: Should I update the langgraph example to use uv image as well?

python/packages/kagent-crewai/src/kagent/crewai/_executor.py
supreme-gg-gg: I've refactored them to an events listener in `_listener.py`

python/samples/crewai/poem_flow/src/poem_flow/main.py
supreme-gg-gg: > Specifically this would be enabling tracing for `crewai`. We bake in tracing for various LLM providers, `http` calls, and ADK, but we don't want to ALWAYS load tracing libraries for agent libraries we aren't actively using. I think I agree with @supreme-gg-gg about `app = KAgentApp(..., tracing=True)`, but maybe it should be `True` by default.

Would CrewAI tracing with this SDK be on top of or replace the existing provider-specific tracing (e.g. OpenAI, Anthropic) in kagent core? it would be placed in the CrewAI package, is that correct

python/packages/kagent-crewai/src/kagent/crewai/_executor.py
EItanya: What will the impact be of running it this way?

supreme-gg-gg: I don't think there are any major risks or downsides for this workaround, the only minor problem I speculate might be that if the enqueue fails we can't log and report it since it's somewhat like "fire and forget"

supreme-gg-gg:
@EItanya could you review my changes to the Dockerfiles for python app and the samples, thanks!

python/samples/crewai/poem_flow/src/poem_flow/main.py
krisztianfekete: Not sure if it would replace those, as what we have in kagent-core utils covers LLM-specific and auxiliary tracing. I'd expect the CrewAI SDK to add additional, CrewAI specific spans that can be used on top of what we have as core tracing.

Have you had a chance to take a look at what spans are you getting with the current changes? If we have LLM calls then I'd say we are fine w/r/t tracing with this PR, and can do add the CrewAI SDK as a potential follow-up, if it's useful.

python/samples/crewai/poem_flow/pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+dependencies = [
+    "kagent-crewai",
+    "crewai[tools]>=0.193.2,<1.0.0"
+]
+
+[project.scripts]
+kickoff = "poem_flow.main:kickoff"
+run_crew = "poem_flow.main:kickoff"
+plot = "poem_flow.main:plot"
+poem_flow = "poem_flow.main:main"
```
EItanya, line 16: Do we need these to be scripts? Scripts end up being global in a `uv` workspace. We can fix in a follow-up

EItanya (approved):

python/samples/crewai/poem_flow/pyproject.toml
supreme-gg-gg: Will remove these in the follow up, they're generated by crewAI

python/samples/crewai/poem_flow/src/poem_flow/main.py
supreme-gg-gg: @krisztianfekete Sorry for the late response, I did a few experiments with CrewAI tracing and validated that the CrewAI spans are indeed sent to the collector. CrewAI has its internal instrumentor, but I found that by adding the CrewAI Instrumentation SDK you suggested the spans will be more comprehensive.

Screenshot 2025-10-01 at 5 27 08 PM

Regarding providers, the CrewAI SDK traces all calls to LiteLLM, but provider-specific SDK is required for spans containing the actual LLM call content e.g. prompts, token count, so I will use an approach similar to Langgraph tracing where both the provider specific + framework specific SDK are used.

I will be adding this to the new follow up PR #951

krisztianfekete: Thanks for following up!

In the core library we have Gemini/Anthropic/OpenAI SDKs already, so by adding the CrewAI one, we should have support for the most popular options I think.

PR #927, opened 2025-09-23
feat: Implement BYO LangGraph tracing

krisztianfekete:
This PR

- Wires `kagent.core` tracing to BYO LangGraph extension to have A2A, FastAPI, and LLM model-specific tracing automatically
- Enriches instrumentation for LangGraph Agents via langsmith-sdk (see screenshot)
- Shows an example implementation by updating the currency agent sample agent

image

As instrumentation libraries evolve rapidly, expect breaking changes to span names, attributes.

### DIFF
--- a/python/samples/langgraph/currency/pyproject.toml
+++ b/python/samples/langgraph/currency/pyproject.toml
@@ -11,7 +11,6 @@ dependencies = [
     "langchain-google-genai>=1.0.0",
     "langgraph-checkpoint-sqlite>=2.0.0",
     "langsmith[otel]>=0.4.30",
-    "opentelemetry-instrumentation-google-generativeai>=0.47.3",
 ]
 
 [build-system]
--- a/python/packages/kagent-core/pyproject.toml
+++ b/python/packages/kagent-core/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
   "opentelemetry-instrumentation-anthropic>=0.44.0",
   "opentelemetry-instrumentation-httpx >= 0.52.0",
   "opentelemetry-instrumentation-fastapi>=0.52.0",
+  "opentelemetry-instrumentation-google-generativeai>=0.47.3",
 ]
 
 [tool.ruff]
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.13"
 
 [manifest]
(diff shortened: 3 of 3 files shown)


python/samples/langgraph/currency/pyproject.toml
```diff
@@ -10,6 +10,8 @@ dependencies = [
     "langchain-core>=0.3.0",
     "langchain-google-genai>=1.0.0",
     "langgraph-checkpoint-sqlite>=2.0.0",
+    "langsmith[otel]>=0.4.30",
+    "opentelemetry-instrumentation-google-generativeai>=0.47.3",
```
EItanya: Why was this added, I thought it doesn't work?

python/samples/langgraph/currency/agent.yaml
```diff
@@ -14,3 +14,18 @@ spec:
...
+        - name: LANGSMITH_TRACING
+          value: "true"
+        - name: LANGSMITH_OTEL_ENABLED
+          value: "true"
+        - name: LANGSMITH_ENDPOINT
+          value: "http://jaeger-collector.jaeger.svc.cluster.local:4317"
+        - name: LANGSMITH_API_KEY
+          value: "langsmith-api-key"
+        - name: LANGSMITH_WORKSPACE_ID
+          value: "kagent-langgraph"
```
EItanya, line 31: Should we set this from our `SDK` if tracing is enabled?

python/samples/langgraph/currency/pyproject.toml
krisztianfekete: I asked the Traceloop folks a couple of weeks ago to migrate it and they were kind enough to do so. It's useful for our agents too.

python/samples/langgraph/currency/agent.yaml
krisztianfekete: We can, but might be better UX for existing langchain-sdk users to use their documented method described in their guides? Basically, making these explicit which should help with troubleshooting if/when they want to customize.

python/packages/kagent-langgraph/src/kagent/langgraph/_a2a.py
```diff
@@ -15,6 +15,7 @@
 from fastapi.responses import PlainTextResponse
 
 from kagent.core import KAgentConfig
+from kagent.core import configure_tracing as _configure_tracing
```
peterj: nit: any reason you're importing as _configure_tracing?

krisztianfekete: No, I was just experimenting with different approaches, it's a leftover.

python/samples/langgraph/currency/agent.yaml
EItanya: What if we set these by default, but if the user already set them then we can use those values.

PR #939, opened 2025-09-23
feat: Merge KMCP crd and controller into Kagent 

jmhbh:
This PR contains the following changes and is the first in two PRs to migrate KMCP into Kagent. The second PR will focus on migrating the KMCP cli functionalities into the Kagent CLI.

- Added KMCP's `mcpServer` CRD to Kagent.
- Added a new controller that watches for the `mcpServer` CRD and handles the deployment lifecycle of the mcp server deployment.
- Refactored existing mcp server controller into `mcp_server_tool_controller` to better reflect its responsibilities of managing tool discovery.
- Added new e2e test for a declarative agent using tools derived from a mcp server deployed using the `mcpServer` CRD.

Manual Testing:
- Interacted with agent using tools derived from the deployed mcp server.
- Tested deleting mcp tool server in the UI. (This is the only portion of the UI that was using kmcp resources)


go/internal/controller/mcp_server_tool_controller.go
```diff
@@ -0,0 +1,65 @@
...
+
+// MCPServerToolController handles reconciliation of a MCPServer object for tool discovery purposes
+type MCPServerToolController struct {
+	Scheme     *runtime.Scheme
+	Reconciler reconciler.KagentReconciler
+}
+
+// +kubebuilder:rbac:groups=kagent.dev,resources=mcpservers,verbs=get;list;watch
+
+func (r *MCPServerToolController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
jmhbh, line 44: This was the original mcp server controller that was implemented as a part of the kmcp integration. I renamed it to `mcp_server_tool_controller` to better represent its responsibility of updating the tool server.

helm/kagent-crds/templates/kagent.dev_agents.yaml
```diff
@@ -6899,7 +6899,8 @@ spec:
                       If not specified, the default value is true.
                     type: boolean
                   systemMessage:
-                    minLength: 1
+                    description: SystemMessage is a string specifying the system message
+                      for the agent
                     type: string
```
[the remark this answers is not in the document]
jmhbh: made optional in https://github.com/kagent-dev/kagent/commit/bbca7dea0d5b5a8265f01c7870ec16c85aceab3b

go/api/v1alpha1/mcpserver_types.go
```diff
@@ -0,0 +1,263 @@
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ObservedGeneration is the most recent generation observed for this MCPServer.
+	// It corresponds to the MCPServer's generation, which is updated on mutation by the API Server.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+}
+
+// MCPServerDeployment
+type MCPServerDeployment struct {
```
EItanya, line 194: Not for this PR, but we should consolidate the deployment types and code for agents + MCP Servers

go/internal/controller/mcp_server_tool_controller.go
EItanya: Why do we need 2 controllers? Now we can re-use since all of the logic is internal

go/internal/controller/mcp_server_controller.go
EItanya, line 1: Same comment about combining controllers

go/internal/controller/translator/mcp/translator_adapter.go
EItanya, line 1: I think this package, `translator` is getting a bit crowded. Can we instead move these into sub-packages: `agent`, `mcp` respectively?

go/internal/controller/translator/labels/constants.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Standard Kubernetes label keys as defined in the Kubernetes documentation
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+const (
+	AppName      = "app.kubernetes.io/name"
+	AppInstance  = "app.kubernetes.io/instance"
+	AppVersion   = "app.kubernetes.io/version"
+	AppComponent = "app.kubernetes.io/component"
+	AppPartOf    = "app.kubernetes.io/part-of"
+	AppManagedBy = "app.kubernetes.io/managed-by"
```
EItanya, line 12: We should proibably use these for agents as well, or at least decide on a shared set of labels. This can also be a TODO

go/test/e2e/invoke_api_test.go
```diff
@@ -298,3 +298,61 @@ func TestInvokeExternalAgent(t *testing.T) {
 		require.Contains(t, text, "kebab for user@example.com", string(jsn))
 	})
 }
+
+func TestInvokeDeclarativeAgentWithMcpServerTool(t *testing.T) {
```
EItanya: Rather than spinning up these agents as a part of the CI job, can we spin up the agent + mcp server at the beginning of the test? That's also how the above: `TestInvokeInlineAgent` works

Makefile
```diff
@@ -138,6 +134,12 @@ push-test-agent: build-kagent-adk
 	$(DOCKER_BUILDER) build --push --build-arg DOCKER_REGISTRY=$(DOCKER_REGISTRY) --build-arg VERSION=$(VERSION) -t $(DOCKER_REGISTRY)/kebab:latest -f go/test/e2e/agents/kebab/Dockerfile ./go/test/e2e/agents/kebab
 	kubectl apply --namespace kagent --context kind-$(KIND_CLUSTER_NAME) -f go/test/e2e/agents/kebab/agent.yaml
 
+# Applies manifests for a declarative agent with a MCP server tool used for testing purposes
+.PHONY: apply-add-numbers-agent
+apply-add-numbers-agent:
+	kubectl apply --namespace kagent --context kind-$(KIND_CLUSTER_NAME) -f go/test/e2e/manifests/everything-mcp-server.yaml
+	kubectl apply --namespace kagent --context kind-$(KIND_CLUSTER_NAME) -f go/test/e2e/manifests/add-numbers-agent.yaml
```
EItanya: I'd prefer not to bloat the `Makefile`. This is just some applies.

EItanya (changes requested):
Overall looking great. The main themes of my review are to clean up the controllers and translation. For future work I would like us to try and unify our approach for mcp server outputs and agent outputs.

go/internal/controller/mcp_server_controller.go
jmhbh: I combined the controllers but the original intention for the separate controllers was because the tool controller would loop forever. Combining the controllers may cause unnecessary churn in the environment wrt unnecessary translation and kube api calls.

go/test/e2e/invoke_api_test.go
jmhbh: done. Refactored the setup functions to make them more reusable while I was at it.

Makefile
jmhbh: removed.

go/internal/controller/mcp_server_controller.go
EItanya: Ohhh interesting, I forgot about that aspect, ok you're right we should revert to having multiple. I hate discovery :laughing:

.github/workflows/ci.yaml
```diff
@@ -119,7 +119,7 @@ jobs:
       - name: Run cmd/main.go tests
         working-directory: go
         run: |
-          go test -skip 'TestInvokeInlineAgent|TestInvokeExternalAgent|TestE2E|TestAutogenClient' -v ./...
+          go test -skip 'TestInvokeInlineAgent|TestInvokeExternalAgent|TestInvokeDeclarativeAgentWithMcpServerTool|TestE2E|TestAutogenClient' -v ./...
```
EItanya: Can we add `TestE2E` prefix to all tests which should not be run here and then just use that regex rather than this growing list?

EItanya (approved):

PR #940, opened 2025-09-23
fix UI lint warnings

peterj:


dobesv:
I notice this removes the timestamp and duration display from the chat messages.  Intentional?  Any thought about bringing those back?

EItanya:
Nope, let's do it!

dobesv:
OK, I will try

dobesv:
I guess the issue is that there are no timestamps in the message metadata, so maybe the timestamp / duration display here was actually dead code.  🤔

PR #941, opened 2025-09-24
CVE fixes, openshift support

dimetron:
Go 1.25.1
Python 3.13.7
Node 22.19

- CVE fixes 202509
- open shift support in UI container
- pull secrets for agents


go/config/rbac/role.yaml
```diff
@@ -4,56 +4,54 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
-  - apiGroups:
-      - ""
-    resources:
-      - configmaps
-      - secrets
-      - services
```
EItanya: Why were secrets and configmaps removed? We need to bet able to watch/read those. If this is necessary to running in OpenShift we need to come up with a different solution.

EItanya (commented):
Everything looks good except this one comment.

go/config/rbac/role.yaml
peterj: these changes come from https://github.com/kagent-dev/kagent/commit/bbca7dea0d5b5a8265f01c7870ec16c85aceab3b

helm/kagent/templates/controller-deployment.yaml
```diff
@@ -73,6 +73,8 @@ spec:
             - {{ coalesce .Values.controller.agentImage.tag .Values.tag .Chart.Version  }}
             - -image-pull-policy
             - {{ .Values.controller.agentImage.pullPolicy | default .Values.imagePullPolicy }}
+            - -image-pull-secret
```
peterj, line 77: should we just not pass in the image-pull-secret if it's not set? (vs. passing in the value `""`)?

python/pyproject.toml
```diff
@@ -6,6 +6,7 @@ dev = [
   "pytest>=8.3.5",
   "pytest-asyncio>=0.25.3",
   "ruff>=0.11.5",
+  "authlib>=1.6.4"
```
peterj, line 9: why was this added?

scripts/kind/kind-config.yaml
```diff
@@ -37,5 +37,7 @@ nodes:
   - role: control-plane
 # Example of port mappings for the control-plane node
 #    extraPortMappings:
-#      - containerPort: 30950
-#        hostPort: 30950
\ No newline at end of file
+#      - containerPort: 30884
```
peterj, line 40: is this needed?

ui/Dockerfile
```diff
@@ -76,14 +76,12 @@ RUN mkdir -p $BUN_INSTALL  \
     && curl -fsSL https://bun.sh/install | bash -s "bun-v$TOOLS_BUN_VERSION" \
     && bun --version
 
-RUN mkdir -p /app/ui/public /run/nginx/ /var/run/nginx/ /var/lib/nginx/tmp/ /var/lib/nginx/logs/  \
-    && addgroup -g 1001    nginx                        \
+RUN mkdir -p /app/ui/public /tmp/nginx/client_temp /tmp/nginx/proxy_temp /tmp/nginx/fastcgi_temp /tmp/nginx/uwsgi_temp /tmp/nginx/scgi_temp  \
```
peterj, line 79: what is the improvement here?  will this work for both openshift and non-openshift?

ui/conf/supervisord.conf
```diff
@@ -1,6 +1,5 @@
 [supervisord]
 nodaemon=true
-user=nextjs
```
peterj, line 3: ?

peterj:
In general it looks good; the only part that's worrying me a bit is the changes to the nginx and that it won't break for non-openshift scenarios

python/pyproject.toml
dimetron: uv run pip-audit
Found 1 known vulnerability in 1 package
Name    Version ID                  Fix Versions
authlib 1.6.1   GHSA-9ggr-2464-2j32 1.6.4

go/config/rbac/role.yaml
dimetron: Looks like it is overwritten by  make -C go manifests

We need to add 
// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete

scripts/kind/kind-config.yaml
dimetron: I use it for local tools testing as alternative to metal-lb, it's commented by default

ui/Dockerfile
dimetron: - both supervisord and nginx run under nextjs docker user instead of root
- for readonly file system require use /tmp directory

ui/conf/supervisord.conf
dimetron: default user from container is used for both - otherwise we need superuser

EItanya (approved):

helm/kagent/values.yaml
```diff
@@ -126,7 +126,7 @@ ui:
   service:
     type: ClusterIP
     ports:
-      port: 80
+      port: 8080
```
onematchfox, line 129: Hey folks, This is a breaking change for any consumers of this `Service`. I know we're still `v0.x` so breaking changes are to be expected but please could we try to at least highlight these sorts of things in the release notes going forward 🙏 At the very least I would have expected a `0.7` release to warn of this sort of thing.

@dimetron @EItanya

EItanya: Omg you’re so right, I’m so so sorry. I will take full responsibility for this. This totally slipped my mind, I’m always so focused on APIs and backends.

@peterj snd @ashleywang1 this is what most likely broke the dashboard command in the CLI.

PR #950, opened 2025-09-24
Fix/argument setup tool server creation

### LINKED ISSUE #936: Fix argument setup in Tool Server creation
# Request

The **AddServerDialog.tsx** has a `commandPrefix` argument _in-code_ which is not in use. We should add it to the creation wizard.

To add to this, the current command setup is `npx [...args] packageName` ([ref](https://github.com/kagent-dev/kagent/blob/1e1850feb6a4a0bf8888f05a9f0dee05152197ec/ui/src/components/AddServerDialog.tsx#L293-L321)). This differs from the preview command which is `npx packageName [...args]`.

## Changes Needed

1. Add a field for `commandPrefix` in UI.
    - Or remove if we don't want to support `commandPrefix`.
3. Update the final `args` setup to properly follow the preview (`[prefix] packageName [...args]`) 
## Context

The idea for this prefix is to act as `npx [prefix] packageName [...args]`, but it's missing in the UI.

rishabh998186:
Adds missing UI field for command prefix functionality.

**Problem:** Command prefix existed in code but no UI field
**Solution:** Added input field with tooltip between executor and package name

**Changes:**
- Add Command Prefix input field
- Placeholder: "E.g. --yes, --force, etc."  
- Help text and form integration

**Result:** Users can now specify prefix flags like `--yes`, `--verbose`

Fixes #936


peterj:
There's a lot of unrelated changes in the PR, so it's hard to say what has actually changed. Can you revert the formatting changes?

rishabh998186:
thanks @peterj for the feedback . you are right there are lot of unrelated changes in my previous commits so as you say ,  I've reverted all the formatting changes and kept only the essential command prefix functionality. i removed all unrelated changes  and duplicate code .

rishabh998186:
@peterj Could you please take a look at my pr when you get a chance , as you say i removed  the unwanted changes let me know if there still any issue .

ui/src/components/AddServerDialog.tsx
```diff
@@ -52,6 +62,51 @@ export function AddServerDialog({ open, onOpenChange, onAddServer, onError }: Ad
   const [sseReadTimeout, setSseReadTimeout] = useState("300s");
   const [terminateOnClose, setTerminateOnClose] = useState(true);
 
+  // Clean up package name for server name
+  const cleanPackageName = (pkgName: string): string => {
```
peterj, line 66: this is not used anywhere and was removed

PR #951, opened 2025-09-25
[Feature] Implement Persistence and Tracing for CrewAI BYO Agents

supreme-gg-gg:
This PR is a follow up for #920. 

- Implement long term memory for CrewAI Crews and state persistence for CrewAI Flow
- Updates samples to show usage of memory and state persistence
- Support tracing with `opentelemetry-instrumentation-crewai` for crewai specific spans as suggested in code review


go/internal/database/models.go
```diff
@@ -179,6 +179,32 @@ type LangGraphCheckpointWrite struct {
...
+	CreatedAt time.Time      `gorm:"autoCreateTime;index:idx_crewai_memory_list" json:"created_at"`
+	UpdatedAt time.Time      `gorm:"autoUpdateTime" json:"updated_at"`
+	DeletedAt gorm.DeletedAt `gorm:"index" json:"deleted_at"`
+	// MemoryData contains JSON serialized memory data including task_description, score, metadata, datetime
+	MemoryData string `gorm:"type:text;not null" json:"memory_data"`
+}
+
+// CrewAIFlowState represents flow state for CrewAI flows
+type CrewAIFlowState struct {
+	ID         uint           `gorm:"primaryKey;autoIncrement" json:"id"`
```
EItanya: Do we need this if we already have the `ThreadID`

go/internal/database/models.go
```diff
@@ -179,6 +179,32 @@ type LangGraphCheckpointWrite struct {
 	DeletedAt    gorm.DeletedAt `gorm:"index" json:"deleted_at"`
 }
 
+// CrewAIAgentMemory represents long-term memory for CrewAI agents
+type CrewAIAgentMemory struct {
+	UserID    string         `gorm:"primaryKey;not null" json:"user_id"`
+	ThreadID  string         `gorm:"primaryKey;not null" json:"thread_id"`
```
EItanya, line 185: Are these enough for a `primary_key`? Can `thread_id` be re-used?

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,175 @@
...
+        try:
+            response = requests.delete(url, params=params, headers={"X-User-ID": self.user_id})
+            response.raise_for_status()
+            logging.info(f"Successfully reset memory for session {self.session_id}")
+        except requests.exceptions.RequestException as e:
+            logging.error(f"Error resetting memory for session {self.session_id}: {e}")
+            raise
+
+
+class KagentFlowPersistence(FlowPersistence):
```
EItanya: Can we put this in a separate file since it's really different from `memory`

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __init__(self, session_id: str, user_id: str, base_url: str = "http://localhost:8080"):
+        self.session_id = session_id
+        self.user_id = user_id
+        self.base_url = base_url
+        self.init_db()
+
+    def init_db(self) -> None:
+        # Nothing to do here as the backend handles DB initialization
+        pass
```
EItanya: Can we just remove this altogether?

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,175 @@
+from typing import Any, Dict, List, Optional, Union
+import logging
+from datetime import datetime, timezone
+
+import requests
```
EItanya: Why aren't we using `httpx` like elsewhere?

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,175 @@
...
+                # We want to return items in the format that LongTermMemory expects
+                results.append(
+                    {
+                        "metadata": memory_data.get("metadata", {}),
+                        "datetime": memory_data.get("datetime", ""),
+                        "score": memory_data.get("score", 0.0),
+                    }
+                )
+
+            # Sort by datetime DESC, then by score ASC (matching SQLite behavior)
```
EItanya: Why isn't the DB sorting this?

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,175 @@
...
+    def reset(self) -> None:
+        """
+        Resets the memory storage by deleting all memories for this session.
+        """
+        url = f"{self.base_url}/api/crewai/memory"
+        params = {"thread_id": self.session_id}
+
+        logging.info(f"Resetting memory for session {self.session_id}")
+        try:
+            response = requests.delete(url, params=params, headers={"X-User-ID": self.user_id})
```
EItanya: `httpx`?

python/samples/crewai/poem_flow/src/poem_flow/main.py
```diff
@@ -22,25 +22,47 @@ class PoemState(BaseModel):
...
+
+        logging.info(f"Poem state is now:\n{self.state.poem}")
 
     @listen(generate_poem)
     def save_poem(self):
         logging.info("Saving poem")
-        with open("output/poem.txt", "w") as f:
-            f.write(self.state.poem)
+        # with open("output/poem.txt", "w") as f:
+        #     f.write(self.state.poem)
```
EItanya: Why is this commented, can we just remove it?

EItanya (changes requested):
Mostly looking good, just nits so far

go/internal/database/models.go
supreme-gg-gg: this ID is different from thread ID, the former is used to sort for the latest flow state since it autoincrements for every state update, but I think we can also use `Timestamp` for this purpose

go/internal/database/models.go
supreme-gg-gg: yes, flowUUID is not necessary and i've removed it

python/samples/crewai/research-crew/agent.yaml
```diff
@@ -10,12 +10,11 @@ spec:
...
       image: localhost:5001/research-crew:latest
       env:
-        # Note that CrewAI LLM expects GEMINI_API_KEY instead of GOOGLE_API_KEY
-        - name: GEMINI_API_KEY
+        - name: OPENAI_API_KEY
           valueFrom:
             secretKeyRef:
-              name: kagent-google
-              key: GOOGLE_API_KEY
+              name: kagent-openai
```
krisztianfekete, line 16: Have you tested with both LLM providers? Did you notice any differences for spans between the two?

supreme-gg-gg: For Gemini it seems that the instrumentation library is not instrumenting all the spans, but for openAI it works fine.

go/internal/database/models.go
supreme-gg-gg: I've changed this to use creation time and removed the ID field

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
supreme-gg-gg: I realised that `init_db` is an abstract method so it must be implemented even tho it does nothing, just added it back

supreme-gg-gg:
@EItanya for the existing E2E tests would it be a good idea to also use Artifacts instead of history when checking agent response since artifact is the usual way of returning task output? I tried it on the ADK inline / declarative agent and it passes

python/packages/kagent-crewai/src/kagent/crewai/_state.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def load_state(self, flow_uuid: str) -> Optional[Dict[str, Any]]:
+        """Loads the flow state from the Kagent backend."""
+        url = f"{self.base_url}/api/crewai/flows/state"
+        params = {"thread_id": self.session_id, "flow_uuid": flow_uuid}
+        logging.info(f"Loading flow state from Kagent backend with params: {params}")
+
+        try:
+            with httpx.Client() as client:
+                response = client.post(url, params=params, headers={"X-User-ID": self.user_id})
```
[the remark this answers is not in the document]
supreme-gg-gg: will fix these

go/test/e2e/invoke_api_test.go
```diff
@@ -359,10 +398,126 @@ func TestE2EInvokeDeclarativeAgentWithMcpServerTool(t *testing.T) {
...
+	// Setup A2A client
+	a2aURL := a2aUrl(agent.Namespace, agent.Name)
+	a2aClient, err := a2aclient.NewA2AClient(a2aURL)
+	require.NoError(t, err)
+
+	t.Run("two_turn_conversation", func(t *testing.T) {
+		// First turn: Generate initial poem
+		// Use artifacts only (true) for CrewAI flows
+		useArtifacts := true
+		taskResult1 := runSyncTest(t, a2aClient, "Generate a poem about CrewAI", "CrewAI", &useArtifacts)
```
EItanya: Nit: can you add more text for the expected response, this string is also in the request.

go/test/e2e/invoke_api_test.go
```diff
@@ -359,10 +398,126 @@ func TestE2EInvokeDeclarativeAgentWithMcpServerTool(t *testing.T) {
...
+
+		// Second turn: Continue poem (tests persistence)
+		// Use the same ContextID to maintain conversation context
+		runSyncTest(t, a2aClient, "Continue the poem", "AI agents", &useArtifacts, taskResult1.ContextID)
+	})
+
+	t.Run("streaming_invocation", func(t *testing.T) {
+		runStreamingTest(t, a2aClient, "Generate a poem about CrewAI", "CrewAI")
```
EItanya: Same comment as above

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
```diff
@@ -27,16 +27,27 @@ def configure(fastapi_app: FastAPI | None = None):
...
             processor = BatchSpanProcessor(OTLPSpanExporter(endpoint=trace_endpoint))
         else:
             processor = BatchSpanProcessor(OTLPSpanExporter())
-        tracer_provider.add_span_processor(processor)
-        trace.set_tracer_provider(tracer_provider)
+
+        # Check if a TracerProvider already exists (e.g., set by CrewAI)
+        current_provider = trace.get_tracer_provider()
+        if isinstance(current_provider, TracerProvider):
+            # TracerProvider already exists, just add our processor to it
```
EItanya, line 41: I think we also need to add our labels here, e.g. `Resource({"service.name": "kagent"})`

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,110 @@
...
+        self.base_url = base_url
+
+    def save(self, task_description: str, metadata: dict, timestamp: str, score: float) -> None:
+        """
+        Saves a memory item to the Kagent backend.
+        The agent_id is expected to be in the metadata.
+        """
+        url = f"{self.base_url}/api/crewai/memory"
+        payload = KagentMemoryPayload(
+            session_id=self.session_id,
```
EItanya: My IDE is complaining that this should be `thread_id` as per pydantic, not sure if yours is also mad.

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
```diff
@@ -0,0 +1,110 @@
...
+    data: List[KagentMemoryPayload]
+
+
+class KagentMemoryStorage:
+    """
+    KagentMemoryStorage is a custom storage class for CrewAI's LongTermMemory.
+    It persists memory items to the Kagent backend, scoped by session_id and user_id.
+    """
+
+    def __init__(self, session_id: str, user_id: str, base_url: str = "http://localhost:8080"):
```
EItanya: Why is this the default?

python/packages/kagent-crewai/src/kagent/crewai/_state.py
```diff
@@ -0,0 +1,71 @@
...
+    data: KagentFlowStatePayload
+
+
+class KagentFlowPersistence(FlowPersistence):
+    """
+    KagentFlowPersistence is a custom persistence class for CrewAI Flows.
+    It saves and loads the flow state to the Kagent backend.
+    """
+
+    def __init__(self, session_id: str, user_id: str, base_url: str = "http://localhost:8080"):
```
EItanya: Same question wondering why this is the default

python/packages/kagent-crewai/src/kagent/crewai/_state.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def init_db(self) -> None:
+        """This is handled by the Kagent backend, so no action is needed here."""
+        pass
+
+    def save_state(self, flow_uuid: str, method_name: str, state_data: Union[Dict[str, Any], BaseModel]) -> None:
+        """Saves the flow state to the Kagent backend."""
+        url = f"{self.base_url}/api/crewai/flows/state"
+        payload = KagentFlowStatePayload(
+            session_id=self.session_id,
```
EItanya: same question about thread_id

python/packages/kagent-crewai/src/kagent/crewai/_memory.py
supreme-gg-gg: This should not be a problem since `thread_id` is an alias to `session_id` which means session_id is renamed to thread_id in the request, but I've replaced all usage of `session_id` with `thread_id` to avoid confusion and match the langgraph checkpointer usage

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
supreme-gg-gg: CrewAI already creates a tracing provider that uses `crewAI-telemetry` as their service name (https://github.com/crewAIInc/crewAI/blob/8d93361cb305e39638cb6b1c257c572e129ac9a1/src/crewai/telemetry/telemetry.py). I tried disabling CrewAI's built in tracing provider so we create our own Kagent one, but that resulted in no spans being sent. 

A reliable workaround I found is to create a custom span processor wrapper that overwrites the resource service name before exporting:

```python
class ResourceOverrideSpanProcessor(SpanProcessor):
    """A span processor that overrides the resource attributes before exporting."""

    def __init__(self, exporter: SpanExporter, resource: Resource):
        self.batch_processor = BatchSpanProcessor(exporter)
        self.override_resource = resource

    def on_start(self, span: "Span", parent_context: Optional[Context] = None) -> None:
        self.batch_processor.on_start(span, parent_context)

    def on_end(self, span: ReadableSpan) -> None:
        # Override the resource before passing to the batch processor
        # This ensures the service.name is always the one in the resource (i.e. kagent)
        span._resource = self.override_resource
        self.batch_processor.on_end(span)

    def shutdown(self) -> None:
        self.batch_processor.shutdown()

    def force_flush(self, timeout_millis: int = 30000) -> bool:
        return self.batch_processor.force_flush(timeout_millis)
```

And we can add it like this:

```python
exporter = OTLPSpanExporter(endpoint=trace_endpoint)
processor = ResourceOverrideSpanProcessor(exporter, resource)
existing_crewai_trace_provider.add_span_processor(processor)
```

This would work for any framework already creating a OTEL trace provider and all their spans will be renamed to `kagent`. How does this sound?

python/packages/kagent-crewai/src/kagent/crewai/_state.py
supreme-gg-gg: Removed default

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
EItanya: It's interesting because this should probably be configurable, I think let's merge this as is and we can potentially revisit.

PR #955, opened 2025-09-26
feat: token propagation

yuval-k:
allow to optionally propagate authorization token from agent to mcp server


go/cli/cmd/kagent/main.go
```diff
@@ -93,6 +93,7 @@ func main() {
 	invokeCmd.Flags().BoolVarP(&invokeCfg.Stream, "stream", "S", false, "Stream the response")
 	invokeCmd.Flags().StringVarP(&invokeCfg.File, "file", "f", "", "File to read the task from")
 	invokeCmd.Flags().StringVarP(&invokeCfg.URLOverride, "url-override", "u", "", "URL override")
+	invokeCmd.Flags().StringSliceVarP(&invokeCfg.Headers, "header", "H", nil, "Headers to include in the requests (can be repeated)")
```
EItanya, line 96: Can you give an example with the delimiter character, I think it's `:` based on what I saw below

go/test/mockllm/server.go
```diff
@@ -13,6 +13,7 @@ import (
 
 // Server is the main mock LLM server
 type Server struct {
+	Address           string
```
EItanya, line 16: Nit, I think this should be a config option rather than settable on the `Server` struct. Otherwise the user has to know to modify the structure after creation.

go/test/mockmain/main.go
```diff
@@ -0,0 +1,56 @@
...
+	baseURL, err := server.Start()
+	if err != nil {
+		log.Fatalf("Failed to start server: %v", err)
+	}
+
+	fmt.Printf("Mock LLM server started at: %s\n", baseURL)
+
+	startMcp()
+	// Keep the server running
+	select {}
```
EItanya, line 55: Will this respond to `ctrl-c`?

go/test/mockmcp/server.go
```diff
@@ -0,0 +1,83 @@
...
+	"net/http"
+	"sync/atomic"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+type requestContextKey struct{}
+type Server struct {
+	Addr        net.Addr
```
EItanya, line 16: Why's this public?

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -273,5 +275,80 @@ async def _prepare_session(self, context: RequestContext, run_args: dict[str, An
...
             # Update run_args with the new session_id
             run_args["session_id"] = session.id
-
+        if context.call_context:
+            if session.state is None:
+                session.state = {}
         return session
+
+
+def wrap_session(session_service, token: str):
```
EItanya, line 284: Nit: can we just define this class normally either in this file or a different one, rather than doing all of the imports and declaration in line? This feels more confusing personally

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -118,9 +118,12 @@ async def execute(
                     final=False,
                 )
             )
-
+        # propagate token to session state for MCP tools. make it temporary so it won't be serialized.
+        token = context.call_context.state.get("token", None)
         # Handle the request and publish updates to the event queue
         runner = await self._resolve_runner()
+        if self._config.propagate_token and token:
+            runner.session_service = wrap_session(runner.session_service, token)
```
EItanya, line 126: Is there any harm to just always using this wrapped service? Might make the code a bit cleaner, also then we could just use normal inheritance

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -273,5 +275,80 @@ async def _prepare_session(self, context: RequestContext, run_args: dict[str, An
...
+                config=config,
+            )
+            if session is not None:
+                if session.state is None:
+                    session.state = {}
+                session.state[State.TEMP_PREFIX + "token"] = self._token
+            return session
+
+        @override
+        async def create_session(
```
EItanya, line 314: Do we need to override this function?

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -273,5 +275,80 @@ async def _prepare_session(self, context: RequestContext, run_args: dict[str, An
...
+                user_id=user_id,
+                session_id=session_id,
+            )
+
+        @override
+        async def append_event(self, session: Session, event: Event) -> Event:
+            return await self._wrapped_service.append_event(
+                session=session,
+                event=event,
+            )
```
EItanya, line 351: Same question about unnecessary overrides

EItanya (changes requested):
The changes make sense to me overall. Mostly just cleanup comments

EItanya:
Closing in favor of #993

PR #962, opened 2025-09-29
remove querydoc from agents

peterj:


EItanya (approved):

onematchfox:
Hey @peterj, I'm intrigued by this change. What's the motivation behind it?

EItanya:
> Hey @peterj, I'm intrigued by this change. What's the motivation behind it?

The truth is that the doc2vec support is awkward at best. I have rarely found it very useful, and it has continued to cause issues. Because of how embeddings work, an OpenAI key is required, which has caused a lot of users pain at install time. 

What we’d like to do is revisit the concept of content which requires embeddings and is fetched from vector DBs in a more holistic way.

Does that make sense?

onematchfox:
Yeah, that makes total sense - can totally see it as a pain point for users of _ecosystems_ other than OpenAI. Thanks for clarifying.

PR #971, opened 2025-09-30
feat: add -race flag for go tests, resolves #970

### LINKED ISSUE #970: Execute go tests with `-race`
### 📋 Prerequisites

### 📝 Feature Summary

Execute go tests with `-race`

Data races may occur in code, and introducing dynamic analysis of code is a standard way to help prevent subtle issues.

### 💡 Proposed Solution

Add the `-race` flag whenever executing go tests

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

EItanya:


.github/workflows/ci.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
           export KAGENT_URL="http://$(kubectl get svc -n kagent kagent-controller -o jsonpath='{.status.loadBalancer.ingress[0].ip}'):8083"
           echo $KAGENT_URL
 
-          go test -v github.com/kagent-dev/kagent/go/test/e2e
+          go test -race -v github.com/kagent-dev/kagent/go/test/e2e
```
sam-heilbron, line 94: Will this cause the tests to fail, or do we need to customize the behavior with GORACE variables? What should the behavior be when race detection is enabled?

sam-heilbron (commented):
lgtm!
In other projects, I've seen a consistent make target used to invoke go tests, and that way we could specify a standard set of flags that are passed. Could we worth considering, or doing as a follow up

sam-heilbron (approved):

go/Makefile
```diff
@@ -99,13 +99,13 @@ run: manifests generate fmt vet ## Run a controller from your host.
...
-	CGO_ENABLED=0 go test ./cli/...
+	CGO_ENABLED=0 go test -race ./cli/...
 #	CGO_ENABLED=0 go test ./controller/...
 
 .PHONY: e2e
 e2e:
-	#UPDATE_GOLDEN=true go test -skip 'TestE2E.*|TestAutogenClient' -v ./...
-	go test -skip 'TestE2E.*|TestAutogenClient' -v ./...
+	#UPDATE_GOLDEN=true go test -race -skip 'TestE2E.*|TestAutogenClient' -v ./...
+	go test -race -skip 'TestE2E.*|TestAutogenClient' -v ./...
```
yuval-k, line 108: is there value for race detector in e2e? can we deploy race images to the cluster instead?
to clarify, using the race detector this way will only apply to test code and not application code, as the app code is running in the cluster

yuval-k:
also, the AI comments seem legitz

go/Makefile
sam-heilbron: It's a good point. I kind of see value in both:
- the race detector in e2e will ensure that if there are issues with the test code/execution, it will be flagged, right?
- the race images running in the cluster will catch runtime issues in the product

wdyt?

PR #993, opened 2025-10-02
(WIP) feat: implement STS client and integration in kagent-adk

jmhbh:


python/packages/kagent-adk/src/kagent/adk/wrapped_session_service.py
```diff
@@ -0,0 +1,71 @@
...
+from google.adk.events.event import Event
+from google.adk.sessions import Session
+from google.adk.sessions.base_session_service import (
+    BaseSessionService,
+    GetSessionConfig,
+    ListSessionsResponse,
+)
+from typing_extensions import override
+
+ACCESS_TOKEN_KEY = "access_token"
```
EItanya: A couple of questions about this? 
1. Is storing an access token in state potentially insecure?
2. Should we prefix this with something unique so there's less of a chance of clashing?

python/packages/agw-sts-client/src/agw_status_client/_client.py
```diff
@@ -0,0 +1,217 @@
...
+        if request.requested_token_type:
+            data["requested_token_type"] = request.requested_token_type.value
+
+        # Add additional parameters
+        if request.additional_parameters:
+            data.update(request.additional_parameters)
+
+        return data
+
+    async def exchange_token(
```
EItanya: This func should be `_exchange_token` since it's private?

EItanya (approved):

PR #996, opened 2025-10-03
add-mcp command and wizard

peterj:


go/cli/internal/agent/frameworks/common/base_generator.go
```diff
@@ -20,90 +17,49 @@ type AgentConfig struct {
...
-
-		if config.Verbose {
-			// print the generated files
-			fmt.Printf("  Generated: %s\n", destPath)
-		}
-
-		return nil
-	})
+func (c AgentConfig) ShouldInitGit() bool {
+	return false // Agent projects don't auto-init git
```
EItanya: Why not?

python/packages/kagent-adk/src/kagent/adk/_a2a.py
```diff
@@ -96,6 +96,60 @@ def create_runner() -> Runner:
...
+                app_name=self.app_name,
+                session_service=session_service,
+            )
+
+        agent_executor = A2aAgentExecutor(
+            runner=create_runner,
+        )
+
+        # Use a simple in-memory task store (no backend needed)
+        from a2a.server.tasks import TaskStore
```
EItanya: Can't we use the upstream `InMemoryTaskStore`

EItanya (commented):
Overall this looks great! I think my comments here are just generally some small UX things.

go/cli/internal/agent/frameworks/common/base_generator.go
peterj: we didn't do it before, nothing preventing us from doing it though

EItanya (dismissed):

EItanya (approved):

PR #999, opened 2025-10-04
fix: show nested agent tool calls with visual hierarchy

### LINKED ISSUE #972: show nested agent tool calls
currently we're only showing the subagent (agent->subagent) calls in the UI -- however, if a subagent calls another subagent (e.g. agent -> subagent -> subagent), we don't display it with a separate UI component.

automationpi:
Add support for displaying nested agent calls (agent -> subagent -> subagent) with proper indentation and visual indicators.

The UI now shows delegated calls in a collapsible section with:
- Indented display for each nesting level
- Level indicators showing nesting depth
- Protection against excessive nesting (max 10 levels)

Includes test coverage for nested call scenarios.

Screenshot: 

image

Fixes #972

### DIFF
--- a/ui/src/components/chat/AgentCallDisplay.tsx
+++ b/ui/src/components/chat/AgentCallDisplay.tsx
@@ -1,11 +1,27 @@
 import { useMemo, useState } from "react";
 import { FunctionCall } from "@/types";
 import { Card, CardHeader, CardTitle, CardContent } from "@/components/ui/card";
-import { convertToUserFriendlyName } from "@/lib/utils";
+import { convertToUserFriendlyName, isAgentToolName } from "@/lib/utils";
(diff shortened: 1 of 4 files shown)


peterj:
Could you add a couple of screenshots to the PR? (I'll try to test it out tomorrow, but I am curious to see how it looks like :))

peterj:
I think there might be some issues with how this works -- I've created a nested agent that references k8s, istio and helm agents -  and the nested display looks incorrect.

Screenshot 2025-10-09 at 11 46 23 AM

so it shows the k8s-agent being called which is correct, but then it shows k8s-agent twice as nested calls and then again at the top level (and nested again) with same inputs.

Screenshot 2025-10-09 at 11 47 38 AM

dimetron:
@peterj how those events are propagated from nested ?

Issue #898, opened 2025-09-16
[BUG] A2A communication emits json response on chat UI


AdityaRajputRana:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

If there is A2A Communication involved then the agent does not output the expected response. It outputs the JSON response from the sub-agent call directly. The calling/parent agent does not parse the response from the sub-agent, it outputs it as it is.

<img width="1726" height="872" alt="Image" src="https://github.com/user-attachments/assets/025aca5f-6e30-43fb-9d47-fbcef9451548" />

### 🔄 Steps To Reproduce

1. Create a parent-agent with 1 or more sub-agents added as tools
2. Query the parent-agent with a query that invokes the sub-agent.

### 🤔 Expected Behavior

Instead of showing the the sub-agent output as JSON, the parent-agent should parse the output JSON and pass the data to LLM and output the final result.

### 📱 Actual Behavior

<img width="982" height="329" alt="Image" src="https://github.com/user-attachments/assets/91a2a2f9-4967-4c2d-9ec8-4af6b14777fc" />
Also the output disappears if the page is refreshed.

### 💻 Environment

- K8s on Minikube:
  * Client Version: v1.33.3
  * Kustomize Version: v5.6.0
  * Server Version: v1.33.1
 
- Kagent Version: 0.6.12
 
- Tested on latest version of Chrome

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

AdityaRajputRana:
Hi @peterj , @EItanya this issue has not been completely fixed.
When there are multi-level calls to sub-agents this issue comes up again.

Example when agent A calls agent B and then agent B calls agent C in for the same query. Then B's JSON response contains the JSON of C's response nested inside. So UI only unpacks the outer JSON, but response C in JSON is always there.

[...]

PR #1004, opened 2025-10-07
Fix sub-agent invocation/multi-agent workflows

onematchfox:
From [docs](https://google.github.io/adk-docs/tools/#controlling-agent-flow):

> skip_summarization: bool: (Default: False) If set to True, instructs the ADK to bypass the LLM call that typically summarizes the tool's output. This is useful if your tool's return value is already a user-ready message.

In other words, this prevents the LLM from actually processing the response from the sub-agent so that it can do something with the data. This blocks any vaguely complex multi-agent use-case. Agents should be able to use the response within their workflow (as they are currently able to do with any other tool).

Take the following simple example where we have a random number generator agent and a calculator agent that uses the random number generator to obtain the numbers on which it must subsequently perform calculations:

```yaml
kind: Agent
metadata:
  name: number-generator
  namespace: kagent
spec:
  declarative:
    stream: true
    systemMessage: |-
      You're a random number generator. Respond with a list of N random numbers between 0 and 100 where N is the number requested by the user. Respond in JSON format.

      ## Example
      Input: 5
      Output: [3, 2, 40, 15, 89]
  description: Generates random numbers between 0 and 100
  type: Declarative
---
kind: Agent
metadata:
  name: calculator
  namespace: kagent
spec:
  declarative:
    stream: true
    systemMessage: "You're a calculator. Perform the mathematical operation requested by the user on a list of 5 random numbers obtained from the number-generator agent."
    tools:
    - agent:
        apiGroup: kagent.dev
        kind: Agent
        name: number-generator
      type: Agent
  description: Performs mathematically operations on numbers
  type: Declarative
```

Prior to this PR chatting with the calculator agent would produce output like:
image

Whereas now, the agent actually processes the response and correctly uses it as it would any other tool

Related to #898 and #900 which, attempted to fix the aforementioned issue by simply duplicating the "final" status update in a more readable text message (which incidentally also disappears if you refresh the conversation since the "duplicated" text message that is displayed is not actually persisted). For evidence on not breaking the functionality in #900 (note that the final comment comes from the root agent `kagent/foo` rather than the sub-agent `kagent/k8s-agent` in the screenshot in #900) - this response is also persisted and will show if the chat is refreshed:

Screenshot from 2025-10-07 13-14-14

### DIFF
--- a/ui/src/lib/messageHandlers.ts
+++ b/ui/src/lib/messageHandlers.ts
@@ -300,7 +300,6 @@ export const createMessageHandlers = (handlers: MessageHandlers) => {
       content: normalizeToolResultToText(toolData),
       is_error: toolData.response?.isError || false
     }];
-    
     const execEvent = createMessage(
       "",
       defaultSource,


peterj (approved):
LGTM. Thanks for fixing it!

PR #1005, opened 2025-10-08
feat(fix): Allow transport adapter version to be overriden

jmhbh:
**Overview**
- Allow transport adapter image's version tag to be overridden via the `controller.env` helm values.  
- Fixes a bug in the CLI where a conflicting shorthand causes the CLI to panic.

**Testing**
- Ran a local helm install of the provider by setting the `controller.env` values, and deploying a mcp server using the kmcp cli to verify the environment variable properly sets the init container image.


EItanya (approved):
Looks good

Issue #952, opened 2025-09-26
the "not ready" badge is squished


peterj:
we should clean this up and make it look better -- I am thinking if we could overlay the card with a "Not ready" or some other message

<img width="391" height="222" alt="Image" src="https://github.com/user-attachments/assets/184364d4-7fe7-4004-88d5-5c7217d6ce4b" />

rutujdhawale:
I would like to work on this issue

peterj:
Perfect! I assigned it to you.

rutujdhawale:
Hey Peter! I like your idea of showing a “Not ready” message on cards. I’m thinking of implementing it by updating the badge component to include a StatusOverlay and just wrapping the elements that need it.

onematchfox:
One thing to remember here. Please make sure that the edit and delete buttons remain accessible even if the agent is not ready.

rutujdhawale:
Hi , I made some small changes to the AgentCard and just wanted to check with you if that’s okay
Removed the “Not Accepted” badge and added ribbon at the card bottom.

1)
<img width="411" height="228" alt="Image" src="https://github.com/user-attachments/assets/341e2651-7770-4837-bcb2-f54f224a162d" />

2)
<img width="411" height="228" alt="Image" src="https://github.com/user-attachments/assets/7d43310d-5028-4aef-95bf-44b4d11c0c0e" />

peterj:
I'd prefer option 2 -- if you make the background a bit more transparent and aligned the "agent not ready" to the right, the model name should be visible as well (it might look good - just guessiing :))

rutujdhawale:
I have made some updates to the AgentCard component and wanted to check in with you to see if the changes look okay.
<img width="428" height="230" alt="Image" src="https://github.com/user-attachments/assets/2db846c0-0c7a-433c-8ed2-7e0a30ac63d9" />
<img width="400" height="230" alt="Image" src="https://github.com/user-attachments/assets/72b0f67f-4397-4c0a-90f7-7ae0c4cb3650" />

peterj:
this looks good to me! (we can play around with making it more transparent and/or change the colors, but I like it)

rutujdhawale:
yeah, we can adjust it later i’m opening a new PR please take a look

inFocus7:
note: This issue should be closed, since it got resolved by killjoy's update, moving away from status pills, to a bottom banner-esque format.

_adding this comment in case anyone else sees this gfi and wants to work on it_

PR #1010, opened 2025-10-09
Squished badge bug

killjoycircuit:
Fixed the squished badge bug with a overlay as mentioned in https://github.com/kagent-dev/kagent/issues/952, also made changes in overlay to display messages depending on scenarios and hide the edit/delete buttons unless the user hovers over the card

### DIFF
--- a/ui/src/components/AgentCard.tsx
+++ b/ui/src/components/AgentCard.tsx
@@ -1,12 +1,14 @@
 import { Card, CardHeader, CardTitle, CardContent } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
+import { Badge } from "@/components/ui/badge";
 import type { AgentResponse } from "@/types";
 import { DeleteButton } from "@/components/DeleteAgentButton";
 import KagentLogo from "@/components/kagent-logo";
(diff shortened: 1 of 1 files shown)


ui/src/components/AgentCard.tsx
```diff
@@ -27,55 +31,63 @@ export function AgentCard({ agentResponse: { agent, model, modelProvider, deploy
...
         </CardTitle>
-        <div className="flex items-center space-x-2 invisible group-hover:visible">
+        
+        <div className="flex items-center space-x-2 relative z-30 opacity-0 group-hover:opacity-100 transition-opacity">
           <Button 
             variant="ghost" 
             size="icon" 
             onClick={handleEditClick} 
             aria-label="Edit Agent"
+            className="bg-white/80 hover:bg-white dark:bg-gray-800/80 dark:hover:bg-gray-800 shadow-sm"
```
peterj, line 52: instead of using specific colors here (e.g. bg-white, bg-gray, and above border-violet), I think you should be able to use the bg-primary/bg-secondary, etc. and then, you don't need the `dark` modifier at all.

PR #1012, opened 2025-10-09
resync controllers on agent/mcpserver label change

ilackarms:


EItanya (approved):

onematchfox:
Hey @ilackarms, I might just be out of touch with all the changes to this code base so I was wondering what motivated this change? Why would an agent/MCP server need to be reconciled based on a label update? Are the values now used within the reconciliation process somehow?

PR #1025, opened 2025-10-18
feat: add HITL with DataPart support and event-based sync

apexlnc:
Improves the LangGraph executor HITL implementation with structured decision support and event-based synchronization, plus fixes for checkpoint duplicate key errors.

Controller (Go):
- Fix checkpoint duplicate key errors using GORM's OnConflict{DoNothing: true}
- Use errors.Is() for proper wrapped error detection

kagent-core:
- Add event-based task save synchronization (wait_for_save())
- Replace arbitrary sleep delays with reactive event signaling

kagent-langgraph:
- DataPart Support: Check structured DataPart for decision_type before text parsing
- Event-based sync: Use wait_for_save() instead of asyncio.sleep(0.5)
- Bug fix: Properly access part.root for RootModel types
- Complete HITL implementation with interrupt/resume logic

Tested end-to-end with live HITL workflows. Follows A2A protocol patterns from deepagents and uses idiomatic GORM patterns for database operations.

### DIFF
--- a/python/packages/kagent-core/src/kagent/core/a2a/_task_store.py
+++ b/python/packages/kagent-core/src/kagent/core/a2a/_task_store.py
@@ -1,7 +1,23 @@
+import asyncio
+
 import httpx
 from a2a.server.tasks import TaskStore
 from a2a.types import Task
+from pydantic import BaseModel
 from typing_extensions import override
+
+
+class KAgentTaskResponse(BaseModel):
+    """Wrapper for KAgent controller API responses.
+
+    The KAgent Go controller wraps all task responses in a StandardResponse envelope
+    with the format: {"error": bool, "data": T, "message": str}.
+    This model unwraps that envelope to extract the actual Task object.
+    """
+
+    error: bool
+    data: Task | None = None
+    message: str | None = None
 
 
 class KAgentTaskStore(TaskStore):
@@ -16,13 +32,16 @@ class KAgentTaskStore(TaskStore):
             client: HTTP client configured with KAgent base URL
         """
         self.client = client
+        # Event-based sync: track pending save operations
+        self._save_events: dict[str, asyncio.Event] = {}
 
     @override
-    async def save(self, task: Task) -> None:
+    async def save(self, task: Task, context=None) -> None:
         """Save a task to KAgent.
 
         Args:
             task: The task to save
+            context: Server call context (unused, for a2a-sdk 0.3+ compatibility)
 
         Raises:
             httpx.HTTPStatusError: If the API request fails
@@ -30,12 +49,17 @@ class KAgentTaskStore(TaskStore):
         response = await self.client.post("/api/tasks", json=task.model_dump(mode="json"))
         response.raise_for_status()
 
+        # Signal that save completed (event-based sync)
+        if task.id in self._save_events:
+            self._save_events[task.id].set()
+
     @override
-    async def get(self, task_id: str) -> Task | None:
+    async def get(self, task_id: str, context=None) -> Task | None:
         """Retrieve a task from KAgent.
 
         Args:
             task_id: The ID of the task to retrieve
+            context: Server call context (unused, for a2a-sdk 0.3+ compatibility)
 
         Returns:
             The task if found, None otherwise
@@ -47,17 +71,43 @@ class KAgentTaskStore(TaskStore):
         if response.status_code == 404:
             return None
         response.raise_for_status()
-        return Task.model_validate(response.json())
+
+        # Unwrap the StandardResponse envelope from the Go controller
+        wrapped = KAgentTaskResponse.model_validate(response.json())
+        return wrapped.data
 
     @override
-    async def delete(self, task_id: str) -> None:
+    async def delete(self, task_id: str, context=None) -> None:
         """Delete a task from KAgent.
 
         Args:
             task_id: The ID of the task to delete
+            context: Server call context (unused, for a2a-sdk 0.3+ compatibility)
 
         Raises:
             httpx.HTTPStatusError: If the API request fails
         """
         response = await self.client.delete(f"/api/tasks/{task_id}")
         response.raise_for_status()
+
+    async def wait_for_save(self, task_id: str, timeout: float = 5.0) -> None:
+        """Wait for a task to be saved (event-based sync).
+
+        This method is used to synchronize with the save operation instead of
+        using arbitrary sleep delays. It's particularly useful after interrupts
+        to ensure the task state is persisted before resuming.
+
+        Args:
+            task_id: The ID of the task to wait for
+            timeout: Maximum time to wait in seconds (default: 5.0)
+
+        Raises:
+            asyncio.TimeoutError: If the save doesn't complete within timeout
+        """
+        event = asyncio.Event()
+        self._save_events[task_id] = event
+        try:
+            await asyncio.wait_for(event.wait(), timeout=timeout)
+        finally:
+            # Clean up the event
+            self._save_events.pop(task_id, None)
--- a/python/packages/kagent-langgraph/src/kagent/langgraph/_checkpointer.py
+++ b/python/packages/kagent-langgraph/src/kagent/langgraph/_checkpointer.py
@@ -5,6 +5,7 @@ for LangGraph checkpoint persistence via HTTP API.
 """
 
 import base64
+import json
 import logging
 import random
 from collections.abc import AsyncIterator, Iterator, Sequence
@@ -141,7 +142,8 @@ class KAgentCheckpointer(BaseCheckpointSaver[str]):
         thread_id, user_id, checkpoint_ns = self._extract_config_values(config)
 
         type_, serialized_checkpoint = self.serde.dumps_typed(checkpoint)
-        serialized_metadata = self.jsonplus_serde.dumps(get_checkpoint_metadata(config, metadata))
+        # Serialize metadata as JSON (simpler, no type needed)
+        serialized_metadata = json.dumps(get_checkpoint_metadata(config, metadata)).encode()
         # Prepare request data
         request_data = KAgentCheckpointPayload(
             thread_id=thread_id,
@@ -233,7 +235,7 @@ class KAgentCheckpointer(BaseCheckpointSaver[str]):
             ),
             metadata=cast(
                 CheckpointMetadata,
-                self.jsonplus_serde.loads(base64.b64decode(checkpoint_tuple.metadata.encode("ascii"))),
+                json.loads(base64.b64decode(checkpoint_tuple.metadata.encode("ascii"))),
             ),
             parent_config=(
                 {
--- a/python/packages/kagent-langgraph/src/kagent/langgraph/_converters.py
+++ b/python/packages/kagent-langgraph/src/kagent/langgraph/_converters.py
@@ -4,6 +4,7 @@ This module implements an agent executor that runs LangGraph workflows
 within the A2A (Agent-to-Agent) protocol, converting graph events to A2A events.
 """
 
+import hashlib
 import uuid
 from datetime import UTC, datetime
 from typing import Any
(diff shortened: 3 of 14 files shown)


apexlnc:
cc @EItanya @yuval-k @ilackarms

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -254,6 +515,14 @@ async def execute(
             )
         except Exception as e:
             logger.error(f"Error during LangGraph execution: {e}", exc_info=True)
+
+            from kagent.core.a2a import get_kagent_metadata_key
+            from ._error_mappings import get_user_friendly_error_message, get_error_metadata
```
EItanya: Can we please just import these at the top of the file with the rest?

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -110,24 +111,49 @@ async def _stream_graph_events(
...
             for a2a_event in a2a_events:
                 task_result_aggregator.process_event(a2a_event)
                 await event_queue.enqueue_event(a2a_event)
 
+        # Check for interrupts after streaming completes
+        interrupt_detected = False
+        if final_state and final_state.get("__interrupt__"):
+            interrupt_detected = True
+            interrupt_data = final_state["__interrupt__"]
+            await self._handle_interrupt(
```
EItanya, line 150: Can't we just return here if an interrupt is detected? the 2 if blocks below don't execute if it is.

python/packages/kagent-langgraph/src/kagent/langgraph/_converters.py
```diff
@@ -56,8 +64,17 @@ async def _convert_langgraph_event_to_a2a(
         if not isinstance(messages, list):
             continue
 
-        # Process each message in the event
         for message in messages:
+            # Deduplicate using content hash (message.id is often None)
+            msg_content = f"{type(message).__name__}:{message.content}"
+            if hasattr(message, 'tool_calls') and message.tool_calls:
+                msg_content += f":tools:{len(message.tool_calls)}"
+            msg_id = hashlib.md5(msg_content.encode()).hexdigest()
```
EItanya, line 65: `hashlib` not imported?

python/packages/kagent-core/src/kagent/core/a2a/_task_store.py
```diff
@@ -17,26 +32,34 @@ def __init__(self, client: httpx.AsyncClient):
...
+            context: Server call context (unused, for a2a-sdk 0.3+ compatibility)
 
         Raises:
             httpx.HTTPStatusError: If the API request fails
         """
         response = await self.client.post("/api/tasks", json=task.model_dump())
         response.raise_for_status()
 
+        # Signal that save completed (event-based sync)
+        if task.task_id in self._save_events:
```
EItanya: Does this work? My IDE claims that the field nameis `id` not `task_id`?

python/packages/kagent-core/src/kagent/core/a2a/_task_store.py
```diff
@@ -48,17 +71,43 @@ async def get(self, task_id: str) -> Task | None:
...
             task_id: The ID of the task to delete
+            context: Server call context (unused, for a2a-sdk 0.3+ compatibility)
 
         Raises:
             httpx.HTTPStatusError: If the API request fails
         """
         response = await self.client.delete(f"/api/tasks/{task_id}")
         response.raise_for_status()
+
+    async def wait_for_save(self, task_id: str, timeout: float = 5.0) -> None:
```
EItanya, line 93: Interesting solution to this problem!

apexlnc:
@EItanya - realized i hadn't pushed some stuff i had locally. pushed and addressed some of your other comments.

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+        self,
+        interrupt_data: list[Any],
+        task_id: str,
+        context_id: str,
+        event_queue: EventQueue,
+        task_store: TaskStore,
+    ) -> None:
+        """Handle interrupt from LangGraph and convert to A2A input_required event."""
+        # Extract interrupt details
+        # Format: [{"value": {"action_requests": [...], "review_configs": [...]}}]
```
peterj: Can we use the typed dicts from the sdk? e.g. https://github.com/langchain-ai/langchain/blob/d9e659ca4f020908a360118e7a2a225d86e4b555/libs/langchain_v1/langchain/agents/middleware/human_in_the_loop.py#L39

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+            tool_name = action.get("name", "unknown")
+            tool_args = action.get("args", {})
+
+            # Escape backticks in tool name to prevent markdown breaking
+            escaped_tool_name = str(tool_name).replace("`", "\\`")
+            parts.append(Part(TextPart(text=f"**Tool**: `{escaped_tool_name}`\n")))
+            parts.append(Part(TextPart(text="**Arguments**:\n")))
+            for key, value in tool_args.items():
+                # Escape backticks in values to prevent markdown breaking
+                escaped_key = str(key).replace("`", "\\`")
```
peterj: extract the backticks escape to a separate func since it's used multiple times here

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+                escaped_value = str(value).replace("`", "\\`")
+                parts.append(Part(TextPart(text=f"  • {escaped_key}: `{escaped_value}`\n")))
+            parts.append(Part(TextPart(text="\n")))
+
+        # Add approval metadata as DataPart for slackbot to parse
```
peterj: nit: replace slackbot with "client" :) ideally we want to implement this in the UI client as well

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+        parts.append(
+            Part(
+                DataPart(
+                    data={
+                        "interrupt_type": INTERRUPT_TYPE_TOOL_APPROVAL,
```
peterj: @EItanya I am assuming we'd want to handle the interrupts/HITL requests uniformly across all agent types (crewai, langgraph, adk, ...) -- I don't have a strong opinion for/against using the langgraph way, but we need to be sure we implement it in the same way everything (e.g. using the interupt type, action requests, etc).

peterj: based on that, we should have the types and the base class/interface in the core kagent SDK and have the individual sdks implement them.

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+        """Resume graph execution after interrupt with user decision."""
+        # Determine decision from message
+        message_text = context.get_user_input().lower()
+
+        if "approved" in message_text or "proceed" in message_text:
+            decision_type = "approve"
+        elif "denied" in message_text or "cancel" in message_text:
+            decision_type = "reject"
+        else:
+            decision_type = "approve"  # Default to approve for safety
```
peterj: defaulting to deny feels safer I think (e.g. "Do you want me to delete stuff?" :)).

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+        resume_input = Command(resume={"decisions": [{"type": decision_type}]})
+
+        # Create graph config with explicit thread_id
+        config = {
+            "configurable": {
+                "thread_id": thread_id,  # Use thread from interrupted task!
+                "app_name": self.app_name,
+            },
+            "project_name": self.app_name,
+            "run_name": "kagent-langgraph-resume",
```
peterj, line 319: are there issues in having the name be constant?

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +202,245 @@ async def _stream_graph_events(
...
+    async def _handle_resume(
+        self,
+        context: RequestContext,
+        event_queue: EventQueue,
+    ) -> None:
+        """Resume graph execution after interrupt with user decision."""
+        # Determine decision from message
+        message_text = context.get_user_input().lower()
+
+        if "approved" in message_text or "proceed" in message_text:
```
peterj: would this be meant for approve/deny scenarios only? what if we want to ask for specific input from the user and resume execution with that? 

I think the decisions (e.g. "approve"/"deny") should be constants coming from the client --  clicking Approve/Deny (or whatever UX we go with) should automatically send True/False or some enum that we know will always signal an approve/deny decision, so we don't have to check for specific strings in message.

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
EItanya: 100% I agree. I'm also ok with the idea of merging this and the unifying, but definitely need to unify at some point.

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
apexlnc: Agreed, but going to leave `TextPart` as a fallback for now, if that's ok

apexlnc:
@peterj @EItanya -- decided to take the time to move the `HITL` stuff to `kagent-core` so it can be reused across the ecosystem. addressed the rest of the comments as well.

apexlnc:
cc @peterj @EItanya can you kick off another GHA?

python/packages/kagent-langgraph/src/kagent/langgraph/_checkpointer.py
```diff
@@ -141,7 +142,8 @@ async def aput(
         thread_id, user_id, checkpoint_ns = self._extract_config_values(config)
 
         type_, serialized_checkpoint = self.serde.dumps_typed(checkpoint)
-        serialized_metadata = self.jsonplus_serde.dumps(get_checkpoint_metadata(config, metadata))
+        # Serialize metadata as JSON (simpler, no type needed)
```
EItanya, line 145: Can you explain this change a little more? I initially used that serializer in order to match the langraph code.

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -171,12 +204,181 @@ async def _stream_graph_events(
...
+        event_queue: EventQueue,
+        task_store: TaskStore,
+    ) -> None:
+        """Handle interrupt from LangGraph and convert to A2A input_required event.
+
+        This is the LangGraph-specific adapter that extracts interrupt data from
+        LangGraph's format and delegates to the generic handler in kagent-core.
+        """
+        # Extract interrupt details from LangGraph format
+        if not interrupt_data:
```
EItanya, line 221: In Python does this make sure it has `len > 0`?

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
EItanya: You didn't answer this

EItanya (commented):
Just a couple of last questions/nits, overall looking awesome!

python/packages/kagent-langgraph/src/kagent/langgraph/_checkpointer.py
apexlnc: - LangGraph 1.0 changed the API: removed .dumps(), now only .dumps_typed() exists, and this was causing: `AttributeError: 'JsonPlusSerializer' object has no attribute 'dumps'`

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
apexlnc: Yeah Python empty lists are falsy, so this checks if the list is empty and returns early if so

python/packages/kagent-langgraph/src/kagent/langgraph/_checkpointer.py
apexlnc: As for why JSON instead of .dumps_typed(): 

- Go backend expects JSON - The database schema comment explicitly says: Metadata string // JSON serialized metadata
- No metadata_type field - Go schema has checkpoint_type for checkpoints but no metadata_type for metadata
- Cross-language - JSON works Python ↔ Go, msgpack would need type info

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
apexlnc: I kept it constant to match the initial execution pattern (line 96 also uses constant "kagent-langgraph-exec").

The run_name is used for LangSmith/tracing to categorize run types. Since tags already include dynamic values (task_id, context_id, thread_id), I figured the run_name should be a constant category label. I'm not 100% certain this is correct. Should run_name include unique identifiers? Or is constant appropriate for grouping trace types? I can change it to f"kagent-langgraph-resume-{task_id}" if that's better for observability.

apexlnc:
> Just a couple of last questions/nits, overall looking awesome!

Let me know if you have any other questions!

PR #1027, opened 2025-10-20
fix(kagent-adk/cli/test): fix "NotADirectoryError: [Errno 20] Not a directory: 'xxx/agent-card.json'"

mozillazg:
## Before

```
$ uv run cli.py test --task 'who are you' --filepath config.json 
...
NotADirectoryError: [Errno 20] Not a directory: 'config.json/agent-card.json'
```

## After

```
$ uv run cli.py test --task 'who are you' --filepath ./
...
INFO:root:Starting KAgent
INFO:kagent.adk._a2a:
>>> User Query: who are you
...
```

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/cli.py
+++ b/python/packages/kagent-adk/src/kagent/adk/cli.py
@@ -98,14 +98,12 @@ async def test_agent(agent_config: AgentConfig, agent_card: AgentCard, task: str
 @app.command()
 def test(
     task: Annotated[str, typer.Option("--task", help="The task to test the agent with")],
-    config_file: Annotated[str, typer.Option("--agent-config", help="The path to the agent config file")],
-    card_file: Annotated[str, typer.Option("--agent-card", help="The path to the agent card file")],
+    filepath: Annotated[str, typer.Option("--filepath", help="The path to the directory of config files")],
 ):
-    with open(config_file, "r") as f:
-        content = f.read()
-        config = json.loads(content)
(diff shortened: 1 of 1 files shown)


mozillazg:
@EItanya 

> Why this change? Why can't both files be in the same directory?

They can be in the same directory. The issue is that `--filepath` currently points to the `config.json` file. The code then joins `agent-card.json` to that file path, producing `/path/to/config.json/agent-card.json`, which raises a `NotADirectoryError`. If you prefer, I can change `--filepath` to accept the config directory instead, and read both files from there.

EItanya:
> @EItanya
> 
> > Why this change? Why can't both files be in the same directory?
> 
> They can be in the same directory. The issue is that `--filepath` currently points to the `config.json` file. The code then joins `agent-card.json` to that file path, producing `/path/to/config.json/agent-card.json`, which raises a `NotADirectoryError`. If you prefer, I can change `--filepath` to accept the config directory instead, and read both files from there.

Yes, let's do that

mozillazg:
> > @EItanya
> > > Why this change? Why can't both files be in the same directory?
> > 
> > 
> > They can be in the same directory. The issue is that `--filepath` currently points to the `config.json` file. The code then joins `agent-card.json` to that file path, producing `/path/to/config.json/agent-card.json`, which raises a `NotADirectoryError`. If you prefer, I can change `--filepath` to accept the config directory instead, and read both files from there.
> 
> Yes, let's do that

@EItanya DONE.

mozillazg:
@EItanya ping ~

Issue #1003, opened 2025-10-07
mismatched tool descriptions


peterj:
I am adding the k8s tool, yet the description (on the right) seems to be from a different tool

<img width="1228" height="210" alt="Image" src="https://github.com/user-attachments/assets/e79c4273-0563-4358-8739-2a8c54cd772d" />

rutujdhawale:
@peterj can I work on this issue? From what I understand, the problem is that the tool description don’t match their actual functionality  is that correct?

peterj:
SO the descriptions on the left side look correct, however, when the tool gets added/selected, for some reason it picks the description of a different tool. I think this might be some issue with incorrectly matching the ids on the tools.

rutujdhawale:
This issue is resolved. Kindly close it

PR #1033, opened 2025-10-20
Mismatched tool description bug

killjoycircuit:
fix(ui): correct tool descriptions and agent card display issues

- Changed server reference comparison from partial name matching to full reference matching, ensuring tools display their correct descriptions and prevent adding the same tool multiple times
- Resolved AgentCard overlay interfering with model name visibility

Fixes incorrect tool descriptions in ToolsSection and SelectToolsDialog components, and improves AgentCard UI rendering.
Closes issue #1003 

Screenshot from 2025-10-20 23-31-02

### DIFF
--- a/ui/src/components/create/ToolsSection.tsx
+++ b/ui/src/components/create/ToolsSection.tsx
@@ -3,13 +3,12 @@ import { Card, CardContent } from "@/components/ui/card";
 import { Plus, FunctionSquare, X } from "lucide-react";
 import { ScrollArea } from "@/components/ui/scroll-area";
 import { useState, useEffect } from "react";
-import { isAgentTool, isMcpTool, getToolDescription, getToolIdentifier, getToolDisplayName } from "@/lib/toolUtils";
+import { isAgentTool, isMcpTool, getToolDescription, getToolIdentifier, getToolDisplayName, serverNamesMatch } from "@/lib/toolUtils";
 import { SelectToolsDialog } from "./SelectToolsDialog";
 import type { Tool, AgentResponse, ToolsResponse } from "@/types";
 import { getAgents } from "@/app/actions/agents";
 import { getTools } from "@/app/actions/tools";
 import KagentLogo from "../kagent-logo";
-import { k8sRefUtils } from "@/lib/k8sUtils";
 
 interface ToolsSectionProps {
   selectedTools: Tool[];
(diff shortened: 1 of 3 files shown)


peterj:
thanks for picking this up! 

There seems to be an issue as I am able to add the same tool multiple times: 

Screenshot 2025-10-21 at 1 29 23 PM

peterj:
another behavior difference is that when I open an existing agent with these changes, I don't see any descriptions. 
Screenshot 2025-10-28 at 2 15 35 PM

Previously, I'd see the actual decsriptions of these tools

rutujdhawale:
i'm getting these issue with the ```npm run lint``` command  i'm not sure how should i fix it  ,might cause the workflow to fail 
Screenshot from 2025-10-30 02-53-20

ui/src/app/models/new/page.tsx
```diff
@@ -56,7 +53,6 @@ const processModelParams = (requiredParams: ModelParam[], optionalParams: ModelP
       return acc;
     }, {} as Record<string, string>);
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
peterj, line 59: add these disable statements back, so the errors are ignored.

rutujdhawale:
@peterj i made the changes can you please take a look

ui/src/components/create/ToolsSection.tsx
```diff
@@ -19,6 +19,25 @@ interface ToolsSectionProps {
   currentAgentName: string;
 }
 
+const serverNamesMatch = (serverName1: string, serverName2: string): boolean => {
```
peterj: this is duplicated in the SelectToolsDialog -- extract it out to the toolUtils.ts in the /utils lib folder, so it can be reused

ui/src/components/AgentCard.tsx
```diff
@@ -69,37 +89,12 @@ export function AgentCard({ agentResponse: { agent, model, modelProvider, deploy
...
+      {statusInfo && (
         <div className={cn(
-          "absolute inset-0 rounded-xl flex flex-col items-center justify-center z-20 backdrop-blur-[2px]",
-          !accepted 
-            ? "bg-destructive/90" 
-            : "bg-secondary/90"
+          "absolute bottom-0 left-0 right-0 z-20 py-1.5 px-4 text-right text-xs font-medium rounded-b-xl",
+          statusInfo.className
         )}>
-          <div className="text-center px-6 py-8 max-w-[80%]">
```
peterj, line 80: why was this removed?

rutujdhawale: my mistake i was testing a different status indicator locally (small bottom ribbon vs full overlay) and accidentally pushed the wrong version. this one is like we discussed in #952

peterj: this is still a change from what is currently checked in -- are we changing this on purpose? (I am fine if we are, I just want to make sure that this is a deliberate change and not a mistake with a merge)

rutujdhawale: yes, this one’s intentional. i realized I’d accidentally pushed my local test version earlier, so this update restores the correct ribbon style . sorry for the mix up

peterj:
almost there :) can you sign the DCO and we’ll get this merged

rutujdhawale:
@peterj i created this new PR to avoid conflicts from pulling changes into my previous branch. all the intended changes are included  here in PR #1079

PR #1034, opened 2025-10-21
feat(agent): default session name when auto-generating session

onematchfox:
Prior to this change, auto-generated sessions - created by invoking `message/send` via A2A - would be displayed in the UI as "Untitled". Now, they will have a default name as per sessions initiated via the UI.

Tested via command:

```bash
$ jq -n --arg id "$(cat /proc/sys/kernel/random/uuid)" --arg messageId "$(cat /proc/sys/kernel/random/uuid)" '{
  "jsonrpc": "2.0",
  "id": $id,
  "method": "message/send",
  "params": {
    "message": {
      "role": "user",
      "parts": [
        {
          "kind": "text",
          "text": "Hi there!"     
        }
      ],
      "messageId": $messageId
    },
    "metadata": {}
  }
}' \
  | curl -H "Content-Type: application/json" http://localhost:8083/api/a2a/kagent/k8s-agent/ -d @-
```

Current behaviour:
image

After change:
image

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
@@ -256,12 +256,35 @@ class A2aAgentExecutor(AgentExecutor):
             session_id=session_id,
         )
         if session is None:
-            session = await runner.session_service.create_session(
-                app_name=runner.app_name,
-                user_id=user_id,
-                state={},
-                session_id=session_id,
-            )
+            # Extract session name from the first TextPart (like the UI does)
+            session_name = None
+            if context.message and context.message.parts:
+                for part in context.message.parts:
+                    # A2A parts have a .root property that contains the actual part (TextPart, FilePart, etc.)
+                    if isinstance(part, Part):
+                        root_part = part.root
+                        if isinstance(root_part, TextPart) and root_part.text:
+                            # Take first 20 chars + "..." if longer (matching UI behavior)
+                            text = root_part.text.strip()
+                            session_name = text[:20] + ("..." if len(text) > 20 else "")
+                            break
+
+            # Prepare kwargs for create_session
+            create_session_kwargs = {
+                "app_name": runner.app_name,
+                "user_id": user_id,
+                "state": {},
+                "session_id": session_id,
+            }
+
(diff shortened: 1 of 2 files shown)


python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -256,11 +256,24 @@ async def _prepare_session(self, context: RequestContext, run_args: dict[str, An
...
+            session_name = None
+            if context.message and context.message.parts:
+                for part in context.message.parts:
+                    # A2A parts have a .root property that contains the actual part (TextPart, FilePart, etc.)
+                    if hasattr(part, "root"):
+                        root_part = part.root
+                        if isinstance(root_part, TextPart) and root_part.text:
+                            # Take first 20 chars + "..." if longer (matching UI behavior)
+                            text = root_part.text.strip()
+                            session_name = text[:20] + ("..." if len(text) > 20 else "")
```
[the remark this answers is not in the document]
onematchfox: Replicates the code in [the UI](https://github.com/kagent-dev/kagent/blob/e87058c73970e1df26d250d7ee7dfafc3e82cf6b/ui/src/components/chat/ChatInterface.tsx#L179).

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -256,11 +256,24 @@ async def _prepare_session(self, context: RequestContext, run_args: dict[str, An
...
+                            break
             session = await runner.session_service.create_session(
                 app_name=runner.app_name,
                 user_id=user_id,
                 state={},
                 session_id=session_id,
+                session_name=session_name,
```
EItanya: nit: this is calling the base class which doesn't have a `session_name` attribute, so this code may fail if running with a different `session_service`. I hate Python, but can we use the `*` argument from the args somehow?

onematchfox: Ah, yeah, fair point and good catch. Will adjust to use kwargs

onematchfox: Done :pray:

PR #1052, opened 2025-10-25
feat: hide Tools & Agents section for BYO agents

### LINKED ISSUE #1047: remove the agents/tool section if we have BYO agent
With the BYO agent we don't know which tools/agents the agent is referencing. in the chat screen we should remove the tools section:

```
Tools & Agents
No tools/agents available
```

Otherwise, it seems like the agent doesn't have any tools.

rishabh998186:
This PR hides the "Tools & Agents" section in the sidebar when viewing BYO agents since they don't expose their tool information to the UI.

closes  #1047

### DIFF
--- a/ui/src/components/sidebars/AgentDetailsSidebar.tsx
+++ b/ui/src/components/sidebars/AgentDetailsSidebar.tsx
@@ -213,7 +213,7 @@ export function AgentDetailsSidebar({ selectedAgentName, currentAgent, allTools 
   };
 
     // Check if agent is BYO type
-  const isBYOAgent = selectedTeam?.agent.spec.type === "BYO";
+  const isDeclarativeAgent = selectedTeam?.agent.spec.type === "Declarative";
   
   return (
     <>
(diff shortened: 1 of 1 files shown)


ui/src/components/sidebars/AgentDetailsSidebar.tsx
```diff
@@ -212,6 +212,9 @@ export function AgentDetailsSidebar({ selectedAgentName, currentAgent, allTools
     );
   };
 
+    // Check if agent is BYO type
+  const isBYOAgent = selectedTeam?.agent.spec.type === "BYO";
```
peterj: This sounds like a nit, but it makes code more readable. 

Instead of checking for a negated value (e.g. `!isBYOAgent`) you should check for a non-negated value instead. so `isBYOAgent` should be `isDeclarativeAgent` and then below you can just use a non-negated value.

rishabh998186:
@peterj 
I' updated the code to use positive logic with `isDeclarativeAgent` instead of a negated value

peterj:
looks good - thanks!

Issue #885, opened 2025-09-10
[FEATURE] Upload files


antonioberben:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

I can upload files through the UI

### ❓ Problem Statement / Motivation

I cannot uploa files as part of the context

### 💡 Proposed Solution

Like Copilot, add the option to upload files as part of the context.

### 🎯 Affected Service(s)

Multiple services / System-wide

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

tanuj-rai:
Hi @antonioberben, I'm interested in this issue. Could you please provide more details to understand the issue better?

antonioberben:
Sure. It is similar functionality that you can have in Copilot. You attach files as part of the context.

It would be interesting that you can do that with Kagent.

Is the picture more clear?

tanuj-rai:
> Sure. It is similar functionality that you can have in Copilot. You attach files as part of the context.
> 
> It would be interesting that you can do that with Kagent.
> 
> Is the picture more clear?

Yes @antonioberben, thank you for the details. I am working on it and will soon raise a PR to fix it.

peterj:
@tanuj-rai  are you working on this issue?

tanuj-rai:
Hi @peterj, I’m not currently working on this issue.

ishansaini194:
Can I work on it if no one is ? @antonioberben @peterj

peterj:
yes! will assign to you. thanks!

ishansaini194:
> yes! will assign to you. thanks!

Just a quick question are we limiting size of the files like to 25 MB or some specific value ?
And also the type of files that can be uploaded ?

antonioberben:
IMO, I would start simple and iterate. It seems that OpenAI's limitation is (for the most basic scenario):
- 10 files in one prompt
- 512MB per file
-  2M tokens per file

peterj:
also let's focus on the text files for now only (e.g. .txt, .yaml, .log etc.) ; the limits can be configurable later as well, so I am fine if we keep them low/small files to start with.

ishansaini194:
[...]

PR #1062, opened 2025-10-28
[Draft] Agent Skills

supreme-gg-gg:
This PR is part of a series of features aiming to introduce agent skills like Claude: https://docs.claude.com/en/docs/agents-and-tools/agent-skills/overview

## Usage

1. Obtain some skills from Anthropic's public repo or write your own

2. Update your agent to use skill and bash tools, if needed the artifacts related tools as well

```python
from kagent.adk.skills import BashTool, SkillsTool

# Define the path to the skills directory, relative to this file.
SKILLS_DIR = Path(__file__).parent.parent / "skills"

agent = Agent(
    model="gemini-2.5-flash",
    name="skills_agent",
    description="An agent using skills to assist the user.",
    instruction="You are a helpful assistant that uses various skills to assist the user.",
    tools=[
        SkillsTool(skills_directory=SKILLS_DIR),  # Discover and load skills
        BashTool(skills_directory=SKILLS_DIR),  # Execute commands
    ],
)
```

3. Alternatively, directly use the plugin in your app this enables these tools for all agents in a multi-agent system:

```python
from kagent.adk.skills import SkillsPlugin
# Optional below
from google.adk.plugins.save_files_as_artifacts_plugin import SaveFilesAsArtifactsPlugin

kagent_app = KAgentApp(
        root_agent=agent,
        agent_card=agent_card,
        kagent_url=config.url,
        app_name=config.app_name,
        plugins=[
            SkillsPlugin(skills_directory=SKILLS_DIR),
            SaveFilesAsArtifactsPlugin(),
        ],
    )
```

## Limitations

- User **cannot** upload or download files, for upload it's technically supported if you write your own script but for the UI we require #885. For download we will need a non-in-memory artifact service, currently we use `InMemoryArtifactService` to save uploaded files and stage them to the working directory as a tool

- The agent is not very good at writing code due to 1. The bash tool is not robust, cd and some other command sometimes fails, this will be solved once we have a better code executor / sandbox 2. (Possibly) not enough tools for the agent to use to write code like write, read, edit, search, etc. commonly given to coding agents, it is reliant on very basic shell commands like cat and printf

- This only works with BYO agents + build from source (not from the `kagent-adk` image)

## Testing

This is tested on the following skills from https://github.com/anthropics/skills and Claude Cookbook

- `brand-guidelines`: covers most simple cases where reading markdown files is enough
- `analyzing-financial-statement`: slightly more difficult, involves finding and running scripts, manages user CSV data
- `slack-gif-creator`: this requires the agent to write and run code with snippets provided, the agent occasionally runs into issues with the bash tool that will probably be solved with the proper sandbox environment

The above performances are subject to the "limitations" discussed. However, some of them can be bypassed by running the agent just locally.


python/packages/kagent-adk/src/kagent/adk/skills/skill_system_prompt.py
supreme-gg-gg, line 1: I will probably remove this skills system prompt function because the agent knows the system well enough just from the tools

EItanya (approved):
Great start! I am going to merge this and then we can continue to iterate

PR #1067, opened 2025-10-29
fix: remove v from kmcp tag

EItanya:


antweiss:
Oops, just opened an identical PR :))

PR #1071, opened 2025-10-29
feat: Use Sandbox Runtime for Bash tool and refactor artifact utils tools

supreme-gg-gg:
This PR expands features from #1062 and integrates with #1077.

- Wrap bash commands with Anthropic Sandbox Runtime ("srt") to replace manually handling security
- Creates working directory for agent based on session id, where `skills/` is symlinked to the `/skills` loaded into the container or any other customised skills base path. This directory is where all commands are run, python imports from skills are handled by modifying the `PYTHONPATH` env variable 
- Added write, read, edit file tools to help the agent perform better at simple coding tasks when skill requires them
- The following enhancement to artifact related utils
  - Tool to "stage" artifact from the storage to working directory so the agent can perform action on them
  - Tool to "return" output artifacts from working directory to the storage so the client can download later
  - Currently, working with artifacts require custom `Runner` loop (run locally) due to frontend limitations

## Testing

Features in this PR will work to the full capacity of the examples listed in #1062, this includes everything from accessing skills to writing code, managing dependencies, and creating files / images, **subject to the limitations below**.

## Limitations

- This only works locally, to work in BYO or any other environment we need to install srt + setup `.srt-config.json` properly, for local testing just adding `pypi.org` to allowed network

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/cli.py
+++ b/python/packages/kagent-adk/src/kagent/adk/cli.py
@@ -13,7 +13,7 @@ from kagent.core import KAgentConfig, configure_tracing
 from .skill_fetcher import fetch_skill
 
 from . import AgentConfig, KAgentApp
-from .skills.skills_plugin import add_skills_tool_to_agent
+from .skills.skills_plugin import SkillsPlugin
 
 logger = logging.getLogger(__name__)
 logging.getLogger("google_adk.google.adk.tools.base_authenticated_tool").setLevel(logging.ERROR)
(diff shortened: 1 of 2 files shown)


python/packages/kagent-adk/src/kagent/adk/cli.py
```diff
@@ -41,9 +41,11 @@ def static(
     skills_directory = os.getenv("KAGENT_SKILLS_FOLDER", None)
     if skills_directory:
         logger.info(f"Adding skills from directory: {skills_directory}")
-        add_skills_tool_to_agent(skills_directory, root_agent)
+        plugins = [SkillsPlugin(skills_directory=skills_directory)]
 
-    kagent_app = KAgentApp(root_agent, agent_card, app_cfg.url, app_cfg.app_name)
```
supreme-gg-gg, line 46: I changed this to use plugin instead of directly adding tools because it handles creating temporary session paths as well

PR #1077, opened 2025-10-30
feat: skills for declarative agents

yuval-k:
pull in skills to the agent from a docker registry. This is done in an init container:
- so it also works for BYO agents (i.e. this part is now not framework dependent)
- so it skills fails to pull the pod fails to start, i think that would give better feedback for users

follow up tasks
- [ ] use a different venv for sandbox with bunch of python packages
- [ ] maybe: support multiple skills in one image
- [ ] validate the skill image has a SKILL.md file and is at most 10MB
- [ ] make running code not require privileged

### DIFF
--- a/go/internal/controller/translator/agent/testdata/inputs/agent_with_code.yaml
+++ b/go/internal/controller/translator/agent/testdata/inputs/agent_with_code.yaml
@@ -34,7 +34,7 @@ objects:
     spec:
       type: Declarative
       declarative:
-        executeCode: true
+        executeCodeBlocks: true
         description: A basic test agent
         systemMessage: You are a helpful assistant.
         modelConfig: basic-model
--- a/go/internal/controller/translator/agent/testdata/inputs/agent_with_skills.yaml
+++ b/go/internal/controller/translator/agent/testdata/inputs/agent_with_skills.yaml
@@ -33,7 +33,7 @@ objects:
       namespace: test
     spec:
       skills:
-        images:
+        refs:
           - foo:latest
       type: Declarative
       declarative:


go/api/v1alpha2/agent_types.go
```diff
@@ -53,6 +53,23 @@ type AgentSpec struct {
...
+type SkillForAgent struct {
+	// Fetch images insecurely from registries (allowing HTTP and skipping TLS verification).
+	// Meant for development and testing purposes only.
+	// +optional
+	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+
+	// The list of skill images to fetch.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=20
+	Images []string `json:"images,omitempty"`
```
EItanya: I don't love the name images because that's an Implementation detail, maybe reference, uri, something like that?

go/api/v1alpha2/agent_types.go
```diff
@@ -85,6 +102,8 @@ type DeclarativeAgentSpec struct {
 
 	// +optional
 	Deployment *DeclarativeDeploymentSpec `json:"deployment,omitempty"`
+
+	ExecuteCode *bool `json:"executeCode,omitempty"`
```
EItanya: Can we be more specific about this field name, specifically this executor will execute code that is returned from the API directly, not files or anything like that. Also, maybe this should be an struct as we may want to support multiple executors in the future?

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1084,16 +1137,19 @@ func getDefaultLabels(agentName string, incoming map[string]string) map[string]s
...
 		"--host",
 		"0.0.0.0",
 		"--port",
 		fmt.Sprintf("%d", port),
 		"--filepath",
 		"/config",
 	}
+	needSandbox := false
+	if ptr.Deref(agent.Spec.Declarative.ExecuteCode, false) {
+		args = append(args, "--code")
```
EItanya: This option should probably be in the agent config.

python/Dockerfile
```diff
@@ -47,6 +47,18 @@ RUN addgroup -g 1001 pythongroup                           && \
     mkdir    -p /python                                    && \
     chown    -vR 1001:1001 /.kagent /python
 
+# Install anthropic sandbox runtime and dependencies
+RUN --mount=type=cache,target=/var/cache/apk,rw \
+    apk add \
+    nodejs npm bubblewrap socat ripgrep
+
+RUN npm install -g @anthropic-ai/sandbox-runtime@0.0.1
```
EItanya: This version doesn't work with the config :(, need to build from source

python/packages/kagent-adk/src/kagent/adk/skill_fetcher.py
```diff
@@ -0,0 +1,104 @@
...
+def fetch_skill(skill_image: str, destination_folder: str, insecure: bool = False) -> None:
+    """
+    Fetch a skill packaged as an OCI/Docker image and write its files to destination_folder.
+
+    Assumptions:
+    - We do not depend on a local Docker daemon.
+    - We pull the manifest directly from the registry (anonymous) and download only the LAST layer.
+    - We read at most 10MB from the registry to avoid OOM/abuse.
+    - If the last layer is a tar+gzip layer, we extract all files to destination_folder.
+    - If the last layer is a custom wasm artifact, we store it as plugin.wasm in destination_folder.
```
EItanya: Wasm?

python/packages/kagent-adk/src/kagent/adk/skill_fetcher.py
```diff
@@ -0,0 +1,104 @@
...
+        name_part, ref = image.split("@", 1)
+        # cover the case of both tag and digest present
+        # Split tag vs registry port: use last ':' after last '/'
+        slash = name_part.rfind("/")
+        colon = name_part.rfind(":")
+        if colon > slash:
+            name_part, ref = name_part[:colon], name_part[colon + 1 :]
+    else:
+        colon = name_part.rfind(":")
+        name_part, ref = name_part[:colon], name_part[colon + 1 :]
```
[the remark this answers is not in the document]
yuval-k: its not a problem

python/packages/kagent-adk/src/kagent/adk/skill_fetcher.py
```diff
@@ -0,0 +1,104 @@
...
+    # Extract the tarball
+    with tarfile.open(tar_path, "r") as tar:
+        tar.extractall(path=destination_folder)
+
+    # Remove the tarball
+    os.remove(tar_path)
+    # remove the Dockerfile if exists, as it's not needed for the agent
+    try:
+        os.remove(os.path.join(destination_folder, "Dockerfile"))
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
yuval-k: i just removed the whole line

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -330,9 +330,54 @@ func (a *adkApiTranslator) buildManifest(
...
+	needSandbox := cfg != nil && cfg.ExecuteCode
+
+	var initContainers []corev1.Container
+
+	if len(skills) > 0 {
+		skillsEnv := corev1.EnvVar{
+			Name:  "KAGENT_SKILLS_FOLDER",
+			Value: "/skills",
+		}
+		needSandbox = true
```
peterj, line 350: by setting this to true here, are saying that if you add skills to the agent you'll always also need a sandbox? Are there agents with skills that won't need to execute code?

yuval-k: there are skills that don't have code... but we don't know it ahead of time...

python/packages/kagent-adk/src/kagent/adk/skill_fetcher.py
yuval-k: copy-paste

go/api/v1alpha2/agent_types.go
yuval-k: change to refs

go/api/v1alpha2/agent_types.go
yuval-k: changed to `ExecuteCodeBlocks` so it's more clear

EItanya (approved):

PR #1084, opened 2025-11-04
docker compose file conventional name should be lowercase - an issue …

eitansuez:
…when running on linux


go/cli/internal/agent/frameworks/adk/python/templates/docker-compose.yaml.tmpl
eitansuez, line 1: renamed from Docker-compose.yaml.tmpl to docker-compose.yaml.tmpl

eitansuez:
not sure what's blocking the pending checks.  also, i don't get why the top of the file says that it's auto-generated by the kagent-adk cli, but i can't seem to find any instructions or makefile targets that would regenerate the file. i suspect manually renaming the template file is safe.

PR #1094, opened 2025-11-06
fix(helm): model config for multiple providers

nlamirault:
generate **multiple ModelConfig resources** (one per provider) instead of just one for the default provider.


EItanya (commented):
This is a very interesting idea I like it! I will try and get to a deeper review ASAP but it may be after KubeCon as there is A LOT happening the next few work days.

dimetron:
maybe  we should add range over additional models of the same provider ?
It's less common to use many provider types, but use model types like (default, thinking, fast) which will be translated into additional model names under the same provider. WDYT ?

PR #1100, opened 2025-11-12
feat: refactor agent skills and support OpenAI Agents SDK

supreme-gg-gg:
Completes #1068 

- Integrates OpenAI SDK (fixed event converters, session service, tools, tracing) and added e2e test for OpenAI SDK
- Refactor skills prompt and tool logic into `kagent-skills`, update ADK and OpenAI SDK to wrap them with their functions tool implementations

### DIFF
--- a/go/test/e2e/mocks/invoke_inline_agent.json
+++ b/go/test/e2e/mocks/invoke_inline_agent.json
@@ -3,7 +3,7 @@
     {
       "name": "initial_request",
       "match": {
-        "match_type": "exact",
+        "match_type": "contains",
         "message": {
           "content": "List all nodes in the cluster",
           "role": "user"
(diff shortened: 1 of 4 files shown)


python/packages/kagent-openai/src/traceloop/__init__.py
supreme-gg-gg, line 1: I used this stub as a temporary solution to fix `opentelemetry-instrumentation-openai-agents` due to some dependency hell. Ideally we would update it to `>0.50.0` since the `traceloop` import is made optional there, but it conflicts with the opentelemtry api pinned by ADK. 

If needed I can look for a more elegant method to fix the issue, but I tried different versions of traceloop and it has some weird version pinning with the instrumentation library and some versions have bugs that just crashed... 😭

PR #1104, opened 2025-11-13
fix(tracing): use a span processor to add custom span attributes

dhaifley:
- Fix custom tracing span attributes for the kagent-adk, kagent-langgraph, kagent-crewai packages.
- Add a custom span processor which can be used to include custom kagent span metrics for all spans created within a particular OTel trace context.
- Use the custom span processor, and update the execute method of each agent framework package to set the custom span attributes in context.

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import inspect
 import logging
 import uuid
@@ -29,6 +28,10 @@ from pydantic import BaseModel
 from typing_extensions import override
 
 from kagent.core.a2a import TaskResultAggregator, get_kagent_metadata_key
+from kagent.core.tracing._span_processor import (
+    clear_kagent_span_attributes,
+    set_kagent_span_attributes,
+)
 
 from .converters.event_converter import convert_event_to_a2a_events
 from .converters.request_converter import convert_a2a_request_to_adk_run_args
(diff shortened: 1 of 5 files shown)


python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -175,14 +189,6 @@ async def _handle_request(
 
         await runner.session_service.append_event(session, system_event)
 
-        current_span = trace.get_current_span()
```
krisztianfekete, line 178: Shouldn't we keep this? I am wondering if we can miss attributes with the processor-only approach for certain spans?

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -262,6 +268,8 @@ async def _handle_request(
                 )
             )
 
+        clear_kagent_span_attributes(context_token)
```
krisztianfekete: Could clearing attributes only here result in keeping stale/invalid attributes for irrelevant spans?

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
dhaifley: The problem is this: trace.get_current_span() doesn't work. We are not in a span here, and this just returns a non-recording span. None of the attributes ever get added in my testing. The only one I ever see is gen_ai.conversation.id, but that is added by the auto-instrumentation packages as well. You can tell because it does not have a typo in it like the code here does.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
dhaifley: These are attached to the otel context. And removed after they should no longer be included for any spans running with that context. The processor just makes sure that any spans created for that otel context get the attributes until this clear_kagent_span_attributes() is called. The best way to do this would probably be with a try/finally block.

krisztianfekete: Yep, a try/finally block was what I had in mind.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
krisztianfekete: Would wrapping `_handle_request` into a `start_as_current_span()` make everything work?

dhaifley: That was my original thought too. But, when I attempted to use start_as_current_span() to make _handle_request have it's own span, that span was not only not becoming the parent of any child spans created, but actually, I do not see that span get emitted at all. So, something more than just using trace.start_as_current_span() here is needed to get this to work correctly. I am still taking a look at this.

But also, that would only result in the spans we are currently using being a child of a span containing these attributes we are setting. The child spans would not have these attributes themselves.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
dhaifley: I have added this.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
krisztianfekete: > actually, I do not see that span get emitted at all

Almost like we are not even configuring our trace_provider. Which might be the case here?

dhaifley: I have modified this to now just use a parent span. The span (see L169), has an instrumentation scope name of "kagent-adk" and a span name of "kagent_a2a_request". This span shows up in my testing, with the custom attributes as the parent span for all subsequent spans related to this request. Unlike the span processor though, this will propagate these attributes to any child spans created. This way is very simple.

dhaifley: I also updated the span name to be execute and move the span creation to a place that can be consistent between kagent-adk, kagent-langgraph, and kagent-crewai.

krisztianfekete: We had a discussion offline, and the direction we will take is not creating a framework-specific root level span (as it's redundant information) but leverage a span processor to make sure the required attributes can be inherited at child span level.

PR #1115, opened 2025-11-20
automatically configure logging/log level when the module loads

peterj:


python/packages/kagent-core/src/kagent/core/__init__.py
```diff
@@ -1,4 +1,36 @@
...
+import os
+
 from ._config import KAgentConfig
 from .tracing import configure as configure_tracing
 
-__all__ = ["KAgentConfig", "configure_tracing"]
+_logging_configured = False
+
+
+def configure_logging() -> None:
```
EItanya: For consistencies sake should this new logic live in `_logging.py`?

EItanya (approved):

PR #1121, opened 2025-11-20
Restart Agents Automatically On Secret Updates

### LINKED ISSUE #1091: Automatic agent redeployment when TLS secret changes
## Problem

Currently, agents do not automatically redeploy when the Kubernetes Secrets referenced in their ModelConfig's TLS configuration change. This means that if a TLS certificate is rotated or updated in a Secret, agents will continue using the old certificate until they are manually redeployed.

This creates operational overhead and potential security issues:
- Certificate rotations require manual intervention
- Agents may fail to connect if old certificates expire
- No automatic propagation of updated CA certificates or client certificates

## Solution

Implement automatic agent redeployment when TLS secrets referenced in ModelConfig resources are updated.

### 1. Extend Secret Watcher in Controller

The controller already watches Secrets for API key changes. Extend this functionality to also monitor TLS-related secrets:

- Watch for changes to Secrets referenced in `ModelConfig.spec.tls.caCertSecretRef`
- Trigger reconciliation when these secrets are updated

### 2. Extend `findModelsUsingSecret` Function

Update the existing `findModelsUsingSecret` function to check TLS secret fields in addition to API key secrets:

```go
// Check both API key secrets and TLS secrets
if model.Spec.APIKeySecretRef == secretName || 
   (model.Spec.TLS != nil && model.Spec.TLS.CACertSecretRef == secretName) {
    // Add to list of models using this secret
}
```

### 3. Trigger Agent Redeployment

When a TLS secret changes:

1. **Option A - Pod Template Hash Update**: Update the agent deployment's pod template annotations with a hash of the secret content, forcing Kubernetes to roll out new pods
2. **Option B - ConfigMap/Secret Version Tracking**: Track secret versions in agent deployment annotations and update when secrets change
3. **Option C - Direct Pod Restart**: Delete existing agent pods to force recreation with updated secret mounts

**Recommended**: Option A (hash-based annotation) as it provides declarative state and proper rollout behavior.

### 4. Ensure Cascading Updates

Make sure the redeployment propagates correctly:
- ModelConfig changes should trigger updates to all dependent Agents
- Secret changes should trigger ModelConfig reconciliation
- ModelConfig reconciliation should trigger Agent reconciliation
- Agent reconciliation should update deployment with new hash/version

## Implementation Notes

- Follow the existing pattern used for API key secret watching
- Ensure reconciliation is idempotent
- Add unit tests for secret change detection
- Add E2E tests for automatic redeployment on secret update
- Consider rate limiting to avoid excessive redeployments
- Document the automatic redeployment behavior for users

## Reference

- PR #1059: feat: Add SSL/TLS Configuration Support to ModelConfig CRD
- Review Comment #11 from @inFocus7
- Related to examples/modelconfig-with-tls.yaml line 384

## Scope Note

This enhancement is **out of scope** for PR #1059. This issue is created for future implementation and tracking purposes.

inFocus7:
note: unsure where the affinity template updates came from, but they get generated with the gen makefile target. maybe from #1085, but surprised it's generating on my pr 3 weeks after the merge 🤔 

# Changes
- Hashing secret alongside config-hash annotation for agent pod, so when a referenced secret updates, it restarts
- Added `SecretHash` status on ModelConfig so that changes to underlying referenced secrets are propagated (resource version updates) to Agent reconciliation

image

With these changes… 
1. When a Secret updates, a ModelConfig will update its status to reflect new hash.
2. ModelConfig updates resource version
3. The agent watching over modelConfig sees resource update
4. Agent reconciles, updating the annotation on the pod.
5. Agent restarts, loading in new secrets

## Golden Test Changes - Notes

The outputs for golden test annotations have _not_ changed, because the annotation hash relies on the modelconfig status which has Secret updates (hash). Modelconfig needs to reconcile for status, and does not reconcile in test, so `byte{}` (no change) is written to the hash.

# Context

With the addition of TLS CA’s to ModelConfigs, it became apparent we’ll need a UX-friendly way for agents to update with the latest Secret (e.g. cert rotation, api key change) without requiring users to manually restart the agent.

Note: We can’t rely on dynamic volume mounting, as the ca cert is read on agent start so that it configures the cached client. The api key also needed a way for its update to propagate to the agent.

## Demo

_steps_

[agent restart validation steps.md](https://github.com/user-attachments/files/23664735/agent.restart.validation.steps.md)

_video_

https://github.com/user-attachments/assets/eca62fb4-2ca2-45eb-94ba-7dfd0db5244b

## Alternative Solutions

_feedback wanted_

### Per-Secret Status

Instead of hashing all secrets into a single final hash to store in the ModelConfig’s status, we could store a status per-Secret.

For example, the status would change from: ```yaml
status:
    […]
    SecretHash: XYZ
```

to something like
```yaml
status:
  […]
  Secrets:
    APIKey:
      Hash/Version: 123
    TLS:
      Hash/Version: 123
```

I avoided this in order to simplify status tracking, less wordy compared to adding a field-per-secret - especially if we expand on referenced secrets in the future. But this manner does provide a better way for users to track where changes occurred exactly, and could avoid needing to do any hashing by using each secret’s resource version for updates.

We would need to see _how_ we’d propagate this to the agent pod annotations: adding annotation-per-secret vs. doing a singular hash for the pod like we do for the status now.

### Avoiding Restart Requirement

We should be able to avoid the restart needed for agents to configure the secrets. For instance, right now we mount a Volume for the TLS CA, and we use its file to configure the client at start which is cached. We could remove the client caching so that updated data from volume mounts are caught and used.

Pros:
- Avoiding restart requirement
Cons:
- Not caching the client would have some performance impact as it would need to be recreated per-call (maybe not a big deal, but noteworthy)
- We won’t be able to do any validation checks like we do now on startup.

---

Resolves #1091

### DIFF
--- a/go/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/internal/controller/translator/agent/adk_api_translator.go
@@ -1189,12 +1189,11 @@ type modelDeploymentData struct {
 // Internal to translator – a unified deployment spec for any agent.
 type resolvedDeployment struct {
 	// Required concrete runtime properties
-	Image                 string
-	Cmd                   string // empty → no explicit command
-	Args                  []string
-	Port                  int32 // container port and Service port
-	ImagePullPolicy       corev1.PullPolicy
-	ModelConfigSecretHash string // hash of any secrets referenced by the model config to handle redeployment on secret changes
+	Image           string
+	Cmd             string // empty → no explicit command
+	Args            []string
+	Port            int32 // container port and Service port
+	ImagePullPolicy corev1.PullPolicy
 
 	// SharedDeploymentSpec merged
 	Replicas         *int32


inFocus7:
Ah golden test failures due to the new hash annotation. Makes sense, will fix.

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -151,7 +151,7 @@ func (a *adkApiTranslator) TranslateAgent(
 			DefaultInputModes:  []string{"text"},
 			DefaultOutputModes: []string{"text"},
 		}
-		return a.buildManifest(ctx, agent, dep, nil, agentCard)
+		return a.buildManifest(ctx, agent, dep, nil, agentCard, "") // TODO: should we pass secret hash on BYO?
```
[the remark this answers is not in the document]
inFocus7: (from my understanding) byo agents define their own agents in deployments, so we can't define their internal clients; there's no reason to compute secret hashes.

EItanya:
Why is this change necessary? The way that secrets are mounted into the agent pods is all via filesystem mounts, can't we just watch for updates on those files?

onematchfox:
> The way that secrets are mounted into the agent pods is all via filesystem mounts, can't we just watch for updates on those files?

This feels like the right way of doing this to me at least. A file watcher would need to be implemented within the agent itself; but also doesn't cater for all use cases - API keys are not mounted as files at least.

EItanya:
> > The way that secrets are mounted into the agent pods is all via filesystem mounts, can't we just watch for updates on those files?
> 
> This feels like the right way of doing this to me at least. A file watcher would need to be implemented within the agent itself; but also doesn't cater for all use cases - API keys are not mounted as files at least.

The more I thought about it the more I think you are correct actually, going to re-review this today

inFocus7:
(copied from earlier message)

They're similar, but the existing config hash annotation is the hashed `adk.AgentConfig` which doesn't include any model information (aside from model type).

It's why i explicitly added `agent` to parameters, since there's also "ModelConfig", and the generic "config" naming confused me for a second

Alternatives:
- I could create a new structure for this to also include secret data and hash it all together {AgentConfig, SecretData).
- I could expand the `AgentConfig` to add secret data, but this config defined in both go & python, so need to look into what this impacts overall

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -317,6 +317,8 @@ type ModelConfigSpec struct {
 type ModelConfigStatus struct {
 	Conditions         []metav1.Condition `json:"conditions"`
 	ObservedGeneration int64              `json:"observedGeneration"`
+	// The secret hash stores a hash of any secrets required by the model config (i.e. api key, tls cert) to ensure agents referencing this model config detect changes to these secrets and restart if necessary.
+	SecretHash string `json:"secretHash,omitempty"`
```
EItanya, line 321: Right now the main reconciler will run for secret changes anyway, so this isn't strictly necessary. Also, right now the agent reconciler doesn't wait for `ModelConfig` resources to be accepted, which would have to change in order for this to work. For now can we just move the Hash computation into the agent reconciler directly and then we can separate it later if we need to?

inFocus7: So prefer the change below:? 

**Current Updates**
1. Agent watches ModelConfigs
2. ModelConfig watches Secrets
3. ModelConfig reconciles if Secret in its spec.TLS/API updates 
4. ModelConfig updates Status using Secret hash (new)
5. Agent reconciles due to update of ModelConfig
6. Agent uses ModelConfig status for pod annotation (new)


"An Agent reconciles because its ModelConfig updates. A ModelConfig updates because a Secret it uses updates."

**New Implementation**
1. Agent watches ModelConfigs **and** Secrets (new)
    - Because the Agent is now also watching Secret, should we update its status with an error if a secret DNE? Or leave that for the ModelConfig resource?
2. Agent reconciles if a Secret in its spec.ModelConfig.TLS/API updates (new)
3. Agent uses calculated Secret hash for pod annotation (new)

"An Agent reconciles because a Secret in its ModelConfig updates."

EItanya: Ok, so I was mistaken about the current code. I assumed that the agent already watched secrets, but it does not, which is I guess part of the bug being fixed 😆. In which case the current code is correct, and this comment can be ignored I think.

PR #1132, opened 2025-11-25
Update controller manifests + add CI check to prevent drift

onematchfox:
Noticed yesterday that the controllers manifests/CRDs were out of date - although it seems like these have since been updated (at least somewhat) by #1111. This PR regenerates them once again and also adds a CI check to try prevent them from drifting. 

Side Note: I've noticed that the category `kagent` seems to be being constantly added/removed - not sure why this is happening but we should probably look into that.

### DIFF
--- a/go/api/v1alpha1/modelconfig_types.go
+++ b/go/api/v1alpha1/modelconfig_types.go
@@ -289,11 +289,10 @@ type ModelConfigStatus struct {
 }
 
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:categories=kagent
+// +kubebuilder:resource:categories=kagent,shortName=mc
 // +kubebuilder:subresource:status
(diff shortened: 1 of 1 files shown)


onematchfox:
> Side Note: I've noticed that the category kagent seems to be being constantly added/removed - not sure why this is happening but we should probably look into that.

Seems like my newly added CI check [is failing](https://github.com/kagent-dev/kagent/actions/runs/19661401118/job/56308343652) because when run on CI, the categories are being removed :thinking:

onematchfox:
> Seems like my newly added CI check [is failing](https://github.com/kagent-dev/kagent/actions/runs/19661401118/job/56308343652) because when run on CI, the categories are being removed 🤔

Fixed in https://github.com/kagent-dev/kagent/pull/1132/commits/c75ec38dffd8201edfbf7248da48f1845066851e

PR #1138, opened 2025-11-26
feat(controller): decouple A2A handler registration from controller reconcilation

onematchfox:
**Decided to split this out of https://github.com/kagent-dev/kagent/pull/1133 to try make review a little easier as it's a chunky commit that can live in isolation of the rest of the changes in that PR**

This change separates A2A handler registration from the main `Agent` controller reconciliation loop by introducing a dedicated `A2ARegistrar` that manages the A2A routing table independently from the main controller.

Currently, A2A handler registration is tightly coupled to the `Agent` controller's reconciliation loop, which performs the following operations:
1. Reconcile Kubernetes resources (Deployment, Service, etc.)
2. Store agent metadata in database
3. Register A2A handler in routing table
4. Update resource status

This coupling is problematic for a number of reasons:
1. Breaks horizontal scaling - with leader election enabled (required to prevent duplicate reconciliation), only the leader pod performs reconciliation and registers A2A handlers. When API requests hit non-leader replicas, they fail because those replicas lack the necessary handler registrations.
2. Could be argued that this violates separation of concerns - the controller handles both cluster resource management (its core responsibility) and API routing configuration (an orthogonal concern).
3. Makes future architectural changes (e.g., splitting API and control plane) unnecessarily complex.

This PR attempts to address those concerns ensuring that all controller replicas, when scaled, will maintain consistent A2A routing tables enabling transparent load balancing across replicas. A2A logic is also consolidated into a dedicated package rather than scattered across controller code ensuring a clean separation of API and control plane such that these could be split into independent deployments without significant refactoring in future.

### DIFF
--- a/go/internal/a2a/a2a_registrar.go
+++ b/go/internal/a2a/a2a_registrar.go
@@ -122,10 +122,7 @@ func (a *A2ARegistrar) Start(ctx context.Context) error {
 
 func (a *A2ARegistrar) upsertAgentHandler(ctx context.Context, agent *v1alpha2.Agent, log logr.Logger) error {
 	agentRef := types.NamespacedName{Namespace: agent.GetNamespace(), Name: agent.GetName()}
-	card, err := a.translator.TranslateAgentCard(ctx, agent)
-	if err != nil {
-		return fmt.Errorf("translate agent %s: %w", agentRef, err)
-	}
+	card := agent_translator.GetA2AAgentCard(agent)
 
 	client, err := a2aclient.NewA2AClient(
 		card.URL,
--- a/go/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/internal/controller/translator/agent/adk_api_translator.go
@@ -68,10 +68,6 @@ type AdkApiTranslator interface {
 		ctx context.Context,
 		agent *v1alpha2.Agent,
 	) (*AgentOutputs, error)
-	TranslateAgentCard(
-		ctx context.Context,
-		agent *v1alpha2.Agent,
-	) (*server.AgentCard, error)
 	GetOwnedResourceTypes() []client.Object
 }
 
@@ -149,16 +145,9 @@ func (a *adkApiTranslator) TranslateAgent(
 		return nil, fmt.Errorf("unknown agent type: %s", agent.Spec.Type)
 	}
 
-	card := a.buildAgentCard(agent)
+	card := GetA2AAgentCard(agent)
 
 	return a.buildManifest(ctx, agent, dep, cfg, card, secretHashBytes)
-}
-
-func (a *adkApiTranslator) TranslateAgentCard(
-	ctx context.Context,
-	agent *v1alpha2.Agent,
-) (*server.AgentCard, error) {
-	return a.buildAgentCard(agent), nil
 }
 
 // GetOwnedResourceTypes returns all the resource types that may be created for an agent.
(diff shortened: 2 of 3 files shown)


go/internal/a2a/a2a_registrar.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if _, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if agent, ok := obj.(*v1alpha2.Agent); ok {
+				if err := a.upsertAgentHandler(ctx, agent, log); err != nil {
+					log.Error(err, "failed to upsert A2A handler", "agent", common.GetObjectRef(agent))
+				}
+			}
+		},
```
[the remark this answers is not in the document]
onematchfox: Current behaviour is desired. Context is only used when registering the handler (not in the handler implementation). Event handlers should fail immediately if triggered during shutdown - no new handlers should be registered.

go/internal/a2a/a2a_registrar.go
```diff
@@ -0,0 +1,171 @@
...
+			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+				var zeroDialer net.Dialer
+				return zeroDialer.DialContext(ctx, network, debugAddr)
+			},
+		}
+		return a2aclient.WithHTTPClient(client)
+	} else {
+		return func(*a2aclient.A2AClient) {}
+	}
+}
```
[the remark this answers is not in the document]
onematchfox: Weak excuse... Replacing existing code that does not have tests :sweat:

go/internal/a2a/a2a_registrar.go
```diff
@@ -0,0 +1,171 @@
...
+			),
+		)...,
+	)
+	if err != nil {
+		return fmt.Errorf("create A2A client for %s: %w", agentRef, err)
+	}
+
+	cardCopy := card
+	cardCopy.URL = fmt.Sprintf("%s/%s/", a.a2aBaseUrl, agentRef)
+
```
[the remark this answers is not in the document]
onematchfox: A2A client does not implement Close

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -178,6 +180,29 @@ func (r *adkApiTranslator) GetOwnedResourceTypes() []client.Object {
 	return ownedResources
 }
 
+func (a *adkApiTranslator) buildAgentCard(agent *v1alpha2.Agent) *server.AgentCard {
```
EItanya: Why is this an interface method and not just a public function somewhere? It doesn't seem to use any items on the struct.

go/internal/a2a/a2a_registrar.go
```diff
@@ -0,0 +1,171 @@
...
+func (a *A2ARegistrar) Start(ctx context.Context) error {
+	log := ctrllog.FromContext(ctx).WithName("a2a-registrar")
+
+	informer, err := a.cache.GetInformer(ctx, &v1alpha2.Agent{})
+	if err != nil {
+		return fmt.Errorf("failed to get cache informer: %w", err)
+	}
+
+	if _, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
EItanya, line 75: Is there a reason you used the cache directly here instead of creating a `Controller` like the rest of the k8s watchers? I usually prefer consistency across the various watchers so the codebase is easier to grok, but definitely open to this if there's a good reason.

go/internal/controller/translator/agent/adk_api_translator.go
onematchfox: Good point. This method specifically could just be a private helper function. However as for "a public function somewhere" - I would say this functionality does belong in this file (as it does so already today) and should be exposed via the `AdkApiTranslator` since it translates from `v1alpha2.Agent` to `server.AgentCard`.

Happy to refactor - can easily make this a private helper function if thats what you were getting at. Otherwise what would you suggest?

go/internal/a2a/a2a_registrar.go
onematchfox: Yeah, I debated this a bit as well. I ended up just an informer implementation for a couple reasons.

1. `Controllers` come with a bunch of overhead that I think is unnecessary for this implementation. We don't need the reconciliation semantics that controllers are designed for. The registrar doesn't need to loop and update to conmverge on desired state; it just needs to react to add/update/delete events to maintain an in-memory routing table. Additionally we don't need all the other overhead that comes with a controller including predicates, owning/watching relationships, woirk queues, rate limiting, retries, etc.
2. I wanted to explicitly break from the existing controllers. This code needs to run on all controller replicas, hence the `A2ARegistrar` deliberately, and explicilty, returns `false` from `NeedLeaderElection()` - rather than making this [configurable](https://github.com/kagent-dev/kagent/blob/d2cdd9dc7489b7ddc46b43c4e7508b39118a4165/go/pkg/app/app.go#L293) when using a `Controller`.

Finally, looking at it another way... if we were in future going to try and extract a `kagent-api` component, then it would feel very weird to me to be running a "`Controller`" within that component - but maybe that's just me :laughing:

onematchfox: Also.... having 2 `AgentController`s in the same pod also feels really strange (and wrong).

EItanya: I think these are all very good reasons!

go/internal/controller/translator/agent/adk_api_translator.go
EItanya: In my opinion, methods are important when you want to either hide the implementation of an interface, or you need to use some long lived item in a struct. In this case it's neither so really it should just be a public utility function

onematchfox: Refactored in https://github.com/kagent-dev/kagent/pull/1138/commits/8f51f114f4080d5bd4114bc94966fa7ecbfb90f7

PR #1139, opened 2025-11-26
Use environment variables for controller configuration

onematchfox:
**Yet another PR split out from #1133 to try reduce review burden** - keeping that one open for now as all of these other PRs are ultimately working towards that goal.

This PR refactors the kagent controller to support the use of environment variables for configuration in addition to command-line arguments. It also updates the Helm chart to make use of env vars instead of command line args and adds the ability for user's to supply their own environment variables with custom configuration. This allows users to supply sensitive configuration (e.g. postgres database url) via secrets instead of exposing these via `args`. Env vars are also easier to patch when working with rendered manifests if needed.

### DIFF
--- a/helm/kagent/templates/controller-deployment.yaml
+++ b/helm/kagent/templates/controller-deployment.yaml
@@ -58,6 +58,9 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "kagent.fullname" . }}-controller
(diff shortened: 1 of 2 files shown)


onematchfox:
> There's now a merge conflict, but overall I approve of the direction and will approve once that's sorted

Fixed.

EItanya (approved):

PR #1146, opened 2025-12-02
Enable leader election on controller when scaled

onematchfox:
This PR enables leader election on the controller if it is configured with one than 1 replica to ensure that only 1 replica is actively reconciling watched manifests. It also ensures that the necessary RBAC manifests are created.

Final part of #1133 (excluding #1138).

### DIFF
--- a/helm/kagent/templates/rbac/leader-election-rolebinding.yaml
+++ b/helm/kagent/templates/rbac/leader-election-rolebinding.yaml
@@ -1,0 +1,17 @@
+{{- if eq (include "kagent.leaderElectionEnabled" .) "true" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "kagent.fullname" . }}-leader-election-rolebinding
+  namespace: {{ include "kagent.namespace" . }}
+  labels:
+    {{- include "kagent.controller.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ include "kagent.fullname" . }}-leader-election-role
+subjects:
+- kind: ServiceAccount
+  name: {{ include "kagent.fullname" . }}-controller
+  namespace: {{ include "kagent.namespace" . }}
+{{- end }}
--- a/helm/kagent/templates/rbac/leader-election-role.yaml
+++ b/helm/kagent/templates/rbac/leader-election-role.yaml
@@ -1,0 +1,28 @@
+{{- if eq (include "kagent.leaderElectionEnabled" .) "true" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "kagent.fullname" . }}-leader-election-role
+  namespace: {{ include "kagent.namespace" . }}
+  labels:
+    {{- include "kagent.controller.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+{{- end }}


helm/kagent/templates/rbac/leader-election-rolebinding.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if eq (include "kagent.leaderElectionEnabled" .) "true" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "kagent.fullname" . }}-leader-election-rolebinding
+  namespace: {{ include "kagent.namespace" . }}
+  labels:
+    {{- include "kagent.controller.labels" . | nindent 4 }}
+roleRef:
```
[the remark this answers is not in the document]
onematchfox: Hallucination

helm/kagent/templates/controller-configmap.yaml
```diff
@@ -15,6 +15,7 @@ data:
   IMAGE_REGISTRY: {{ .Values.controller.agentImage.registry | default .Values.registry  | quote }}
   IMAGE_REPOSITORY: {{ .Values.controller.agentImage.repository | quote }}
   IMAGE_TAG: {{ coalesce .Values.controller.agentImage.tag .Values.tag .Chart.Version | quote }}
+  LEADER_ELECT: {{ include "kagent.leaderElectionEnabled" . | quote }}
```
EItanya, line 18: Nit: `LEADER_ELECTION_ENABLED`

helm/kagent/templates/_helpers.tpl
```diff
@@ -107,6 +107,13 @@ Engine labels
 app.kubernetes.io/component: engine
 {{- end }}
 
+{{/*
+Check if leader election should be enabled (more than 1 replica)
+*/}}
+{{- define "kagent.leaderElectionEnabled" -}}
+{{- gt (.Values.controller.replicas | int) 1 -}}
+{{- end -}}
```
EItanya, line 115: Relying on replicas feels potentially a tad dangerous to me. In the future we could potentially use an HPA or something like that. Maybe let's just set it to `true` by default?

helm/kagent/templates/controller-configmap.yaml
onematchfox: Needs to match the flag in the controller https://github.com/kagent-dev/kagent/blob/main/go%2Fpkg%2Fapp%2Fapp.go#L127

helm/kagent/templates/_helpers.tpl
onematchfox: Not sure how dangerous having this configured based on replicas really is... And, I suspect the underlying concern/issue may lie in the fact that the controller is currently responsible for both resource reconiliation and serving the API. 

From a controller perspective, leader election is only needed when there's more than 1 controller pod - to prevent duplicate reconiliation of resources. As for HPA, controllers are _never_[^1] scaled based on HPA. 

However, in the case of kagent I acknowledge that that might not hold true given that the API is exposed via the controller. I would hope however that if it gets to the point of really needing to scale out/in the API, we'd consider splitting this out into it's own deployment instead.[^2]

> Maybe let's just set it to true by default?

Totally OK with adding this as an explicit configuration that defaults to true if that's what you prefer. But I would also then add another error that checks that someone doesn't disable this when replicas > 1. 

Also, just so you are aware, enabling this with only 1 replica will add a short-delay when rolling out updates since the new pod will only become the leader after a short hand-off delay. WDYT?

[^1]: I challenge you to find me an example in the wild - would love to see how they dynamically shard ownership of resources.
[^2]: Even in the case of HPA introduced on the controller, this could be changed to activate based on `maxReplicas` instead.

helm/kagent/templates/controller-configmap.yaml
EItanya: Fair, maybe at some point in the future we can clarify

helm/kagent/templates/_helpers.tpl
EItanya: Ok I buy all of this, the reality is that the A2A gateway should probably be decoupled from the controller at some point, so this is a good first pass :)

EItanya (approved):

PR #1161, opened 2025-12-07
feat: Enable configurable streaming support for OpenAI models

### LINKED ISSUE #1099: Agent requests OpenAI-compatible modelservice with stream set to false
Hi,
I noticed that when using the agent to call an OpenAI-compatible model service, the http request always includes "stream": false.
I’d like to ask whether this parameter can be configured or overridden. Currently, it seems that the agent doesn’t provide an option to enable streaming responses (i.e., "stream": true").

syedazeez337:
Description:

This PR addresses the issue where requests to OpenAI-compatible model
services were hardcoded with "stream": false.

I have introduced a new stream configuration field to the Agent API. This gives
  developers explicit control over whether they want streaming responses enabled or
disabled for their agents.

  Key Changes:
```
   - API Update: Added an optional stream field (boolean) to the Agent spec in both
     v1alpha1 and v1alpha2.
   - Default Behavior: To ensure backward compatibility and safety, the field defaults
     to false if not specified.
   - ADK Integration: Updated the Python ADK and the translator logic to propagate
     this setting all the way down to the model runner. When enabled, it correctly
     sets the internal state to StreamingMode.SSE.
   - Documentation: Updated the Helm chart descriptions to reflect the new field.
```
  Testing:
```
   - Added new Python unit tests to verify that the boolean flag correctly maps to the
     StreamingMode enum.
   - Verified that existing OpenAI model integration tests pass.
   - Updated the Go "golden files" to match the new API schema.
```
  Everything has been linted and verified against the contribution guidelines. Let me
  know if you have any questions
  
  Fixes #1099

### DIFF
--- a/go/api/v1alpha1/agent_types.go
+++ b/go/api/v1alpha1/agent_types.go
@@ -37,7 +37,7 @@ type AgentSpec struct {
 	// +optional
 	ModelConfig string `json:"modelConfig,omitempty"`
 	// Whether to stream the response from the model.
-	// If not specified, the default value is true.
+	// If not specified, the default value is false.
 	// +optional
 	Stream *bool `json:"stream,omitempty"`
 	// +kubebuilder:validation:MaxItems=20
--- a/go/api/v1alpha2/agent_types.go
+++ b/go/api/v1alpha2/agent_types.go
@@ -86,7 +86,7 @@ type DeclarativeAgentSpec struct {
 	// +optional
 	ModelConfig string `json:"modelConfig,omitempty"`
 	// Whether to stream the response from the model.
-	// If not specified, the default value is true.
+	// If not specified, the default value is false.
 	// +optional
 	Stream *bool `json:"stream,omitempty"`
 	// +kubebuilder:validation:MaxItems=20
--- a/go/config/crd/bases/kagent.dev_agents.yaml
+++ b/go/config/crd/bases/kagent.dev_agents.yaml
@@ -2204,7 +2204,7 @@ spec:
               stream:
                 description: |-
                   Whether to stream the response from the model.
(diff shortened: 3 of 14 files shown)


syedazeez337:
Hi @EItanya, I have tested and passed the e2e tests locally. Let me know if it is working now. Thank you.

supreme-gg-gg:
> Thanks so much for the contribution, and this change definitely makes sense. The one thing i'd like to note is that I actually originally tried to make `stream: true` the default, but ran into issues. Have you done e2e testing with `stream = true`?

@EItanya was this also the issue you ran into when trying streaming?

syedazeez337:
Hi @supreme-gg-gg  I have reviewed your changes in jetc/fix-streaming - great work on the streaming handler and partial event filtering. You have covered almost all of the issue. I would like to propose where collaboration is needed, since you are one of the maintainers as well, let me know in what capacity I can be helpful here. Thank you.

supreme-gg-gg:
> Hi @supreme-gg-gg I have reviewed your changes in jetc/fix-streaming - great work on the streaming handler and partial event filtering. You have covered almost all of the issue. I would like to propose where collaboration is needed, since you are one of the maintainers as well, let me know in what capacity I can be helpful here. Thank you.

How about I'll open a PR from my branch since this branch is somewhat out of date as well and I'll add you as coauthor? there are some work that needs to be finished, like updating types in go, e2e tests with streaming (that requires new mockLLM though), perhaps making streaming false by default (the original behaviour), etc.

syedazeez337:
Hi @supreme-gg-gg Let's do that

EItanya:
This sounds like a great plan, @syedazeez337 would you mind closing this PR as it is now superseded by #1202

PR #1172, opened 2025-12-12
Fix npm audit issue

Charlesthebird:
Fixes https://github.com/advisories/GHSA-w37m-7fhw-fmv9

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -33,7 +33,7 @@
         "cmdk": "^1.1.1",
         "date-fns": "^4.1.0",
         "lucide-react": "^0.539.0",
-        "next": "^15.5.8",
+        "next": "^15.5.9",
         "next-themes": "^0.4.6",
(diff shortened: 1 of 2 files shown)


reneleonhardt:
Glad that security fixes have been released!

Does "Fix npm audit issue" mean that `npm audit` wasn't working correctly?

I can only find that images are being scanned once per week, wouldn't a daily `npm audit` help to reduce the response time (like in this case 4 weeks)?
https://github.com/kagent-dev/kagent/blob/main/.github/workflows/image-scan.yaml#L6

And the fix implemented here (bumping dependencies) could be created by Dependabot automatically instead of waiting for contributors:
https://docs.github.com/en/code-security/dependabot/dependabot-security-updates

In addition, Dependabot could automatically update the outdated GitHub Actions and maybe even the Dockerfiles (although Renovatebot offers much better options).

PR #1174, opened 2025-12-16
allow rbac rolebindings to be disabled

tomhaynes:
allow rbac rolebindings to be disabled
- this allows for external configuration of these, for instance for more locked-down permissioning

### DIFF
--- a/helm/kagent/templates/rbac/clusterrole.yaml
+++ b/helm/kagent/templates/rbac/clusterrole.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.clusterRoleEnabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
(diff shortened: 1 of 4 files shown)


dimetron:
if .Values.rbac.clusterRoleEnabled -> can we create namespace level rbac instead ?

tomhaynes:
@dimetron yes this is what I'm doing locally.. with this I thought it might be easier to just disable rbac management rather than trying to pre-suppose how others want it configured.

EItanya:
> @dimetron yes this is what I'm doing locally.. with this I thought it might be easier to just disable rbac management rather than trying to pre-suppose how others want it configured.

So I think it may be helpful for us to understand how you're configuring Namespaced RBAC, and then maybe we can design a system for that as a complement to cluster scoped.

PR #1177, opened 2025-12-23
Passthrough UserId for open ai requests

kevin-shelaga:


python/packages/kagent-adk/src/kagent/adk/models/_openai.py
```diff
@@ -394,6 +395,11 @@ async def generate_content_async(
             "messages": messages,
         }
 
+        # Add user parameter for usage tracking if user_id is available in context
+        user_id = get_user_id()
```
supreme-gg-gg, line 399: It seems like the chat completions API is deprecating the user's field https://platform.openai.com/docs/api-reference/chat

```
user: This field is being replaced by `safety_identifier` and `prompt_cache_key`. Use
              `prompt_cache_key` instead to maintain caching optimizations. A stable
              identifier for your end-users. Used to boost cache hit rates by better bucketing
              similar requests and to help OpenAI detect and prevent abuse.
              [Learn more](https://platform.openai.com/docs/guides/safety-best-practices#safety-identifiers).
```

PR #1183, opened 2025-12-30
Add proxy support for kagent-adk

iplay88keys:
## Description

Add global Proxy configuration to allow for defining proxies for declarative `adk` agents.

This proxy configuration is used for:
- Agents as tools
- Services as MCP Tools
- MCPServer resources

Each of these proxied requests includes a header with the original destination hostname: `x-kagent-host`.

Notes:
- The proxy needs to be configured to route based on the `x-kagent-host` header.
- Traffic from controller -> agent, agent -> controller, and agent -> LLM are not included.
- Only declarative `adk` agents are configured with the proxies

#### Global Proxy Configuration
Configure proxy URLs globally via Helm values:

```
proxy:
  # Proxy URL for internally-built k8s URLs
  url: "http://proxy.kagent.svc.cluster.local:8080"
```

#### Proxy setup:
- A gateway proxy
- If defining the A2A proxy, the following need to be routed through the proxy:
  - Each agent referenced as a tool
- If defining the egress proxy, the following need to be routed through the proxy:
  - Each mcp server/tool

#### URL Rewriting & Host Header Support
- When a proxy is configured, target URLs are rewritten to use the proxy while preserving the original path
- Automatically sets the `x-kagent-host` header to the original target hostname (without port) for proxy routing purposes
- Original request paths are preserved when routing through proxies
- This is important when it comes to agent cards:
  - We configure the second agent as a tool with the proxy url and host header
  - The initial request to the agent (for the agent card) goes through the proxy
  - By default, `adk` uses the url returned in the agent card
  - This URL will be rewritten automatically to match the proxy with the original url as a host header

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/types.py
+++ b/python/packages/kagent-adk/src/kagent/adk/types.py
@@ -177,7 +177,6 @@ class AgentConfig(BaseModel):
                 if base_url and event_hooks:
                     client = httpx.AsyncClient(
                         timeout=timeout,
-                        trust_env=False,
                         headers=headers,
                         base_url=base_url,
                         event_hooks=event_hooks,
@@ -185,13 +184,11 @@ class AgentConfig(BaseModel):
                 elif headers:
                     client = httpx.AsyncClient(
                         timeout=timeout,
-                        trust_env=False,
                         headers=headers,
                     )
                 else:
                     client = httpx.AsyncClient(
                         timeout=timeout,
-                        trust_env=False,
                     )
 
                 remote_a2a_agent = RemoteA2aAgent(


go/internal/controller/translator/agent/testdata/outputs/agent_with_proxy.json
```diff
@@ -0,0 +1,311 @@
...
+    "skills": null,
+    "url": "http://agent-with-proxy.test:8080",
+    "version": ""
+  },
+  "config": {
+    "description": "",
+    "http_tools": [
+      {
+        "params": {
+          "headers": {
```
yuval-k, line 25: i don't think we want any visible changes in the agent card. ideally using a proxy should be internal to the agent

iplay88keys: The only place we are updating the proxy is in the configuration of what tools an agent has access to. The [event hook](https://github.com/kagent-dev/kagent/pull/1183/changes#diff-e189d8a3bff494aa64559465b66ae6ef4dede37407ab738d306e900baaaa4a17R152-R165) allows us to make all agent to agent request through the proxy even though the [agent card](https://github.com/kagent-dev/kagent/pull/1183/changes#diff-64c41d57781c7b89b063ddad8d4f067866535d956e02a93027d18ad27cac80dbR17) returns the non-proxied URL:

```
    "url": "http://agent-with-proxy.test:8080",
```

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -555,15 +559,36 @@ func (a *adkApiTranslator) translateInlineAgent(ctx context.Context, agent *v1al
...
+					if err != nil {
+						return nil, nil, nil, fmt.Errorf("failed to parse proxy URL %q: %w", a.globalProxyURL, err)
+					}
+					// Use proxy URL with original path
+					targetURL = fmt.Sprintf("%s://%s%s", proxyURLParsed.Scheme, proxyURLParsed.Host, originalURLParsed.Path)
+					// Set X-Kagent-Host header to original hostname (without port) for Gateway API routing
+					if headers == nil {
+						headers = make(map[string]string)
+					}
+					headers["X-Kagent-Host"] = originalURLParsed.Hostname()
```
EItanya: Please make this a constant, also why is it capitalized?

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -943,14 +989,35 @@ func (a *adkApiTranslator) translateStreamableHttpTool(ctx context.Context, tool
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse proxy URL %q: %w", proxyURL, err)
+		}
+		// Use proxy URL with original path
+		targetURL = fmt.Sprintf("%s://%s%s", proxyURLParsed.Scheme, proxyURLParsed.Host, originalURL.Path)
+		// Set X-Kagent-Host header to original hostname (without port) for Gateway API routing
+		if headers == nil {
+			headers = make(map[string]string)
+		}
+		headers["X-Kagent-Host"] = originalURL.Hostname()
```
EItanya: This block of logic is repeated 3 times, can we make it a function instead?

go/internal/controller/translator/agent/adk_translator_golden_test.go
```diff
@@ -74,20 +78,61 @@ func runGoldenTest(t *testing.T, inputFile, outputsDir, testName string, updateG
...
+						hostParts := strings.Split(hostPart, ":")
+						hostname := hostParts[0]
+						hostnameParts := strings.Split(hostname, ".")
+						if len(hostnameParts) == 2 {
+							namespacesSeen[hostnameParts[1]] = true
+						}
+					}
+				}
+			}
+		}
```
EItanya: Nit: This feels super hacky, is there a cleaner way we can do this?

python/packages/kagent-adk/src/kagent/adk/types.py
```diff
@@ -112,30 +112,64 @@ def to_agent(self, name: str, sts_integration: Optional[ADKTokenPropagationPlugi
...
+
+                            request.url = httpx.URL(new_url)
+                            # Preserve X-Kagent-Host header for Gateway API routing
+                            request.headers["X-Kagent-Host"] = target_host
+
+                        return rewrite_url_to_proxy
+
+                    client_kwargs["event_hooks"] = {"request": [make_rewrite_url_to_proxy(proxy_base, target_host)]}
+
+                client = httpx.AsyncClient(**client_kwargs)
```
EItanya: This logic sense, but its there any way we can rewrite it to not use arg expansion. I find that very hard to read, and makes type hints nearly impossible.

EItanya: Also, same comment about the `x-kagent-host`, please use a const or global var of some kind

go/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Absolutely, will do!

go/internal/controller/translator/agent/adk_translator_golden_test.go
iplay88keys: I will look into it

go/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Will do. When looking at examples of `x-` prefixed headers, I had found many capitalized. I don't have a preference either way, though, so I can make them lowercase.

Issue #1190, opened 2026-01-09
[BUG] kagent-tools may not have fullnameOverride default value


ricky9408:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

Seems like the default helm values of "kagent" chart has an unnecessary fullnameOverride.

I am facing error on v0.7.8.

My controller is logging error that it can't access to remote MCP server.

In the log, the controller was trying to access to `http://kagent-app-tools.kagent:8084/mcp\`

※My helm release name is `kagent-app`.

However, the list of service doesn't contain `kagent-app-tools`.

```bash
kubectl get svc -n kagent
```

```
NAME                                                 TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
kagent-app-controller                                ClusterIP   10.96.5.231     <none>        8083/TCP   3h24m
kagent-app-grafana-mcp                               ClusterIP   10.96.134.5     <none>        8000/TCP   3h24m
kagent-app-kmcp-controller-manager-metrics-service   ClusterIP   10.96.179.253   <none>        8443/TCP   3h16m
kagent-app-querydoc                                  ClusterIP   10.96.87.218    <none>        8080/TCP   3h24m
kagent-app-ui                                        ClusterIP   10.96.49.104    <none>        8080/TCP   3h24m
kagent-tools                                         ClusterIP   10.96.60.175    <none>        8084/TCP   3h24m
promql-agent                                         ClusterIP   10.96.199.27    <none>        8080/TCP   3h23m
```

Only `kagent-tools` doesn't contain the release name.

Then I found there was a default fullnameOverride [here](https://github.com/kagent-dev/kagent/blob/83af77bb25771d1962dfc15a5746628217582d00/helm/kagent/values.yaml#L216)

I reckon this default fullnameOverride is blocking the usage of custom release name.

So that I think it should be removed.

※Below is my custom values.yaml to deploy "kagent" with helm.

```yaml
providers:
  default: gemini
[...]

PR #1192, opened 2026-01-09
Fix kagent-tools fullnameOverride

ricky9408:
FIx this issue https://github.com/kagent-dev/kagent/issues/1190

### DIFF
--- a/helm/kagent/templates/rbac/clusterrolebinding.yaml
+++ b/helm/kagent/templates/rbac/clusterrolebinding.yaml
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ include "kagent.fullname" . }}-getter-rolebinding
+  name: {{ include "kagent.fullname" . }}-controller-getter-rolebinding
   labels:
     {{- include "kagent.labels" . | nindent 4 }}
 roleRef:
@@ -16,7 +16,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ include "kagent.fullname" . }}-writer-rolebinding
+  name: {{ include "kagent.fullname" . }}-controller-writer-rolebinding
   labels:
     {{- include "kagent.labels" . | nindent 4 }}
 roleRef:


EItanya:
IIRC this used to be necessary, but I can't for the life of me remember why, I will respond again if I can think of it 😆

ricky9408:
@EItanya 
Thanks for the response!

During the CI testing, 1 CI has failed.

```
Error: 2 errors occurred:
	* clusterrolebindings.rbac.authorization.k8s.io "kagent-getter-rolebinding" already exists
	* clusterrolebindings.rbac.authorization.k8s.io "kagent-writer-rolebinding" already exists


make: *** [Makefile:327: helm-install-provider] Error 1
Error: Process completed with exit code 2.
```

I investigated the CI logs and I found that both `kagent` chart and `kagent-tools` chart are creating same clusterrolebindings.

Reference: https://github.com/kagent-dev/tools/blob/main/helm/kagent-tools/templates/clusterrolebinding.yaml

So that I added [this commit](https://github.com/kagent-dev/kagent/pull/1192/commits/21c7f60297b49a45dac51f7db02661c480edfde0) for this conflict.

May I ask for this to be checked as well? Thanks.

ricky9408:
@EItanya 

The CI failed again.
However, previous problem was resolved.

Seems there are dockerhub rate limit exceeded issue.

May I ask for the retry of the failed job? Thanks.

Log:

```
bash ./scripts/kind/setup-kind.sh
Unable to find image 'registry:2' locally
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/library/registry/manifests/2": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Alibrary%2Fregistry%3Apull&service=registry.docker.io": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

Run 'docker run --help' for more information
make: *** [Makefile:180: create-kind-cluster] Error 125
```

EItanya:
> May I ask for the retry of the failed job? Thanks.

Re-running now

EItanya:
@onematchfox I'd love to get your eyes on this. It seems simple but it's the type of change with potential hiding dragons and you're an expert at seeing those

helm/kagent/values.yaml
```diff
@@ -213,7 +213,7 @@ kmcp:
 # ==============================================================================
 
 kagent-tools:
-  fullnameOverride: kagent-tools
```
onematchfox, line 216:
``` suggestion
  nameOverride: tools
```

Removing `fullnameOverride` will resulting in any resources created by the `kagent-tools` chart being created using just the release name supplied when installing this chart. See https://github.com/kagent-dev/tools/blob/9cc60665518f6c0d0bf11f10fbc07f220efa019f/helm/kagent-tools/templates/_helpers.tpl#L8-L21

I.e. in the default scenario this will result in thge deployment `kagent-tools` being deleted and a new deployment called `kagent` being created. To preserve the existing name whilst also honouring the release name supplied when installing this chart we should make use of `nameOverride` instead. This is how this is done for the [kmcp chart](https://github.com/kagent-dev/kagent/blob/7202591e282e5e8828d62ba815d20796069a11f1/helm/kagent/values.yaml#L204) as well. 

FYI: This is also why you ended up having to change the name of the cluster role bindings. So, you should be able to revert that change after making this one.

Issue #1173, opened 2025-12-12
[Feature Request] Add context window management and conversation summarization support


jeanlopezxyz:
## Summary

When using kagent with models that have limited context windows (e.g., `gpt-4o-mini` with 128k tokens), agents with many MCP tools can exceed the context limit, causing "Session terminated" errors.

Currently, kagent does not expose ADK's context management features through the Agent CRD.

## Problem

With an agent configured with ~40+ MCP tools, the context quickly exceeds limits:
- Each tool definition adds tokens to the context
- Conversation history accumulates
- No way to truncate, summarize, or manage context programmatically

Error observed:
```
context_length_exceeded: This model's maximum context length is 128000 tokens. 
However, your messages resulted in 155XXX tokens.
```

## Proposed Solution

Expose [ADK's context management capabilities](https://google.github.io/adk-docs/context/) through the Agent CRD:

### Option 1: Context Compression (Compaction)
ADK has [Context Compression](https://google.github.io/adk-docs/context/compaction/) that summarizes older events:

```yaml
apiVersion: kagent.dev/v1alpha2
kind: Agent
spec:
  declarative:
    contextCompression:
      enabled: true
      compactionInterval: 5  # Compress every 5 events
      overlapSize: 1
      summarizer:
        type: llm  # Use LlmEventSummarizer
        model: gpt-4o-mini
```

### Option 2: Context Caching
ADK has [Context Caching](https://google.github.io/adk-docs/context/caching/) for optimizing repeated requests:

```yaml
spec:
  declarative:
    contextCache:
      enabled: true
      minTokens: 1000
      ttlSeconds: 3600
      cacheIntervals: 10
```

### Option 3: Context Window Compression Config
From the [ADK API](https://google.github.io/adk-docs/api-reference/python/google-adk.html), there's `ContextWindowCompressionConfig`:

```yaml
spec:
  declarative:
    contextWindowCompression:
      enabled: true
      maxLength: 100000  # Max context length before compression
```

## Alternatives Considered

1. **Reduce tools**: Works but limits agent capabilities
2. **Use larger context models**: More expensive (gpt-4o vs gpt-4o-mini)
3. **Memory/RAG**: Currently not supported in ADK (v0.6 release notes confirm this)

## Additional Context

- ADK already implements these features, they just need to be exposed in the CRD
- This is critical for production deployments with complex agents
- ADK documentation: https://google.github.io/adk-docs/context/

## Environment

- kagent version: v0.7.6
[...]

PR #1198, opened 2026-01-11
add feature: config summarisation, context compression, caching

Insomniac2904:
adds feature for #1173.

example use case:
```
kind: Agent
metadata:
  name: storage-agent
spec:
  declarative:
    context:
      cache:
        enabled: true
        cacheIntervals: 5
        minTokens: 200
        ttlSeconds: 3600

      compression:
        enabled: true
        compactionInterval: 10
        overlapSize: 50
        summarizer:
          type: llm
          model: gpt-4o-mini

      windowCompression:
        enabled: true
        maxLength: 4096

```


peterj:
this only adds the configuration options to the resources - are the features actually implemented in the ADK library too?

Insomniac2904:
> this only adds the configuration options to the resources - are the features actually implemented in the ADK library too?

Hey @peterj you were right. I could not find the context config for golang, only for python. I have revert back the original changes and made the changes to types.py and _a2a.py files. Also i am not sure for the imports that needs to me made for `windowCompression` and thus i have not implemented it as of now. Also could you please point me to the right test file, so that i can check my changes.
Thanks.

dobesv:
This seems like a very important feature to have, otherwise long sessions with agents might be impossible (?).  Any idea what is needed to move this forward?  Can I help in any way?

dobesv:
Sorry for the AI slop here. I need to go over this a bit more in detail before it can be merge ready. I'd appreciate some feedback on the approach, though.  If I fix all the issues, would you accept it?

EItanya:
Closing this in favor of #1298

ronbutbul:
any updates on that matter?
(compression and caching feature)

can not really work like this currently
im hitting token limit very quickly cause every prompt history is being added to the context

dobesv:
Working on this in  https://github.com/kagent-dev/kagent/pull/1340 now

PR #1207, opened 2026-01-14
Chore: upgrade the aquasecurity/trivy-action action

felipefrocha:
Simple update for the actions


EItanya (commented):
Thanks for the update, you'll need to sign your commits to pass DCO

PR #1201, opened 2026-01-12
Feat/expose kagent agents in mcp

riturajFi:
<html><head></head><body><hr><p>solves #1160</p><hr><h2>Title</h2><p>Expose kagent agents as dynamic MCP tools over stdio (<code inline="">kagent mcp serve-agents</code>)</p><hr><h2>Summary</h2><p>This PR adds a new CLI subcommand, <strong><code inline="">kagent mcp serve-agents</code></strong>, which runs an <strong>MCP server over stdio</strong> and dynamically exposes <strong>kagent agents as MCP tools</strong>.</p><p>Each eligible kagent agent is surfaced as a discoverable MCP tool at runtime. MCP-capable clients (IDEs, agent frameworks, assistants) can list and invoke these tools without any static tool definitions or manual schema maintenance.</p><p>All execution continues to flow through the existing <strong>A2A (agent-to-agent) API</strong>, preserving current agent semantics, permissions, and behavior.</p><hr><h2>What This Enables</h2><ul><li><p>MCP clients can automatically discover kagent agents via <code inline="">tools/list</code></p></li><li><p>Agents can be invoked using standard MCP <code inline="">tools/call</code></p></li><li><p>No static MCP schemas need to be authored or updated</p></li><li><p>Agent additions/removals are reflected immediately at runtime</p></li></ul><p>This turns kagent into a <strong>dynamic MCP tool provider</strong> backed by real agents.</p><hr><h2>New CLI Command</h2><pre><code class="language-bash">kagent mcp serve-agents
</code></pre><h3>Behavior</h3><ul><li><p>Starts an MCP server over <strong>stdin/stdout</strong></p></li><li><p>Intended to be launched and managed by an MCP client</p></li><li><p>Dynamically registers one MCP tool per eligible kagent agent</p></li><li><p>Does <strong>not</strong> open network ports or run an HTTP server</p></li></ul><hr><h2>Tool Model</h2><h3>Tool Naming</h3><ul><li><p>One tool per agent</p></li><li><p>Tool name is a deterministic, MCP-safe identifier derived from the agent ID</p></li></ul><h3>Tool Description</h3><pre><code>kagent agent &lt;namespace&gt;/&lt;name&gt;
</code></pre><h3>Tool Inputs</h3>
Field | Type | Required | Description
-- | -- | -- | --
task | string | yes | User prompt / task sent to the agent
context_id | string | no | A2A context ID for conversational continuity
history_length | number | no | History length forwarded to A2A configuration

[...]

PR #1210, opened 2026-01-14
feat: expose a2a agents via mcp in controller http server and migrate to mcp go-sdk

supreme-gg-gg:
The previous PR #1201 adds a CLI option to serve A2A agents via stdio / http transport MCP server. This PR extends this feature to the Go controller's HTTP server by adding a new `/mcp` route and handler for Streamable HTTP server. The MCP bridge exposes the `list_agents` and `invoke_agent` tools. 

e2e test has been added using a MCP client. Manual verification was done using MCP inspector and MCP clients like Cursor as well.

This PR also migrates all usages of `mark3labs/mcp-go` to the new official `go-sdk` library from Anthropic for MCP protocol and server / client implementation. Most changes are direct and syntax-only. The only exception is that I cannot find a built-in option to set headers for transport interfaces (SSE/Streamable HTTP) in `go-sdk`, so I had to create a custom HTTP client to make sure these headers are set in `reconciler.go`

### DIFF
--- a/go/internal/controller/reconciler/reconciler.go
+++ b/go/internal/controller/reconciler/reconciler.go
@@ -690,7 +690,7 @@ func (a *kagentReconciler) createMcpTransport(ctx context.Context, s *v1alpha2.R
 // so we need to create a custom HTTP client that adds headers to all requests.
 func newHTTPClient(headers map[string]string) *http.Client {
 	if len(headers) == 0 {
-		return nil
+		return http.DefaultClient
 	}
 	return &http.Client{
 		Transport: &headerTransport{


go/internal/httpserver/server.go
```diff
@@ -35,6 +36,7 @@ const (
 	APIPathMemories        = "/api/memories"
 	APIPathNamespaces      = "/api/namespaces"
 	APIPathA2A             = "/api/a2a"
+	APIPathMCP             = "/api/mcp"
```
EItanya: Typically the path for streamable HTTP mcp servers is `/mcp`, any reason we can't do that?

go/internal/mcp/mcp_handler.go
```diff
@@ -0,0 +1,255 @@
...
+	var contextIDPtr *string
+	if prior, ok := h.contextBySessionAndAgent.Load(contextKey); ok {
+		if priorStr, ok := prior.(string); ok && priorStr != "" {
+			contextIDPtr = &priorStr
+		}
+	}
+
+	// Create A2A client
+	a2aURL := fmt.Sprintf("%s/%s/", h.a2aBaseURL, agentRef)
+	a2aClient, err := a2aclient.NewA2AClient(a2aURL, a2aclient.WithTimeout(30*time.Second))
```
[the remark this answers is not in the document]
EItanya: This is correct, we should have the same authentication as the `a2a_registrar`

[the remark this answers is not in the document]
EItanya: Does this change make sense to you?

go/internal/mcp/mcp_handler.go
```diff
@@ -0,0 +1,255 @@
...
+
+// MCPHandler handles MCP requests and bridges them to A2A endpoints
+type MCPHandler struct {
+	kubeClient    client.Client
+	a2aBaseURL    string
+	authenticator auth.AuthProvider
+	httpServer    *mcpserver.StreamableHTTPServer
+	lock          sync.RWMutex
+	// Map to store context IDs per session and agent
+	contextBySessionAndAgent sync.Map
```
EItanya: Might be worth adding a TODO that this will not work across replicas. We may be able to store any necessary data in an external store if need be. Not sure we have any data specific to MCP here though, since the A2A data is already stored.

supreme-gg-gg:
> Overally looks good to me, just some minor changes

Yes the changes make sense, I haven't had the time to update them will do that now

go/internal/mcp/mcp_handler.go
supreme-gg-gg: This map is only for session "routing" so the agent using this MCP tool will have proper conversation continuation with the A2A agent by tracking the MCP context ID and the A2A context ID.

supreme-gg-gg: I have removed this session caching map since it won't work across replicas. The same functionality is achieved by just returning the context ID to the MCP client and most agents using it are capable of remembering the proper context id to call the A2A agent if it wants to continue the conversation. This makes the code simpler and avoids needing storage on server side.

PR #1214, opened 2026-01-15
fix(controller): resolve mutex deadlock between agent and remotemcpserver reconcilers

joshadambell:
## Summary

Fixes controller reconciliation deadlock by removing the application-level `upsertLock` mutex. Database-level concurrency control (atomic upserts and transactions) now handles synchronization.

## Problem

The shared `upsertLock` mutex blocked all Agent reconciliations when RemoteMCPServer reconciliation performed slow network I/O while holding the lock.

## Solution

**Phase 1**: Reduced lock scope to exclude network I/O
**Phase 2**: Fixed Agent predicate to ensure Create events are always processed
**Phase 3**: Removed mutex entirely - database handles concurrency via:
- Atomic upserts: `INSERT ... ON CONFLICT DO UPDATE`
- Transactions: Atomic tool replacement in `RefreshToolsForServer()`

**Phase 4**: Documentation and tests
- Added architecture doc explaining concurrency model
- Added idempotence tests for `StoreAgent` and `StoreToolServer`
- Added IMPORTANT comment to `RefreshToolsForServer` about transaction scope

### DIFF
--- a/go/internal/controller/reconciler/reconciler_test.go
+++ b/go/internal/controller/reconciler/reconciler_test.go
@@ -207,4 +207,3 @@ func TestAgentIDConsistency(t *testing.T) {
 
 	assert.Equal(t, storeID, deleteID)
 }
-
--- a/go/internal/database/client_test.go
+++ b/go/internal/database/client_test.go
@@ -25,10 +25,10 @@ func TestConcurrentAgentUpserts(t *testing.T) {
 	// All goroutines upsert to the same agent ID - this tests conflict handling
 	agentID := "test-agent"
 
-	for i := 0; i < numGoroutines; i++ {
+	for i := range numGoroutines {
 		go func(goroutineID int) {
 			defer wg.Done()
-			for j := 0; j < numUpserts; j++ {
+			for j := range numUpserts {
 				agent := &Agent{
 					ID:   agentID,
 					Type: fmt.Sprintf("type-%d-%d", goroutineID, j),
@@ -63,10 +63,10 @@ func TestConcurrentToolServerUpserts(t *testing.T) {
 	serverName := "test-server"
 	groupKind := "RemoteMCPServer"
 
-	for i := 0; i < numGoroutines; i++ {
+	for i := range numGoroutines {
 		go func(goroutineID int) {
 			defer wg.Done()
-			for j := 0; j < numUpserts; j++ {
+			for j := range numUpserts {
 				toolServer := &ToolServer{
 					Name:        serverName,
 					GroupKind:   groupKind,
@@ -110,7 +110,7 @@ func TestConcurrentRefreshToolsForServer(t *testing.T) {
 	var wg sync.WaitGroup
 	wg.Add(numGoroutines)
 
-	for i := 0; i < numGoroutines; i++ {
+	for i := range numGoroutines {
 		go func(goroutineID int) {
 			defer wg.Done()
 			// Each goroutine refreshes with a different set of tools


joshadambell:
> Great catch on the predicate issues with agents! That's so weird, I'll need to dig into the impl of those predicates.
> 
> In terms of the lock, I hate to say it but I think we can actually get rid of those locks altogether. Those are a relic of an older impl which didn't have a DB. The DB should function as the atomic record for us.

@EItanya you are right. I completely removed the app level locks.

go/internal/database/client.go
```diff
@@ -261,59 +260,36 @@ func (c *clientImpl) ListToolsForServer(serverName string, groupKind string) ([]
...
-			existingTool.GroupKind = groupKind
-			existingTool.Description = tool.Description
-			err = save(c.db, &existingTool)
-			if err != nil {
-				return err
-			}
-		} else {
-			err = save(c.db, &Tool{
+	return c.db.Transaction(func(tx *gorm.DB) error {
+		// Delete all existing tools for this server in the transaction
```
EItanya, line 272: Will deleting ALL and then recreating them potentially change the ID/creation time? Not a big deal overall but just wondering if you thought about it?

EItanya (commented):
This looks freaking awesome! Just the one question about the ID/creation time

go/internal/database/client.go
joshadambell: From what I gather, the tools are saved with an ID of tool.Name and the CreatedAt/UpdatedAt timestamp isn't used for anything meaningful. So I think this Delete-Recreate is simpler and guarantees a clean state (no orphan tools left) even with the metadata lost.

EItanya:
The code is failing lint, can merge after that is fixed

onematchfox:
> Great catch on the predicate issues with agents! That's so weird, I'll need to dig into the impl of those predicates.

Was just going through this PR and I'm not sure I follow the need for the change to the predicates. At the very least this change seems highly speculative.

The implementation of the replaced predicates can be seen at:
- [`GeneratedChangedPredicate`](https://github.com/kubernetes-sigs/controller-runtime/blob/6422ed031ac07ae85c8bd234657e77ddf5cace6e/pkg/predicate/predicate.go#L213-L224)
- [`LabelChangedPredicate`](https://github.com/kubernetes-sigs/controller-runtime/blob/6422ed031ac07ae85c8bd234657e77ddf5cace6e/pkg/predicate/predicate.go#L279C12-L290)

and 
- `predicate.Or` can be found [here](https://github.com/kubernetes-sigs/controller-runtime/blob/6422ed031ac07ae85c8bd234657e77ddf5cace6e/pkg/predicate/predicate.go#L337-L380)
- the underlying `Create` implementation can be found [here](https://github.com/kubernetes-sigs/controller-runtime/blob/6422ed031ac07ae85c8bd234657e77ddf5cace6e/pkg/predicate/predicate.go#L78-L84)

I do not see how the justification for the change can be true :confused:

> // - For Update events, only processes if generation or labels changed
> // This fixes a bug where GenerationChangedPredicate combined with LabelChangedPredicate
> // could inconsistently filter out Create events for some agents

The only real difference between the implementations is that the [new implementation](https://github.com/kagent-dev/kagent/blob/5633597bb1682a2d60ef8eb9fa2c5293f4a79ae1/go/internal/controller/agent_controller.go#L361-L371) of "label changed" is less resilient than using [`maps.Equal`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.6:src/maps/maps.go;l=17) since it will compare the default zero value even if the key is not present. So, IIUC, now, if I remove a blank label `domain.com/foo: ""`, then this will not result in the agent being reconciled.

Perhaps this code should be reverted? My guess is that the other code in this PR correctly handled the "deadlock" - perhaps that was the cause of the missing reconciliation to begin with.

@EItanya

EItanya:
Yes I completely agree, I'll give @joshadambell the rest of today to respond and then I'll merge your fix

joshadambell:
@onematchfox is correct. I overshot on this PR and the predicate logic changes. I agree on the revert to original in that part of the code.

Issue #1043, opened 2025-10-22
[FEATURE] make max payload size configurable for a2a


kevin-shelaga:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add support for configuring the max payload size for a2a

### ❓ Problem Statement / Motivation

- When sending context over 7-8mb (context sometimes includes documents or images) an error can be seen in the agents logs 

WARNING:a2a.server.apps.jsonrpc.jsonrpc_app:Request Error (ID: req-1761157821573): Code=-32600, Message='Payload too large'

### 💡 Proposed Solution

Make this setting configurable for larger payloads

### 🔄 Alternatives Considered

None. Currently blocked. 

### 🎯 Affected Service(s)

Not Sure

### 📚 Additional Context

https://github.com/a2aproject/a2a-python/blob/9c21e1a68fe579f906272b18fd47932c9b871180/src/a2a/server/apps/jsonrpc/jsonrpc_app.py#L94

https://github.com/a2aproject/a2a-python/blob/9c21e1a68fe579f906272b18fd47932c9b871180/src/a2a/server/apps/jsonrpc/jsonrpc_app.py#L368

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

PR #1216, opened 2026-01-17
make a2a max payload size configurable

kevin-shelaga:
Add ability to make a2a max content length configurable.

Resolves https://github.com/kagent-dev/kagent/issues/1043

### DIFF
--- a/python/packages/kagent-core/src/kagent/core/a2a/_config.py
+++ b/python/packages/kagent-core/src/kagent/core/a2a/_config.py
@@ -5,6 +5,7 @@ import os
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


python/packages/kagent-core/src/kagent/core/a2a/_config.py
```diff
@@ -0,0 +1,51 @@
...
+    Environment variable:
+        A2A_MAX_CONTENT_LENGTH: Maximum payload size in bytes.
+                                Default: 10485760 (10MB, a2a-sdk default)
+                                Example: 52428800 (50MB)
+                                Set to "0" or "none" for unlimited.
+
+    Returns:
+        The max content length in bytes, or None for unlimited.
+    """
+    max_content_length_str = os.getenv("A2A_MAX_CONTENT_LENGTH")
```
EItanya: Nit: Can this be a "const".

PR #1217, opened 2026-01-18
add list of allowed headers to be passed to tool calls

kevin-shelaga:
Add new allowedHeaders option on agent mcp server config. Its an array of string (case insensitive) of headers to be passed along to tool calls.

This is useful for things outside of jwt, like passing in tenant id and user email for cloud provider calls.

### DIFF
--- a/go/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/internal/controller/translator/agent/adk_api_translator.go
@@ -1131,7 +1131,7 @@ func (a *adkApiTranslator) translateMCPServerTarget(ctx context.Context, agent *
 			return err
 		}
 
-		return a.translateRemoteMCPServerTarget(ctx, agent, remoteMcpServer, toolServer.ToolNames, toolServer.AllowedHeaders, agentHeaders, proxyURL)
+		return a.translateRemoteMCPServerTarget(ctx, agent, remoteMcpServer, toolServer, agentHeaders, proxyURL)
 
 	case schema.GroupKind{
 		Group: "",
@@ -1157,7 +1157,7 @@ func (a *adkApiTranslator) translateMCPServerTarget(ctx context.Context, agent *
 			proxyURL = a.globalProxyURL
 		}
 
-		return a.translateRemoteMCPServerTarget(ctx, agent, remoteMcpServer, toolServer.ToolNames, toolServer.AllowedHeaders, agentHeaders, proxyURL)
+		return a.translateRemoteMCPServerTarget(ctx, agent, remoteMcpServer, toolServer, agentHeaders, proxyURL)
 	case schema.GroupKind{
 		Group: "",
 		Kind:  "Service",
@@ -1180,7 +1180,7 @@ func (a *adkApiTranslator) translateMCPServerTarget(ctx context.Context, agent *
 			return err
 		}
 
-		return a.translateRemoteMCPServerTarget(ctx, agent, remoteMcpServer, toolServer.ToolNames, toolServer.AllowedHeaders, agentHeaders, proxyURL)
+		return a.translateRemoteMCPServerTarget(ctx, agent, remoteMcpServer, toolServer, agentHeaders, proxyURL)
 	default:
 		return fmt.Errorf("unknown tool server type: %s", gvk)
 	}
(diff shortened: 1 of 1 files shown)


python/packages/kagent-adk/src/kagent/adk/types.py
```diff
@@ -25,15 +26,61 @@
...
+        sts_header_provider: Optional STS header provider for token propagation.
+
+    Returns:
+        A header provider function, or None if no headers need to be propagated.
+    """
+    if not allowed_headers and not sts_header_provider:
+        return None
+
+    # Normalize header names to lowercase for case-insensitive matching
+    normalized_allowed = [h.lower() for h in allowed_headers] if allowed_headers else []
```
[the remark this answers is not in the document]
kevin-shelaga: To me this is acceptable, since users must opt in to select the headers they want to send.

go/internal/controller/translator/agent/adk_translator_golden_test.go
```diff
@@ -37,6 +37,10 @@ type TestInput struct {
 
 // TestGoldenAdkTranslator runs golden tests for the ADK API translator
 func TestGoldenAdkTranslator(t *testing.T) {
+	// Set OTEL env var for deterministic output since collectOtelEnvFromProcess()
+	// reads from os.Environ() and we need consistent golden file comparisons
+	t.Setenv("OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE", "delta")
```
EItanya: Why was this change necessary?

go/api/v1alpha2/agent_types.go
```diff
@@ -221,6 +221,12 @@ type McpServerTool struct {
 	// For a list of all the tools provided by the server,
 	// the client can query the status of the ToolServer object after it has been created
 	ToolNames []string `json:"toolNames,omitempty"`
+
+	// AllowedHeaders specifies which headers from the A2A request should be
```
EItanya, line 233: Add a comment that `Authorization` has different semantics and will be skipped here. Or maybe we should discuss that here 😆

EItanya: Also, in the go code you called this `PropogateHeaaders`, but here it's `AllowedHeaders`, why the difference?

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1110,7 +1110,7 @@ func (a *adkApiTranslator) translateMCPServerTarget(ctx context.Context, agent *
 
 		spec.HeadersFrom = append(spec.HeadersFrom, toolHeaders...)
 
-		return a.translateRemoteMCPServerTarget(ctx, agent, agentNamespace, spec, toolServer.ToolNames, proxyURL)
+		return a.translateRemoteMCPServerTarget(ctx, agent, agentNamespace, spec, toolServer.ToolNames, proxyURL, toolServer.AllowedHeaders)
```
EItanya: Now that we have multiple `toolServer` fields can we just pass in a pointer to the struct.

python/packages/kagent-adk/src/kagent/adk/types.py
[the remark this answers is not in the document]
EItanya: As I mentioned above I think we should explicitly skip headers that we have semantics for like `Authorization`.

python/packages/kagent-adk/src/kagent/adk/types.py
```diff
@@ -25,15 +26,68 @@
...
+
+    def header_provider(readonly_context: Optional[ReadonlyContext]) -> dict[str, str]:
+        headers: dict[str, str] = {}
+
+        # Add allowed headers from session state first
+        if normalized_allowed and readonly_context:
+            request_headers = readonly_context.state.get(HEADERS_STATE_KEY, {})
```
EItanya, line 63: Is `headers` something set in the upstream library. I see you use it as a getter here, but wondering where they're set.

EItanya (commented):
LGTM overall, just a nit and a couple questions

go/internal/controller/translator/agent/adk_translator_golden_test.go
kevin-shelaga: I believe the golden tests were non-deterministic in how they wrote the otel env var. I attempted to address that while testing locally.

go/api/v1alpha2/agent_types.go
kevin-shelaga: This was because i started with sending everything then changed my mind, i will fix.

go/internal/controller/translator/agent/adk_translator_golden_test.go
EItanya: Very weird, I've never seen that. Is it possible that you have it set in your env or something?

python/packages/kagent-adk/src/kagent/adk/types.py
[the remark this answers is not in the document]
kevin-shelaga: i think if a user does auth a different way, then passes the auth header along with a jwt, they should be allowed to pass it through if its in the list. 

a comment pointing out that with sts the auth header is handled differently might be good enough?

python/packages/kagent-adk/src/kagent/adk/types.py
kevin-shelaga: https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/apps/jsonrpc/jsonrpc_app.py#L138

go/internal/controller/translator/agent/adk_translator_golden_test.go
kevin-shelaga: i think the issue is different developers may have different OTEL env vars set (or none at all).                                  
    - Developer A (has delta): test passes                                                                                                           
    - Developer B (has cumulative): test fails - golden mismatch                                                                                     
    - Developer C (no OTEL vars): test fails - missing env var in output

PR #1232, opened 2026-01-22
Fix for kagent controller locking up with bad mcp server config

tzhukov:
# Description

- **Motivation:**
  kagent-controller locks up when an MCPServer with bad configuration(ie MCP server is up but cannot communicate with the control plane) is applied. This blocks new agents passing from `Not Accepted` state into `Ready/Not Ready`

Error seen in kagent-controller: 
```
"msg":"Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes requeuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler"
```

- **What changed:**
  * Resolving reconcile function bug in MCPServerToolController that wasn`t allowing controller-runtime to hangle exponential backoff properly. 
  * Corrected error handling where error was logged but not returned, this was making MCPServer not registered in the tool server db, but reconcile appeared successful. '
  * Added explicit validation in validateAgent to check that MCP server reference exists and has valid port configurations before accepting agent.
  * Added testing for this behavior.

# Testing

To replicate the issue I applied this MCPServer + RemoteMCPServer manifest:
```
---
kind: MCPServer
metadata:
  name: example-mcp-server
  namespace: kagent
  labels:
    app: example-mcp-server
spec:
  deployment:
    image: alpine:latest
    port: 3001
    cmd: sh
    args:
      - "-c"
      - "while true; do sleep 3600; done"
  transportType: stdio
  stdioTransport: {}

---
kind: RemoteMCPServer
metadata:
  name: example-remote-mcp-server
  namespace: kagent
  labels:
    app: example-remote-mcp-server
spec:
  description: "Example remote MCP server"
  protocol: STREAMABLE_HTTP
  url: "http://example-mcp-server.kagent.svc.cluster.local:8081/mcp"
  timeout: 30s
  sseReadTimeout: 5m0s
  terminateOnClose: true
```

After this I tried creating a new agent and observed it being stuck in `Not Accepted` state and the error message mentioned in motivation. The same workflow after the PR change result in agents being accepted.

Instead we are appropriately see error logs about the new RemoteMCPServer:
```
"failed to upsert tool server for remote mcp server","remoteMCPServer":"kagent/example-remote-mcp-server","error":"failed to fetch tools for toolServer kagent /example-remote-mcp-server: failed to connect client for toolServer kagent/example-remote-mcp-server: calling \"initialize\": rejected by transport: sending \"initialize\":  Post \"http://example-mcp-server.kagent.svc.cluster.local: 8081/mcp\": dial tcp 10.96.47.128:8081: i/o timeout"
```

### DIFF
--- a/go/internal/controller/reconciler/reconciler.go
+++ b/go/internal/controller/reconciler/reconciler.go
@@ -164,7 +164,7 @@ func (a *kagentReconciler) reconcileAgentStatus(ctx context.Context, agent *v1al
 	if conditionChanged || agent.Status.ObservedGeneration != agent.Generation {
 		agent.Status.ObservedGeneration = agent.Generation
 		if err := a.kube.Status().Update(ctx, agent); err != nil {
-			return fmt.Errorf("failed to update agent status: %v", err)
+			return fmt.Errorf("failed to update agent status: %w", err)
 		}
 	}
 
@@ -189,7 +189,7 @@ func (a *kagentReconciler) ReconcileKagentMCPService(ctx context.Context, req ct
 			}
 			return nil
 		}
-		return fmt.Errorf("failed to get service %s: %v", req.Name, err)
+		return fmt.Errorf("failed to get service %s: %w", req.Name, err)
 	}
 
 	dbService := &database.ToolServer{
@@ -227,7 +227,7 @@ func (a *kagentReconciler) ReconcileKagentModelConfig(ctx context.Context, req c
 			return nil
 		}
 
-		return fmt.Errorf("failed to get model %s: %v", req.Name, err)
+		return fmt.Errorf("failed to get model %s: %w", req.Name, err)
 	}
 
 	var err error
@@ -337,7 +337,7 @@ func (a *kagentReconciler) reconcileModelConfigStatus(ctx context.Context, model
 	if conditionChanged || modelConfig.Status.ObservedGeneration != modelConfig.Generation || secretHashChanged {
 		modelConfig.Status.ObservedGeneration = modelConfig.Generation
 		if err := a.kube.Status().Update(ctx, modelConfig); err != nil {
-			return fmt.Errorf("failed to update model config status: %v", err)
+			return fmt.Errorf("failed to update model config status: %w", err)
 		}
 	}
 	return nil
@@ -361,7 +361,7 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
 			}
 			return nil
 		}
-		return fmt.Errorf("failed to get mcp server %s: %v", req.Name, err)
+		return fmt.Errorf("failed to get mcp server %s: %w", req.Name, err)
 	}
 
 	dbServer := &database.ToolServer{
@@ -413,7 +413,7 @@ func (a *kagentReconciler) ReconcileKagentRemoteMCPServer(ctx context.Context, r
 			return nil
 		}
 
-		return fmt.Errorf("failed to get remote mcp server %s: %v", serverRef, err)
+		return fmt.Errorf("failed to get remote mcp server %s: %w", serverRef, err)
 	}
 
 	dbServer := &database.ToolServer{
@@ -441,7 +441,7 @@ func (a *kagentReconciler) ReconcileKagentRemoteMCPServer(ctx context.Context, r
 		tools,
 		err,
 	); err != nil {
-		return fmt.Errorf("failed to reconcile remote mcp server status %s: %v", req.NamespacedName, err)
+		return fmt.Errorf("failed to reconcile remote mcp server status %s: %w", req.NamespacedName, err)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -190,32 +190,66 @@ func (a *adkApiTranslator) validateAgent(ctx context.Context, agent *v1alpha2.Ag
...
-		}
+			err = a.validateAgent(ctx, toolAgent, state.with(agent))
+			if err != nil {
+				return err
+			}
 
-		err = a.validateAgent(ctx, toolAgent, state.with(agent))
-		if err != nil {
-			return err
+		case v1alpha2.ToolProviderType_McpServer:
```
EItanya: Is this part actually necessary? later on in translation we return an error for this exact scenario. iirc the point of this function is really just to pre-check recursion depth so that we don't have to do the entire translation. and fail towards the end, but honestly it's a bit confusing and maybe we should just check the recursion depth during the actual translation. Either way I don't think this actually changes anything

tzhukov: Yeah, it's redundant here, we do it later again in `translateMCPServerTarget`. It does fail fast this way, but that doesn't really add any extra benefit, I'm going to remove it.

go/internal/controller/reconciler/reconciler.go
```diff
@@ -358,12 +373,36 @@ func (a *kagentReconciler) ReconcileKagentMCPServer(ctx context.Context, req ctr
...
+	if _, err := a.upsertToolServerForRemoteMCPServer(ctx, dbServer, remoteSpec, mcpServer.Namespace); err != nil {
+		// Transient error (database, network, etc.) - update status and retry
+		reconcileLog.Error(err, "failed to upsert tool server for mcp server", "mcpServer", utils.GetObjectRef(mcpServer))
+		status.ReconcileMCPServerStatus(ctx, a.kube, mcpServer, err)
+		return fmt.Errorf("failed to upsert tool server for remote mcp server %s: %v", utils.GetObjectRef(mcpServer), err)
+	}
+
+	// Success - update status
+	if _, err := status.ReconcileMCPServerStatus(ctx, a.kube, mcpServer, nil); err != nil {
+		return err
```
EItanya: We can't do this here, we don't own this resource, it's owned by `kmcp`

go/internal/controller/reconciler/reconciler.go
```diff
@@ -192,12 +193,26 @@ func (a *kagentReconciler) ReconcileKagentMCPService(ctx context.Context, req ct
...
-	if remoteService, err := agent_translator.ConvertServiceToRemoteMCPServer(service); err != nil {
-		reconcileLog.Error(err, "failed to convert service to remote mcp service", "service", utils.GetObjectRef(service))
-	} else {
-		if _, err := a.upsertToolServerForRemoteMCPServer(ctx, dbService, remoteService, service.Namespace); err != nil {
-			return fmt.Errorf("failed to upsert tool server for mcp service %s: %v", utils.GetObjectRef(service), err)
+	// Convert Service to RemoteMCPServer spec
+	remoteService, err := agent_translator.ConvertServiceToRemoteMCPServer(service)
+	if err != nil {
+		// Check if this is a validation error
+		var validationErr *agent_translator.ValidationError
```
EItanya: As I mentioned in that other comment, we can't update the status of this resource in `kagent` at all because this is owned by `kmcp`. Also, I think this error checking/logic should happen in the controller, not the reconciler.

go/internal/controller/service_controller.go
```diff
@@ -41,7 +43,21 @@ type ServiceController struct {
 
 func (r *ServiceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	_ = log.FromContext(ctx)
-	return ctrl.Result{}, r.Reconciler.ReconcileKagentMCPService(ctx, req)
+
```
tzhukov, line 47: Moved validation/error handling logic into controller

go/internal/controller/service_controller.go
```diff
@@ -41,7 +43,21 @@ type ServiceController struct {
...
+		if errors.As(err, &validationErr) {
+			// Validation error - don't retry until Service annotations/ports are fixed
+			// Return empty result with no error to avoid exponential backoff
+			return ctrl.Result{}, nil
+		}
+		// Transient error - return error to trigger exponential backoff retry
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
```
EItanya: Shouldn't this also have the `60s` backoff?

tzhukov: Yeah, it needs that

PR #1240, opened 2026-01-25
fix for Ollama ModelConfig options not propagating to runtime

### LINKED ISSUE #1218: [BUG] Ollama ModelConfig `options` (e.g. `num_ctx`) are defined but not propagated to runtime
### 📋 Prerequisites

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

The `OllamaConfig.Options` field in the `ModelConfig` CRD allows users to specify Ollama-specific parameters such as `num_ctx`, `temperature`, etc. However, these options are currently **not forwarded to the runtime or Ollama API calls**.

Only the `host` (via `OLLAMA_API_BASE`) and `defaultHeaders` are passed through the ADK translator. As a result, any values defined under `ollama.options` are silently ignored at runtime, despite being part of the public API schema.

This creates a mismatch between the CRD definition and actual behavior.

### 🔄 Steps To Reproduce

Define an Ollama ModelConfig Resource with the options spec:

```yaml
# ModelConfig for granite4:3b (for general tasks, not agents)
kind: ModelConfig
metadata:
  name: granite4-3b-ollama
  namespace: kagent
spec:
  model: "granite4:3b"
  provider: Ollama
  apiKeySecret: kagent-secrets
  apiKeySecretKey: litellm-api-key
  ollama:
    host: "http://ollama:11434"
    options:
      num_ctx: "128000" # bug with passing the options to runtime agent runner
```

### 🤔 Expected Behavior

When a user defines `ollama.options` in a `ModelConfig`, those values should be passed to the Ollama provider client (LiteLLM / HTTP request) so they affect model execution. This should result in the Ollama request being executed with the specified context size and sampling parameters.

## Actual Behavior

- The `options` map is accepted by the CRD and appears in test data
- The ADK translator **does not forward `options`**
- The Python runtime (LiteLLM) never receives `num_ctx` or other Ollama parameters
- Only keys passed through:
  - `OLLAMA_API_BASE`
  - `defaultHeaders`

  
This makes `ollama.options` effectively a no-op.

### 💻 Environment

- OS: TalosOS
- Kuberenets: 1.34.1
- kagent Helm Version: 0.7.8

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

[DeepWiki](https://deepwiki.com/search/modelconfig-for-usage-with-the_1128bc20-a383-420d-8e80-83710468c71e?mode=deep)

Relevant code paths:

- CRD defines `OllamaConfig.Options`  
  `go/api/v1alpha2/modelconfig_types.go`
- Test YAML includes `num_ctx`  
  `go/internal/controller/translator/agent/testdata/inputs/ollama_agent.yaml`
- Translator only sets `OLLAMA_API_BASE` and headers  
  `go/internal/controller/translator/agent/adk_api_translator.go`
- Python runtime initializes LiteLLM with headers only  
  `python/packages/kagent-adk/src/kagent/adk/types.py`

### 📋 Logs

```shell
Ollama:

time=2026-01-19T17:20:45.615Z level=DEBUG source=sched.go:290 msg="runner with non-zero duration has gone idle, adding timer" runner.name=registry.ollama.ai/library/granite4:3b runner.inference="[{ID:GPU-aafce598-4e49-4405-5ed6-7d8c9af80b95 Library:CUDA}]" runner.size="7.5 GiB" runner.vram="7.5 GiB" runner.parallel=1 runner.pid=1199 runner.model=/root/.ollama/models/blobs/sha256-6c02683809a8dc4eb05c78d44bc63bcd707703b078998fa58829c858ab337bb0 runner.num_ctx=40000 duration=5m0s


Manually set context via Ollama env variable to reach 40000
```

### 📷 Screenshots

_No response_

Insomniac2904:
fixes #1218 

changes made:

Modified 
go/internal/adk/types.go to add Options map[string]string to the Ollama struct and Updated MarshalJSON to include options in the JSON output.

Modified 
go/internal/controller/translator/agent/adk_api_translator.go translateModel to populate ollama.Options from the ModelConfig spec.
Python ADK

Modified 
python/packages/kagent-adk/src/kagent/adk/types.py  to include options and updated AgentConfig.to_agent to pass options as kwargs to the LiteLlm constructor.

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
@@ -185,7 +185,7 @@ class A2aAgentExecutor(AgentExecutor):
                     logger.error("Failed to publish failure event: %s", enqueue_error, exc_info=True)
         finally:
             clear_kagent_span_attributes(context_token)
-            # close the runner which cleans up the mcptoolsets 
+            # close the runner which cleans up the mcptoolsets
             # since the runner is created for each a2a request
             # and the mcptoolsets are not shared between requests
             # this is necessary to gracefully handle mcp toolset connections
--- a/python/packages/kagent-adk/src/kagent/adk/types.py
+++ b/python/packages/kagent-adk/src/kagent/adk/types.py
@@ -241,7 +241,9 @@ class AgentConfig(BaseModel):
         elif self.model.type == "gemini_anthropic":
             model = ClaudeLLM(model=self.model.model)
         elif self.model.type == "ollama":
-            model = LiteLlm(model=f"ollama_chat/{self.model.model}", extra_headers=extra_headers, **(self.model.options or {}))
+            model = LiteLlm(
+                model=f"ollama_chat/{self.model.model}", extra_headers=extra_headers, **(self.model.options or {})
+            )
         elif self.model.type == "azure_openai":
             model = OpenAIAzure(
                 model=self.model.model,


dimetron:
I see 2 issues

1. OLLAMA_API_BASE doesn't work without http - looks like regression because of some recent updates. Requires small fix in go/internal/controller/translator/agent/adk_api_translator.go
```go
		host := model.Spec.Ollama.Host
		if !strings.HasPrefix(host, "http://") && !strings.HasPrefix(host, "https://") {
			host = "http://" + host
		}
		modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
			Name:  "OLLAMA_API_BASE",
			Value: host,
		})
```

2. type conversion issue in options

Our values using string helm values:

```yaml
  ollama:
    provider: Ollama
    model: "llama3.2"
    config:
      host: host.docker.internal:11434
      options:
        num_ctx: "64000"
```

Ollama API expects sometimes integer or float https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion

```json
kagent/istio-agent
litellm.APIConnectionError: Ollama_chatException - {"error":"option "num_ctx" must be of type integer"}
```

I suggest to use official SDK and convert it when creating model using 

```python
def _convert_ollama_options(options: dict[str, str] | None) -> dict[str, Any]:
    """Convert Ollama options from string values to their correct types.

    Uses type annotations from the official Ollama Python SDK Options class
    to determine the correct type for each option.

    Since the options come from YAML/JSON config as strings, we need to convert them
    to the types expected by the Ollama API.
    """
    if not options:
        return {}

    # Import Ollama SDK Options class for type introspection
    from typing import get_args, get_origin

    from ollama import Options as OllamaOptions

    # Get type hints from the Ollama SDK Options class
    type_hints = OllamaOptions.model_fields

    converted: dict[str, Any] = {}
    for key, value in options.items():
        if key in type_hints:
            field_info = type_hints[key]
            # Get the annotation and unwrap Optional[T] -> T
            annotation = field_info.annotation
            origin = get_origin(annotation)
            if origin is Union:
                # Optional[T] is Union[T, None], get the non-None type
                args = [arg for arg in get_args(annotation) if arg is not type(None)]
                if args:
                    target_type = args[0]
                else:
                    target_type = str
            else:
                target_type = annotation

            # Convert based on the target type
            if target_type is int:
                converted[key] = int(value)
            elif target_type is float:
                converted[key] = float(value)
            elif target_type is bool:
                converted[key] = value.lower() == "true"
            else:
                # Keep as string for other types (e.g., Sequence[str])
                converted[key] = value
        else:
            # Unknown option - keep as string
            converted[key] = value

    return converted
```

see https://github.com/kagent-dev/kagent/pull/1245/commits/136167193dc474255398ac735b153c514aa09532

EItanya:
@dimetron took these commits and took them over the line in #1245

Issue #1249, opened 2026-01-27
[FEATURE] Provider abstraction with model discovery


nujragan93:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Introduce a **Provider abstraction with model discovery** in Kagent, allowing operators to configure multiple providers and users to select a provider and discover available models dynamically during model creation.

### Current Limitations / Pain Points

- Provider details are effectively static and managed through fixed configuration rather than being dynamically extensible.
- Users cannot choose between multiple providers (e.g., OpenAI, Bedrock, internal vLLM).
- Model creation today relies on a **fixed set of stock model options** exposed by the UI/CLI, rather than dynamically discovering models from the selected provider.

### Who Is Affected

- Platform / Infra teams managing Kagent as a shared service
- Application teams consuming models across different providers
- Operators who need RBAC, auditability, and safe multi-provider support

### Why This Is Needed

Without a provider abstraction:

- UX is poor and error-prone

### Option 1: Default Provider via ConfigMap and Kagent refering default apikey for models

Allow operators to define a default provider using a ConfigMap variable, with credentials supplied to the Kagent controller pod via a mounted Secret or environment variables, enabling Kagent to authenticate and fetch available models for simpler or single-provider setups.

```yaml
# Example ConfigMap for configuring LLM providers with model discovery
#
# This ConfigMap defines the providers available for model discovery.
# The controller reads this ConfigMap and exposes the providers via the
# /api/providers/configured endpoint. Models can be discovered dynamically
# via /api/providers/configured/{name}/models.
#
# Prerequisites:
# - Create Secrets for each provider's API key (see examples below)
#
# Usage:
#   kubectl apply -f kagent-providers-configmap.yaml
#   kubectl apply -f provider-secrets.yaml  # Create secrets first
#
apiVersion: v1
kind: ConfigMap
metadata:
  name: kagent-providers
  namespace: kagent
  labels:
    app.kubernetes.io/name: kagent
    app.kubernetes.io/component: provider-config
data:
  # YAML list of provider configurations
  # Each provider has:
  #   - name: unique identifier for this provider instance
  #   - type: provider type (OpenAI, Anthropic, AzureOpenAI, Ollama, Gemini, etc.)
[...]

PR #1251, opened 2026-01-27
split provider and Models in UI

nujragan93:
Image
This is an UI change to faciliate https://github.com/kagent-dev/kagent/issues/1249


nujragan93:
full feature implementation requested

PR #1252, opened 2026-01-27
propagate headers from the agent to the mcp tools

peterj:
HeaderCaptureMiddleware intercepts the incoming HTTP requests and copies headers into a context variable. The KAgentCallContextBuilder reads that and stores the headers in `ServerCallContext.state["headers"]`. 

Once the headers are in the context, the agent executor can read them and store them in the sessions state for any plugins to extract them/use them.


felipefrocha:
Why this PR was closed? Currently did something really similar to make it work with impersonation and it works really well

EItanya:
There's now an official way to do this here: https://github.com/kagent-dev/kagent/pull/1217

PR #1254, opened 2026-01-29
feat: support custom service account override in agent crds

### LINKED ISSUE #1233: [FEATURE] Support k8s Service Account overrides in Agent templates
### 📋 Prerequisites

### 📝 Feature Summary

Support using an existing or custom k8s Service Account in helm overrides in Agent templates

In order to allow agent deployments have required permissions on the cluster as well as through machine/workload-identities to cloud provider services, service accounts in kubernetes are used. Currently all agents as part of their deployment spawn with a k8s service account of their own. In certain cases there are platform level service accounts with all required permissions to external APIs etc. that need to be potentially used instead of doing a per agent per k8s-SA permission assignment. Currently the agent template specs do not allow to override the name of the service account to use an existing SA in the agent's deployment and pod. In previous versions before v0.6, patching the SA name in the agent's deployment used to be a workaround, but it is not possible anymore as the controller puts it back to the default service account name in the deployment and pod.

### 💡 Proposed Solution

Introduce the ability to pass in helm values overrides in the agent spec templates to choose an existing service account to use for the agent's deployment and pod so that permissions and identities for external APIs can be managed centrally if needed. Default can be same as for now, but the override ability can be introduced for the agent CRD or agent declarative spec, like done for many other parameters.

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

_No response_

supreme-gg-gg:
Set `serviceAccountName` to use existing SA instead of creating a new SA for each agent pod. If not provided, will default to agent name and create a new SA. Alternatively, provide `serviceAccountConfig` that will be applied to every SA created by the agent controller. Close #1233

### DIFF
--- a/go/config/crd/bases/kagent.dev_agents.yaml
+++ b/go/config/crd/bases/kagent.dev_agents.yaml
@@ -4092,6 +4092,11 @@ spec:
                             type: object
                         type: object
                       serviceAccountConfig:
+                        description: |-
+                          ServiceAccountConfig configures the ServiceAccount created by the Agent controller.
+                          This field can only be used when ServiceAccountName is not set.
+                          If ServiceAccountName is not set, a default ServiceAccount (named after the agent)
+                          is created, and this config will be applied to it.
                         properties:
                           annotations:
                             additionalProperties:
@@ -4103,6 +4108,10 @@ spec:
                             type: object
                         type: object
                       serviceAccountName:
+                        description: |-
+                          ServiceAccountName specifies the name of an existing ServiceAccount to use.
+                          If this field is set, the Agent controller will not create a ServiceAccount for the agent.
+                          This field is mutually exclusive with ServiceAccountConfig.
                         type: string
                       tolerations:
                         items:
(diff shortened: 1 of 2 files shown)


go/api/v1alpha2/agent_types.go
```diff
@@ -173,6 +173,17 @@ type SharedDeploymentSpec struct {
 	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
 	// +optional
 	PodSecurityContext *corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`
+	// +optional
+	ServiceAccountName *string `json:"serviceAccountName,omitempty"`
+	// +optional
+	ServiceAccountConfig *ServiceAccountConfig `json:"serviceAccountConfig,omitempty"`
```
EItanya, line 187: Can you add a cel rule to ensure that only one of these can be set. If the `ServiceAccountName` is set, than we won't create the `ServiceAccount`. And if the config is set, then the name can't be changed.

Also can you add more comments to these fields to explain a bit about this interaction

EItanya (changes requested):
One main comment, also can you add a golden test

Issue #1256, opened 2026-01-30
[FEATURE] Support long-term memory for agents


supreme-gg-gg:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add support for long-term memory

### Why do we need long term memory?

Sessions can track history (events) and temporary data for a single ongoing conversation. But in order for agents to recall information from past conversations (from the same or different session), we need a memory service to support "long-term knowledge", similar to a searchable archive or knowledge library that the agent can consult, potentially containing information from many past chats or other sources.

### System Overview

- Agent (Python ADK): Uses KagentMemoryService to send/retrieve memories via HTTP and generate embeddings, configured from the CRDs
- Controller (Go): Exposes /api/memory endpoints to manage vector storage
- Storage (Postgres): Stores embeddings and metadata in the agent_memory table.

### Components

1. HTTP server endpoints (Go)

- POST /api/memory/sessions: Accepts (text, vector) pairs. Stores in agent_memory.
- POST /api/memory/search: Accepts (query_vector). Performs cosine similarity search in pgvector

2. Kagent ADK (Python)

`KagentMemoryService` will `BaseMemoryService` with two methods:  

- `add_session_to_memory`: Uses LiteLLM to generate embeddings using the configured embedding model, send vectors to Kagent API
- `search_memory`: Embed query and send `query_vector` to Kagent API

This will be created and passed to the Runner if memory is enabled on the agent. The agent will use tools to either optionally or preload memory. After each agent execution, a callback will save the session to memory using the memory service. These are documented further in the ADK memory interface: https://google.github.io/adk-docs/sessions/memory/

3. Embedding Model Config

```
apiVersion: kagent.dev/v1alpha2
kind: ModelConfig
spec:
  embedding:
    provider: OpenAI # or AzureOpenAI, Vertex, Ollama
    model: "text-embedding-3-small"
```

4. Agent Configuration 

```yaml
spec:
  declarative:
    memory:
      enabled: true
      modelConfig: "embedding-model-config"
```

5. Data Model

The user can configure whether to extend the database to use vector-extensions in Helm Chart:

```
// helm/kagent/values.yaml
database:
  postgres:
    vector:
      enabled: true # Default: true
```

[...]

PR #1266, opened 2026-02-04
feat: memory store and agent long-term memory

supreme-gg-gg:
Initial design: #1256 
Detailed design doc: [design/EP-1256-memory.md](https://github.com/supreme-gg-gg/kagent/blob/feat/memory/design/EP-1256-memory.md)

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_memory_service.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_memory_service.py
@@ -1,5 +1,6 @@
 """Kagent Memory Service implementation conforming to ADK BaseMemoryService interface."""
 
+import asyncio
 import json
 import logging
 from typing import Any, Dict, List, Optional, Union
@@ -49,7 +50,20 @@ class KagentMemoryService(BaseMemoryService):
         self.ttl_days = ttl_days
 
     async def add_session_to_memory(self, session: Session, model: Optional[Any] = None) -> None:
-        """Add a session's content to long-term memory.
+        """Add a session's content to long-term memory (non-blocking).
+
+        Schedules the actual work as a background task so the caller returns
+        immediately. Memory saving (summarization, embedding, storage) happens
+        asynchronously and does not block the agent from handling the next request.
+
+        Args:
+            session: The session to add to memory
+            model: Optional ADK model object (e.g., LiteLlm, OpenAI) to use for summarization.
+        """
+        asyncio.create_task(self._add_session_to_memory_background(session, model))
+
+    async def _add_session_to_memory_background(self, session: Session, model: Optional[Any] = None) -> None:
+        """Background implementation of add_session_to_memory.
 
         Extracts text from session events, summarizes it using LLM, generates
         an embedding, and stores it in the Kagent backend with TTL support.
@@ -58,64 +72,64 @@ class KagentMemoryService(BaseMemoryService):
             session: The session to add to memory
             model: Optional ADK model object (e.g., LiteLlm, OpenAI) to use for summarization.
         """
-        # Extract content from session events
-        raw_content = self._extract_session_content(session)
-        if not raw_content:
-            logger.debug("No content to add to memory from session %s", session.id)
-            return
-
-        logger.debug("Adding session %s to memory for user %s", session.id, session.user_id)
-
-        # Summarize content before embedding
-        # Returns a list of strings (individual facts/memories)
-        contents = await self._summarize_session_content_async(raw_content, model=model)
-
-        # Filter out empty content items
-        valid_contents = [c for c in contents if c]
-        if not valid_contents:
-            return
-
-        logger.debug("Generating embeddings for %d content items", len(valid_contents))
-
-        # Batch generate embeddings
-        vectors = await self._generate_embedding_async(valid_contents)
-        if not vectors:
-            logger.warning("Failed to generate embeddings for session %s", session.id)
-            return
-
-        if not isinstance(vectors[0], (list, np.ndarray)):
-            # vectors is a flat list of floats (single vector); wrap it
-            vectors = [vectors]
-
-        # Prepare batch items
-        batch_items = []
-
-        # Iterate over synced content and vectors
-        for content_item, vector in zip(valid_contents, vectors, strict=True):
-            if not vector:
-                continue
-
-            item: Dict[str, Any] = {
-                "agent_name": self.agent_name,
-                "user_id": session.user_id,
-                "content": content_item,
-                "vector": vector,
-            }
-            if self.ttl_days > 0:
-                item["ttl_days"] = self.ttl_days
-            batch_items.append(item)
-
-        if not batch_items:
-            return
-
         try:
+            # Extract content from session events
+            raw_content = self._extract_session_content(session)
+            if not raw_content:
+                logger.debug("No content to add to memory from session %s", session.id)
+                return
+
+            logger.debug("Adding session %s to memory for user %s", session.id, session.user_id)
+
+            # Summarize content before embedding
+            # Returns a list of strings (individual facts/memories)
+            contents = await self._summarize_session_content_async(raw_content, model=model)
+
+            # Filter out empty content items
+            valid_contents = [c for c in contents if c]
+            if not valid_contents:
+                return
+
+            logger.debug("Generating embeddings for %d content items", len(valid_contents))
+
+            # Batch generate embeddings
+            vectors = await self._generate_embedding_async(valid_contents)
+            if not vectors:
+                logger.warning("Failed to generate embeddings for session %s", session.id)
+                return
+
+            if not isinstance(vectors[0], (list, np.ndarray)):
+                # vectors is a flat list of floats (single vector); wrap it
+                vectors = [vectors]
+
+            # Prepare batch items
+            batch_items = []
+
+            # Iterate over synced content and vectors
+            for content_item, vector in zip(valid_contents, vectors, strict=True):
+                if not vector:
+                    continue
+
+                item: Dict[str, Any] = {
+                    "agent_name": self.agent_name,
+                    "user_id": session.user_id,
+                    "content": content_item,
+                    "vector": vector,
+                }
+                if self.ttl_days > 0:
+                    item["ttl_days"] = self.ttl_days
+                batch_items.append(item)
+
+            if not batch_items:
+                return
+
             response = await self.client.post("/api/memories/sessions/batch", json={"items": batch_items})
             if response.status_code >= 400:
                 logger.error("Response body: %s", response.text)
             response.raise_for_status()
             logger.info("Successfully saved %d memory items via batch API", len(batch_items))
         except Exception as e:
-            logger.error("Failed to save session %s memory items: %s", session.id, e)
+            logger.error("Failed to save session %s to memory in background: %s", session.id, e)
 
     async def add_memory(
         self,
@@ -380,7 +394,6 @@ class KagentMemoryService(BaseMemoryService):
 Focus on:
 - User preferences, decisions, and explicit requests
 - Important facts mentioned (names, dates, project names, etc.)
-- Action items or commitments made
 - Contextual information that provides background
 - Lessons learned from the conversation
 
--- a/design/EP-1256-memory.md
+++ b/design/EP-1256-memory.md
@@ -116,7 +116,7 @@ spec:
 
 ### 4. UI
 
-The user can view memories for an agent and clear it manually.
+The user can view memories for an agent and clear it manually. UI section to configure agent with memory.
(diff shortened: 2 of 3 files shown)


dimetron:
Hi @supreme-gg-gg 

- Do you plan to add GenerateMemories func to do LLM based extraction ? 
- How I can configure default InMemoryMemoryService, VertexAiMemoryBankService options ?

supreme-gg-gg:
> * Do you plan to add GenerateMemories func to do LLM based extraction ?

Yea there is a LLM based extraction step here: https://github.com/kagent-dev/kagent/blob/7c1409b2a7a5908b0457b0cb09aac0287d050a80/python/packages/kagent-adk/src/kagent/adk/_memory_service.py#L358

However there isn't a "consolidation" step which means before inserting memory directly after extraction, find relevant entries in the db, show them to another LLM and ask it to update / add these entries, this is commonly done in e.g. Vertex AI memory bank. I marked that in the limitations section but that will be more complicated to do.

> * How I can configure default InMemoryMemoryService, VertexAiMemoryBankService options ?

With this PR if you enable memory on declarative agents you only use the kagent memory store, for those two options you can use BYO for now. But I'm working on refactoring the interface for extending to other memory stores I'll also think about if we can allow things like Vertex memory service.

python/packages/kagent-adk/src/kagent/adk/_memory_service.py
```diff
@@ -0,0 +1,463 @@
...
+            for item in response.data:
+                embedding = item["embedding"]
+
+                # LiteLLM does not truncate embeddings by default if the model doesn't support it
+                # However, truncating embeddings is still valid (for most models, see OpenAI's docs and this research https://arxiv.org/html/2508.17744v1)
+                if len(embedding) > 768:
+                    embedding = embedding[:768]
+                    # if we change dimension manually, we need to re-normalize the embeddings
+                    embedding = self._normalize_l2(embedding)
+
```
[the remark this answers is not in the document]
supreme-gg-gg: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/kagent-dev/kagent/pull/1266#discussion_r2849404794)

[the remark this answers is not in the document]
supreme-gg-gg: apparently that did not work so I'll fix it manually

go/api/v1alpha2/agent_types.go
```diff
@@ -118,6 +118,21 @@ type DeclarativeAgentSpec struct {
...
+
+type MemorySpec struct {
+	// Enabled determines whether memory is enabled for the agent.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
+
+	// The name of the model config to use for embeddings.
+	// If not specified, it defaults to the agent's main ModelConfig.
+	// +optional
+	ModelConfig string `json:"modelConfig,omitempty"`
```
EItanya: Can this be a `*corev1.LocalObjectReference`?

EItanya: Please use `*corev1.LocalObjectReference`

go/api/v1alpha2/agent_types.go
```diff
@@ -118,6 +118,21 @@ type DeclarativeAgentSpec struct {
...
+
+	// Memory configuration for the agent.
+	// +optional
+	Memory *MemorySpec `json:"memory,omitempty"`
+}
+
+type MemorySpec struct {
+	// Enabled determines whether memory is enabled for the agent.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
EItanya: Why do we need this field? If this struct exists it's set.

go/api/v1alpha2/agent_types.go
EItanya: Also, should be required

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -574,6 +574,40 @@ func (a *adkApiTranslator) translateInlineAgent(ctx context.Context, agent *v1al
...
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to resolve embedding config: %w", err)
+		}
+
+		cfg.Embedding = adk.ModelToEmbeddingConfig(embeddingCfg)
+
+		// Merge EnvVars from embedding config (e.g. API Keys)
+		mdd.EnvVars = append(mdd.EnvVars, embMdd.EnvVars...)
+		mdd.Volumes = append(mdd.Volumes, embMdd.Volumes...)
+		mdd.VolumeMounts = append(mdd.VolumeMounts, embMdd.VolumeMounts...)
```
EItanya: What about duplicates? I believe they may make the Deployment invalid. e.g. if the `OPENAI_API_KEY` is the LLM and the embedding model.

go/Dockerfile
```diff
@@ -34,9 +34,10 @@ RUN --mount=type=cache,target=/root/go/pkg/mod,rw             \
     CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -ldflags "$LDFLAGS" -o manager cmd/controller/main.go
 
 ### STAGE 2: final image
-# Use distroless as minimal base image to package the manager binary
+# Use distroless/cc-debian12 which includes C/C++ runtime libraries
+# This is required for turso-go's purego library loading
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/cc-debian12:nonroot
```
EItanya, line 40: This is a pretty big change. Why do we need this if we don't need `CGO`?

go/pkg/database/models.go
```diff
@@ -127,79 +131,96 @@ type Feedback struct {
...
+	UpdatedAt   time.Time `gorm:"autoUpdateTime" json:"updated_at"`
+	DeletedAt   time.Time `gorm:"index" json:"deleted_at"`
+	Description string    `json:"description"`
 }
 
 // ToolServer represents a tool server that provides tools
 type ToolServer struct {
-	CreatedAt     time.Time      `gorm:"autoCreateTime" json:"created_at"`
-	UpdatedAt     time.Time      `gorm:"autoUpdateTime" json:"updated_at"`
-	DeletedAt     gorm.DeletedAt `gorm:"index" json:"deleted_at"`
```
EItanya, line 143: Will this break folks currently running in postgres?

supreme-gg-gg: Oops these changes are not necessary I fixed them but forgot to push that commit

go/Dockerfile
supreme-gg-gg: I believe we do since otherwise I ran into some error from Turso

python/packages/kagent-adk/src/kagent/adk/_memory_service.py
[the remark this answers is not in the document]
supreme-gg-gg: fixed

go/api/v1alpha2/agent_types.go
supreme-gg-gg: > Please use `*corev1.LocalObjectReference`

Why should we not use string since the upper level `ModelConfig` uses that

go/internal/controller/translator/agent/adk_api_translator.go
supreme-gg-gg: Added deduplication of env vars and volumes when translating embedding config, so that the user can use a different provider for embedding vs LLM but avoids duplication

go/api/v1alpha2/agent_types.go
supreme-gg-gg: Fixed

supreme-gg-gg:
Fixed most comments above 😄 will be adding e2e test to memory together with this one

go/pkg/adk/types.go
```diff
@@ -296,26 +296,94 @@ type RemoteAgentConfig struct {
...
+	Description   string                `json:"description"`
+	Instruction   string                `json:"instruction"`
+	HttpTools     []HttpMcpServerConfig `json:"http_tools"`
+	SseTools      []SseMcpServerConfig  `json:"sse_tools"`
+	RemoteAgents  []RemoteAgentConfig   `json:"remote_agents"`
+	ExecuteCode   bool                  `json:"execute_code,omitempty"`
+	Stream        bool                  `json:"stream"`
+	MemoryEnabled bool                  `json:"memory_enabled,omitempty"`
+	MemoryTTLDays int                   `json:"memory_ttl_days,omitempty"`
+	Embedding     *EmbeddingConfig      `json:"embedding,omitempty"`
```
EItanya: Can we move these into a `*Memory` struct so we don't need an explciit enabled flag?

go/internal/httpserver/server.go
```diff
@@ -107,6 +107,22 @@ func (s *HTTPServer) Start(ctx context.Context) error {
...
+			select {
+			case <-ticker.C:
+				if err := s.config.DbClient.PruneExpiredMemories(); err != nil {
+					log.Error(err, "Failed to prune expired memories")
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
```
EItanya: We only wanna do this in the leader

go/pkg/database/models.go
```diff
@@ -202,6 +204,36 @@ type CrewAIFlowState struct {
...
+}
+
+// BeforeCreate generates a UUID for the Memory ID if one has not been set,
+// making ID generation database-agnostic (works for both Postgres and SQLite).
+func (m *Memory) BeforeCreate(tx *gorm.DB) error {
+	if m.ID == "" {
+		m.ID = uuid.New().String()
+	}
+	return nil
+}
```
EItanya, line 229: I think the DBs can do this automatically?

go/pkg/database/models.go
```diff
@@ -213,5 +245,6 @@ func (Tool) TableName() string                     { return "tool" }
 func (ToolServer) TableName() string               { return "toolserver" }
 func (LangGraphCheckpoint) TableName() string      { return "lg_checkpoint" }
 func (LangGraphCheckpointWrite) TableName() string { return "lg_checkpoint_write" }
-func (CrewAIAgentMemory) TableName() string        { return "crewai_agent_memory" }
+func (CrewAIAgentMemory) TableName() string        { return "crewai_agentp_memory" }
```
EItanya: ?

EItanya: ?

go/pkg/database/models.go
supreme-gg-gg: Not for sqlite, it requires an extension for uuid function.

python/packages/kagent-adk/src/kagent/adk/_memory_service.py
```diff
@@ -0,0 +1,472 @@
...
+        if not self.memory_enabled:
+            return
+
+        # Extract content from session events
+        raw_content = self._extract_session_content(session)
+        if not raw_content:
+            logger.debug("No content to add to memory from session %s", session.id)
+            return
+
+        logger.info("Adding session %s to memory for user %s", session.id, session.user_id)
```
EItanya: Just a general nit, I think there's A LOT of `info` level logging in the new code, I think we should probably be a little more sparing with that level of logging

supreme-gg-gg: Removed some logging and changed most to debug

supreme-gg-gg:
Short demo (really simple example):

https://github.com/user-attachments/assets/bf699f62-a15a-49be-b697-9df7f2e8fb1d

This does not show auto memory extraction, auto prefetching memory, and auto memory cleanup.

go/internal/httpserver/server.go
supreme-gg-gg: Done by implementing `LeaderElectionRunnable` interface and adding the runnable to controller manager

supreme-gg-gg:
A small UI change to add memory section, here's how it looks like:

Screenshot 2026-02-25 at 6 44 06 PM

Also tested long term memory with some somewhat-complex and multi-turn use cases, performs as expected and made several enhancements based on that.

go/pkg/adk/types.go
```diff
@@ -296,6 +296,72 @@ type RemoteAgentConfig struct {
 	Description string            `json:"description,omitempty"`
 }
 
+// EmbeddingConfig is the embedding model config for memory tools.
+// JSON uses "provider" to match Python EmbeddingConfig; unmarshaling accepts "type" for backward compat.
+type EmbeddingConfig struct {
```
EItanya, line 301: Please add more unit tests for this in a follow-up. The custom marshalling in this function is starting to get pretty complex

PR #1268, opened 2026-02-05
fix: avoid 500s when a2a cleanup is cancelled

lithammer:
We hit a case where a client retried on `5xx` responses, and the agent's Slack tool had already executed before the server returned `500`. The request itself was effectively successful, but cleanup raised `CancelledError`, which bubbled into a `500` and triggered the retry. That led to duplicate Slack messages.

This change treats `CancelledError` during `_cleanup_producer` as non‑fatal so a completed request doesn't get turned into a `500` during cleanup. The result is that clients don't retry after work has already been done, avoiding duplicate side effects.

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_safe_request_handler.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_safe_request_handler.py
@@ -1,0 +1,24 @@
+import asyncio
+import logging
+
+from a2a.server.request_handlers import DefaultRequestHandler
+
+logger = logging.getLogger(__name__)
+
+
+class SafeRequestHandler(DefaultRequestHandler):
+    """Request handler that avoids turning cleanup cancellation into 500s."""
+
+    async def _cleanup_producer(self, producer_task, task_id):
+        try:
+            await super()._cleanup_producer(producer_task, task_id)
+        except asyncio.CancelledError:
+            logger.debug(
+                "A2A cleanup cancelled",
+                extra={"task_id": task_id},
+                exc_info=True,
+            )
+            # Make a best-effort attempt to clean up resources.
+            await self._queue_manager.close(task_id)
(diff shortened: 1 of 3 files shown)


python/packages/kagent-adk/src/kagent/adk/_safe_request_handler.py
```diff
@@ -0,0 +1,24 @@
...
+        except asyncio.CancelledError:
+            logger.debug(
+                "A2A cleanup cancelled",
+                extra={"task_id": task_id},
+                exc_info=True,
+            )
+            # Make a best-effort attempt to clean up resources.
+            await self._queue_manager.close(task_id)
+            async with self._running_agents_lock:
+                self._running_agents.pop(task_id, None)
```
lithammer, line 24: A bit unsure about this part.

lithammer:
Or would you prefer I try to solve this upstream in https://github.com/a2aproject/a2a-python?

lithammer:
> Or would you prefer I try to solve this upstream in https://github.com/a2aproject/a2a-python?

https://github.com/a2aproject/a2a-python/pull/669

EItanya:
Agreed, I think the upstream fix is much much better than this one. Let's give it a few days and see if they respond

lithammer:
> Agreed, I think the upstream fix is much much better than this one. Let's give it a few days and see if they respond

Seems like it was fixed in https://github.com/a2aproject/a2a-python/pull/614.

lithammer:
> > Agreed, I think the upstream fix is much much better than this one. Let's give it a few days and see if they respond
> 
> Seems like it was fixed in [a2aproject/a2a-python#614](https://github.com/a2aproject/a2a-python/pull/614).

https://github.com/kagent-dev/kagent/pull/1326

PR #1274, opened 2026-02-09
Feature: UI dynamic provider model discovery

nujragan93:
Summary

  Introduces a new Provider Custom Resource Definition (CRD) that enables automatic model discovery from LLM providers. Users can configure provider endpoints (e.g., AI gateways, self-hosted LLMs) and kagent will dynamically discover available models.

  Changes

  Backend (Go)

  - New CRD: Provider type (go/api/v1alpha2/provider_types.go) with fields for provider type, endpoint, and secret reference
  - Controller: ProviderController watches Provider CRs and triggers model discovery
  - Model Discovery: provider/discoverer.go - discovers models from OpenAI, Anthropic, Ollama, Azure, Gemini, and Bedrock APIs
  - Provider Manager: provider/manager.go - caches discovered models and manages provider state
  - API Endpoints:
    - GET /api/providers/configured - list configured providers
    - GET /api/providers/configured/{name}/models - get models for a provider

  Frontend (React/Next.js)

  - ProviderCombobox: New dropdown grouping "Stock Providers" and "Configured Providers"
  - ModelCombobox: Displays discovered models with function-calling indicator
  - Dynamic Model Fetching: Fetches models from configured provider endpoints
  - Auto-selection: Defaults to stock OpenAI provider on page load

  Infrastructure

  - CRD manifest: kagent.dev_providers.yaml
  - Helm chart templates for CRD and RBAC
  - Sample Provider manifests for OpenAI and Anthropic

  Testing

  - Unit tests for model discoverer (17 tests)
  - Unit tests for provider manager
  - Provider selection behavior tests

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -114,6 +114,12 @@ jobs:
         uv build --package kagent-crewai
         uv version $VERSION --package kagent-skills
         uv build --package kagent-skills
+        uv version $VERSION --package kagent-openai
+        uv build --package kagent-openai
+        uv version $VERSION --package agentsts-adk
+        uv build --package agentsts-adk
+        uv version $VERSION --package agentsts-core
+        uv build --package agentsts-core
         uv publish --token ${{ secrets.PYPI_TOKEN }}
       
   release:
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,8 @@ LDFLAGS := "-X github.com/$(DOCKER_REPO)/go/internal/version.Version=$(VERSION) 
             -X github.com/$(DOCKER_REPO)/go/internal/version.BuildDate=$(BUILD_DATE)"
 
 #tools versions
-TOOLS_UV_VERSION ?= 0.9.2
-TOOLS_BUN_VERSION ?= 1.3.6
+TOOLS_UV_VERSION ?= 0.10.4
+TOOLS_BUN_VERSION ?= 1.3.9
 TOOLS_NODE_VERSION ?= 24.13.0
 TOOLS_PYTHON_VERSION ?= 3.13
 
--- a/design/EP-476-OIDC-Authentication.md
+++ b/design/EP-476-OIDC-Authentication.md
@@ -1,0 +1,762 @@
+# EP-476: OIDC Authentication Integration
+
+* Issue: [#476](https://github.com/kagent-dev/kagent/issues/476)
+
+## Background
+
+KAgent currently uses an unsecure authentication mechanism (`UnsecureAuthenticator`) that accepts any user ID provided via query parameters or headers, with no validation. This development-grade authentication is suitable only for trusted environments and poses significant security risks for production deployments.
+
+This proposal adds enterprise-grade authentication to KAgent by implementing a standard OIDC (OpenID Connect) client that works with any compliant OIDC provider. This enables integration with enterprise identity systems including Keycloak, Auth0, Okta, Azure AD, Google, or Dex. KAgent will not bundle an identity provider; instead, users deploy and manage their OIDC provider separately, following infrastructure separation principles.
+
+**Sponsors**: Collin Walker (@lets-call-n-walk)
+
+**Origin**: Security hardening initiative to enable production deployments
+
+## Motivation
+
+### Current State
+- **No real authentication**: System trusts any user ID provided by the client
+- **No user management**: Users are just string identifiers with no profiles or credentials
+- **Client-side identity**: User ID stored in browser localStorage can be trivially modified
+- **No authorization**: NoopAuthorizer allows all users to perform all operations
+- **Impersonation vulnerability**: Anyone can impersonate any user by changing query parameters
+- **Not production-ready**: Suitable only for development/testing environments
+
+### Why OIDC with External Provider?
+1. **Standards-based**: OIDC is an industry-standard protocol supported by all major identity providers
+2. **Enterprise integration**: Works with existing identity infrastructure (Okta, Azure AD, Keycloak, Auth0, Google, etc.)
+3. **Flexibility**: Organizations choose their preferred identity provider based on requirements
+4. **Separation of concerns**: KAgent focuses on agent orchestration, not identity management
+5. **Scalability**: External providers can be deployed with HA and persistent storage
+6. **Multi-backend support**: Providers like Dex aggregate multiple identity sources (LDAP, SAML, GitHub, etc.) through a single OIDC interface
+7. **Operational simplicity**: No identity provider lifecycle management within KAgent deployments
+
+### Goals
+
+1. **Add secure authentication** using OAuth2/OIDC flow with proper token-based authentication
+2. **Support any OIDC provider** via standard OIDC discovery and token validation
+3. **Work with existing identity infrastructure** including Keycloak, Auth0, Okta, Azure AD, Google, Dex, and other OIDC-compliant providers
+4. **Implement session management** with secure token storage, refresh, and revocation
+5. **Add RBAC foundation** with group-based role mapping from OIDC claims
+6. **Maintain backward compatibility** during migration with feature flags
+7. **Provide migration path** from unsecure auth to OIDC with clear documentation
+8. **Support both UI and CLI** authentication flows (web flow for UI, device code or browser flow for CLI)
+
+### Non-Goals
+
+1. **Deploying or managing identity providers**: Users must deploy and manage their own OIDC provider (Dex, Keycloak, etc.)
+2. **Identity provider configuration**: KAgent does not configure upstream identity providers; that's the operator's responsibility
+3. **Custom user database**: Users will be managed by external identity providers, not stored locally
+4. **Fine-grained permissions**: Initial implementation focuses on authentication; detailed RBAC policies come in future iterations
+5. **Multi-tenancy isolation**: Organization/workspace-level isolation is out of scope
+6. **User provisioning API**: No self-service user registration or profile management
+7. **Social login widgets**: Only identity providers configured by administrators are supported
+8. **Advanced token encryption**: Basic token protection will be implemented; advanced features like HSM integration or key rotation are deferred
+
+## Implementation Details
+
+### Architecture Overview
+
+**Component Structure** (External OIDC Provider):
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│                         User Browser                         │
+└────┬──────────────────────────────────────────────────┬─────┘
+     │                                                    │
+     │ 1. /auth/login                                    │ 4. Set cookie
+     │                                                    │    with JWT
+     ▼                                                    │
+┌─────────────────────────────────────────────────────────────┐
+│               kagent-server (Go HTTP Server)                │
+│                                                              │
+│  • /auth/login    → Redirect to OIDC OAuth2 authorize      │
+│  • /auth/callback → Exchange code for tokens               │
+│  • /auth/logout   → Revoke tokens, clear session           │
+│  • /api/*         → Protected API endpoints (authN/authZ)  │
+│                                                              │
+│  Middleware:                                                │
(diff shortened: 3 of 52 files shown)


nujragan93:
Screenshot 2026-02-09 at 8 04 53 AM
Screenshot 2026-02-09 at 8 05 01 AM

go/api/v1alpha2/provider_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// Type is the model provider type (OpenAI, Anthropic, etc.)
+	// +required
+	// +kubebuilder:validation:Required
+	Type ModelProvider `json:"type"`
+
+	// Endpoint is the API endpoint URL for the provider
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^https?://.*`
+	Endpoint string `json:"endpoint"`
```
EItanya: Can we have a default endpoint based on the provider? For example OpenAI can go to openai, and Anthropic to anthropic?

go/api/v1alpha2/provider_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// SecretRef references the Kubernetes Secret containing the API key
+	// +required
+	// +kubebuilder:validation:Required
+	SecretRef SecretReference `json:"secretRef"`
```
EItanya: What if the endpoint isn't secured with a token, like local ollama or something

go/internal/controller/provider/discoverer.go
```diff
@@ -0,0 +1,217 @@
...
+
+// OllamaTagsResponse represents Ollama's /api/tags response format
+type ollamaTagsResponse struct {
+	Models []struct {
+		Name string `json:"name"`
+	} `json:"models"`
+}
+
+// DiscoverOllamaModels handles Ollama's different response format.
+func (d *ModelDiscoverer) DiscoverOllamaModels(ctx context.Context, endpoint string) ([]string, error) {
```
EItanya: Why is this public? If it's just for tests then we should probably test from within the package or something

go/internal/controller/reconciler/reconciler.go
```diff
@@ -910,3 +912,194 @@ func convertTool(tool *database.Tool) (*v1alpha2.MCPTool, error) {
...
+		p.Status.LastDiscoveryTime = &now
+	}
+
+	// Clear force-discovery annotation if set
+	if p.Annotations != nil && p.Annotations[v1alpha2.ProviderAnnotationForceDiscovery] == "true" {
+		delete(p.Annotations, v1alpha2.ProviderAnnotationForceDiscovery)
+		if err := a.kube.Update(ctx, p); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to clear force-discovery annotation: %w", err)
+		}
+	}
```
EItanya: I don't like doing this. This resource is not owned by us. For a force reconciliation can we use an endpoint or something like that?

go/internal/controller/reconciler/reconciler.go
```diff
@@ -910,3 +912,194 @@ func convertTool(tool *database.Tool) (*v1alpha2.MCPTool, error) {
...
+}
+
+// discoverProviderModels calls the model discoverer to fetch models
+func (a *kagentReconciler) discoverProviderModels(ctx context.Context, p *v1alpha2.Provider, secret *corev1.Secret) ([]string, error) {
+	if secret == nil {
+		return nil, fmt.Errorf("cannot discover models: secret not available")
+	}
+
+	// Get API key from secret
+	apiKey, ok := secret.Data[p.Spec.SecretRef.Key]
```
EItanya: We're doing this twice, in validate and discover

go/internal/controller/provider/types.go
```diff
@@ -0,0 +1,59 @@
...
+
+package provider
+
+import (
+	v1alpha2 "github.com/kagent-dev/kagent/go/api/v1alpha2"
+)
+
+// ProviderConfig represents a configured LLM provider instance.
+// Multiple ProviderConfigs can exist for the same provider type (e.g., two OpenAI instances).
+type ProviderConfig struct {
```
EItanya: Why not just use the actual k8s object?

go/internal/controller/provider/manager.go
```diff
@@ -0,0 +1,148 @@
...
+	provider := &v1alpha2.Provider{}
+	if err := m.client.Get(ctx, types.NamespacedName{
+		Name: providerName, Namespace: m.namespace,
+	}, provider); err != nil {
+		return nil, fmt.Errorf("provider %s not found: %w", providerName, err)
+	}
+
+	// Force refresh by setting annotation
+	if forceRefresh {
+		if provider.Annotations == nil {
```
EItanya: Same comment, we should not be modifying the actual provider object

go/internal/controller/provider/manager.go
```diff
@@ -0,0 +1,148 @@
...
+	// No models discovered - provide helpful message
+	if forceRefresh {
+		return nil, fmt.Errorf("no models discovered for provider %s", providerName)
+	}
+	return nil, fmt.Errorf("no models discovered for provider %s, try refreshing", providerName)
+}
+
+// ClearCache is a no-op for CRD-based implementation.
+// Models are cached in Provider.Status.DiscoveredModels and cleared by the controller.
+func (m *Manager) ClearCache(providerName string) {
```
EItanya: Why is this here if it's unused?

go/internal/controller/provider/manager.go
```diff
@@ -0,0 +1,148 @@
...
+	// No-op - cache is managed by Provider controller in CRD status
+}
+
+// HasProviders returns true if any Ready providers are configured.
+func (m *Manager) HasProviders() bool {
```
EItanya: This function feels unecessary, especially if uncached

go/internal/controller/provider/manager.go
```diff
@@ -0,0 +1,148 @@
...
+		if provider.Annotations == nil {
+			provider.Annotations = make(map[string]string)
+		}
+		provider.Annotations[v1alpha2.ProviderAnnotationForceDiscovery] = "true"
+		if err := m.client.Update(ctx, provider); err != nil {
+			return nil, fmt.Errorf("failed to trigger discovery: %w", err)
+		}
+
+		// Wait for controller to refresh (with timeout)
+		deadline := time.Now().Add(10 * time.Second)
```
EItanya: I think this system is a bit overcomplicated, can we just call the reconciler and trigger that logic manually

go/internal/controller/provider/manager.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/kagent-dev/kagent/go/api/v1alpha2"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// DefaultNamespace is the default namespace for kagent resources
+	DefaultNamespace = "kagent"
```
EItanya: Why do we need this? Can we just use the injected `POD_NAMESPACE` variable. I think there's a helper function for that somewhere

go/api/v1alpha2/provider_types.go
```diff
@@ -0,0 +1,158 @@
...
+// +kubebuilder:printcolumn:name="Type",type="string",JSONPath=".spec.type"
+// +kubebuilder:printcolumn:name="Endpoint",type="string",JSONPath=".spec.endpoint"
+// +kubebuilder:printcolumn:name="Models",type="integer",JSONPath=".status.modelCount"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:storageversion
+
+// Provider is the Schema for the providers API.
+// It represents a model provider configuration with automatic model discovery.
+type Provider struct {
```
EItanya: I think we should rename this to model provider so it's more clear what the purpose is, what do youthink?

go/api/v1alpha2/provider_types.go
```diff
@@ -0,0 +1,158 @@
...
+
+// SecretReference contains information to locate a secret.
+type SecretReference struct {
+	// Name is the name of the secret in the same namespace as the Provider
+	// +required
+	Name string `json:"name"`
+
+	// Key is the key within the secret that contains the API key or credential
+	// +required
+	Key string `json:"key"`
```
EItanya: Another API nit, why is this required? If there's only one key in the secret we can use that. Just small usability things.

go/internal/controller/reconciler/reconciler.go
```diff
@@ -910,3 +913,234 @@ func convertTool(tool *database.Tool) (*v1alpha2.MCPTool, error) {
...
+		now := metav1.Now()
+		p.Status.LastDiscoveryTime = &now
+	}
+
+	// Update status subresource only - never modify the Provider object itself
+	if err := a.kube.Status().Update(ctx, p); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update provider status: %w", err)
+	}
+
+	// No periodic requeue - discovery only on-demand via HTTP API
```
EItanya, line 1101: Why not?

EItanya (commented):
These changes are looking great! I would love to test it all locally but it looks like there are some build failures

go/api/v1alpha2/provider_types.go
nujragan93: Agree, provider seems too vague, I will change this to modelprovider similar to modelconfig

go/internal/controller/reconciler/reconciler.go
nujragan93: llms behind a provider would not change very frequently, this is why I opted to have a force refresh

go/api/v1alpha2/modelproviderconfig_types.go
```diff
@@ -0,0 +1,155 @@
...
+type ModelProviderConfigSpec struct {
+	// Type is the model provider type (OpenAI, Anthropic, etc.)
+	// +required
+	// +kubebuilder:validation:Required
+	Type ModelProvider `json:"type"`
+
+	// Endpoint is the API endpoint URL for the provider.
+	// If not specified, the default endpoint for the provider type will be used.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https?://.*`
```
EItanya, line 73: Does `ollama` also have to be `https`?

EItanya: Also, might be worth preemptively creating an issue for custom TLS config because eventually someone will probably want it 😆

nujragan93: We mount custom CA cert into the kagent controller pod to get over this issue

nujragan93:
> Just tested this locally and this is super exciting! One comment I have for the UX is that when creating a `ModelConfig` from a provider you still need to specify an api key, what do you think about allowing the user to re-use the key which is being used for the provider by default?

I had thought about that, but if we allow  the user to reuse the apikey in the UI, then it will be difficult for operator to stop people from using it.

EItanya:
> > Just tested this locally and this is super exciting! One comment I have for the UX is that when creating a `ModelConfig` from a provider you still need to specify an api key, what do you think about allowing the user to re-use the key which is being used for the provider by default?
> 
> I had thought about that, but if we allow the user to reuse the apikey in the UI, then it will be difficult for operator to stop people from using it.

Sounds good, we can start here!

go/api/v1alpha2/modelproviderconfig_types.go
nujragan93: The pattern ^https?://.* allows both http:// and https://

EItanya (approved):
HELL YEA 🎉

Issue #1272, opened 2026-02-09
[BUG] Default MCP server connection timeout (5s) is too low for sidecar gateway deployments, causing intermittent "Failed to create MCP session" errors


skhedim:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

When an agent references MCP tools via `MCPServer` CRDs, the agent pod intermittently fails with:

```
Failed to create MCP session: Failed to connect to MCP server at http://localhost:8081/mcp: [Errno 111] Connection refused
```

or times out silently, causing the entire tool call to fail.

**Root cause:** The Python ADK's `StreamableHTTPConnectionParams` has a hardcoded default `timeout: float = 5.0` (see [mcp_session_manager.py](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_session_manager.py)). When users deploy a `MCPServer` CRD, the kagent controller automatically generates a corresponding `RemoteMCPServer` resource. However:

1. The `MCPServer` CRD **does not expose a `timeout` field** for users to configure
2. The auto-generated `RemoteMCPServer` therefore has `spec.timeout: nil`
3. The controller does **not** populate a default timeout in the agent's `config.json`
4. The Python ADK falls back to its own 5-second default

**Why 5 seconds is insufficient:** MCP servers deployed via the `MCPServer` CRD use a sidecar [agentgateway](https://github.com/kagent-dev/agentgateway) container that spawns a new stdio process (e.g. via `uvx` or `npx`) for each MCP session. Process startup time depends heavily on:

- **Package cache state:** Cold starts (first invocation, new pod, or cache miss) require downloading/installing the MCP server package via `uvx` or `npx`. Even well-known MCP servers like `awslabs.eks-mcp-server`  can take **2–8 seconds** on cold starts.
[...]

PR #1275, opened 2026-02-10
fix: make default MCP server timeout configurable via Helm chart

skhedim:
# Summary

Closes https://github.com/kagent-dev/kagent/issues/1272

This PR makes the default MCP server timeout configurable via the Helm chart, instead of being hardcoded. When a RemoteMCPServer resource does not have an explicit Spec.Timeout set, the controller now applies a configurable default rather than leaving it unset (which caused the ADK to fall back to its own built-in default of 5 seconds).

## Changes

- **Helm chart**: Added `controller.defaultMCPServerTimeout` value (default: "30s") and corresponding `DEFAULT_MCP_SERVER_TIMEOUT` env var in the controller configmap.
- **Controller flag**: New `--default-mcp-server-timeout` CLI flag (auto-populated from env via LoadFromEnv), with a default of 30 seconds.
- **ADK API translator**: `NewAdkApiTranslator` now accepts a `defaultMCPServerTimeout` parameter. When `RemoteMCPServer.Spec.Timeout` is nil, the configurable default is applied. Also defaults `terminate_on_close` to true when not explicitly set (matches ADK expected behavior).
- **Config threading**: The timeout is threaded through Config → ServerConfig → NewHandlers → Base → agents.go, following the same pattern as ProxyURL.

## Backward Compatibility

- The default value is 30s, which preserves the same behavior as the previous hardcoded timeout.
- When `Spec.Timeout` is explicitly set on a `RemoteMCPServer`, that value takes precedence.
- No CRD changes required.

## Testing

- All existing unit tests updated and passing.
- Golden test outputs regenerated.
- `go build` and `go test` pass cleanly.

### DIFF
--- a/go/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/internal/controller/translator/agent/adk_api_translator.go
@@ -112,19 +112,19 @@ func NewAdkApiTranslator(kube client.Client, defaultModelConfig types.Namespaced
 		defaultMCPServerTimeout = DefaultMCPServerTimeout
 	}
 	return &adkApiTranslator{
-		kube:                   kube,
-		defaultModelConfig:     defaultModelConfig,
-		plugins:                plugins,
-		globalProxyURL:         globalProxyURL,
+		kube:                    kube,
+		defaultModelConfig:      defaultModelConfig,
+		plugins:                 plugins,
+		globalProxyURL:          globalProxyURL,
 		defaultMCPServerTimeout: defaultMCPServerTimeout,
 	}
 }
 
 type adkApiTranslator struct {
-	kube                   client.Client
-	defaultModelConfig     types.NamespacedName
-	plugins                []TranslatorPlugin
-	globalProxyURL         string
+	kube                    client.Client
+	defaultModelConfig      types.NamespacedName
+	plugins                 []TranslatorPlugin
+	globalProxyURL          string
 	defaultMCPServerTimeout time.Duration
 }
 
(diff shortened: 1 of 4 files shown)


EItanya:
First off, thanks for the PR. I think this PR is important but a bit of a workaround, to me it feels like the correct fix would be to add this field to the MCPServer CRD: https://github.com/kagent-dev/kmcp/blob/main/api/v1alpha1/mcpserver_types.go#L157, and to give it a default value there. What do you think?

skhedim:
Thanks for the feedback @EItanya — you're absolutely right. Adding the timeout to the MCPServer CRD is the proper fix.

I initially went with the Helm chart approach to keep it contained in a single PR and get a quicker merge, but I agree it's a workaround: threading the value through Helm → ConfigMap → env var → controller flag → handlers → translator is a lot of plumbing for something that really belongs on the resource itself.

I've implemented the CRD-based approach:

1. **kmcp PR**: [`skhedim/kmcp#feat/mcpserver-default-timeout`](https://github.com/kagent-dev/kmcp/pull/121) — adds a `Timeout` field to `MCPServerSpec` with `+kubebuilder:default="30s"`, following the same pattern as `Port` and `Replicas`.
2. **kagent PR** (will replace this one): `fix/mcpserver-default-timeout-crd` — `ConvertMCPServerToRemoteMCPServer` propagates the CRD timeout to the generated `RemoteMCPServer`, and also sets `TerminateOnClose: true` by default.

I'll close this PR once the new ones are ready for review. Do you think this is the right solution?

skhedim:
Resolved by #1277

PR #1277, opened 2026-02-10
fix: propagate MCPServer CRD timeout to RemoteMCPServer

### LINKED ISSUE #1272: [BUG] Default MCP server connection timeout (5s) is too low for sidecar gateway deployments, causing intermittent "Failed to create MCP session" errors

skhedim:
Closes #1272
Supersedes #1275

### Summary

Propagates the new `Timeout` field from the MCPServer CRD (added in [kagent-dev/kmcp#121](https://github.com/kagent-dev/kmcp/pull/121)) to the generated `RemoteMCPServer` resources, fixing intermittent `"Failed to create MCP session"` errors for sidecar gateway deployments.

### Context

[#1272](https://github.com/kagent-dev/kagent/issues/1272) reported that the default 5s MCP connection timeout is too low for MCPServer CRD deployments where the sidecar gateway spawns stdio processes (via `uvx`/`npx`) with 2–8s cold starts.

An initial approach ([#1275](https://github.com/kagent-dev/kagent/pull/1275)) made the timeout configurable via Helm chart values, threading it through 6+ layers (Helm → ConfigMap → env → flag → handlers → translator). Per [maintainer feedback](https://github.com/kagent-dev/kagent/pull/1275#issuecomment-), the proper fix is to add the timeout directly to the MCPServer CRD.

### Approach

Instead of global controller-level configuration, the timeout is now a **per-resource field** on the MCPServer CRD with a kubebuilder default of `30s` (see [kagent-dev/kmcp#121](https://github.com/kagent-dev/kmcp/pull/121)).

`ConvertMCPServerToRemoteMCPServer` now propagates:
- **`Timeout`** from `MCPServerSpec` → `RemoteMCPServerSpec` (defaults to 30s via CRD)
- **`TerminateOnClose: true`** (matches ADK expected behavior for managed servers)

### Changes

| File | Description |
|------|-------------|
| `go/go.mod` | Bump kmcp dependency to version with `Timeout` field (kmcp version 0.2.6) |
| `go/internal/controller/translator/agent/adk_api_translator.go` | `ConvertMCPServerToRemoteMCPServer` propagates `Timeout` and sets `TerminateOnClose: true` |
| `go/internal/controller/translator/agent/testdata/outputs/agent_with_proxy_mcpserver.json` | Golden test updated |

### Backward Compatibility

- Default timeout is `30s`, matching the previous hardcoded behavior from #1275.
- When `MCPServer.Spec.Timeout` is explicitly set, that value takes precedence.
- `RemoteMCPServer` resources created directly (not via MCPServer CRD) are unaffected — they continue using their own `Spec.Timeout` field.

### Depends on

- [x] [kagent-dev/kmcp#121](https://github.com/kagent-dev/kmcp/pull/121) must be merged and released first
- [x] Update `go.mod` to replace the `replace` directive with the actual released version

### Testing

- All existing unit tests pass
- Golden test outputs regenerated
- `go build ./...` and `go test ./internal/controller/...` pass cleanly

### DIFF
--- a/go/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/internal/controller/translator/agent/adk_api_translator.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/kagent-dev/kagent/go/api/v1alpha2"
 	"github.com/kagent-dev/kagent/go/internal/adk"
@@ -45,6 +46,12 @@ const (
 	MCPServiceProtocolDefault = v1alpha2.RemoteMCPServerProtocolStreamableHttp
 
 	ProxyHostHeader = "x-kagent-host"
+
+	// DefaultMCPServerTimeout is the fallback connection timeout applied when
+	// an MCPServer CRD resource does not have an explicit Timeout set (e.g.
+	// objects created before the field was introduced). This value mirrors
+	// the kubebuilder default on MCPServerSpec.Timeout in the kmcp CRD.
+	DefaultMCPServerTimeout = 30 * time.Second
 )
 
 // ValidationError indicates a configuration error that requires user action to fix.
@@ -1259,7 +1266,7 @@ func ConvertMCPServerToRemoteMCPServer(mcpServer *v1alpha1.MCPServer) (*v1alpha2
 		return nil, NewValidationError("cannot determine port for MCP server %s", mcpServer.Name)
 	}
 
-	return &v1alpha2.RemoteMCPServer{
+	remoteMCP := &v1alpha2.RemoteMCPServer{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      mcpServer.Name,
 			Namespace: mcpServer.Namespace,
@@ -1268,7 +1275,20 @@ func ConvertMCPServerToRemoteMCPServer(mcpServer *v1alpha1.MCPServer) (*v1alpha2
 			URL:      fmt.Sprintf("http://%s.%s:%d/mcp", mcpServer.Name, mcpServer.Namespace, mcpServer.Spec.Deployment.Port),
 			Protocol: v1alpha2.RemoteMCPServerProtocolStreamableHttp,
 		},
-	}, nil
+	}
+
+	// Propagate the timeout from the MCPServer CRD to the generated
+	// RemoteMCPServer spec. Fall back to DefaultMCPServerTimeout for
+	// MCPServer objects created before the CRD default was introduced,
+	// so the ADK never uses its own 5s built-in which is too short for
+	// sidecar gateway cold starts.
+	if mcpServer.Spec.Timeout != nil {
+		remoteMCP.Spec.Timeout = mcpServer.Spec.Timeout
+	} else {
+		remoteMCP.Spec.Timeout = &metav1.Duration{Duration: DefaultMCPServerTimeout}
+	}
+
+	return remoteMCP, nil
 }
 
 func (a *adkApiTranslator) translateRemoteMCPServerTarget(ctx context.Context, agent *adk.AgentConfig, remoteMcpServer *v1alpha2.RemoteMCPServer, mcpServerTool *v1alpha2.McpServerTool, agentHeaders map[string]string, proxyURL string) error {
--- a/go/go.mod
+++ b/go/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/jedib0t/go-pretty/v6 v6.7.8
-	github.com/kagent-dev/kmcp v0.2.5
+	github.com/kagent-dev/kmcp v0.2.6
 	github.com/kagent-dev/mockllm v0.0.3
 	github.com/modelcontextprotocol/go-sdk v1.2.0
 	github.com/muesli/reflow v0.3.0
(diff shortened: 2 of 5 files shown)


go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1259,16 +1266,30 @@ func ConvertMCPServerToRemoteMCPServer(mcpServer *v1alpha1.MCPServer) (*v1alpha2
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      mcpServer.Name,
 			Namespace: mcpServer.Namespace,
 		},
 		Spec: v1alpha2.RemoteMCPServerSpec{
-			URL:      fmt.Sprintf("http://%s.%s:%d/mcp", mcpServer.Name, mcpServer.Namespace, mcpServer.Spec.Deployment.Port),
-			Protocol: v1alpha2.RemoteMCPServerProtocolStreamableHttp,
+			URL:              fmt.Sprintf("http://%s.%s:%d/mcp", mcpServer.Name, mcpServer.Namespace, mcpServer.Spec.Deployment.Port),
+			Protocol:         v1alpha2.RemoteMCPServerProtocolStreamableHttp,
+			TerminateOnClose: ptr.To(true),
```
EItanya: What does this do? Is this something we want to automatically enable for every MCP connection?

skhedim: Good catch, it's out of scope for this PR (which focuses on timeout). Removed in the latest force-push.

For context: it was added to match the `+kubebuilder:default=true` on `RemoteMCPServerSpec.TerminateOnClose`, since this struct is built in-memory and never goes through API server defaulting. 

That said, the sibling `ConvertServiceToRemoteMCPServer` doesn't set it either, so leaving it `nil` is consistent with existing behavior.

PR #1284, opened 2026-02-11
GO ADK declarative image

yanivmn:
go-adk directory maintained separately until community approval.

Tested AzureOpenIA and Anthropic Models.

### DIFF
--- a/go/api/v1alpha2/agent_types.go
+++ b/go/api/v1alpha2/agent_types.go
@@ -146,9 +146,6 @@ type ByoDeploymentSpec struct {
 
 // +kubebuilder:validation:XValidation:message="serviceAccountName and serviceAccountConfig are mutually exclusive",rule="!(has(self.serviceAccountName) && has(self.serviceAccountConfig))"
 type SharedDeploymentSpec struct {
-	// Image overrides the default repository (e.g. "kagent-dev/kagent/app"). When set, used with ImageRegistry and tag to form the full image.
-	// +optional
-	Image string `json:"image,omitempty"`
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 	// +optional
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,3 @@
-### Makefile local overrides (e.g. proxy)
-config.mk
-buildx/config.mk
-
 ### dotenv template
 python/.env
 
--- a/Makefile
+++ b/Makefile
@@ -20,34 +20,10 @@ KUBECONFIG_PERM ?= $(shell \
   fi)
 
 
-# Optional config overrides
--include config.mk
-# Buildx proxy config: copy buildx/config.mk.example to buildx/config.mk and set
-# HTTP_PROXY/HTTPS_PROXY so the buildx builder can load base image metadata.
--include buildx/config.mk
-
-# Proxy for Docker buildx (BuildKit). Set in buildx/config.mk or env so the builder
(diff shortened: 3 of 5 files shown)


EItanya:
Hey @yanivmn, this PR is really coming together. I think I have 3 main comments remaining. 
1. I would like to see if we can use the [upstream A2A server](https://github.com/google/adk-go/tree/main/server/adka2a) completely. This should work now that [my PR](https://github.com/kagent-dev/kagent/pull/1294) has merged to support both metadata prefixes.
2. I think we need to figure out a layout of files and folders for go which actually makes sense rather than the current approach. I think a `go.work` is probably in order to split up the modules. If we do this I also think we need to have passing end to end tests for this feature. If we don't do this I think we should temporarily move all of the code into `contrib`, and then do the proper setup as a follow-up

EItanya:
> Hey @yanivmn, this PR is really coming together. I think I have 3 main comments remaining.
> 
> 1. I would like to see if we can use the [upstream A2A server](https://github.com/google/adk-go/tree/main/server/adka2a) completely. This should work now that [my PR](https://github.com/kagent-dev/kagent/pull/1294) has merged to support both metadata prefixes.
> 2. I think we need to figure out a layout of files and folders for go which actually makes sense rather than the current approach. I think a `go.work` is probably in order to split up the modules. If we do this I also think we need to have passing end to end tests for this feature. If we don't do this I think we should temporarily move all of the code into `contrib`, and then do the proper setup as a follow-up

I see a lot of changes have been made, but I don't see either of these questions really answered.

yanivmn:
Hi, i didnt had the time for it yet. 
Just rebased to latest. Will try to test it tomorrow

yanivmn:
@EItanya 
I tried to refactor and use upstream A2A server.
but Google ADK upstream response is mush different so i endup in converting responses with much new conversion code.
For example, upstream Google ADK executor places the agent's response into artifacts, not into history. So when the page reloads and calls extractMessagesFromTasks, it only extracts the user's message from history. The agent's complete response (which is in artifacts) is never extracted.

EItanya:
> @EItanya I tried to refactor and use upstream A2A server. but Google ADK upstream response is mush different so i endup in converting responses with much new conversion code. For example, upstream Google ADK executor places the agent's response into artifacts, not into history. So when the page reloads and calls extractMessagesFromTasks, it only extracts the user's message from history. The agent's complete response (which is in artifacts) is never extracted.

Ok, hopefully we can move in that direction in the future

yanivmn:
@EItanya 
Just pushed your requested changes to use upstream a2a package

go/api/v1alpha2/agent_types.go
```diff
@@ -146,6 +146,9 @@ type ByoDeploymentSpec struct {
 
 // +kubebuilder:validation:XValidation:message="serviceAccountName and serviceAccountConfig are mutually exclusive",rule="!(has(self.serviceAccountName) && has(self.serviceAccountConfig))"
 type SharedDeploymentSpec struct {
+	// Image overrides the default repository (e.g. "kagent-dev/kagent/app"). When set, used with ImageRegistry and tag to form the full image.
+	// +optional
+	Image string `json:"image,omitempty"`
```
EItanya: This can't be here because it's also in the `ByoDeploymentSpec` above.

EItanya (commented):
Can you please revert all changes not under `contrib/go-adk` to merge this. I also want to think through the API by which we enable this.

Issue #1270, opened 2026-02-06
[FEATURE] User RBAC Authorization for Kagent


nujragan93:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add role-based access control (RBAC) authorization for Kagent API, UI, and CLI with OIDC group integration

## Current Limitation
        Kagent currently lacks authorization for its UI and API. All users have full access to all resources across all namespaces. The `NoopAuthorizer` allows everything - there are no permission checks.

        ## Security Risks
        This creates security risks in multi-tenant environments where:
        - Users should only see resources in their assigned namespaces
        - Different roles need different permission levels (admins vs readonly users)
        - Sensitive data (API keys, secrets) should be masked from non-admins
        - Unauthorized users can create, modify, or delete critical resources

        ## Who Is Affected
        - **Platform teams** running Kagent in shared/multi-tenant environments
        - **Security teams** requiring audit trails and access control
        - **End users** who need read-only access without modification permissions
        - **Administrators** who need to enforce least-privilege access

        ## Why This Is Needed
        - Compliance requirements for production deployments
        - Prevent accidental or malicious resource modifications
        - Enable safe delegation of access to junior team members or contractors
        - Support enterprise OIDC/SSO integration for unified access control

## Overview
        Implement Casbin-based RBAC authorization enforced at the API layer (applies to UI, CLI, and direct API calls).

        ## Core Components

        ### 1. Role System
        - **2 built-in roles**: `admin` (full access) and `readonly` (read-only)
        - Support for **multiple roles per user**
        - **Default policy**: authenticated users without explicit roles get `admin` access
        - Extensible architecture for custom roles via Casbin policy configuration

        ### 2. Permission Model
        ```
        ┌─────────────────┬───────┬──────────┐
        │ Resource/Action │ admin │ readonly │
        ├─────────────────┼───────┼──────────┤
        │ Agent           │       │          │
        │   create        │ ✓     │ ✗        │
        │   get/list      │ ✓     │ ✓        │
        │   update        │ ✓     │ ✗        │
        │   delete        │ ✓     │ ✗        │
        │   invoke (chat) │ ✓     │ ✗        │
        ├─────────────────┼───────┼──────────┤
[...]

PR #1293, opened 2026-02-13
OIDC proxy auth

lets-call-n-walk:
## Summary

Adds optional OIDC proxy-based authentication to kagent, enabling integration with enterprise identity providers (Cognito, Okta, Dex, etc.) via an [oauth2-proxy](https://oauth2-proxy.github.io/oauth2-proxy/) subchart.

When `controller.auth.mode` is set to `"proxy"`, the controller trusts JWT tokens from the `Authorization` header injected by oauth2-proxy, extracting user identity from configurable JWT claims. When set to `"unsecure"` (the default), behavior is unchanged — no auth is required.

**Note:** This does not implement Access Control. It is purely the authentication mechanism.  Access control is being discussed in this issue: #1270 

### What's included

**Backend (Go)**
- `ProxyAuthenticator` that extracts user identity (email, name, groups) from JWT claims with configurable claim mapping
- `AUTH_MODE` env var to switch between `unsecure` and `proxy` authentication at startup
- `/api/me` endpoint returning the current user's identity from the auth context
- Nil-safety fix for database manager shutdown

**Frontend (Next.js)**
- `AuthContext` provider that decodes the JWT from the `Authorization` header and exposes user state
- `/login` page with branded SSO redirect
- `UserMenu` component showing current user info with sign-out
- Auth header forwarding in all server actions and the A2A proxy route
- `jose` dependency for client-side JWT decoding

**Helm**
- oauth2-proxy subchart dependency (disabled by default)
- `controller.auth` values for auth mode and JWT claim mapping
- `ui.auth.ssoRedirectPath` for configurable SSO redirect
- `NetworkPolicy` templates that lock down UI/controller access when auth is enabled
- Custom oauth2-proxy templates for branded login redirect
- `NOTES.txt` additions documenting auth configuration

### How it works

```
Browser → oauth2-proxy → (OIDC flow with IdP) → Next.js UI → Go Controller
                ↓                                     ↓              ↓
         Session cookie                    AuthContext decodes   ProxyAuthenticator
         + JWT injection                   JWT for user state    extracts claims
```

NetworkPolicies ensure UI and controller only accept traffic from oauth2-proxy when auth mode is `proxy`.

### Enabling

```yaml
controller:
  auth:
    mode: proxy

oauth2-proxy:
  enabled: true
  extraEnv:
    - name: OIDC_ISSUER_URL
      value: "https://your-idp.example.com"
    - name: OIDC_REDIRECT_URL
      value: "https://kagent.example.com/oauth2/callback"
```

See `docs/OIDC_PROXY_AUTH_ARCHITECTURE.md` for full architecture details.

### DIFF
--- a/ui/src/components/Header.stories.tsx
+++ b/ui/src/components/Header.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, StoryObj } from "@storybook/nextjs-vite";
 import { Header } from "./Header";
+import { AuthProvider } from "@/contexts/AuthContext";
 
 const meta = {
   title: "Components/Header",
@@ -7,6 +8,13 @@ const meta = {
   parameters: {
     layout: "fullscreen",
   },
+  decorators: [
+    (Story) => (
+      <AuthProvider>
+        <Story />
+      </AuthProvider>
+    ),
+  ],
 } satisfies Meta<typeof Header>;
 
 export default meta;


helm/kagent/values.yaml
```diff
@@ -380,6 +380,114 @@ querydoc:
...
+# They restrict access to UI and Controller to only allow traffic from oauth2-proxy
+# and internal cluster communication
+
+networkPolicy:
+  # Set to false to disable network policies even when auth is enabled
+  enabled: true
+  # Additional pod selector labels for oauth2-proxy (if using custom labels)
+  oauth2ProxySelector: {}
+  # Additional namespaces to allow traffic from (e.g., for monitoring)
+  additionalAllowedNamespaces: []
```
EItanya: Just for sheer size and breadth of this PR, can we move `NetworkPolicy` into a follow-up?

ui/src/app/login/page.tsx
```diff
@@ -0,0 +1,149 @@
+import Link from "next/link";
+
+// SSO redirect path - defaults to oauth2-proxy's start endpoint
+const SSO_REDIRECT_PATH = process.env.SSO_REDIRECT_PATH || "/oauth2/start";
+
+// Kagent logo SVG - extracted from official logo
+function KagentLogo({ size = 32 }: { size?: number }) {
```
peterj: there's a logo here - https://github.com/kagent-dev/kagent/blob/706a170c8e8c84dd2c1b5c5c64398c0fd2a6cc86/ui/src/components/kagent-logo.tsx#L4 -- can that be re-used instead of creating a new one?

ui/src/app/login/page.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export default function LoginPage() {
+  return (
+    <>
+      {/* Preload background image for faster rendering */}
+      <link rel="preload" href="/login-bg.webp" as="image" type="image/webp" fetchPriority="high" />
+
+      {/* CSS to hide header/footer and override main layout for fullscreen login */}
+      <style>{`
+        body:has(.login-page) > div header,
```
peterj: the project is using tailwind, can you convert this to tailwind classes instead?

ui/src/app/login/page.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+        {/* Main Content */}
+        <main className="relative z-10 h-full flex flex-col justify-center items-center text-center p-5">
+          <div className="bg-white/5 backdrop-blur-md border border-white/10 rounded-2xl px-6 py-8 md:px-12 md:py-10 max-w-[700px] flex flex-col items-center animate-in fade-in duration-500 delay-150 fill-mode-backwards">
+            <h1 className="flex items-center gap-3 md:gap-5 font-extrabold text-4xl md:text-[6rem] leading-tight tracking-tighter mb-4 text-white [text-shadow:0_0_20px_rgba(168,85,247,0.6),0_0_60px_rgba(168,85,247,0.4)]">
+              <KagentLogo size={80} />
+              <span>kagent</span>
+            </h1>
+            <p className="text-lg md:text-2xl text-gray-300 max-w-[600px] font-normal mb-10 leading-relaxed">
+              Orchestrating Your AI Future with Kubernetes and Beyond
```
peterj: since we have "Bringing Agentic AI to cloud native" on the website -- perhaps we do the same here?

ui/src/lib/jwt.ts
```diff
@@ -0,0 +1,62 @@
...
+export function decodeJWT(token: string): JWTPayload | null {
+  try {
+    return decodeJwt(token);
+  } catch {
+    return null;
+  }
+}
+
+export function extractUserFromClaims(claims: JWTPayload) {
+  return {
```
peterj: can we introduce a type for the { user, email, name, groups }?

ui/src/app/login/page.tsx
```diff
@@ -0,0 +1,149 @@
...
+      />
+    </svg>
+  );
+}
+
+export default function LoginPage() {
+  return (
+    <>
+      {/* Preload background image for faster rendering */}
+      <link rel="preload" href="/login-bg.webp" as="image" type="image/webp" fetchPriority="high" />
```
peterj, line 11: if the jpg version of the file is not used, we should remove it from the /public folder

go/internal/httpserver/auth/proxy_authn.go
```diff
@@ -0,0 +1,173 @@
...
+
+var ErrUnauthenticated = errors.New("unauthenticated: missing or invalid Authorization header")
+
+// ClaimsConfig holds configurable JWT claim names
+type ClaimsConfig struct {
+	UserID string // Default: "sub"
+	Email  string // Default: "email"
+	Name   string // Default: tries "name", "preferred_username"
+	Groups string // Default: tries "groups", "cognito:groups", "roles"
+}
```
EItanya: Do we need these to be hardcoded? Can we just passthrough the raw claims map as `map[string]any`?

go/cmd/controller/main.go
```diff
@@ -38,3 +41,18 @@ func main() {
 		}, nil
 	})
 }
+
+func getAuthenticator() pkgauth.AuthProvider {
+	switch os.Getenv("AUTH_MODE") {
```
EItanya: Can we use some sort of config struct here rather than a bunch of env vars? I'm a little worried about the amount of env var sprawl we have.

go/internal/httpserver/handlers/current_user.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/kagent-dev/kagent/go/pkg/auth"
+)
+
+type CurrentUserResponse struct {
+	User   string   `json:"user"`
+	Email  string   `json:"email"`
+	Name   string   `json:"name"`
+	Groups []string `json:"groups"`
+}
```
EItanya: Not every set of claims will have these values. Can we do this in a way that's not as prescriptive about the data available?

lets-call-n-walk:
@EItanya @peterj let me know if this addresses your feedback, thanks!

go/pkg/app/app.go
EItanya, line 389: Why do we need a webhook now?

go/core/internal/httpserver/auth/proxy_authn.go
```diff
@@ -0,0 +1,113 @@
...
+		P: auth.Principal{
+			User: auth.User{
+				ID: userID,
+			},
+			Agent: auth.Agent{
+				ID: agentID,
+			},
+		},
+		authHeader: authHeader,
+	}, nil
```
EItanya, line 85: We should probably create a follow-up for this to make sure it's more secure

go/core/internal/httpserver/auth/proxy_authn.go
```diff
@@ -0,0 +1,113 @@
...
+func (a *ProxyAuthenticator) Authenticate(ctx context.Context, reqHeaders http.Header, query url.Values) (auth.Session, error) {
+	authHeader := reqHeaders.Get("Authorization")
+
+	// Always read agent identity from X-Agent-Name header (used by agents calling back)
+	agentID := reqHeaders.Get("X-Agent-Name")
+
+	// If we have a Bearer token, parse JWT
+	if tokenString, ok := strings.CutPrefix(authHeader, "Bearer "); ok {
+		// Parse JWT without validation (oauth2-proxy or k8s service account already validated)
+		rawClaims, err := parseJWTPayload(tokenString)
```
EItanya, line 37: Can we call this mode `trusted-proxy` instead of just proxy because it's explicitly not re-validating

go/pkg/app/app.go
EItanya, line 111: Still not sure why we need this

EItanya, line 155: Same

go/pkg/app/app.go
```diff
@@ -463,17 +459,6 @@ func Start(getExtensionConfig GetExtensionConfig) {
 		os.Exit(1)
 	}
```
EItanya, line 564: I feel like the git diff is a bit broken, why is this saying it's added here?

lets-call-n-walk:
@EItanya renamed proxy to trusted proxy. Most of the other comments were on code that wasn't mine, but the diff had gotten messed up from being behind - I merged and the diff is correct now.

dobesv:
This is a nice feature, however ...

One issue I think you run into once you actually have different users is there's no administrative or sharing or permissions functions in kagent.  What if your team wants kagent sessions to be visible/shared team wide?  Currently in kagent it's very strict about not letting you see any other session except the ones you initiated (even sub-agent sesssions are inaccessible in the kagent permission model).

dobesv:
I did have an alternative approach here just looking at headers.  You can set an env var on kagent like TRUSTED_USER_HEADERS and specify the headers to look at.  AWS ALB and oauth2 proxy will give you a header with a User ID in it that you can use.  kagent can be configured to use given header(s) as the user ID if detected.  I think it's simpler (doesn't require JWT parsing) and also supports more reverse proxies (ones that don't use JWT)

lets-call-n-walk:
> I did have an alternative approach here just looking at headers. You can set an env var on kagent like TRUSTED_USER_HEADERS and specify the headers to look at. AWS ALB and oauth2 proxy will give you a header with a User ID in it that you can use. kagent can be configured to use given header(s) as the user ID if detected. I think it's simpler (doesn't require JWT parsing) and also supports more reverse proxies (ones that don't use JWT)

The reason why JWTs are desired is so that the organization can give more data than just USER_ID, such as groups. Then that JWT can be passed to an RBAC service (which is also being worked on), and used to verify permissions for said user. (which would then start to address the first concern of allowing users to share sessions.).

lets-call-n-walk:
@EItanya Just following up - have you had a chance to review this? It looks like this is out of date again.

go/pkg/app/app.go
```diff
@@ -165,10 +169,13 @@ func (cfg *Config) SetFlags(commandLine *flag.FlagSet) {
 
 	commandLine.Var(&cfg.Streaming.MaxBufSize, "streaming-max-buf-size", "The maximum size of the streaming buffer.")
 	commandLine.Var(&cfg.Streaming.InitialBufSize, "streaming-initial-buf-size", "The initial size of the streaming buffer.")
-	commandLine.DurationVar(&cfg.Streaming.Timeout, "streaming-timeout", 600*time.Second, "The timeout for the streaming connection.")
+	commandLine.DurationVar(&cfg.Streaming.Timeout, "streaming-timeout", 60*time.Second, "The timeout for the streaming connection.")
 
 	commandLine.StringVar(&cfg.Proxy.URL, "proxy-url", "", "Proxy URL for internally-built k8s URLs (e.g., http://proxy.kagent.svc.cluster.local:8080)")
 
+	commandLine.StringVar(&cfg.Auth.Mode, "auth-mode", "unsecure", "Authentication mode: unsecure or proxy")
```
jsonmp-k8: `getAuthenticator` only recognizes `trusted-proxy`, but this help text says `unsecure or proxy`. If an operator sets `AUTH_MODE=proxy` from `--help`, the controller silently falls back to `UnsecureAuthenticator` and starts insecurely. This should either document `trusted-proxy` here or reject unknown modes at startup.

ui/src/app/actions/feedback.ts
```diff
@@ -1,20 +1,18 @@
...
+import { fetchApi } from "./utils";
 
 /**
  * Submit feedback to the server
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 async function submitFeedback(feedbackData: FeedbackData): Promise<any> {
-    const userID = await getCurrentUserId();
-    const body = { 
+    const body = {
```
jsonmp-k8, line 11: Removing `user_id` from this payload looks like a regression unless the backend now injects it from the auth session. `HandleCreateFeedback` still persists the request body verbatim, so `database.Feedback.UserID` is stored as empty here and `ListFeedback(userID)` will never return these rows.

helm/kagent/values.yaml
```diff
@@ -387,6 +401,99 @@ querydoc:
...
+  # Cluster-specific OIDC settings - override these per deployment
+  # These are set as env vars and referenced in args for easy patching
+  extraEnv:
+    - name: OIDC_ISSUER_URL
+      value: ""
+    - name: OIDC_REDIRECT_URL
+      value: ""
+    # Default assumes release name "kagent". Override if using different release name.
+    - name: UPSTREAM_URL
+      value: "http://kagent-ui:8080"
```
jsonmp-k8, line 568: Hard-coding `http://kagent-ui:8080` only works for the default release name. The chart templates the actual service name as `{{ include "kagent.fullname" . }}-ui`, so `fullnameOverride` or a non-default release name will break oauth2-proxy unless the user manually overrides `UPSTREAM_URL`.

go/core/pkg/app/app.go
```diff
@@ -165,10 +169,13 @@ func (cfg *Config) SetFlags(commandLine *flag.FlagSet) {
 
 	commandLine.Var(&cfg.Streaming.MaxBufSize, "streaming-max-buf-size", "The maximum size of the streaming buffer.")
 	commandLine.Var(&cfg.Streaming.InitialBufSize, "streaming-initial-buf-size", "The initial size of the streaming buffer.")
-	commandLine.DurationVar(&cfg.Streaming.Timeout, "streaming-timeout", 600*time.Second, "The timeout for the streaming connection.")
+	commandLine.DurationVar(&cfg.Streaming.Timeout, "streaming-timeout", 60*time.Second, "The timeout for the streaming connection.")
```
jsonmp-k8, line 171: This changes the default streaming timeout from 10 minutes to 1 minute. `A2ARegistrar` wires this value straight into `a2aclient.WithTimeout`, so long-running streams and tool calls will now terminate after 60s unless the operator overrides the flag. Was that intentional for this auth PR?

go/core/internal/httpserver/auth/proxy_authn.go
```diff
@@ -0,0 +1,113 @@
...
+			},
+			Agent: auth.Agent{
+				ID: agentID,
+			},
+		},
+		authHeader: authHeader,
+	}, nil
+}
+
+func (a *ProxyAuthenticator) UpstreamAuth(r *http.Request, session auth.Session, upstreamPrincipal auth.Principal) error {
```
jsonmp-k8, line 88: This drops `X-User-Id` on controller-to-agent A2A calls. The downstream A2A runtimes still derive the caller from that header, so in `trusted-proxy` mode the real user becomes `A2A_USER_<context_id>` after the first hop. We likely need to forward the resolved user id here in addition to the bearer token.

jholt96:
> > I did have an alternative approach here just looking at headers. You can set an env var on kagent like TRUSTED_USER_HEADERS and specify the headers to look at. AWS ALB and oauth2 proxy will give you a header with a User ID in it that you can use. kagent can be configured to use given header(s) as the user ID if detected. I think it's simpler (doesn't require JWT parsing) and also supports more reverse proxies (ones that don't use JWT)
> 
> The reason why JWTs are desired is so that the organization can give more data than just USER_ID, such as groups. Then that JWT can be passed to an RBAC service (which is also being worked on), and used to verify permissions for said user. (which would then start to address the first concern of allowing users to share sessions.).

In addition to the reason above, passing JWT directly also allows for supporting user identity propagation to downstream services. Some enterprise Cyber practices don't like OBO and prefer propagation. I think it's great to be able to support both with kagent directly being able to propagate identity, and then having agent gateway integration for OBO and more fine-grained control.

christensenjairus:
> This pull request has been marked as stale because of no activity in the last 15 days. It will be closed in the next 5 days unless it is tagged "no stalebot" or other activity occurs.

Not stale! Would love to see this completed

jholt96:
This is a must have for the project! Please let me know what is needed to get this merged. Happy to assist @lets-call-n-walk @EItanya

lets-call-n-walk:
@EItanya @jsonmp-k8 updated per review and fixed merge conflicts

EItanya:
Hey there, sorry for the delay, I updated the PR so it should now pass in CI. I will do a final review tomorrow

PR #1297, opened 2026-02-14
fix: propagate trace context end-to-end for agent Services

### LINKED ISSUE #1295: Trace context (traceparent) not propagated from controller to agent pods
## Summary

W3C TraceContext headers (`traceparent`, `tracestate`) are not propagated from the kagent controller to agent pods, breaking distributed tracing across the controller-agent boundary.

When an upstream service (e.g., AgentGateway) sends a request with a `traceparent` header to the kagent controller, the trace context is lost when the controller forwards the request to the agent pod via the A2A client. This results in disconnected traces — the upstream service's spans appear under one trace ID while the agent's spans appear under a completely different trace ID.

## Root Cause

There are two independent issues:

### 1. Go Controller: A2A client drops incoming HTTP headers

The `A2ARequestHandler` in `go/internal/httpserver/auth/authn.go` (line 82-117) only injects `X-User-Id` and `Authorization` headers into the outgoing request to agent pods. The trpc-a2a-go A2A client constructs a new HTTP request from the JSON-RPC payload, so the original `traceparent`/`tracestate` headers from the incoming request are not carried over.

**Flow where headers are lost:**
```
Incoming HTTP request (with traceparent)
  → A2A Server (trpc-a2a-go) receives request with headers ✅
  → PassthroughManager.OnSendMessage() — only passes protocol.SendMessageParams (no HTTP headers)
  → A2AClient.SendMessage() — constructs NEW HTTP request
  → A2ARequestHandler — only adds auth headers, NOT trace headers ❌
  → Agent pod receives request WITHOUT traceparent
```

### 2. Python Agent: No W3C TraceContext propagator configured

In `python/packages/kagent-core/src/kagent/core/tracing/_utils.py`, the `configure()` function sets up `FastAPIInstrumentor` but does not call `set_global_textmap_propagator(TraceContextTextMapPropagator())`. Even if the `traceparent` header reached the agent pod, the Python OTEL SDK would not extract it.

Note: The `OTEL_PROPAGATORS` environment variable does **not** automatically configure the propagator in the Python OTEL SDK — explicit code is required.

### Go Controller (authn.go)

Store trace headers from the incoming request in context, then inject them in `A2ARequestHandler`:

```go
// In middleware.go or similar — capture trace headers into context
type traceHeadersKey struct{}

func traceHeaderMiddleware(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        ctx := r.Context()
        if tp := r.Header.Get("traceparent"); tp != "" {
            ctx = context.WithValue(ctx, traceHeadersKey{}, map[string]string{
                "traceparent": tp,
                "tracestate":  r.Header.Get("tracestate"),
            })
        }
        next.ServeHTTP(w, r.WithContext(ctx))
    })
}

// In A2ARequestHandler — inject trace headers into outgoing request
if headers, ok := ctx.Value(traceHeadersKey{}).(map[string]string); ok {
    if tp := headers["traceparent"]; tp != "" {
        req.Header.Set("traceparent", tp)
    }
    if ts := headers["tracestate"]; ts != "" {
        req.Header.Set("tracestate", ts)
    }
}
```

### Python Agent (_utils.py)

Add propagator setup after TracerProvider configuration:

```python
from opentelemetry.propagate import set_global_textmap_propagator
from opentelemetry.propagators.composite import CompositePropagator
from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator

# Inside configure(), after trace.set_tracer_provider():
set_global_textmap_propagator(
    CompositePropagator([TraceContextTextMapPropagator()])
)
```

## Reproduction

1. Deploy kagent with tracing enabled (OTEL_TRACING_ENABLED=true)
2. Send a request with a known traceparent header:
   ```bash
   curl -H "traceparent: 00-aaaabbbbccccddddeeee111122223344-1234567890abcdef-01" \
     -H "Content-Type: application/json" \
     -X POST http://<gateway>/api/a2a/<namespace>/<agent>/ \
     -d '{"jsonrpc":"2.0","method":"message/send","params":{"message":{"role":"user","parts":[{"kind":"text","text":"hello"}]}},"id":"test"}'
   ```
3. Check traces in your OTEL backend — the agent spans will have a different trace ID than `aaaabbbbccccddddeeee111122223344`

## Environment

- kagent version: latest main branch (as of 2026-02-13)
- Upstream: AgentGateway (sends traceparent headers correctly)
- OTEL backend: Langfuse (via OTEL Collector)
- Python OTEL SDK: opentelemetry-sdk
- trpc-a2a-go: used for A2A client/server in Go controller

syn-zhu:
## Summary

Three fixes to enable end-to-end W3C TraceContext propagation across the controller→agent boundary:

1. **AppProtocol on agent Services** — Set `appProtocol: kgateway.dev/a2a` on the Service port created for each Agent CR so AgentGateway's A2A plugin can discover agent Services directly via protocol matching, rather than proxying through the kagent controller (which drops HTTP headers including `traceparent`).

2. **W3C TraceContext propagator in Python SDK** — Configure the W3C TraceContext propagator in `kagent-core` tracing setup so agent pods correctly extract incoming `traceparent` headers and propagate them on outgoing requests.

3. **Trace header propagation in Go controller** — The A2A server deserializes incoming HTTP requests into JSON-RPC params, discarding the original HTTP headers. When the controller forwards requests to agent pods via the A2A client, `traceparent`/`tracestate` are lost. Fix: capture W3C trace context headers from the incoming request into the Go context in the A2A auth middleware (`A2AAuthenticator.Wrap`), then inject them into outgoing requests in `A2ARequestHandler`.

All golden test outputs have been updated to include the new `appProtocol` field, including `agent_with_passthrough` (added in #1327).

Incorporates changes from https://github.com/opspawn/kagent/commit/d9f2a3a45a26ec11dd4a9a4cf18e4374a374b03f.

## Test plan

- [x] Golden tests updated and passing — all `testdata/outputs/*.json` files include `appProtocol: "kgateway.dev/a2a"` on Service ports
- [x] `go test ./internal/httpserver/auth/...` passes
- [ ] Deploy agent, verify `kubectl get svc <agent> -o jsonpath='{.spec.ports[0].appProtocol}'` returns `kgateway.dev/a2a`
- [ ] Verify AgentGateway A2A plugin discovers the Service
- [ ] Send request with `traceparent` header through gateway → controller → agent pod, verify trace ID is preserved end-to-end

Closes #1295

### DIFF
--- a/go/core/pkg/auth/auth.go
+++ b/go/core/pkg/auth/auth.go
@@ -58,12 +58,8 @@ type Authorizer interface {
 // context utils
 
 type sessionKeyType struct{}
-type traceHeadersKeyType struct{}
 
-var (
-	sessionKey      = sessionKeyType{}
-	traceHeadersKey = traceHeadersKeyType{}
-)
+var sessionKey = sessionKeyType{}
 
 func AuthSessionFrom(ctx context.Context) (Session, bool) {
 	v, ok := ctx.Value(sessionKey).(Session)
@@ -72,19 +68,6 @@ func AuthSessionFrom(ctx context.Context) (Session, bool) {
 
 func AuthSessionTo(ctx context.Context, session Session) context.Context {
 	return context.WithValue(ctx, sessionKey, session)
-}
-
-// TraceHeadersFrom retrieves W3C trace context headers (traceparent, tracestate)
-// that were captured from an incoming request.
-func TraceHeadersFrom(ctx context.Context) (http.Header, bool) {
-	v, ok := ctx.Value(traceHeadersKey).(http.Header)
-	return v, ok && v != nil
-}
-
-// TraceHeadersTo stores W3C trace context headers in the context so they can
-// be propagated to outgoing requests (e.g., from the controller to agent pods).
-func TraceHeadersTo(ctx context.Context, headers http.Header) context.Context {
-	return context.WithValue(ctx, traceHeadersKey, headers)
 }
 
 func AuthnMiddleware(authn AuthProvider) func(http.Handler) http.Handler {
--- a/python/packages/kagent-core/src/kagent/core/tracing/_utils.py
+++ b/python/packages/kagent-core/src/kagent/core/tracing/_utils.py
@@ -5,9 +5,6 @@ from fastapi import FastAPI
 from opentelemetry import _logs, trace
 from opentelemetry.exporter.otlp.proto.grpc._log_exporter import OTLPLogExporter
 from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
-from opentelemetry.propagate import set_global_textmap
-from opentelemetry.propagators.composite import CompositeHTTPPropagator
-from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
 from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
 from opentelemetry.instrumentation.openai import OpenAIInstrumentor
@@ -80,9 +77,6 @@ def configure(name: str = "kagent", namespace: str = "kagent", fastapi_app: Fast
     # Configure tracing if enabled
     if tracing_enabled:
         logging.info("Enabling tracing")
-        # Set up W3C TraceContext propagator so incoming traceparent headers
-        # are extracted and outgoing requests carry them forward.
-        set_global_textmap(CompositeHTTPPropagator([TraceContextTextMapPropagator()]))
         # Check standard OTEL env vars: signal-specific endpoint first, then general endpoint
         trace_endpoint = (
             os.getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")
--- a/go/core/internal/httpserver/auth/authn.go
+++ b/go/core/internal/httpserver/auth/authn.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/kagent-dev/kagent/go/core/internal/tracecontext"
 	"github.com/kagent-dev/kagent/go/core/pkg/auth"
 	"k8s.io/apimachinery/pkg/types"
 )
@@ -81,7 +82,7 @@ func (p *A2AAuthenticator) Wrap(next http.Handler) http.Handler {
 			if ts := r.Header.Get("tracestate"); ts != "" {
 				traceHeaders.Set("tracestate", ts)
 			}
-			r = r.WithContext(auth.TraceHeadersTo(r.Context(), traceHeaders))
+			r = r.WithContext(tracecontext.HeadersTo(r.Context(), traceHeaders))
 		}
 		authn.ServeHTTP(w, r)
 	})
@@ -122,7 +123,7 @@ func A2ARequestHandler(authProvider auth.AuthProvider, agentNns types.Namespaced
 		}
 
 		// Propagate W3C trace context headers captured from the incoming request.
-		if traceHeaders, ok := auth.TraceHeadersFrom(ctx); ok {
+		if traceHeaders, ok := tracecontext.HeadersFrom(ctx); ok {
 			for key, values := range traceHeaders {
 				for _, v := range values {
 					req.Header.Set(key, v)
--- a/go/core/internal/tracecontext/tracecontext.go
+++ b/go/core/internal/tracecontext/tracecontext.go
@@ -1,0 +1,23 @@
+package tracecontext
+
+import (
+	"context"
+	"net/http"
+)
+
+type traceHeadersKeyType struct{}
+
+var traceHeadersKey = traceHeadersKeyType{}
+
+// HeadersFrom retrieves W3C trace context headers (traceparent, tracestate)
+// that were captured from an incoming request.
+func HeadersFrom(ctx context.Context) (http.Header, bool) {
+	v, ok := ctx.Value(traceHeadersKey).(http.Header)
+	return v, ok && v != nil
+}
+
+// HeadersTo stores W3C trace context headers in the context so they can
+// be propagated to outgoing requests (e.g., from the controller to agent pods).
+func HeadersTo(ctx context.Context, headers http.Header) context.Context {
+	return context.WithValue(ctx, traceHeadersKey, headers)
+}


EItanya:
Hey there, thanks for the PR, this is a great idea! You will need to update the goldens as well as sign your commits for us to merge this

syn-zhu:
> Hey there, thanks for the PR, this is a great idea! You will need to update the goldens as well as sign your commits for us to merge this

Thanks! Just saw this now, but it looks like the issue was already fixed by https://github.com/opspawn/kagent/commit/d9f2a3a45a26ec11dd4a9a4cf18e4374a374b03f :)

Gonna just close this PR, thanks!

syn-zhu:
Oops @EItanya i realized the commit I linked wasn't actually a merged commit, but rather a branch. I've updated and reopened my PR to address the things you mentioned. Please lmk if there's anything else!

krisztianfekete:
Hi @syn-zhu, Could you please test this locally on your end first? As it's Claude-generated code, a brief manual validation, e.g. such as posting before/after screenshots from a tracing tool would be the minimum step to ensure it's ready for contribution.

syn-zhu:
## End-to-End Test Results: Trace Propagation

Tested on a live EKS cluster running kagent v0.7.13 with Langfuse (OTLP-backed) as the trace backend. Each test sends an A2A `tasks/send` request with a known `traceparent` header and verifies what Langfuse receives.

### Setup

- **Agent**: `platform_assistant` (Declarative agent with Anthropic via agentgateway-proxy)
- **LLM path**: Agent → agentgateway-proxy → Anthropic API
- **OTEL path**: Agent → gRPC OTLP → otel-collector → HTTP OTLP → Langfuse
- **Controller**: kagent-controller proxies A2A requests to agent pods

---

### Stage 0 — Baseline (upstream v0.7.13, no patches)

| Metric | Value |
|--------|-------|
| Trace ID | `00000000aaaa11112222333344445555` |
| Observations | 66 |
| Services in trace | `platform_assistant` only |
| AgentGateway spans | **0** ❌ |

**Problem**: Agent uses the default `CompositeTextMapPropagator` but never injects `traceparent` into outbound HTTP headers. LLM calls to agentgateway-proxy start a new disconnected trace.

---

### Stage 1 — Commit 1 only (Python SDK: W3C propagator + AioHttpClientInstrumentor)

Sent request **directly to agent pod** (bypassing controller) to isolate the Python SDK changes.

| Metric | Value |
|--------|-------|
| Trace ID | `11111111bbbb22223333444455556666` |
| Observations | 130 |
| Services in trace | `platform_assistant` + `agentgateway-proxy` |
| AgentGateway spans | **4** ✅ |

**Result**: `TraceContextTextMapPropagator` + `AioHttpClientInstrumentor` causes the agent's outbound LLM calls (via litellm's aiohttp transport) to inject `traceparent`/`tracestate` headers. AgentGateway sees the trace context and its spans now join the same trace.

---

### Stage 2 — Both commits (Python SDK + Go controller trace propagation)

A2A request sent **through the controller** (the production path).

#### Before (upstream controller, patched agent image)

| Metric | Value |
|--------|-------|
| Trace ID sent | `33333333dddd44445555666677778888` |
| Trace found in Langfuse | **NOT FOUND** ❌ |
| Orphan trace ID | `1159632b4b40815148c345bc9ead6dbf` |
| Orphan observations | 64 |
| Orphan services | `platform_assistant` + `agentgateway-proxy` + `agentgateway-waypoint` |

**Problem**: The upstream controller strips `traceparent` when proxying A2A requests to agent pods. The agent creates a new root trace (`1159632b...`), disconnected from the caller's trace. The agent's outbound calls to agentgateway DO propagate correctly (thanks to Commit 1), but on the wrong trace ID.

#### After (patched controller + patched agent)

| Metric | Value |
|--------|-------|
| Trace ID sent | `22222222cccc33334444555566667777` |
| Trace found in Langfuse | **YES** ✅ |
| Observations | 69 |
| Services in trace | `platform_assistant` + `agentgateway-proxy` |

**Result**: The patched controller extracts `traceparent`/`tracestate` from the incoming A2A request and re-injects them when proxying to the agent pod. The full chain — caller → controller → agent → agentgateway → LLM — now shares a single unified trace ID.

---

### Summary

| What | Before | After |
|------|--------|-------|
| Agent → AgentGateway propagation | ❌ Disconnected | ✅ Same trace |
| Controller → Agent propagation | ❌ traceparent dropped | ✅ traceparent forwarded |
| End-to-end trace unity | ❌ 2-3 disconnected traces | ✅ Single trace across all services |

syn-zhu:
Screenshot 2026-02-25 at 10 26 47 PM
Screenshot 2026-02-25 at 11 02 49 PM
Screenshot 2026-02-25 at 10 58 28 PM
Screenshot 2026-02-25 at 10 46 38 PM
Screenshot 2026-02-25 at 10 46 31 PM
Screenshot 2026-02-25 at 10 27 28 PM
Screenshot 2026-02-25 at 10 26 57 PM

syn-zhu:
> Hi @syn-zhu, Could you please test this locally on your end first? As it's Claude-generated code, a brief manual validation, e.g. such as posting before/after screenshots from a tracing tool would be the minimum step to ensure it's ready for contribution.

updated

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
```diff
@@ -63,6 +80,9 @@ def configure(name: str = "kagent", namespace: str = "kagent", fastapi_app: Fast
     # Configure tracing if enabled
     if tracing_enabled:
         logging.info("Enabling tracing")
+        # Set up W3C TraceContext propagator so incoming traceparent headers
+        # are extracted and outgoing requests carry them forward.
+        set_global_textmap(CompositeHTTPPropagator([TraceContextTextMapPropagator()]))
```
krisztianfekete: Are you sure this is necessary? If it is (but I don't think it is), we should at least preserve the existing propagators that this overrides.

go/core/pkg/auth/auth.go
krisztianfekete, line 1: These are unrelated to auth. Can we move these into internal/ or somewhere tracing-specific?

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
syn-zhu: Good catch, no it isn't necessary anymore. addressed in most recent commit

go/core/pkg/auth/auth.go
syn-zhu: addressed in most recent commits!

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -544,9 +544,10 @@ func (a *adkApiTranslator) buildManifest(
 		Spec: corev1.ServiceSpec{
 			Selector: selectorLabels,
 			Ports: []corev1.ServicePort{{
-				Name:       "http",
-				Port:       dep.Port,
-				TargetPort: intstr.FromInt(int(dep.Port)),
+				Name:        "http",
+				Port:        dep.Port,
+				TargetPort:  intstr.FromInt(int(dep.Port)),
+				AppProtocol: ptr.To("kgateway.dev/a2a"),
```
onematchfox, line 550: Whilst I'm completely for integrating with other projects, should this really be a hard-coded value that is applied to all Agents (regardless of whether users actually make use of KGateway/AgentGateway)? 

What happens when this value changes; there seems to be a bit of a [split](https://github.com/kgateway-dev/kgateway/issues/13492) going on between kgateway and agentgateway - so what would happen if this changes from `kgateway.dev/a2a` to `agentgateway.dev/a2a`? And what happens if I want to use a different `appProtocol` for whatever reason? Or, what happens if I only want some agents (not all) to be picked up via the A2A plugin?

Maybe a better approach here would be to allow users to configure this via the `Agent` CRD instead?

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
```diff
@@ -91,6 +105,7 @@ def configure(name: str = "kagent", namespace: str = "kagent", fastapi_app: Fast
             logging.info("Created new TracerProvider")
 
         HTTPXClientInstrumentor().instrument()
+        _instrument_aiohttp_client()
```
onematchfox, line 108: What value do we get out of this? Outbound LLM calls are already instrumented from what I can tell - see `openai.chat` span for example.

go/core/internal/controller/translator/agent/adk_api_translator.go
syn-zhu: > Maybe a better approach here would be to allow users to configure this via the Agent CRD instead?

Completely agree, and was just thinking to make this change myself. Will update the PR

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
syn-zhu: See the Stage 0 vs Stage 1 comparison in my earlier comment: https://github.com/kagent-dev/kagent/pull/1297#issuecomment-3963860371

The problem is correlation with the *incoming* request; if the incoming request to the agent had trace headers, they currently don't get propagated. the outbound calls start a completely disjoint trace.

Adding this change makes it possible to trace long, complex agent request flows (e.g. user calls Agent 1 who calls Agent 2 and Agent 3, who each call MCP server etc etc) with a single root trace ID

Does this make sense?

syn-zhu: Having said that, I actually wanna triple-check this earlier comment: https://github.com/kagent-dev/kagent/pull/1297/changes/a48fbcf42cb250ade0bf4b5a4d59509b39645a87#r2859471141

Need to make sure that change which I removed is *really* not needed. Might need to add it back.

PR #1323, opened 2026-02-17
Memory, context cache/compaction, latest ADK

dobesv:
Update to the latest ADK and support configuration of new ADK features:

- context compaction: automatically based on the number of events or tokens
- context cache config
- memory, either using provided in-memory or vertex memory, or delegating to an MCP server
- resumability - ability to enable resumability

After this I am hoping get these working:

- Show in the UI if compaction is running (similar UI as "thinking...") ideally
- Add a button to request manual compaction
- Attempt to resume sessions automatically if the pod was restarted (and maybe in some other cases?  Not sure)


dobesv:
Most of this is just allowing configuration of  built-in ADK features so it doesn't seem like too much innovation.  The memory proxy maybe is a bit more of a design question.  

Context compression was my main pain point with kagent so far, it's pretty useless without it.  Do you think you would accept a pull request with just the context options?

Or maybe just rip out the memory proxy from this but leave all the built-in ADK features?

When are the meetings?

dobesv:
Will open a new pull request with only context configs (cache and compression) and revisit the resumability and memory features later.

PR #1326, opened 2026-02-18
chore(python): bump a2a-sdk to 0.3.23

lithammer:
The main motivation is that it includes this fix:

- https://github.com/a2aproject/a2a-python/pull/614

It's described in more detail from a kagent perspective here:

- https://github.com/kagent-dev/kagent/pull/1268


lithammer:
I updated the version across the board, not sure if that's correct/wanted or not.

EItanya:
Thanks so much for the PR, you will need to sign your commits for us to merge it

lithammer:
> Thanks so much for the PR, you will need to sign your commits for us to merge it

Oops, forgot about that. My bad. Should be fixed now.

PR #1327, opened 2026-02-18
feat: add API key passthrough for ModelConfig

EItanya:
Allow organizations with federated identity to forward the Bearer token from incoming A2A requests directly to LLM providers as the API key, eliminating the need for separate secret management.

Changes:
- Add apiKeyPassthrough field to ModelConfig CRD with mutual exclusion validation against apiKeySecret
- Add api_key_passthrough to Go ADK BaseModel types and MarshalJSON
- Update translator to skip API key env vars when passthrough is enabled
- Create LLMPassthroughPlugin using a SupportsPassthroughAuth protocol to set the API key from the Bearer token before each LLM call
- Add KAgentLiteLlm subclass to properly support passthrough on LiteLLM models without monkey-patching
- Add set_passthrough_key() to BaseOpenAI for typed model mutation
- Add e2e test with mockllm header matching to verify token forwarding
- Fix flaky TLS e2e test by using OS-assigned ephemeral ports

For posterity, I've been enjoying posting the costs of these sessions
```
❯ /cost                                                                                                           
  ⎿  Total cost:            $29.50
     Total duration (API):  52m 58s                                                                               
     Total duration (wall): 2h 30m 3s                                                                           
     Total code changes:    1098 lines added, 385 lines removed                                                   
     Usage by model:                                                                                              
         claude-haiku-4-5:  195.2k input, 42.1k output, 5.8m cache read, 548.7k cache write ($1.67)
          claude-opus-4-6:  32.7k input, 132.2k output, 30.6m cache read, 1.5m cache write ($27.83)   
```

### DIFF
--- a/go/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/internal/controller/translator/agent/adk_api_translator.go
@@ -865,7 +865,7 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 			Name:  "GOOGLE_GENAI_USE_VERTEXAI",
 			Value: "true",
 		})
-		if !model.Spec.APIKeyPassthrough && model.Spec.APIKeySecret != "" {
+		if model.Spec.APIKeySecret != "" {
 			modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
 				Name:  "GOOGLE_APPLICATION_CREDENTIALS",
 				Value: "/creds/" + model.Spec.APIKeySecretKey,
@@ -906,7 +906,7 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 			Name:  "GOOGLE_CLOUD_LOCATION",
 			Value: model.Spec.AnthropicVertexAI.Location,
 		})
-		if !model.Spec.APIKeyPassthrough && model.Spec.APIKeySecret != "" {
+		if model.Spec.APIKeySecret != "" {
 			modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
 				Name:  "GOOGLE_APPLICATION_CREDENTIALS",
 				Value: "/creds/" + model.Spec.APIKeySecretKey,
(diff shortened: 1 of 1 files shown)


go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -805,17 +807,19 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 		if model.Spec.AzureOpenAI == nil {
 			return nil, nil, nil, fmt.Errorf("AzureOpenAI model config is required")
 		}
-		modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
-			Name: "AZURE_OPENAI_API_KEY",
-			ValueFrom: &corev1.EnvVarSource{
-				SecretKeyRef: &corev1.SecretKeySelector{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: model.Spec.APIKeySecret,
+		if !model.Spec.APIKeyPassthrough {
```
jmhbh, line 810: do we need to validate that `model.spec.APIKeySecret != ""` here as well?

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -949,20 +956,23 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
...
 
 		return ollama, modelDeploymentData, secretHashBytes, nil
 	case v1alpha2.ModelProviderGemini:
-		modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
-			Name: "GOOGLE_API_KEY",
-			ValueFrom: &corev1.EnvVarSource{
-				SecretKeyRef: &corev1.SecretKeySelector{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: model.Spec.APIKeySecret,
+		if !model.Spec.APIKeyPassthrough {
```
jmhbh: nit: same question as above

python/packages/kagent-adk/src/kagent/adk/_llm_passthrough_plugin.py
```diff
@@ -0,0 +1,60 @@
...
+    async def before_model_callback(
+        self, *, callback_context: CallbackContext, llm_request: LlmRequest
+    ) -> Optional[LlmResponse]:
+        token = _extract_bearer_token(callback_context)
+        if not token:
+            return None
+
+        model = callback_context._invocation_context.agent.model
+        if not isinstance(model, SupportsPassthroughAuth):
+            return None
```
jmhbh, line 54: nit: might be worth adding debug logs here indicating the model does not support passthrough

go/internal/controller/translator/agent/adk_api_translator.go
EItanya: I think they didn't check right now because it's assumed to be required. I don't want to mess with it right now because I don't want to break anything, but I think we should think about adding better validation either in code or in CEL in the future

go/internal/controller/translator/agent/adk_api_translator.go
EItanya: Same answer as above

PR #1333, opened 2026-02-18
Restore chat status/subscription when loading a chat

dobesv:
Previously the UI would just assume a chat was idle when loading it.  This fetches the chat status and displays an appropriate state.  It will resubscribe for live updates in some cases as well.

This makes the experience a lot nicer when switching between chats or reloading the page while looking at a chat.

### DIFF
--- a/ui/src/components/chat/ChatInterface.tsx
+++ b/ui/src/components/chat/ChatInterface.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import type React from "react";
-import { useState, useRef, useEffect } from "react";
+import { useState, useRef, useEffect, useCallback, useMemo } from "react";
 import { ArrowBigUp, X, Loader2, Mic, Square } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import {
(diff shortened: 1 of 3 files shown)


dobesv:
I've been using this for a while today and it's a big improvement although I still feel like I'm not always getting updates to the chat unless I actually sent the last message without navigating away and back.  So maybe this still has something missing.

EItanya:
Verifying locally now

EItanya:
Screenshot 2026-02-19 at 18 17 31

So it doesn't quite work. When I hit refresh during some tool calls it does get the stream back, but then it won't allow me to send another message. So probably just need to figure out how to update that state properly

dobesv:
Yeah I think I have seen the same thing. I can click cancel though and then I can chat again.  So maybe it thinks the agent is working even though it is done.

EItanya:
Going to move into draft until that's fixed

dobesv:
Sure, no problem.  Any suggestions or hints on what you think it might be?  I'm new to the project so I don't have the same intuition a veteran might have.

PR #1334, opened 2026-02-18
Add session tools to built-in kagent-agents MCP server

dobesv:
This allows an agent to create an kagent session to run a task instead of simple delegation.  This can be used to "hand off" a task to a different agent.

For example you can have an interactive session with a planning agent, it can hand that off to the implementation agent, and you can have an interactive session with the implementation agent.

The interactive session is desirable so the agent can ask questions or get corrections.  Doing it via multiple agents is desirable so that you can pick your model parameters and system prompt separately for the task.

This combines the two together - you can setup a RemoteMCPServer that points to the kagent-controller and give the planning agent the power to create a session and give you a link to it to continue the work.

In principle this might move in the direction of having multi-agent coordination where you can interact with the agents and they interact with each other if you give them each others' session IDs or have a coordinating agent that starts multiple agents and then gives them each others' session IDs.

### DIFF
--- a/go/internal/mcp/mcp_handler.go
+++ b/go/internal/mcp/mcp_handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -11,8 +12,10 @@ import (
 	"github.com/kagent-dev/kagent/go/api/v1alpha2"
 	"github.com/kagent-dev/kagent/go/internal/a2a"
 	authimpl "github.com/kagent-dev/kagent/go/internal/httpserver/auth"
+	"github.com/kagent-dev/kagent/go/internal/utils"
 	"github.com/kagent-dev/kagent/go/internal/version"
 	"github.com/kagent-dev/kagent/go/pkg/auth"
+	dbpkg "github.com/kagent-dev/kagent/go/pkg/database"
 	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


EItanya:
Hey there, thanks for the contribution. I definitely think this is moving in an awesome direction, but I'm not sure this is the level of abstraction we want to expose to the agents. Perhaps we can look further into the patterns that ADK recommends: https://google.github.io/adk-docs/agents/multi-agents/

dobesv:
My main goal here is handing off some work from one interactive agent session to another.  

The multi-agent patterns you linked seem to be a more advanced version of the current agent delegation system which isn't really what I wanted.

I thought enabling some interesting multi-agent workflows beyond my simple use case could be built on top this tool perhaps.  But that's not really the main point of this relatively small change.

For now I am assuming I will have to maintain my own fork of kagent more or less indefinitely.  I thought I would share some of my changes back upstream in case people have a use for them.  If not, feel free to close my pull requests.  Even if you do not merge them, others who want these capabilities and are also maintaining their own fork can use them as a reference.  Having my changes upstream and not maintaining a customized fork seems a distant dream at this point, I'm not counting on it.

dobesv:
Note that these MCP additions also can be a way to transfer a workflow into kagent from an external MCP client as well, not just internally within kagent.  If you use `invoke_agent` then the session is not interactive within kagent, you still have to continue the work within the original tool to manage it.

Think of the feature in Gemini where you can hand-off work to Jules, or in Claude where you can spawn a job into Claude for the Web.

EItanya:
> My main goal here is handing off some work from one interactive agent session to another.
> 
> The multi-agent patterns you linked seem to be a more advanced version of the current agent delegation system which isn't really what I wanted.
> 
> I thought enabling some interesting multi-agent workflows beyond my simple use case could be built on top this tool perhaps. But that's not really the main point of this relatively small change.
> 
> For now I am assuming I will have to maintain my own fork of kagent more or less indefinitely. I thought I would share some of my changes back upstream in case people have a use for them. If not, feel free to close my pull requests. Even if you do not merge them, others who want these capabilities and are also maintaining their own fork can use them as a reference. Having my changes upstream and not maintaining a customized fork seems a distant dream at this point, I'm not counting on it.

As I said in your other PR, we're very much open to new features, but it's also important that we keep in mind the rest of the community, and the maintainability and health of the project overall. For example, I am also very very much interested in background tasks, and how we can enable those sorts of workflows, but the best first step would be to create a design and attend the community meeting so that we can all be aligned before the feature is added. The most important thing is to make features which are useful for everyone, and also maintainable by the maintainers.

The link for the community meeting can be found here: https://kagent.dev/community

The meeting happens at 12:00 EST on Tuesdays, looking forward to seeing you there

dobesv:
It did occur to me today that it might be possible to allow opening a delegated agent session in kagent UI in which case I could still use the A2A protocol to create the session and just offer a link to it to the user to continue the work.  That might be a more elegant solution, if it's compatible with the way kagent works.

dobesv:
I think I can just use my own MCP server and call the agents from there using A2A instead of modifying the core.

PR #1337, opened 2026-02-19
feat: centralize environment variables for documentation 

### LINKED ISSUE #1335: [feat/docs] Consolidate all ENV vars in a way that can be easily documented
Right now env vars in `kagent` are implemented in a haphazard localized way which can make it very hard to figure out what they do without looking at the code. We should come up with a centralized way to define them so that we can also generate docs for them to publish. Istio does this

EItanya:
Consolidate all 70+ environment variables into a centralized registry inspired by Istio's pkg/env package. Each variable is self-registering with metadata (name, default, description, type, component), enabling automatic documentation generation.

Go side:
- Add go/pkg/env/ package with typed registration (StringVar, BoolVar, IntVar, DurationVar) and a global registry
- Add declaration files organized by domain: kagent, providers, otel, cli, database, testing
- Add `kagent env` cobra subcommand for generating markdown/JSON docs
- Migrate all callsites to use registry constants instead of raw os.Getenv() string literals

Python side:
- Add kagent.core.env module with EnvVar dataclass and register helpers
- Migrate callsites in kagent-core, kagent-adk, kagent-skills, and kagent-openai to use the registry

Fixes #1335

### DIFF
--- a/go/cli/internal/cli/envdoc/envdoc.go
+++ b/go/cli/internal/cli/envdoc/envdoc.go
@@ -18,7 +18,7 @@ func NewEnvCmd() *cobra.Command {
 		Use:    "env",
 		Hidden: true,
 		Short:  "List all kagent environment variables",
-		Long:   "Generate documentation for all kagent environment variables in markdown, JSON, or table format.",
+		Long:   "Generate documentation for all kagent environment variables in markdown or JSON format.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			switch format {
 			case "markdown", "md":
--- a/go/pkg/env/var.go
+++ b/go/pkg/env/var.go
@@ -9,6 +9,7 @@ package env
 
 import (
 	"cmp"
+	"encoding/json"
 	"fmt"
 	"os"
 	"slices"
@@ -47,6 +48,11 @@ func (v VarType) String() string {
 	}
 }
 
+// MarshalJSON serializes VarType as its string representation.
+func (v VarType) MarshalJSON() ([]byte, error) {
+	return json.Marshal(v.String())
+}
+
 // Component identifies which part of the kagent system consumes the variable.
 type Component string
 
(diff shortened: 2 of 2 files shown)


go/pkg/env/var.go
```diff
@@ -0,0 +1,375 @@
...
+	if err != nil {
+		return d.defaultValue, false
+	}
+	return parsed, true
+}
+
+// Name returns the environment variable name.
+func (d DurationVar) Name() string { return d.v.Name }
+
+// ---------- Formatting ----------
```
peterj, line 306: ?

go/pkg/env/var.go
```diff
@@ -0,0 +1,375 @@
...
+		if !first {
+			sb.WriteString(",\n")
+		}
+		first = false
+		fmt.Fprintf(&sb, `  {"name": %q, "type": %q, "default": %q, "description": %q, "component": %q, "deprecated": %v}`,
+			v.Name, v.Type, v.DefaultValue, v.Description, v.Component, v.Deprecated)
+	}
+	sb.WriteString("\n]\n")
+	return sb.String()
+}
```
iplay88keys, line 378: What about this?

Update `Var` above to:
```
type Var struct {
	Name         string    `json:"name"`
	DefaultValue string    `json:"default"`
	Description  string    `json:"description"`
	Type         VarType   `json:"type"`
	Component    Component `json:"component"`
	Hidden       bool      `json:"-"` // internal only
	Deprecated   bool      `json:"deprecated"`
}
```

add this so `VarType` doesn't serialize as a number:
```
func (v VarType) MarshalJSON() ([]byte, error) {
	return json.Marshal(v.String())
}
```

and update export to not be hand-written json:
```
func ExportJSON(component string) string {
	vars := VarDescriptions()
	out := make([]Var, 0, len(vars))
	for _, v := range vars {
		if v.Hidden {
			continue
		}
		if component != "" && component != "all" && string(v.Component) != component {
			continue
		}
		out = append(out, v)
	}

	b, err := json.MarshalIndent(out, "", "  ")
	if err != nil {
		return "[]\n"
	}
	return string(b) + "\n"
}
```

go/cli/internal/cli/envdoc/envdoc.go
```diff
@@ -0,0 +1,39 @@
...
+		Long:   "Generate documentation for all kagent environment variables in markdown, JSON, or table format.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			switch format {
+			case "markdown", "md":
+				fmt.Fprint(cmd.OutOrStdout(), env.ExportMarkdown(component))
+			case "json":
+				fmt.Fprint(cmd.OutOrStdout(), env.ExportJSON(component))
+			default:
+				return fmt.Errorf("unknown format %q: use markdown or json", format)
+			}
```
iplay88keys, line 30: I noticed the output for markdown is to group the env vars by component, but the output for json is a single array. Not a problem, just wanted to see if that was intentional.

PR #1341, opened 2026-02-20
Add context management options

dobesv:
Adds options to the Agent CRD for context compression and caching.  These map directly to ADK built-in features for this purpose.


dobesv:
Sorry somehow opened an extra PR here ... agents going wild on me

PR #1351, opened 2026-02-20
Fix backend error handler

dobesv:
I noticed that when kagent-controller was down, the UI showed the wrong error "Agent not found".  According to claude code, this change should fix it to show a more specific error.  Basically it seems like some code was expecting `error` to be set for errors, so this sets error and message both to handle those cases.


peterj (approved):

dobesv:
I notice you approved but did not merge - are you thinking I would merge?  Or is there some automated process that would merge it?  I don't have the ability to merge it.

lwencel:
@peterj could you help to merge this change?

PR #1360, opened 2026-02-23
Improve display of agent/tool cards in chat session

dobesv:
- Parse JSON payloads and render basic tree view
- Render strings as markdown
- Copy to clipboard button copies original text
- Add view source button in case of render/parse issues
- Prevent ScrollArea from overflow horizontally
- Prevent Tool Call ID from making cards too wide for the container


dobesv:
@peterj @EItanya are you guys interested in having this change?

dobesv:
@peterj can you take a look at this one?

ui/src/components/chat/AgentCallDisplay.tsx
```diff
@@ -19,13 +22,22 @@ interface AgentCallDisplayProps {
   tokenStats?: TokenStats;
 }
 
+const AGENT_TOOL_NAME_RE = /^(.+)__NS__(.+)$/;
```
supreme-gg-gg: There's a function to check if it's an agent tool in `ui/src/lib/utils.ts` `isAgentToolName`

ui/src/components/chat/AgentCallDisplay.tsx
```diff
@@ -78,56 +96,49 @@ const AgentCallDisplay = ({ call, result, status = "requested", isError = false,
             <KagentLogo className="w-4 h-4 mr-2" />
             {agentDisplay}
           </div>
-          <div className="font-light">{call.id}</div>
+          <div className="font-light">
+            {functionCallLink ? (
+              <Link href={functionCallLink} className="text-blue-500 hover:underline">
```
supreme-gg-gg: What is the function call link page supposed to do? I don't think the route is present in the app it always redirects to 404

ui/src/components/chat/CollapsibleSection.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <Icon className="w-3.5 h-3.5 shrink-0 mt-0.5 text-muted-foreground" />
+        <div className="flex-1 min-w-0">
+          <div className={`relative rounded-md ${errorStyle ? "bg-red-50 dark:bg-red-950/10" : ""}`}>
+            <ScrollArea className="max-h-96 overflow-y-auto p-2 w-full rounded-md bg-muted/50">
+              {expandedContent}
+            </ScrollArea>
+          </div>
+        </div>
+      </div>
+      <button
```
supreme-gg-gg, line 56: IMO it would be nice if clicking on the text area just toggles the collapse state, since right now to close it I will need to click on the small button, and that's also not consistent with the expand behaviour where if you click on the text area it opens

ui/src/components/chat/AgentCallDisplay.tsx
dobesv: Ah it might be something from another change that snuck in here

ui/src/components/chat/CollapsibleSection.tsx
dobesv: I wanted to be able to copy paste things from the content when expanded

dobesv:
I rebased this on main, hopefully it still works OK.  I did a quick test and it looked OK.

peterj (approved):

PR #1362, opened 2026-02-23
Add the ability to click through to sub-agent session

dobesv:
When debugging agent behavior it is very useful to be able to drill down into the sub-agent activity and what tool calls they had - or even drill down into *their* sub-agent calls, and so on.

I find this feature very useful in agent development with sub-agents and sub-sub-agents.

I'm not super happy with this implementation, there are some hacks in here.  It does basically work, though (at least it does in my main branch with all my customizations in it).  I'd be happy to get some tips on a better way to achieve this.


go/internal/database/client.go
```diff
@@ -491,6 +491,27 @@ func (c *clientImpl) DeleteCheckpoint(userID, threadID string) error {
 	})
 }
 
+func (c *clientImpl) FindSessionByParentFunctionCallID(functionCallID string) (*dbpkg.Session, error) {
+	var event dbpkg.Event
+	// Use a wildcard between the key and value to handle potential whitespace (e.g. ": " vs ":")
+	searchPattern := fmt.Sprintf("%%\"parent_function_call_id\"%%\"%s\"%%", functionCallID)
+	err := c.db.Where("data LIKE ?", searchPattern).First(&event).Error
```
EItanya, line 498: I agree with copilot here, I think searching through JSON like this is pretty fragile. Can we instead add more keys to the events, or think about organizing the DB data differently?

dobesv: Yeah I def agree ... need some guidance on the right solution I guess.

PR #1365, opened 2026-02-23
feat: add git-based skill fetching with shared auth and lightweight init image

EItanya:
## Summary

- Add support for fetching agent skills from **Git repositories** alongside existing OCI image refs
- Introduce a single shared `gitAuthSecretRef` on `SkillForAgent` that applies credentials to all `gitRefs` entries (supports both HTTPS token and SSH key auth)
- Replace two separate init containers (alpine/git for git + the full ~1GB kagent runtime for OCI) with a **single lightweight `skills-init` image** (~30MB) containing only `git` and `krane`
- Harden the init script against shell injection using heredoc variable assignments, and add server-side validation for `subPath` traversal and duplicate skill names

## New CRD Fields

```yaml
spec:
  skills:
    # Existing: OCI skill image refs
    refs:
      - ghcr.io/org/my-skill:v1

    # NEW: shared git auth secret (HTTPS token or SSH key)
    gitAuthSecretRef:
      name: my-git-creds

    # NEW: git repositories to clone as skills
    gitRefs:
      - url: https://github.com/org/my-skills.git
        ref: v2.0.0           # branch, tag, or commit SHA (default: main)
        path: skills/k8s      # optional subdirectory
        name: k8s-skill       # optional custom directory name
```

### Auth Secret Format

The secret referenced by `gitAuthSecretRef` should contain either:
- `token` key — for HTTPS credential helper (`x-access-token`)
- `ssh-privatekey` key — for SSH-based cloning

### CRD (`go/api/v1alpha2/agent_types.go`)
- Add `GitRepo` type with `url`, `ref`, `path`, and `name` fields
- Add `gitAuthSecretRef` and `gitRefs` fields to `SkillForAgent`
- Regenerate CRDs and Helm chart CRD templates

### Controller / Translator (`go/internal/controller/translator/agent/`)
- Add `buildSkillsInitContainer` — unified init container for both git and OCI skills
- Add `skills-init.sh.tmpl` — shell script template using heredoc variables to prevent injection
- Add `validateSubPath()` — rejects absolute paths and `..` traversal
- Add duplicate skill directory name detection across git and OCI refs
- Add `gitSkillName()` with `url.Parse()` to strip query params/fragments
- Add `ociSkillName()` for extracting names from OCI refs
- Add `DefaultSkillsInitImageConfig` and `ImageConfig.Image()` method

### Lightweight Init Image (`docker/skills-init/`)
- New multi-stage Dockerfile: builds `krane` from source, produces a minimal Alpine image with just `git` + `krane`
- Replaces use of the full kagent runtime image for OCI skill pulling

### Helm
- Add `skillsInitImage` config (`registry`, `repository`, `tag`, `pullPolicy`) to `values.yaml`
- Wire through controller ConfigMap

### Tests
- Golden tests for `agent_with_git_skills` (git + OCI + auth + subpath + commit SHA)
- Integration tests for all skill init container scenarios
- Unit tests for `ociSkillName`, `gitSkillName`, `validateSubPath`, and duplicate name detection

## Test plan

- [ ] `cd go && go build ./...` compiles
- [ ] `cd go && go test ./internal/controller/translator/agent/... -count=1` — all tests pass
- [ ] Golden files match expected output (`UPDATE_GOLDEN=true` to regenerate)
- [ ] Deploy to Kind cluster with an agent using `gitRefs` and verify skills are cloned
- [ ] Verify OCI skills still work with the new unified init container
- [ ] Verify auth secret (HTTPS token) works for private repos

### DIFF
--- a/go/internal/controller/translator/agent/adk_translator_golden_test.go
+++ b/go/internal/controller/translator/agent/adk_translator_golden_test.go
@@ -37,6 +37,16 @@ type TestInput struct {
 
 // TestGoldenAdkTranslator runs golden tests for the ADK API translator
 func TestGoldenAdkTranslator(t *testing.T) {
+	// Clear all OTEL_ env vars so host environment doesn't leak into
+	// golden outputs via collectOtelEnvFromProcess().
+	for _, env := range os.Environ() {
+		if strings.HasPrefix(env, "OTEL_") {
+			key, _, _ := strings.Cut(env, "=")
+			t.Setenv(key, "")
+			os.Unsetenv(key)
+		}
+	}
+
 	// Skip if running in CI without update flag
 	updateGolden := os.Getenv("UPDATE_GOLDEN") == "true"
 
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_allowed_headers.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_allowed_headers.json
@@ -173,10 +173,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_code.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_code.json
@@ -165,10 +165,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_cross_namespace_tools.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_cross_namespace_tools.json
@@ -178,10 +178,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_custom_sa.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_custom_sa.json
@@ -133,10 +133,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_git_skills.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_git_skills.json
@@ -165,10 +165,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_http_toolserver.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_http_toolserver.json
@@ -172,10 +172,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_mcp_service.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_mcp_service.json
@@ -168,10 +168,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
--- a/go/internal/controller/translator/agent/testdata/outputs/agent_with_nested_agent.json
+++ b/go/internal/controller/translator/agent/testdata/outputs/agent_with_nested_agent.json
@@ -166,10 +166,6 @@
                     }
                   },
                   {
-                    "name": "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE",
-                    "value": "delta"
-                  },
-                  {
                     "name": "KAGENT_NAMESPACE",
                     "valueFrom": {
                       "fieldRef": {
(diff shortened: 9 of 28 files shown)


go/api/v1alpha2/agent_types.go
```diff
@@ -69,16 +69,49 @@ type AgentSpec struct {
 	AllowedNamespaces *AllowedNamespaces `json:"allowedNamespaces,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="size(self.refs) > 0 || size(self.gitRefs) > 0",message="at least one of refs or gitRefs must be specified"
```
timflannagan: nit: You can use AtLeastOneOf here.

docker/skills-init/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN apk add --no-cache git && \
+    git clone --depth 1 --branch $KRANE_VERSION \
+    https://github.com/google/go-containerregistry.git
+
+WORKDIR /build/go-containerregistry/cmd/krane
+
+RUN CGO_ENABLED=0 go build -trimpath -ldflags="-s -w" -o /build/krane .
+
+### Stage 1: runtime
```
timflannagan: nit: No need for comments here. They're implicit based on FROM structuring.

go/api/v1alpha2/agent_types.go
```diff
@@ -69,16 +69,49 @@ type AgentSpec struct {
...
 type SkillForAgent struct {
 	// Fetch images insecurely from registries (allowing HTTP and skipping TLS verification).
 	// Meant for development and testing purposes only.
 	// +optional
 	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
 
 	// The list of skill images to fetch.
-	// +kubebuilder:validation:MinItems=1
```
timflannagan, line 79: Why remove this? If it's +optional, you can still enforce "at least one item should be present".

go/api/v1alpha2/agent_types.go
```diff
@@ -69,16 +69,49 @@ type AgentSpec struct {
...
+	// +kubebuilder:default="main"
+	Ref string `json:"ref,omitempty"`
+
+	// Subdirectory within the repo to use as the skill root.
+	// +optional
+	Path string `json:"path,omitempty"`
+
+	// Name for the skill directory under /skills. Defaults to the repo name.
+	// +optional
+	Name string `json:"name,omitempty"`
```
timflannagan, line 116: Take a peek at https://github.com/kubernetes-sigs/kube-api-linter/. All +optional fields should be a pointer to a string. We can also take this opportunity to define simple validation checks too, e.g. minLength, patterns, etc.

go/api/v1alpha2/agent_types.go
```diff
@@ -69,16 +69,49 @@ type AgentSpec struct {
...
+
+	// Reference to a Secret containing git credentials.
+	// Applied to all gitRefs entries.
+	// The secret should contain a `token` key for HTTPS auth,
+	// or `ssh-privatekey` for SSH auth.
+	// +optional
+	GitAuthSecretRef *corev1.LocalObjectReference `json:"gitAuthSecretRef,omitempty"`
+
+	// Git repositories to fetch skills from.
+	// +kubebuilder:validation:MaxItems=20
```
timflannagan, line 93: MinItems?

go/api/v1alpha2/agent_types.go
```diff
@@ -69,16 +69,49 @@ type AgentSpec struct {
...
 	// +kubebuilder:validation:MaxItems=20
+	// +optional
 	Refs []string `json:"refs,omitempty"`
+
+	// Reference to a Secret containing git credentials.
+	// Applied to all gitRefs entries.
+	// The secret should contain a `token` key for HTTPS auth,
+	// or `ssh-privatekey` for SSH auth.
+	// +optional
+	GitAuthSecretRef *corev1.LocalObjectReference `json:"gitAuthSecretRef,omitempty"`
```
timflannagan, line 90: This is slightly awkward to me. The intention is that this field decorates the gitRefs field? Couldn't we achieve the same thing with a top-level field where we decorate references with auth instead?

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -422,6 +427,17 @@ func (a *adkApiTranslator) buildManifest(
...
+	// Unified skills init container (handles both git and OCI skills)
+	if hasSkills {
+		insecure := agent.Spec.Skills != nil && agent.Spec.Skills.InsecureSkipVerify
+		container, skillsVolumes, err := buildSkillsInitContainer(gitRefs, gitAuthSecretRef, skills, insecure, dep.SecurityContext)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build skills init container: %w", err)
+		}
+		initContainers = append(initContainers, container)
+		volumes = append(volumes, skillsVolumes...)
+	}
```
jmhbh: nit: can't we combine these two `if hasSkills` conditionals?

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1345,6 +1361,220 @@ func collectOtelEnvFromProcess() []corev1.EnvVar {
...
+		return nil
+	}
+	if path.IsAbs(p) {
+		return fmt.Errorf("skill subPath must be relative, got %q", p)
+	}
+	if slices.Contains(strings.Split(p, "/"), "..") {
+		return fmt.Errorf("skill subPath must not contain '..', got %q", p)
+	}
+	return nil
+}
```
jmhbh, line 1401: I feel like this is the sort of thing that if we use kube validation for on the API would be much more explicit and reduce the need to maintain and test such code.

go/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1345,6 +1361,220 @@ func collectOtelEnvFromProcess() []corev1.EnvVar {
...
+		gitRef := ref.Ref
+		if gitRef == "" {
+			gitRef = "main"
+		}
+		ref.Ref = gitRef
+
+		name := gitSkillName(ref)
+		if seen[name] {
+			return skillsInitData{}, fmt.Errorf("duplicate skill directory name %q", name)
+		}
```
jmhbh, line 1491: Would it be possible to use kube validation here to ensure that the individual gitRef items are unique? This should allow us to never have to worry about duplicate skill directory names and simplify this code.

go/api/v1alpha2/agent_types.go
EItanya: I agree that this API is awkward. I think unfortunately this will be a `v1alpha3` type of thing. Ideally we'd have a list of items which have a `oneof`. Either OCI refs or GitHub refs. The issue is that the auth mechanisms are different so I don't think we can realistically use the same secret for both which is very annoying

go/api/v1alpha2/agent_types.go
EItanya: No, because it can be either `gitRefs` or `refs`

go/internal/controller/translator/agent/adk_api_translator.go
EItanya: I agree in theory, but the logic here is a bit too involved for cel I think. Too much string manipulation

go/internal/controller/translator/agent/adk_api_translator.go
EItanya: I think combining data from 2 lists and then checking uniqueness is a bit too complex for CEL. Most likely we should do best effort here, but I think we can start with this. What do you think?

iplay88keys: Do the directory names need to be defined by the user? What if we just generated them for the user in a similar way to how go mod downloaded libraries work. Splitting the ref into a folder structure like: `/skills/<org>/<repo>@<tag>`?

EItanya: It needs to be `skills/<skill-name>` because of how skills are organized on the filesystem

jmhbh: makes sense to me

iplay88keys: Ok, so it sounds like it's a limitation of how skills are discovered and loaded by the python ADK. That makes sense, then.

go/api/v1alpha2/agent_types.go
timflannagan: Right, but in this case, the field is optional. So when you specify it, you're enforcing at least one item has been configured by the user (vs. an empty list).

PR #1383, opened 2026-02-26
feat: sub-agent session viewing

dobesv:
- Parent agent assigns session IDs and stores in the event, for linking.  
- UI can see child contextId stored in the event and show a link to it.
- Allow anyone who knows the ID of a session to access it

Note that this does kind of remove a permission check.  This seems OK to me because in its current state kagent does not have multi-user support anyway.  Later if multi-user becomes a thing, the permission check could be adjusted somehow.

### DIFF
--- a/go/internal/database/client.go
+++ b/go/internal/database/client.go
@@ -109,8 +109,7 @@ func (c *clientImpl) GetTaskMessages(taskID int) ([]*protocol.Message, error) {
 // GetSession retrieves a session by name and user ID
 func (c *clientImpl) GetSession(sessionName string, userID string) (*dbpkg.Session, error) {
 	return get[dbpkg.Session](c.db,
-		Clause{Key: "id", Value: sessionName},
-		Clause{Key: "user_id", Value: userID})
+		Clause{Key: "id", Value: sessionName})
 }
 
 // GetAgent retrieves an agent by name and user ID
--- a/go/internal/database/fake/client.go
+++ b/go/internal/database/fake/client.go
@@ -21,7 +21,7 @@ type InMemoryFakeClient struct {
 	mu                sync.RWMutex
 	feedback          map[string]*database.Feedback
 	tasks             map[string]*database.Task    // changed from runs, key: taskID
-	sessions          map[string]*database.Session // key: sessionID_userID
+	sessions          map[string]*database.Session // key: sessionID
 	agents            map[string]*database.Agent   // changed from teams
 	toolServers       map[string]*database.ToolServer
 	tools             map[string]*database.Tool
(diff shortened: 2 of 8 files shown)


dobesv:
I don't think I removed it from the session. I just changed the session lookup not to require a match on user ID.  This is because the sub agent sessions don't have the user's ID.

The session ID should be unique enough to identify the session, we don't need to scope it to the user as well.

An alternative might be to somehow figure out the right user ID to use for the lookup, maybe we know it's a sub-agent session and calculate the user ID based on the session ID.

I also looked at propagation of the user ID to sub-agent sessions but it was getting complex and I wanted to keep the change small.  I could attempt that again, though, if you prefer.

dobesv:
Propagating the user's ID to the sub agent sessions does have the side effect that you can see the sub agent sessions along with your own chats in your chat history.  I couldn't quite decide if that's a good thing or not, but my feeling is that it would be a bit noisy as a default.

dobesv:
Oh to clarify your question... The sub-agent calls have user_id set to a string based on the session ID, so it's a new user every session.  It's not blank, but it's also not the user's actual ID.  So it doesn't match if the user requests that session by ID.

EItanya:
Ok, those are all very good points. Let's step back for 2 seconds and think about the data model from first principles. If you could organize them however you want what do you think you would do. 1 requirement is that it's still important for sessions to be user-scoped, even if that's just the root

dobesv:
Where does the user scoped session requirement come from?  Do you think two users would generate the same session ID?

EItanya:
> Where does the user scoped session requirement come from? Do you think two users would generate the same session ID?

No, but we need to be able to separate the sessions by the user who created them, otherwise users will be able to see other users sessions

dobesv:
Another trick I could do here is match a session if the user ID matches OR the user_id on the session is a generated A2A session user_id.

Personally I think that if you know a session ID you can see the session, at least for now since kagent is effectively single user and session IDs are hard enough to guess. 

The logic to get a list of sessions is still scoped to the user, you can only search for your own chat sessions.

Fully solving the question of access control to sub-agent sessions might blow up the pull request.

dobesv:
> > Where does the user scoped session requirement come from? Do you think two users would generate the same session ID?
> 
> No, but we need to be able to separate the sessions by the user who created them, otherwise users will be able to see other users sessions

That part is not affected by this change, I only changed it so when you look by session ID you don't have to match by user ID as well.  Getting a list of chats by user ID still matches by user.

dobesv:
seems like this broke after merging latest, I'll have to work on a fix

PR #1392, opened 2026-02-27
fix: resolve Trivy image scan CVEs in app image

jsonmp-k8:
## Summary
- **CVE-2026-26996** (minimatch, HIGH): Add `npm audit fix` after sandbox-runtime npm install to update transitive minimatch dependency from 9.0.5 to 9.0.6
- **CVE-2026-23949** (jaraco.context, HIGH): Upgrade setuptools in uv-managed Python installations to update vendored jaraco.context from 5.3.0 to >=6.1.0
- **CVE-2026-24049** (wheel, HIGH): Same setuptools upgrade updates vendored wheel from 0.45.1 to >=0.46.2

The `jaraco.context` and `wheel` vulnerabilities are in packages **vendored inside setuptools** in the uv-managed Python 3.10 installation (`python/cpython-3.10.19-linux-x86_64-gnu/lib/python3.10/site-packages/setuptools-/_vendor/`). Upgrading setuptools replaces these vendored copies with patched versions.

## Test plan
- [ ] Verify the `Scan images` CI workflow passes for the `app` image
- [ ] Verify the `CI Build` workflow continues to pass
- [ ] Verify `npm audit fix` successfully updates minimatch 9.0.5 → 9.0.6
- [ ] Verify setuptools upgrade resolves vendored jaraco.context and wheel CVEs


python/Dockerfile
```diff
@@ -138,6 +140,11 @@ RUN --mount=type=cache,target=/.kagent/cache,uid=1001,gid=1001                \
     && uv cache prune                                                         \
     && echo "Installation complete."
 
+# Fix CVE-2026-23949 (jaraco.context) and CVE-2026-24049 (wheel):
+# Upgrade setuptools in managed Python installations to update vendored packages
+# Pin to 82.0.0 which vendors jaraco.context 6.1.0 and wheel 0.46.3
+RUN find /python -name 'pip3' -path '*/bin/pip3' -exec {} install "setuptools==82.0.0" \;
```
EItanya: Can we solve this from the `pyproject.toml`? Why do we need to do this weird `Dockerfile` patching?

jsonmp-k8:
Re: @EItanya's [comment](https://github.com/kagent-dev/kagent/pull/1392#discussion_r2864239231) — unfortunately `pyproject.toml` can't fix this one.

The CVEs (CVE-2026-23949 and CVE-2026-24049) are in the `setuptools` package bundled inside uv's managed Python base installations under `/python`. These are standalone Python builds downloaded by `uv` — they come pre-bundled with their own `setuptools` that internally vendors `jaraco.context` and `wheel`.

The project's `constraint-dependencies` in `pyproject.toml` only control dependency resolution in the project venv (`/.kagent/.venv`), and `setuptools` isn't even a project dependency (it's not in `uv.lock`). So there's no way to reach the vulnerable copies through `pyproject.toml`.

Updated the Dockerfile comment to explain why this approach is necessary.

EItanya:
Re: the `pyproject.toml` vs Dockerfile discussion — I agree `pyproject.toml` can't reach the setuptools bundled in uv's managed Python installations. But rather than patching the symptom in the Dockerfile, we should fix the root cause.

The reason these CVEs exist is that `.python-version` is pinned to `3.10`, and Python 3.10 ships with a bundled setuptools (which vendors the vulnerable `jaraco.context` and `wheel`). **Python 3.12+ no longer bundles setuptools at all** (PEP 632), so upgrading eliminates both CVEs without any Dockerfile hacking.

### Suggested fix

1. Update `.python-version` from `3.10` to `3.13` (to match the existing `TOOLS_PYTHON_VERSION` arg in the Dockerfile)
2. Drop the `find /python -name 'pip3' ... setuptools` RUN step entirely

This is safe because:
- All packages already declare `requires-python = ">=3.10"` (only `kagent-crewai` has `<3.14`)
- The Dockerfile already creates venvs with `--python=python3.13`
- We want to support 3.10–3.13 for users, but there's no reason not to ship our official image with the latest

This way we're not carrying a Dockerfile workaround that we'll need to maintain every time a new setuptools CVE drops.

PR #1394, opened 2026-02-27
fix: escape curly braces in agent instructions to prevent context variable errors

### LINKED ISSUE #1382: [BUG] Context variable references in agent prompts cause errors
### 📋 Prerequisites

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

If you write some text like `{repo}` in your agent prompt you can get an error like:

```
2026-02-25 19:56:39,014 - kagent_adk.kagent.adk._agent_executor - ERROR - Error handling A2A request: 'Context variable not found: repo.'
  File "/.kagent/.venv/lib/python3.13/site-packages/google/adk/utils/instructions_utils.py", line 131, in _replace_match
    raise KeyError(f'Context variable not found: {var_name}.')
KeyError: 'Context variable not found: repo.'
```

I don't think there's any practical way to set these context variables in kagent, so this ADK feature should probably be disabled or the {}'s should be automatically escaped.

### 🔄 Steps To Reproduce

1. Create an agent with a prompt with `{repo}` or similar in it
2. Try to use it
3. 

### 🤔 Expected Behavior

Does not try to process `{...}`

### 📱 Actual Behavior

Error

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

fl-sean03:
## Summary

Fixes #1382

When an agent's instructions contain curly braces (e.g., `{repo}`, `{branch}`), ADK's `inject_session_state` tries to resolve them as context variable references. Since kagent doesn't use ADK context variables, this causes a `KeyError`.

**Root cause:** ADK's `LlmAgent.canonical_instruction()` returns `bypass_state_injection=False` when `instruction` is a string, triggering `inject_session_state` which uses regex `{+[^{}]*}+` to find and resolve `{text}` patterns as state variables.

**Fix:** Wrap the instruction string in a callable (`InstructionProvider`) before passing to the ADK `Agent` constructor. When instruction is callable, ADK sets `bypass_state_injection=True`, completely skipping the template processing. The `_configure_memory` method is also updated to properly handle callable instructions when appending memory usage instructions.

## Changes

- `python/packages/kagent-adk/src/kagent/adk/types.py`: Wrap `self.instruction` in an `InstructionProvider` callable in `to_agent()`. Update `_configure_memory()` to compose callables when appending memory instructions.
- `python/packages/kagent-adk/tests/unittests/test_instruction_escaping.py`: 7 unit tests covering curly braces, JSON-like content, nested braces, empty instructions, and verifying the callable bypass mechanism.

## Test plan

- [x] All 7 new unit tests pass
- [x] Existing test suite unaffected (pre-existing env import issues for other tests are unrelated)
- [ ] CI pipeline validation

Signed-off-by: opspawn <agent@opspawn.com>

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/types.py
+++ b/python/packages/kagent-adk/src/kagent/adk/types.py
@@ -405,10 +405,18 @@ class AgentConfig(BaseModel):
             agent.tools.append(LoadMemoryTool())
             agent.tools.append(SaveMemoryTool())
 
-            agent.instruction = (
-                "You have long-term memory: use save_memory to store important findings, learnings, and user preferences. "
+            memory_suffix = (
+                "\n\nYou have long-term memory: use save_memory to store important findings, learnings, and user preferences. "
                 "When you need more context or are unsure, use load_memory to search past conversations for relevant information."
             )
+            # Append memory instructions to static_instruction so they stay
+            # in the system prompt (same as the main instruction) instead of
+            # being injected as a separate user message, which would confuse
+            # the LLM conversation flow and break mock-based e2e tests.
+            if agent.static_instruction:
+                agent.static_instruction += memory_suffix
+            else:
+                agent.instruction += memory_suffix
 
             # Define auto-save callback
             async def auto_save_session_to_memory_callback(callback_context: CallbackContext):


python/packages/kagent-adk/src/kagent/adk/types.py
```diff
@@ -420,12 +420,22 @@ async def rewrite_url_to_proxy(request: httpx.Request) -> None:
...
+        raw_instruction = self.instruction
+
+        def instruction_provider(_: ReadonlyContext) -> str:
+            return raw_instruction
 
         agent = Agent(
             name=name,
             model=model,
             description=self.description,
-            instruction=self.instruction,
```
supreme-gg-gg, line 376: Perhaps using this new field in ADK will solve the issue, I believe it would not cause any regression either:

```python
static_instruction: Optional[types.ContentUnion] = Field(
      default=None,
      description=(
          'Optional. LlmAgent.static_instruction. Static content sent literally'
          ' at position 0 without placeholder processing. When set, changes'
          ' instruction behavior to go to user content instead of'
          ' system_instruction. Supports context caching. Accepts'
          ' types.ContentUnion (str, types.Content, types.Part,'
          ' PIL.Image.Image, types.File, or list[PartUnion]).'
      ),
  )
```

fl-sean03: @supreme-gg-gg Thanks for the suggestion! I looked into `static_instruction` — it's an interesting field but it wouldn't fully solve this issue on its own.

The key problem: `static_instruction` changes where the instruction content goes (from `system_instruction` to user content), but ADK still runs `_process_agent_instruction()` on the `instruction` field, which calls `canonical_instruction()` → `inject_session_state()`. Since a raw string instruction returns `bypass_state_injection=False`, the `{curly_brace}` KeyError would still occur.

The callable approach directly sets `bypass_state_injection=True` via ADK's own mechanism — when `instruction` is callable, `canonical_instruction()` returns `(text, True)`, completely skipping state injection. This is simpler (one wrapping layer vs two-step setup) and semantically correct for kagent's use case since we don't use ADK context variables at all.

Rebased against main to resolve the merge conflicts. The Copilot review feedback on tests and linting has been addressed in the follow-up commits.

supreme-gg-gg: So I looked deeper and that doesn't seem to be the case, I think static instructions are not preprocessed: 
> Static instruction content sent literally as system instruction at the beginning....This field is for content that never changes and doesn't contain placeholders. It's sent directly to the model without any processing or variable substitution. 

in `llm_agent.py` 

and also in the code in `instructions.py`. I think this pattern aligns better with how ADK is designed, but yea indeed either works

fl-sean03:
@supreme-gg-gg You're absolutely right — I dug deeper into the ADK source and `static_instruction` is the correct approach here. Looking at `instructions.py`, `static_instruction` goes through `_transformers.t_content()` directly without any call to `_process_agent_instruction()` or `inject_session_state()`, so it completely bypasses placeholder processing.

I've updated the PR to use `static_instruction` instead of the callable wrapper:

- Agent instructions are now passed as `static_instruction` (sent literally, no template substitution)
- The memory suffix (which has no curly braces) goes into the `instruction` field
- This is simpler and aligns with how ADK is designed

Thanks for pushing back on this — much cleaner solution.

fl-sean03:
Thanks for digging into the ADK source — you're right that `static_instruction` bypasses the template preprocessing entirely (`instructions.py` doesn't touch it). I've updated the PR to use this approach: memory instructions now append to `static_instruction` instead of setting `agent.instruction` separately.

All 21 CI checks pass including both e2e suites (sqlite + postgres). Would you be open to approving this? 🙏

EItanya (approved):

PR #1412, opened 2026-03-01
Enable /metrics endpoint with prometheus metrics

dobesv:
Promethteus metrics are mostly supported by the libraries we are using so it is not hard to expose these.  It allows simple creation of a grafana dashboard to monitor agent LLM usage.

image


peterj:
@krisztianfekete when you get a chance, can you take a look?

dobesv:
OK I might have to revisit this at a later date.

PR #1414, opened 2026-03-01
Built-in tool for agents to get session info

dobesv:
This tool is essential for my multi-agent development workflows.

It allows agents to generate a link to the kagent session for review when creating commits.

This also allows agents to register for automatic ping-backs from a tool I have that monitors to-do lists and periodically pings the agent session if it hasn't touched the to-do list in a while.

It probably has many other uses.

An alternative to this tool approach could be to auto-inject this information into the prompt somehow.  I wasn't sure if I should do it that way or not. The tool is working well enough for me.

I couldn't do this without modifying kagent itself because the agent name and session ID aren't available to MCP servers.

Signed-off-by: Dobes Vandermeer <dobes.vandermeer@newsela.com>

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/cli.py
+++ b/python/packages/kagent-adk/src/kagent/adk/cli.py
@@ -15,7 +15,7 @@ from google.adk.cli.utils.agent_loader import AgentLoader
 from kagent.core import KAgentConfig, configure_logging, configure_tracing
 
 from . import AgentConfig, KAgentApp
-from .tools import add_skills_tool_to_agent
+from .tools import add_session_tool, add_skills_tool_to_agent
 
 logger = logging.getLogger(__name__)
 logging.getLogger("google_adk.google.adk.tools.base_authenticated_tool").setLevel(logging.ERROR)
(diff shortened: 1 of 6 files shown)


EItanya:
> An alternative to this tool approach could be to auto-inject this information into the prompt somehow. I wasn't sure if I should do it that way or not. The tool is working well enough for me.
> 
> I couldn't do this without modifying kagent itself because the agent name and session ID aren't available to MCP servers.


I'm a little worried about adding a tool just to inject 3 very small pieces of information. Models can still behave strangely with too many tools. Would injecting it into the System prompt be difficult?

dobesv:
> Would injecting it into the System prompt be difficult?

Currently the agent system prompt is set on agent startup to a static string generated by the controller.

But, it might be possible to inject a message into the session when it is created?  I don't really know ... I'm still a bit lost in the kagent / ADK stuff, just trying to figure out how to make things work.

Another thought I had was maybe the session metadata could be attached to tool calls automatically and the MCP tool could just pluck it from a header or metadata field.  For example in https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta maybe add `kagent_caller_app_name`, `kagent_caller_session_id`, `kagent_caller_user_id` .

Then I could make my own tool to do this if you wanted to avoid having it in kagent core.

dobesv:
It doesn't look like python adk provides any mechanism to set meta fields on MCP calls, though - it's just always set to otel trace context: https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_tool.py#L336

😢

Issue #1415, opened 2026-03-02
[FEATURE] Built-in "question' tool


EItanya:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add a built-in question tool which enables the model/agent to ask the user questions

### ❓ Problem Statement / Motivation

Tools like `opencode`, and `claude code` have showed the power of the model/agent being able to ask the user questions. Kagent would benefit from a similar tool.

This [blog](https://www.atcyrus.com/stories/claude-code-ask-user-question-tool-guide) shows an example of the tool and how useful it can be.

### 💡 Proposed Solution

I don't have a proposed solution, research needs to be done, most likely it should follow the work started in https://github.com/kagent-dev/kagent/pull/1398

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

PR #1420, opened 2026-03-03
feat: human-in-the-loop (user input)

supreme-gg-gg:
- Ask user questions tool #1415 (added by default)
- Enhancements to the HITL logic (regarding rejection)

### DIFF
--- a/ui/src/components/chat/ToolCallDisplay.tsx
+++ b/ui/src/components/chat/ToolCallDisplay.tsx
@@ -76,8 +76,12 @@ const extractToolCallRequests = (message: Message): FunctionCall[] => {
     if (part.metadata) {
       if (getMetadataValue<string>(part.metadata as Record<string, unknown>, "type") === "function_call") {
         const data = part.data as unknown as FunctionCall;
-        // Skip ADK internal function calls (confirmation/auth) — not real tool calls
-        if (data.name === "adk_request_confirmation" || data.name === "adk_request_credential") {
+        // Skip ADK internal function calls (confirmation/auth) and ask_user (has its own display)
+        if (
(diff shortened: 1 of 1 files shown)


supreme-gg-gg:
Screenshot 2026-03-03 at 4 24 21 PM

^ This is how the tool looks like in the new UI component. The flow follows pretty much exactly as existing HITL.

EItanya:
2 quick things:
1. I think we can make the UX slightly better for the forms, but that can happen in a follow-up
2. I think it's a bit confusing that the `ask_user` tool shows up in the list of called tools, it's really an internal detail and we should just surface the questions, not the tool call.

supreme-gg-gg:
> 2 quick things:
> 
> 1. I think we can make the UX slightly better for the forms, but that can happen in a follow-up
> 2. I think it's a bit confusing that the `ask_user` tool shows up in the list of called tools, it's really an internal detail and we should just surface the questions, not the tool call.

yea fixed 2, skipped showing ask user function call. 

For 1 do you have any specific ideas for what we need for better UX? I mainly copied the design from open code / claudecode where there's multi-select, single-select, or type your own option.

EItanya:
> > 2 quick things:
> > 
> > 1. I think we can make the UX slightly better for the forms, but that can happen in a follow-up
> > 2. I think it's a bit confusing that the `ask_user` tool shows up in the list of called tools, it's really an internal detail and we should just surface the questions, not the tool call.
> 
> yea fixed 2, skipped showing ask user function call.
> 
> For 1 do you have any specific ideas for what we need for better UX? I mainly copied the design from open code / claudecode where there's multi-select, single-select, or type your own option.

Sorry, I meant UI, not UX, the feature makes sense as written

PR #1433, opened 2026-03-05
feat: instrument controller with distributed tracing and A2A trace propagation

### LINKED ISSUE #1295: Trace context (traceparent) not propagated from controller to agent pods

onematchfox:
This PR adds OpenTelemetry distributed tracing to the kagent controller API, fixes trace context propagation across A2A agent calls, and cleans up noise in the existing Python agent tracing.

Fixes #1295 essentially replacing a chunk of #1297 (it does not address being able to set the `appProtocol` on the `Agent` `Service` - that is a separate concern IMO).


Makefile
```diff
@@ -440,7 +440,7 @@ open-dev-container:
 otel-local:
 	docker rm -f jaeger-desktop || true
 	docker run -d --name jaeger-desktop --restart=always -p 16686:16686 -p 4317:4317 -p 4318:4318 jaegertracing/jaeger:2.7.0
-	open http://localhost:16686/
+	@echo "Jaeger UI available at http://localhost:16686/"
```
onematchfox, line 443: `open` is MacOS specific

python/packages/kagent-core/tests/test_tracing_configure.py
```diff
@@ -76,3 +79,14 @@ def fake_instrument_google():
     assert instrument_calls["instrument_kwargs"] == {}
     assert instrument_calls["anthropic_event_logger_provider"] is None
     assert instrument_calls["google_instrumented"] is True
+
+
+def test_otel_sdk_default_propagator_includes_w3c_tracecontext():
```
onematchfox, line 84: Added this test to ensure that text propagation remains in place across underlying SDK upgrades - without needing any `set_global_textmap` call.

helm/kagent/values.yaml
```diff
@@ -397,6 +397,7 @@ otel:
     exporter:
       otlp:
         endpoint: ""
+        protocol: "grpc"
```
onematchfox, line 402: `autoexport` defaults to `http/protobuf` so might be worth explicitly calling this out in release notes as anyone who uses http will need to ensure this is set correctly.

onematchfox:
Don't think `test-e2e` failure is related to this PR but let me know if there's something I need to fix.

go/core/internal/httpserver/server.go
```diff
@@ -94,10 +95,18 @@ func (s *HTTPServer) Start(ctx context.Context) error {
...
+	// and W3C TraceContext propagation on every incoming request.
 	s.httpServer = &http.Server{
-		Addr:    s.config.BindAddr,
-		Handler: s.router,
+		Addr: s.config.BindAddr,
+		Handler: otelhttp.NewHandler(s.router, "http.server",
+			otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
+				return r.Method + " " + r.URL.Path
+			}),
+			otelhttp.WithFilter(func(r *http.Request) bool {
```
krisztianfekete, line 106: Would it make sense to filter A2A spans as well to reduce redundancy? Not sure how noisy this can get.

go/core/internal/telemetry/tracing.go
```diff
@@ -0,0 +1,83 @@
+package telemetry
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/google/uuid"
+	"go.opentelemetry.io/contrib/exporters/autoexport"
```
krisztianfekete, line 9: Cannot we just initiate an exporter based on `OTEL_EXPORTER_OTLP_PROTOCOL`? The rest would would be unused and using them would require code changes anyway due to how we set `OTEL_EXPORTER_OTLP_PROTOCOL` in Helm.

go/core/internal/a2a/trace.go
```diff
@@ -0,0 +1,102 @@
...
+		mcName = "default-model-config"
+	}
+	mc := &v1alpha2.ModelConfig{}
+	if err := cache.Get(ctx, types.NamespacedName{Namespace: agent.GetNamespace(), Name: mcName}, mc); err != nil {
+		return semconv.GenAIProviderNameKey.String("kagent")
+	}
+	return genAIProviderName(mc.Spec.Provider)
+}
+
+// genAIProviderName maps kagent's ModelProvider values to the standard
```
krisztianfekete, line 78: Very nice!

go/core/internal/httpserver/server.go
krisztianfekete: Hm checking the middleware, looks like if we do this the `invoke_agent` span would become a root span instead of a child.

helm/kagent/templates/controller-configmap.yaml
```diff
@@ -34,11 +34,13 @@ data:
   OTEL_EXPORTER_OTLP_TRACES_TIMEOUT: {{ .Values.otel.tracing.exporter.otlp.timeout | quote }}
   OTEL_EXPORTER_OTLP_LOGS_INSECURE: {{ .Values.otel.logging.exporter.otlp.insecure | quote }}
   OTEL_EXPORTER_OTLP_LOGS_TIMEOUT: {{ .Values.otel.logging.exporter.otlp.timeout | quote }}
+  OTEL_EXPORTER_OTLP_PROTOCOL: {{ .Values.otel.tracing.exporter.otlp.protocol | quote }}
```
krisztianfekete, line 37: Can we move this inside `{{- if $tracesEndpoint }}`?

go/core/internal/httpserver/server.go
onematchfox: Yeah, we want the nesting in this case.

helm/kagent/templates/controller-configmap.yaml
onematchfox: It is already there albeit slightly differently named as per the current convention of either handling traces and logs [together](https://github.com/onematchfox/kagent/blob/61153eea714a5aad1f86cd6d653ec936f108c29e/helm/kagent/templates/controller-configmap.yaml#L31) or [individually](https://github.com/onematchfox/kagent/blob/61153eea714a5aad1f86cd6d653ec936f108c29e/helm/kagent/templates/controller-configmap.yaml#L39).

go/core/internal/telemetry/tracing.go
onematchfox: I guess we could, although I feel like we're just going to end up duplicating what `autoexport` does since users will probably expect us to fully comply with the standard hierarchy of env var configuration . E.g. `OTEL_EXPORTER_OTLP_PROTOCOL` applying to both logs and traces vs the use of `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` and `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL` individually. See also https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/ and https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables

Personally, I'd prefer relying on the upstream here rather than doing our own thing.

PR #1443, opened 2026-03-05
agentsts: allow dynamic fetching of actor token and enable caching

shashankram:
- Allows fetching the actor token using a dynamic callback.
- Enables using cached actor and subject tokens if they haven't expired. This allows skipping the token exchange if the cached token has not expired.


python/packages/agentsts-adk/src/agentsts/adk/_base.py
```diff
@@ -86,41 +127,145 @@ async def before_run_callback(
...
+
+        # Clean up expired actor token cache
+        if self.actor_token_cache and _has_token_expired(self.actor_token_cache.expiry):
+            logger.debug("Removing expired actor token from cache")
+            self.actor_token_cache = None
+
         return None
 
 
+def _has_token_expired(expiry: Optional[int], buffer_seconds: int = 5) -> bool:
```
EItanya, line 250: Should we use a percentage rather than a hardcoded value?

shashankram: How would a percentage help here? Isn't it intuitive to define buffer in terms of seconds relative to the expiration time in seconds?

PR #1447, opened 2026-03-05
feature/kagent-mcp-app-plugins

dimetron:
WIP - for reference only


Charlesthebird:
I looked through the proposal, and I like the idea of being able to configure the site. But this plan seems much more complicated than it should be.

I would think:

For iframes in new sidebar items, this needs:
- A `label` for the sidebar item label
- A `url` for the iframe

For iframes to show up inline in the chat
- This should be a new artifact type or task type (I forget the exact the term), that renders a custom component (iframe), with the loaded URL. 
  - The data in this chat response would need
    - `url` for the iframe
    - `page-height` (optional) for the pixel height of the iframe, to render it well, and this could be optional with some default value.


For the `postMessage` idea, we could a standard set of messages that can be posted to/from the iframe, reused for both the inline chat, and the sidebar page iframes.

---
 
What I'd request for this feature is:

To please remove all or almost all backend changes, so that adding support for this includes:
- A way to configure the UI sidebar pages (could be a JSON object with the properties I mentioned above)
- A `postMessage` TypeScript type that defines a shared contract between all custom iframes and the UI.

To add the inline-iframe chat feature in a follow-up PR, since both features are fairly involved and it will be easier to identify what is needed for each feature this way.

Thank you!

dimetron:
Thanks @Charlesthebird - I think this is the direction it goes
see design/EP-2004-dynamic-mcp-ui-routing.md

## Implementation Details

- **Discovery:** `GET /api/plugins` returns registered plugins for sidebar
- **UI:** `ui/src/app/plugins/[name]/[[...path]]/page.tsx` — iframe host with postMessage bridge

Inspired by MCP-APP implementation https://modelcontextprotocol.io/extensions/apps/overview

Charlesthebird (commented):
Looking at the new design plan more, it's an interesting idea, and thank you for the links!

I have a couple of notes:

- This is a large PR, and I think the scope of it could be trimmed down significantly.
- Having only one or two examples showing the different `postMessage` and other features instead of all the different examples in the `specs` folder might help explain this feature better.

And some questions:

1. After looking through the site you linked, I see why `RemoteMCPServer` metadata was chosen as the way to configure the plugin applications. Should this be added for `MCPServer` resources as well?
2. Is the main reason for adding this feature so that we can get MCP App iframes in the chat responses? A source of confusion was that you mentioned updating the kagent app's sidebar as well as extending the `RemoteMCPServer` CRD to support MCP App plugins. What if the scope of this PR was reduced to just adding MCP App plugins to the chat (not the sidebar)?

PR #1493, opened 2026-03-12
fix(bash tool) implement proper command execution to not escape the sandbox

jholt96:
It is currently possible for users to prompt an agent using skills to escape srt and run bash commands without file or network restriction. This change enforces proper bash command where it is passed as a single argument without shell interpolation


EItanya (approved):
Amazing catch, thanks so much :)

Issue #1495, opened 2026-03-12
[BUG] File Tools from Skills provide full file system access


jholt96:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

The read, write, and edit file tools provide full access to the filesystem. Users can ask the agent to use these tools to access files outside of their chat sessions or skills which can result in secret leakage if secrets are mounted. A security prompt may block it but there are ways around that and it would be better for these tools to only have access to the working directory of the session (/tmp/kagent/ctx-***) or in the case of the read_file tool, the skills directory. 

### 🔄 Steps To Reproduce

1. Create an agent with a skill
2. ask it to read/write/edit a file outside of the working directory or skills directory
3. It will perform the action unless a security prompt blocks it (can be inconsistent)

### 🤔 Expected Behavior

These tools should not provide access outside of the working directory or skills repo. 

### 📱 Actual Behavior

agents can read, write, and edit any file the agent uid has access to

### 📋 Logs

```shell

```

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR to fix this issue

supreme-gg-gg:
Completed in #1496

PR #1496, opened 2026-03-12
fix(file tools) implement directory restrictions to file tools

jholt96:
This limits the file tools to either the current working directory or the skills repo

write_file & edit_file - restricted to just working directory

read_file - restricted to working directory & skills directory

implements: [#1495](https://github.com/kagent-dev/kagent/issues/1495)

### DIFF
--- a/python/packages/kagent-skills/src/kagent/skills/shell.py
+++ b/python/packages/kagent-skills/src/kagent/skills/shell.py
@@ -136,7 +136,7 @@ def _get_command_timeout_seconds(command: str) -> float:
         return 30.0  # 30 seconds for other commands
 
 
-async def execute_command(command: str, working_dir: Path, skills_dir: Path) -> str:
+async def execute_command(command: str, working_dir: Path, skills_dir: Path = Path("/skills")) -> str:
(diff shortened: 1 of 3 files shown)


python/packages/kagent-skills/src/kagent/skills/shell.py
```diff
@@ -110,16 +136,13 @@ def _get_command_timeout_seconds(command: str) -> float:
         return 30.0  # 30 seconds for other commands
 
 
-async def execute_command(
-    command: str,
-    working_dir: Path,
-) -> str:
+async def execute_command(command: str, working_dir: Path, skills_dir: Path) -> str:
```
supreme-gg-gg: Can you also update the usage of `execute_command` in `test_skill_execution.py` and `kagent/openai/tools/_tools.py` to pass in the `skills_dir` argument or make the skills dir argument optional?

jholt96: just pushed changed for this! lmk if there is anything else. I made the openai have a default of /skills and try to pull in that env

PR #1518, opened 2026-03-18
fix: LangGraph Executor HITL and E2E Tests

supreme-gg-gg:
See https://docs.langchain.com/oss/python/langgraph/interrupts for how HITL works in langgraph.

Tested the new HITL flow with a langgraph BYO agent and it works. However, unlike ADK declarative where you can just specify tools requiring approval in CRD, the developer using `kagent-langgraph` will need to setup their graph in a specific way following the above documentation. Most existing graphs using `interrupt()` mechanism should be compatible, but a more comprehensive testing would help to fully verify it especially for complex graphs. 

I have provided an example in the `samples` folder to show how to do that for starters.


python/samples/langgraph/hitl-tools/hitl_tools/agent.py
```diff
@@ -0,0 +1,177 @@
...
+
+    results: list[ToolMessage] = []
+
+    for tool_call in last_message.tool_calls:
+        tool_name = tool_call["name"]
+        tool_args = tool_call["args"]
+        tool_call_id = tool_call["id"]
+
+        if tool_name in TOOLS_REQUIRING_APPROVAL:
+            # Pause execution and ask the user for approval.
```
supreme-gg-gg, line 105: **Note to users**: `interrupt` allows the user to set arbitrary data in the graph state, this can literally be anything you want, as long as it matches the response in `Command` (this is unlike the HITL mechanism in ADK which only has specific fields). However, in order for our executor to work with our HITL flow we require your `interrupt` function to be called with the specific structure below. 

This is not a limitation, you are welcome to use any interrupt mechanisms, you will just need to modify the executor to handle that specific response and convert that to a protocol message in Kagent's HITL flow documented in `docs/architecture/human-in-the-loop.md`. The default Kagent HITL flow is not as versatile as langgraph because it's designed around ADK rather than langgraph.

In ADK, HITL will propagate through subagents, but since our langgraph SDK does not yet support "subgraphs", that feature will not be available here. You can implement this yourself using the same logic.

PR #1528, opened 2026-03-19
feat(helm): allow to specify nodePort of UI Service

leomichalski:
## Problem
Currently, when exposing the UI Service using the `NodePort` service type, there is no way to explicitly define the `nodePort` value in the Helm chart. Users are forced to either accept a dynamically assigned port from the Kubernetes control plane or modify the manifests post-rendering.

## Solution
This PR adds the ability to specify a `nodePort` for the UI Service. 
- Added a `nodePort` field to the UI service configuration block in `values.yaml`.
- The field is only used if service type is already set to "NodePort".

## Testing
To verify this change:

### Scenario: default

Command: `helm install kagent ./helm/kagent/ --namespace kagent-dev --dry-run | grep "targetPort: 8080" -A 5 -B 5`

Output 

```yaml
   app.kubernetes.io/component: ui
spec:
  type: ClusterIP
  ports:
    - port: 8080
      targetPort: 8080
      
      protocol: TCP
      name: ui
  selector:
    app.kubernetes.io/name: kagent
```

### Scenario: specify nodePort of NodePort Service

Command: `helm install kagent ./helm/kagent/ --namespace kagent-dev --dry-run --set ui.service.ports.nodePort=30100 --set ui.service.type="NodePort" | grep "targetPort: 8080" -A 5 -B 5`

Output

```yaml
    app.kubernetes.io/component: ui
spec:
  type: NodePort
  ports:
    - port: 8080
      targetPort: 8080
      
      nodePort: 30100
      
      protocol: TCP
      name: ui
```

### Scenario: specify nodePort of ClusterIP Service

Command: `helm install kagent ./helm/kagent/ --namespace kagent-dev --dry-run --set ui.service.ports.nodePort=30100 | grep "targetPort: 8080" -A 5 -B 5`

Output 

```yaml
    app.kubernetes.io/component: ui
spec:
  type: ClusterIP
  ports:
    - port: 8080
      targetPort: 8080
      
      protocol: TCP
      name: ui
  selector:
    app.kubernetes.io/name: kagent
```


EItanya (approved):

helm/kagent/values.yaml
```diff
@@ -204,6 +204,7 @@ ui:
     ports:
       port: 8080
       targetPort: 8080
+      nodePort: null
```
EItanya, line 221: Why is this `null` and not `0`?

EItanya (changes requested):
Thanks so much for the PR. Can you please add helm unit tests?

PR #1529, opened 2026-03-20
fix: support ssh auth for git-based skills

### LINKED ISSUE #1523: [BUG] git-based skills can't use ssh-based authentication
### 📋 Prerequisites

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

Adding an agent that uses skill from a git repo, results in a dependency missing in the init container:
```
skills-init /bin/sh: ssh-keyscan: not found
stream closed EOF for kagent/git-skills-ssh-agent-547458ddf6-d9g49 (skills-init)
```

### 🔄 Steps To Reproduce

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: gitea-ssh-credentials
  namespace: kagent
type: Opaque
stringData:
  ssh-privatekey: "$SSH_PRIVATE_KEY"
---
kind: Agent
metadata:
  name: git-skills-ssh-agent
  namespace: kagent
spec:
  type: Declarative
  skills:
    gitAuthSecretRef:
      name: gitea-ssh-credentials
    gitRefs:
    - url: ssh://git@gitea-ssh.gitea:22/gitops/ssh-skills-repo.git
      ref: main
      path: skills/baklava-maker
      name: baklava-maker
  declarative:
    systemMessage: You are a helpful assistant with skills from a private git repo accessed via SSH.
    tools: []
```

### 🤔 Expected Behavior

_No response_

### 📱 Actual Behavior

_No response_

### 💻 Environment

0.8.0-beta8

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

jsonmp-k8:
## Summary
- add `openssh-client` to the `skills-init` image so SSH host key discovery is available
- derive SSH hosts from skill `gitRefs` and populate `known_hosts`, including custom `ssh://` ports
- add translator, unit, and golden coverage for SSH-based git skill auth

### Behavioral change: dynamic host scanning replaces hardcoded defaults
The old template unconditionally ran `ssh-keyscan github.com gitlab.com bitbucket.org`. This PR replaces that with dynamic host extraction from `gitRefs` URLs (both `ssh://` and scp-style like `git@github.com:...`). This is more precise — it avoids scanning unused hosts and correctly handles self-hosted instances with custom ports. The common case (github.com, gitlab.com, bitbucket.org via scp-style URLs) is still covered by the `gitSSHHost` parser.

## Testing
- `go test ./core/internal/controller/translator/agent`
- `CGO_ENABLED=0 make lint`

Fixes #1523


EItanya (commented):
## Review

Comment left by Claude on behalf of @EItanya.

Overall the core logic is well-written — extracting `gitSSHHost` as a standalone tested function, handling scp-style URLs, adding `-H` to hash hostnames, proper `chmod` on `.ssh`, and the injection test coverage are all real improvements. A few concerns though:

### 1. Behavioral regression: Removed hardcoded host scanning

The old template unconditionally scanned the big three:
```sh
ssh-keyscan github.com gitlab.com bitbucket.org >> ~/.ssh/known_hosts
```

This PR removes that entirely and derives hosts *only* from `gitRefs` URLs. While more precise, this is a **silent breaking change** for existing users. The new `gitSSHHost` handles scp-style URLs so most cases should work, but the old code provided a safety net. This should at minimum be called out in the PR description as a deliberate behavioral change.

### 2. `ssh-keyscan` failures are swallowed

```sh
ssh-keyscan -H "{{ .Host }}" >> ~/.ssh/known_hosts 2>/dev/null || echo "WARNING: ..." >&2
```

With `set -e` at the top, the `|| echo` prevents the script from exiting on keyscan failure. The git clone will then fail later with a confusing "host key verification failed" error instead of failing fast. The `2>/dev/null` also hides the actual ssh-keyscan error output that would help debug DNS/network issues. Keyscan failure should arguably be **fatal** (remove the `||`), or at minimum not suppress stderr.

### 3. Empty `if` block after refactor

After the refactor, `prepareSkillsInitData` has:
```go
if authSecretRef != nil {
    data.AuthMountPath = "/git-auth"
}
```

The SSH host collection was moved into the `for` loop below, gated separately on `authSecretRef != nil`. The empty block isn't wrong but reads like something was accidentally deleted. A comment or consolidation would help.

### 4. Minor: `sshHostData` as a map key

```go
seenSSHHosts := make(map[sshHostData]bool)
```

Works because both fields are strings, but fragile — if someone adds a non-comparable field to `sshHostData` later, this becomes a compile error. A string key like `host + ":" + port` (which the old code used) is more robust.

### 5. Security boundary comment

The `validHostPattern`/`validPortPattern` regexes are the security boundary preventing shell injection in the template. A comment on the regexes noting this (and that they must not be relaxed without auditing the template) would help prevent future rot.

jsonmp-k8:
Thanks for the thorough review @EItanya! Pushed fixes addressing all points:

**1. Behavioral regression (hardcoded host scanning):** This is a deliberate change — `gitSSHHost` extracts hosts from scp-style URLs (e.g. `git@github.com:org/repo.git` → `github.com`), so the common case is covered dynamically. The old blanket scan of github.com/gitlab.com/bitbucket.org would run `ssh-keyscan` against hosts the user may not even use, and would miss self-hosted instances entirely. Updated the PR description to call this out.

**2. `ssh-keyscan` failures swallowed:** Fixed — removed `2>/dev/null` and `|| echo WARNING`. Failures now trigger `set -e` and abort the script immediately. Stderr from ssh-keyscan is preserved for debugging DNS/network issues.

**3. Empty `if` block:** Added a comment clarifying that `AuthMountPath` is set once at the top, while SSH host collection runs per-ref inside the loop — they can't be consolidated.

**4. `sshHostData` as map key:** Changed to `map[string]bool` keyed by `host + ":" + port`.

**5. Security boundary comment:** Added an explicit comment on `validHostPattern`/`validPortPattern` noting they are the shell injection boundary and must not be relaxed without auditing the template.

EItanya (approved):

PR #1534, opened 2026-03-23
feat: add --default-agent-pod-labels flag for global agent pod labels

EItanya:
## Summary
- Adds a new `--default-agent-pod-labels` CLI flag / `DEFAULT_AGENT_POD_LABELS` env var that applies a set of labels to all agent pod templates created by kagent
- Labels are specified as comma-separated `key=value` pairs (e.g. `team=platform,environment=production`)
- Fully wired through Helm via `controller.agentDeployment.podLabels` map in values.yaml
- Label precedence: built-in defaults → global default labels → per-agent `spec.deployment.labels`

## Test plan
- [x] Unit tests for `ParseLabels` (8 cases covering valid input, edge cases, and errors)
- [x] Unit test for `DEFAULT_AGENT_POD_LABELS` env var loading
- [x] Existing translator golden tests pass
- [x] 4 new Helm unit tests (labels set, empty, not configured, single label)
- [x] Deployed to Kind cluster with `podLabels: {team: platform, environment: production}` — verified all 11 agent pod templates contain both labels
- [x] `kubectl get pods -l team=platform` returns all agent pods

Comment left by Claude on behalf of @EItanya

### DIFF
--- a/go/core/pkg/app/app_test.go
+++ b/go/core/pkg/app/app_test.go
@@ -359,6 +359,18 @@ func TestMapValueWithLoadFromEnv(t *testing.T) {
 	assert.Equal(t, map[string]string{"team": "platform", "env": "prod"}, target)
 }
 
+func TestMapValueWithLoadFromEnvEqualsInValue(t *testing.T) {
+	t.Setenv("DEFAULT_AGENT_POD_LABELS", "token=abc=def,team=platform")
+
+	var target map[string]string
+	fs := flag.NewFlagSet("test", flag.ContinueOnError)
+	fs.Var(&MapValue{Target: &target}, "default-agent-pod-labels", "test flag")
+
+	err := LoadFromEnv(fs)
+	assert.NoError(t, err)
+	assert.Equal(t, map[string]string{"token": "abc=def", "team": "platform"}, target)
+}
+
 func TestLoadFromEnvIntegration(t *testing.T) {
 	envVars := map[string]string{
 		"METRICS_BIND_ADDRESS":           ":9090",


go/core/pkg/app/app_test.go
```diff
@@ -265,6 +265,92 @@ func TestDatabaseUrlFileFlag(t *testing.T) {
 	assert.Equal(t, "/etc/credentials/db-url", cfg.Database.UrlFile)
 }
 
+func TestMapValue(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   string
+		want    map[string]string
+		wantErr bool
+	}{
```
iplay88keys, line 274: Missing the test case where an env var contains an equals (such as: `key=val=ue`). This could occur with a base64 value, for example, so we should have a test case for it.

go/core/pkg/app/app_test.go
```diff
@@ -265,6 +265,92 @@ func TestDatabaseUrlFileFlag(t *testing.T) {
...
+	}
+}
+
+func TestMapValueString(t *testing.T) {
+	var target map[string]string
+	mv := &MapValue{Target: &target}
+	assert.Equal(t, "", mv.String())
+
+	target = map[string]string{"team": "platform"}
+	assert.Equal(t, "team=platform", mv.String())
```
iplay88keys, line 344: Worth noting that iterating over a map in Go is random, so if there were multiple keys here, the output would be inconsistent. It's worth considering sorting before ranging over the keys in the implementation.

PR #1535, opened 2026-03-24
fix: redact database credentials in startup logs

fabidick22:
## Summary
- The controller logs the full Config struct at startup, exposing `Database.Url` with plaintext credentials
- Implement `logr.Marshaler` on `Config` to redact the password via `url.Redacted()` before logging


fabidick22:
@EItanya Could someone review these changes?
Thanks.

PR #1539, opened 2026-03-24
feat(skills-init): add securityContext + resources configuration to i…

jholt96:
…nitContainer

there are issues with the initContainer where if a limitRanger is set in the namespace then it can add a limit that is too low for the skills to be added. There is also an issue where the securityContext is not set for enterprise controls and slows adoption

### DIFF
--- a/go/core/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/core/internal/controller/translator/agent/adk_api_translator.go
@@ -1773,7 +1773,6 @@ func buildSkillsInitContainer(
 	if err != nil {
 		return corev1.Container{}, nil, err
 	}
-	
 	initSecCtx := securityContext
 	if initSecCtx != nil {
 		initSecCtx = initSecCtx.DeepCopy()


jholt96:
Just to add more context on this, We have strict securityContext requirements for all containers and this was blocking us from being able to properly deploy this out for our clusters. The same for the resources block being a requirement. Lmk if there are any changes!

jholt96:
> So I think a much simpler fix here could just use a restricted security context and small resources as the default rather than making this configurable. What do you think?

I think a standard securityContext would be totally fine with me if thats the direction we want. I was trying to keep it configurable as that was the behavior before. I can update the PR with that!

For the resources, I actually saw the initContainer go OOM from the limit ranger setting too low of limits. We were having it pull about 10 skills folders which is when we were seeing that happen. Granted they were unnecessarily bloated folders but I think it might make sense to keep this configurable because eventually someone will hit the limits we set. 

The other change in here is populating the envs for the initContainer. I can break that out into a separate PR if you want but we have github access behind corporate proxy so we needed to set those envs. 

Lmk what you think and I can adjust the PR from there!

EItanya:
> > So I think a much simpler fix here could just use a restricted security context and small resources as the default rather than making this configurable. What do you think?
> 
> I think a standard securityContext would be totally fine with me if thats the direction we want. I was trying to keep it configurable as that was the behavior before. I can update the PR with that!
> 
> For the resources, I actually saw the initContainer go OOM from the limit ranger setting too low of limits. We were having it pull about 10 skills folders which is when we were seeing that happen. Granted they were unnecessarily bloated folders but I think it might make sense to keep this configurable because eventually someone will hit the limits we set.
> 
> The other change in here is populating the envs for the initContainer. I can break that out into a separate PR if you want but we have github access behind corporate proxy so we needed to set those envs.
> 
> Lmk what you think and I can adjust the PR from there!

Ok, I buy that, but I do think we should probably split the env vars, and maybe have separate vars because I don't think we necessarily need to set all of the env vars from the main container, what do you think?

jholt96:
> > > So I think a much simpler fix here could just use a restricted security context and small resources as the default rather than making this configurable. What do you think?
> > 
> > 
> > I think a standard securityContext would be totally fine with me if thats the direction we want. I was trying to keep it configurable as that was the behavior before. I can update the PR with that!
> > For the resources, I actually saw the initContainer go OOM from the limit ranger setting too low of limits. We were having it pull about 10 skills folders which is when we were seeing that happen. Granted they were unnecessarily bloated folders but I think it might make sense to keep this configurable because eventually someone will hit the limits we set.
> > The other change in here is populating the envs for the initContainer. I can break that out into a separate PR if you want but we have github access behind corporate proxy so we needed to set those envs.
> > Lmk what you think and I can adjust the PR from there!
> 
> Ok, I buy that, but I do think we should probably split the env vars, and maybe have separate vars because I don't think we necessarily need to set all of the env vars from the main container, what do you think?

Makes sense to me will get the PR updated

jholt96:
> > > So I think a much simpler fix here could just use a restricted security context and small resources as the default rather than making this configurable. What do you think?
> > 
> > 
> > I think a standard securityContext would be totally fine with me if thats the direction we want. I was trying to keep it configurable as that was the behavior before. I can update the PR with that!
> > For the resources, I actually saw the initContainer go OOM from the limit ranger setting too low of limits. We were having it pull about 10 skills folders which is when we were seeing that happen. Granted they were unnecessarily bloated folders but I think it might make sense to keep this configurable because eventually someone will hit the limits we set.
> > The other change in here is populating the envs for the initContainer. I can break that out into a separate PR if you want but we have github access behind corporate proxy so we needed to set those envs.
> > Lmk what you think and I can adjust the PR from there!
> 
> Ok, I buy that, but I do think we should probably split the env vars, and maybe have separate vars because I don't think we necessarily need to set all of the env vars from the main container, what do you think?

Sorry for the delay, but the PR is updated. The changes are now:
1. add Resource Config
2. add Env Config
3. revert the securityContext as I think this [PR](https://github.com/kagent-dev/kagent/commit/476a03c69d161d6fac97316ab6dc91acedeade09) actually fixes the issues seen.

PR #1545, opened 2026-03-25
Add postgres service account

iplay88keys:
## Description

The bundled PostgreSQL deployment was missing a `ServiceAccount`. Unlike the controller and UI components, which each have a dedicated `ServiceAccount` and reference it via `serviceAccountName` in their pod specs, the PostgreSQL pod was running without one, falling back to the `default` service account in the namespace.

Running pods against the `default` service account is a security best practice violation. It conflates identity across unrelated workloads and makes it harder to scope RBAC or audit pod-level API access. Each component should have a dedicated, least-privilege identity.


iplay88keys:
> Why not colocate this with `postgres.yaml`?

No particular reason. That works for me.

PR #1551, opened 2026-03-26
fix: don't set Privileged when AllowPrivilegeEscalation is false for skills

skhedim:
## Problem

When an `Agent` spec includes a container `securityContext` with
`allowPrivilegeEscalation: false` (PSS Restricted profile) **and** the agent
has `spec.skills.refs` configured, the controller generates an invalid pod spec:

```
cannot set allowPrivilegeEscalation to false and privileged to true
```

Kubernetes rejects this combination at admission time, leaving the agent stuck
in a reconciliation loop with all pods failing to start.

## Root cause

`buildManifest()` sets `needSandbox = true` when skills are present (because
skills use `BashTool` which calls `srt` → bubblewrap for sandboxing), then
blindly sets `Privileged: true` on whatever `securityContext` is provided —
including one that already has `AllowPrivilegeEscalation: false`.

The securityContext merge did not check for this conflict before setting
`Privileged: true`.

## Fix

Add a helper `allowPrivilegeEscalationExplicitlyFalse()` that detects when the
user has opted out of privilege escalation, and skip `Privileged: true` in
that case.

When `Privileged` is withheld, `srt` degrades gracefully: on modern kernels
(EKS, GKE, AL2023 ≥ 5.10) that have unprivileged user namespaces enabled
(`user.max_user_namespaces > 0`), bubblewrap can still create sandboxes using
`clone(CLONE_NEWUSER)` without requiring a privileged seccomp profile.

> **Note on seccomp**: The containerd `RuntimeDefault` seccomp profile blocks
> `clone(CLONE_NEWUSER)` without `CAP_SYS_ADMIN`. Users who need full bwrap
> sandbox execution (running bash scripts inside skills) must additionally
> provide a custom `seccompProfile: Localhost` with a profile that allows
> user namespace syscalls. This PR makes agent deployment possible for
> PSS-Restricted namespaces; the seccomp tuning is an operational concern
> separate from this bug.

## Behaviour matrix

| Agent spec | needSandbox | Result |
|---|---|---|
| skills or executeCode, **no** custom securityContext | `true` | `Privileged: true` — full bwrap sandbox (unchanged) |
| skills + `allowPrivilegeEscalation: false` | `true` | No `Privileged` — srt uses user-namespace mode |
| executeCode + `allowPrivilegeEscalation: false` | `true` | No `Privileged` — srt uses user-namespace mode |
| No skills, no executeCode | `false` | No securityContext created (unchanged) |

### `go/core/internal/controller/translator/agent/adk_api_translator.go`

- Add comment explaining *why* `needSandbox = true` for skills (BashTool → srt → bwrap)
- Fix securityContext merge: guard `Privileged: true` with `allowPrivilegeEscalationExplicitlyFalse()`
- Add `allowPrivilegeEscalationExplicitlyFalse()` helper

### `go/core/internal/controller/translator/agent/security_context_test.go`

- Replace `TestSecurityContext_WithSandbox` (which tested an internally
  contradictory state) with two focused tests:
  - `TestSecurityContext_SkillsDefaultPrivilegedSandbox`: no custom
    securityContext → `Privileged: true` (default sandbox path)
  - `TestSecurityContext_SkillsPSSRestricted`: `allowPrivilegeEscalation: false`
    → no `Privileged`, original securityContext fields preserved intact

## Testing

```bash
cd go
go test -race -skip 'TestE2E.*' -v ./core/internal/controller/translator/agent/...
```

All existing tests pass. Two new tests cover both code paths.

## Related

- PSS Restricted policy: https://kubernetes.io/docs/concepts/security/pod-security-standards/
- Bubblewrap user namespaces: https://github.com/containers/bubblewrap#usage

### DIFF
--- a/python/packages/agentsts-adk/src/agentsts/adk/_base.py
+++ b/python/packages/agentsts-adk/src/agentsts/adk/_base.py
@@ -122,17 +122,22 @@ class ADKTokenPropagationPlugin(BasePlugin):
         Add the plugin to an ADK LLM agent by updating its MCP toolset
         Call this once when setting up the agent; do not call it at runtime.
         """
+        agent_name = getattr(agent, "name", "unknown")
+        logger.debug(f"add_to_agent called for agent {agent_name}")
+
         if not isinstance(agent, LlmAgent):
+            logger.debug(f"add_to_agent: agent {agent_name} is not LlmAgent, skipping")
             return
 
         if not agent.tools:
+            logger.debug(f"add_to_agent: agent {agent_name} has no tools, skipping")
             return
 
         for tool in agent.tools:
             if isinstance(tool, McpToolset):
                 mcp_toolset = tool
                 mcp_toolset._header_provider = self.header_provider
-                logger.debug("Updated tool connection params to include access token from STS server")
+                logger.debug(f"add_to_agent: updated MCP tool's header provider for agent {agent_name}")
 
     def header_provider(self, readonly_context: Optional[ReadonlyContext]) -> Dict[str, str]:
         # access saved token
@@ -140,6 +145,7 @@ class ADKTokenPropagationPlugin(BasePlugin):
         if not cache_entry:
             return {}
 
+        logger.debug("Using cached access token for tool invocation")
         return {
             "Authorization": f"Bearer {cache_entry.token}",
         }


skhedim:
Hello @EItanya What do you think of this fix ?

PR #1611, opened 2026-04-01
feat(adk): Allow session name generation using LLM

jmhbh:
- Allows session names to be optionally generated based on an interval configurable via the `Agent` CRD as shown below. By default, this feature is disabled only kicks in after the first message has been sent when the env var is configured. This behavior is identical in both the go and python agent runtimes. 
```yaml
spec:
  declarative:
    deployment:
      env:
      - name: KAGENT_SESSION_NAME_UPDATE_INTERVAL
        value: 1m
```

example:
Screenshot 2026-04-01 at 2 59 00 PM

closes #1555


EItanya (changes requested):
Great feature — LLM-generated session names are a big UX improvement. I like the overall approach and the implementation is solid across both Go and Python runtimes. However, I have concerns about the configuration mechanism and a few other items.

**Main concern: env var configuration**

The `KAGENT_SESSION_NAME_UPDATE_INTERVAL` env var approach doesn't feel right for this. The original issue (#1555) specifically calls for this to be configured in the agent spec:

> "In the kagent agent spec you should be able to optionally specify a model to use for session name generation, and an optional interval for session name updates."

This should be a first-class field on the Agent CRD (e.g., under `spec.sessionNameGeneration` or similar), not an env var injected via `spec.declarative.deployment.env`. Benefits:
- **Discoverability**: users can see it in the CRD schema / docs rather than needing to know a magic env var name
- **Validation**: the CRD can validate the interval format and model reference at admission time
- **Consistency**: aligns with how other agent behaviors are configured (model, tools, etc.)
- **Separation of concerns**: `deployment.env` is for runtime environment tuning, not feature configuration

The CRD field would be translated by the controller into the appropriate runtime config, keeping the env var as an internal implementation detail if needed, or better yet passing it through the agent config that's already being loaded.

Comment left by Claude on behalf of @EItanya

PR #1612, opened 2026-04-01
improve session resilience

joshadambell:
## Summary

Adds session resilience: automatic session recovery on backend and stream reconnection on frontend for seamless experience when sessions are deleted or pages reload during active tasks.

- **Backend session recovery**: Automatically recreates sessions on 404 errors during event append, with in-flight task detection
- **Frontend stream resumption**: Reconnects to in-flight tasks (`working`/`submitted` state) when page is reloaded
- **Bug fix**: Session events now properly populated on load (`events=events` instead of `events=[]`)

### Backend (`python/packages/kagent-adk`)
- **`_session_service.py`**: New `_recreate_session()` method handles 404 during `append_event()` — recreates the session preserving ID/user/agent, handles 409 conflicts from concurrent recreation, fetches and logs in-flight tasks after recovery. Single retry prevents infinite recursion.
- **`_session_service.py`**: Bug fix — `get_session()` now returns `events=events` instead of empty list
- **`test_session_service.py`**: 9 new tests covering append 404 recovery, session recreation, conflict handling, and in-flight task detection

### Frontend (`ui/src`)
- **`a2aClient.ts`**: New `resubscribeTask()` method sends `tasks/resubscribe` JSON-RPC to resume SSE streams
- **`messageHandlers.ts`**: `extractMessagesFromTasks()` now returns `TaskExtractionResult` with optional `pendingTask` for in-flight task detection
- **`ChatInterface.tsx`**: On page load, detects pending tasks and auto-resubscribes to their streams. Properly cleans up abort controller on unmount. HITL approval takes priority over resubscribe when both apply.
- **`AgentCallDisplay.tsx`**: Updated for new `extractMessagesFromTasks` return type

### Tests
- `a2aClient.test.ts`: New test suite for `resubscribeTask()` and existing client methods
- `messageHandlers.test.ts`: New tests for pending task detection (working, submitted, completed states)

## How It Works

1. **Session deleted during active task** → Backend 404 handler recreates session, logs in-flight tasks
2. **Page reload during active task** → Frontend detects pending task via `extractMessagesFromTasks`
3. **Frontend calls `resubscribeTask()`** → Reconnects to SSE stream for the in-flight task
4. **Stream resumes** → User sees continued updates without interruption
5. **If HITL approval is pending** → Approval takes priority over resubscribe (gated by `!hasPendingApproval`)


EItanya (changes requested):
Review by Claude on behalf of @EItanya

## Overall

Good concept — session resilience is a real need. However, there are several issues ranging from a critical regression bug to missing test coverage that should be addressed before merging.

---

## Critical

### 1. `events=events` bug fix introduces event duplication regression
**`_session_service.py:117`** — Changing `events=[]` to `events=events` while keeping the `for event in events: await super().append_event(session, event)` loop (lines 122-123) will cause every event to appear **twice** in the session — once from the constructor, once from the append loop. `BaseSessionService.append_event` appends to `session.events`, so this is a duplication bug, not a fix.

**Fix options:**
- Change to `events=events` AND remove the append loop (but then state deltas won't be applied)
- Keep `events=[]` with the loop (current working behavior)
- Or use `events=events` and only call state-delta application without re-appending

### 2. All existing `get_session` tests deleted
The previous test file had 8 tests covering `get_session` behavior including critical regression tests like `test_get_session_events_not_duplicated` and `test_get_session_state_delta_applied_once`. These are all gone in the rewrite. The `events=events` change specifically needs the duplication regression test.

---

## Medium

### 3. `setChatStatus('ready')` in `finally` block overwrites HITL status
**`ChatInterface.tsx`** — If the resubscribe stream ends with a `handleMessageEvent` that sets status to `input_required` (HITL), the `finally` block immediately overwrites it to `ready`. Compare with `sendApprovalDecision` which uses `setChatStatus(prev => prev === "thinking" ? "ready" : prev)`. The fix should similarly be conditional.

### 4. No stream timeout on resubscribe
The normal `sendA2AMessage` flow has a 10-minute inactivity timeout. The resubscribe `for await` loop has none — if the server accepts but never sends events, the stream hangs indefinitely.

### 5. Multiple pending tasks — only last one returned
**`messageHandlers.ts`** — The loop overwrites `pendingTask` on each match, so if multiple tasks are `working`/`submitted`, all but the last are silently ignored. Either return an array, or document that only one concurrent in-flight task per session is supported.

### 6. `source` field not preserved during session recreation
**`_session_service.py`** — `create_session` handles the `source` field from session state (line 46), but `_recreate_session` does not. Sessions with a `source` will lose it on recreation. The state-loss warning also doesn't account for `source` as a known field, so it would fire a misleading warning.

### 7. `any` return type on `resubscribeTask`
**`a2aClient.ts`** — `Promise<AsyncIterable<any>>` violates the project's "Don't use `any` type in TypeScript" convention. The SSE events should be typed.

### 8. Task fetching after recreation is noisy
**`_session_service.py:~220-250`** — The post-recreation task fetch adds an HTTP round-trip, and the log messages reference "UI should resubscribe" which leaks UI concerns into the backend session service. Consider removing or gating behind debug logging.

---

## Low

### 9. Shared `abortControllerRef` between resubscribe and send
**`ChatInterface.tsx`** — If `handleSendMessage` fires during the async gap of `initializeChat()`, the shared `abortControllerRef` could conflict. Consider a local `AbortController` for the resubscribe flow.

### 10. Code duplication with `streamA2AMessage`
The resubscribe flow manually reimplements the stream consumption, error handling, and abort management that `streamA2AMessage` already provides (including timeout). Consider refactoring to share the logic.

### 11. Missing test coverage
- No test for `_recreate_session` raising (e.g., server 500 during recreation)
- No test for retry-also-returning-404 (verifying no infinite recursion)
- No happy-path SSE test for `resubscribeTask`
- No test for multiple pending tasks scenario
- No test for state-loss warning when session has extra fields

### 12. `_recreate_session` exceptions lose 404 context
If recreation fails with a 500, the caller sees "session creation failed" instead of "event append failed because session was deleted and recreation also failed." Consider wrapping with context.

python/packages/kagent-adk/src/kagent/adk/_session_service.py
```diff
@@ -114,7 +114,7 @@ async def get_session(
             session = Session(
                 id=session_data["id"],
                 user_id=session_data["user_id"],
-                events=[],
```
EItanya, line 117: Comment left by Claude on behalf of @EItanya

**Critical: This change introduces event duplication.**

Changing `events=[]` to `events=events` while keeping the `for event in events: await super().append_event(session, event)` loop below (lines 122-123) will cause every event to appear **twice** in the session — once from the constructor pre-populating `session.events`, and again from `super().append_event()` which appends to the same list.

The previous test `test_get_session_events_not_duplicated` was specifically a regression test for this exact scenario.

**Fix options:**
1. Change to `events=events` AND remove the append loop (but state deltas won't be applied)
2. Keep `events=[]` with the loop (current working behavior)
3. Use `events=events`, skip appending, and apply state deltas separately

python/packages/kagent-adk/src/kagent/adk/_session_service.py
```diff
@@ -157,6 +157,94 @@ async def delete_session(self, *, app_name: str, user_id: str, session_id: str)
...
+        - user_id: Preserved
+        - agent_ref: Preserved
+        - session_name: Preserved (from session.state["session_name"])
+        - Other session.state fields: NOT preserved (lost on recreation)
+
+        Note: Only session_name is currently used by the application.
+        If additional state fields are added in the future, they must be
+        explicitly preserved here.
+
+        Args:
```
EItanya, line 177: Comment left by Claude on behalf of @EItanya

**`source` field not preserved during recreation.** `create_session` (line 46) handles the `source` field from session state, but `_recreate_session` does not. Sessions with a `source` will lose it on recreation.

Also, the state loss warning (`len(session.state) > 1`) only excludes `session_name` — a session with `source` in state would trigger a misleading warning even though `source` is a known field. Consider maintaining a `PRESERVED_STATE_FIELDS` set for both the preservation logic and the warning filter.

python/packages/kagent-adk/src/kagent/adk/_session_service.py
```diff
@@ -157,6 +157,94 @@ async def delete_session(self, *, app_name: str, user_id: str, session_id: str)
...
+            logger.warning(
+                "Session %s has additional state fields that will not be preserved during recreation: %s. "
+                "Update _recreate_session() if these fields are critical.",
+                session.id,
+                extra_fields,
+            )
+
+        response = await self.client.post(
+            "/api/sessions",
+            json=request_data,
```
EItanya, line 208: Comment left by Claude on behalf of @EItanya

**Task fetching here adds latency and leaks UI concerns into the backend.**

This block adds an extra HTTP round-trip on every session recreation purely for informational logging. The log messages reference "UI should resubscribe" which is a cross-layer concern — the backend session service shouldn't be aware of or advise the UI. The UI handles resubscription independently via its own pending task detection.

Consider removing this block entirely, or gating it behind `logger.debug` level.

python/packages/kagent-adk/src/kagent/adk/_session_service.py
```diff
@@ -157,6 +157,94 @@ async def delete_session(self, *, app_name: str, user_id: str, session_id: str)
...
+            if tasks_data.get("data"):
+                logger.info(
+                    "Session %s has %d existing task(s) after recreation",
+                    session.id,
+                    len(tasks_data["data"]),
+                )
+                # Log info about in-flight tasks
+                for task in tasks_data["data"]:
+                    task_status = task.get("status", {})
+                    task_state = task_status.get("state", "unknown")
```
EItanya: Comment left by Claude on behalf of @EItanya

If `_recreate_session` raises (e.g., 500 from the server), the exception propagates without the original 404 context. The caller sees "session creation failed" rather than "event append failed because session was deleted and recreation also failed."

Consider wrapping:
```python
try:
    await self._recreate_session(session)
except Exception as e:
    raise RuntimeError(f"Session {session.id} not found and recreation failed") from e
```

python/packages/kagent-adk/tests/unittests/test_session_service.py
```diff
@@ -1,166 +1,276 @@
-"""Tests for KAgentSessionService."""
```
EItanya, line 1: Comment left by Claude on behalf of @EItanya

**All 8 existing `get_session` tests were deleted.** The previous test file had critical regression tests including:
- `test_get_session_events_not_duplicated`
- `test_get_session_single_event_not_duplicated`
- `test_get_session_state_delta_applied_once`
- `test_get_session_multiple_state_deltas_applied_once`
- `test_get_session_returns_none_on_404`
- `test_get_session_event_ids_preserved`

These tests need to be preserved — especially the duplication regression tests, given this PR modifies the very code they guard.

Missing test coverage for the new code:
- No test for `_recreate_session` raising (server 500 during recreation)
- No test for retry-also-returning-404 (verifying no infinite recursion)
- No test for the state-loss warning when session has extra fields

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -155,6 +155,45 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
...
+              const stream = await kagentA2AClient.resubscribeTask(
+                selectedNamespace,
+                selectedAgentName,
+                pendingTask.taskId,
+                abortControllerRef.current.signal
+              );
+
+              for await (const event of stream) {
+                handleMessageEvent(event);
+                if (abortControllerRef.current?.signal.aborted) break;
```
EItanya: Comment left by Claude on behalf of @EItanya

**`setChatStatus('ready')` in `finally` can overwrite HITL status.**

If the resubscribe stream ends with a `handleMessageEvent` that sets status to `input_required` (e.g., task enters HITL), the `finally` block immediately overwrites it to `ready`.

Compare with how `sendApprovalDecision` handles this:
```ts
setChatStatus(prev => prev === "thinking" ? "ready" : prev);
```

Suggest using a conditional here too:
```ts
setChatStatus(prev => ['working', 'submitted'].includes(prev) ? 'ready' : prev);
```

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -155,6 +155,45 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
           if (hasPendingApproval) {
             setChatStatus("input_required");
           }
+
+          // If there's a pending task (and no pending approval taking priority), reconnect to its stream
+          if (pendingTask && !hasPendingApproval) {
+            setIsLoading(false);
+            setChatStatus(mapA2AStateToStatus(pendingTask.state as TaskState));
```
EItanya: Comment left by Claude on behalf of @EItanya

**No stream timeout.** The normal `sendA2AMessage` flow has a 10-minute inactivity timeout via `streamA2AMessage`. This `for await` loop has none — if the server accepts the resubscribe but never sends events, the stream hangs indefinitely.

Also, the shared `abortControllerRef` could conflict if `handleSendMessage` fires during the async gap. Consider using a local `AbortController` for the resubscribe flow, or extracting the timeout logic from `streamA2AMessage` to reuse here.

ui/src/lib/a2aClient.ts
```diff
@@ -76,6 +76,47 @@ export class KagentA2AClient {
     return this.processSSEStream(response.body);
   }
 
+  /**
+   * Resubscribe to an in-flight task to resume receiving streaming events
```
EItanya, line 81: Comment left by Claude on behalf of @EItanya

Nit: `Promise<AsyncIterable<any>>` — project convention says "Don't use `any` type in TypeScript." The SSE events could be typed as a union or at minimum use `unknown`.

ui/src/lib/messageHandlers.ts
```diff
@@ -4,12 +4,25 @@ import { convertToUserFriendlyName, isAgentToolName, messageUtils } from "@/lib/
...
 
+// Result type for extractMessagesFromTasks
+export interface TaskExtractionResult {
+  messages: Message[];
+  pendingTask?: { taskId: string; state: string };
+}
+
 // Helper functions for extracting data from stored tasks
-export function extractMessagesFromTasks(tasks: Task[]): Message[] {
+export function extractMessagesFromTasks(tasks: Task[]): TaskExtractionResult {
```
EItanya, line 14: Comment left by Claude on behalf of @EItanya

**Multiple pending tasks: only the last one wins.** If a session has multiple tasks in `working` or `submitted` state, `pendingTask` is overwritten on each iteration, silently dropping all but the last. The `ChatInterface` only resubscribes to one task.

Either return an array of pending tasks, or document/assert that only one concurrent in-flight task per session is supported.

PR #1633, opened 2026-04-06
fix(ui): restore tool selection feedback in onboarding and render selected tool names in review

whtssub:
Normalize server-name matching for Step 3 checkboxes and use unique per-tool keys in Step 4 to fix duplicate key issues.

refs #1614 

I have rendered/tested the updated docker image locally and found consistent desired results. Refer the below images. 

image

image


ui/src/components/onboarding/steps/ReviewStep.tsx
```diff
@@ -74,12 +74,34 @@ export function ReviewStep({ onboardingData, isLoading, onBack, onSubmit }: Revi
...
-                                    <Badge variant="secondary" key={tool.mcpServer?.name} className="flex items-center gap-1">
-                                        <FunctionSquare className="h-3 w-3" />
-                                        {tool.mcpServer?.name}
-                                    </Badge>
-                                ))}
+                                {onboardingData.selectedTools.flatMap((tool, toolIndex) => {
+                                    const serverName = tool.mcpServer?.name || "unknown-server";
+                                    const toolNames = tool.mcpServer?.toolNames || [];
+
+                                    if (toolNames.length === 0) {
```
peterj, line 81: shouldn't we just ignore this and not even show anything? if there's no tools, the MCP server is useless

peterj (changes requested):
one comment that I think should be resolved before we merge

ui/src/components/onboarding/steps/ReviewStep.tsx
whtssub: makes sense, no point in showing it in the review if there are no tools. I'm updating the code.

PR #1636, opened 2026-04-07
define a2a mcp timeout

lbi22:
# Configurable A2A MCP Timeout

## Description
This MR introduces a configurable timeout for the internal A2A connections established between the KAgent Controller's MCP handler and the destination Agent Pods.

Previously, when an MCP client requested the `invoke_agent` tool, the controller established a network channel to the agent with a hard-coded constraint of `30 * time.Second`. If the agent's LLM generation or tool execution took longer than 30 seconds, the connection forcefully terminated, preventing the use of long-running skills.

This change paramaterizes that constraint, making it fully configurable down from Helm charts to the Go application runtime, while enforcing a definitive ceiling of 30 minutes to prevent zombie connections.

## Changes Included
* **Go Application (`app.go`, `mcp_handler.go`)**
  * Added `--a2a-mcp-timeout` CLI flag with backward compatibility defaulting to `30s`.
  * Updated the controller's `MCPHandler` to dynamically adopt the new configuration value.
  * Implemented an absolute ceiling check: if a system administrator passes a timeout longer than `30m`, it correctly caps at `30m` max. 
* **Helm Deployment (`values.yaml`, `controller-configmap.yaml`)**
  * Exposed `controller.a2aMcpTimeout` in `values.yaml` (default: `"30s"`).
  * Mapped the helm configuration value seamlessly into the controller's pod context under the `A2A_MCP_TIMEOUT` environment variable.

## Verification & Testing
* Validated that a provided `a2aMcpTimeout` smoothly propagates across the deployment components to the lowest networking level in the `a2aclient`.
* Verified backward compatibility (retains 30s timeout without interference if no custom param is provided).

### DIFF
--- a/go/core/internal/mcp/mcp_handler.go
+++ b/go/core/internal/mcp/mcp_handler.go
@@ -59,6 +59,9 @@ type InvokeAgentOutput struct {
 // NewMCPHandler creates a new MCP handler
 // Wraps the StreamableHTTPHandler and adds A2A bridging and context management.
 func NewMCPHandler(kubeClient client.Client, a2aBaseURL string, authenticator auth.AuthProvider, a2aTimeout time.Duration) (*MCPHandler, error) {
+	if a2aTimeout <= 0 {
+		return nil, fmt.Errorf("a2aTimeout must be greater than 0 and no more than %s", 30*time.Minute)
+	}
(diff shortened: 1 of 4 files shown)


lbi22:
Closing as it is a duplicate

PR #1640, opened 2026-04-07
feat: agent sandbox support 

peterj:
adds a new SandboxAgent CRD that creates a sandbox agent using agent-sandbox CRD. 

Install: 

```
export VERSION="v0.3.10"                              
kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/${VERSION}/manifest.yaml
kubectl apply -f https://github.com/kubernetes-sigs/agent-sandbox/releases/download/${VERSION}/extensions.yaml
```

Then create SandboxAgent resource (same as Agent resource).


go/core/pkg/sandboxbackend/apis_available.go
```diff
@@ -0,0 +1,38 @@
...
+	extensionsv1alpha1 "sigs.k8s.io/agent-sandbox/extensions/api/v1alpha1"
+)
+
+// EnsureAgentSandboxAPIsRegistered checks that the apiserver exposes the agent-sandbox
+// resources kagent needs (SandboxTemplate, SandboxClaim, Sandbox). Call this before
+// creating or reconciling SandboxAgent when a sandbox backend is configured.
+//
+// When CRDs are missing, the apiserver returns a *meta.NoKindMatchError (or similar);
+// that surfaces as a clear prerequisite error instead of a late reconcile failure.
+func EnsureAgentSandboxAPIsRegistered(ctx context.Context, c client.Client) error {
```
EItanya, line 19: You should use the discovery API here so we don't need to actually check CRDs

EItanya (approved):

PR #1649, opened 2026-04-10
fix: dereference symlinks when copying git skill subpaths

samskelton1:
## Summary
When a git skill uses a path, the init script copies the selected subdirectory and then removes the original clone root. With cp -a, symlinks inside that subdirectory are preserved as symlinks, so repo-internal links can break once the clone root is deleted.

This switches that copy step to cp -rL so symlink targets are materialized before the source repo is removed.

## Verification
- GOCACHE=/tmp/go-build GOMODCACHE=/tmp/go-mod-cache go test ./core/internal/controller/translator/agent/...


jholt96:
@EItanya bumping this PR as it would help us tremendously with our current skills setup. follow symlinks allow us to maintain single kb folder for skills that are for different agents e.g claude code vs agents in kagent

EItanya (approved):

EItanya (approved):

Issue #1645, opened 2026-04-09
[BUG] Runtime: go | Agent fails to call sub-agents | Missing "request" field


sidritic:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

I have a multi-agent architecture: 1 orchestrator delegating tasks to the sub-agents.
Given the better performance of `runtime: go`, I switched all of the Agents to it. The single agents seem to work as before, while the main agent (the orchestrator) can not delegate tasks. E.g. error
```
required: missing properties: [\"request\"]"}
```
Direct tools seem to be working properly, e.g. datetime_get_current_time

Workaround: reverted the orchestrator agent to the default runtime (python) and kept all sub-agents with `runtime: go`

### 🔄 Steps To Reproduce

1. create 2 agents, using `runtime: go`
2. config one of them to use the other as a sub-agent
3. ask smth to the "main" agent that would delegate to the sub-agent

### 🤔 Expected Behavior

Agents with `runtime: go` should be able to delegate task to sub-agents.
The proper field "request" is filled during the delegation.

### 📱 Actual Behavior

Agent with `runtime: go` can't deletate tasks to sub-agents.
Intput during delegate has empty `request` field

### 💻 Environment

AWS EKS - eks.29
Kubernetes - 1.33
Kagent: 0.8.6
Model: Bedrock - Anthropic Claude Sonnet 4.5

### 📋 Logs

```shell

```

### 📷 Screenshots

It send an empty request field to the sub-agents:

<img width="974" height="330" alt="Image" src="https://github.com/user-attachments/assets/c8070087-9488-4bf5-baff-dd34890a9183" />

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR to fix this issue

supreme-gg-gg:
[...]

Issue #1643, opened 2026-04-08
Track Go runtime parity gaps


supreme-gg-gg:
## Summary

This issue tracks **user-facing differences** between Go and Python runtimes. Solutions to the following issues can either be enhanced documentation or feature additions to the Go runtime. Please comment below if any more differences are found when you are using the Go runtime.

## Action Items

- [x] Tighten embedding model compatibility for Go runtime.
  Python memory embeddings support `openai`, `azure_openai`, `ollama`, `gemini`/Vertex AI, and `bedrock`. Go currently supports only `openai` and `azure_openai`, even though translated embedding configs can point at other providers. Either validation or additional support could work here. Python embedding models also can use a refactor to move them out from `_memory_service.py`.

- [x] Clarify or fix Bedrock model support in Go runtime.
  Python Bedrock uses the Bedrock Converse API and supports Bedrock model families broadly. Go currently routes Bedrock through the Anthropic Bedrock client, so Bedrock is effectively Anthropic-only today.

- [x] Implement `apiKeyPassthrough` in the Go runtime.
  Users can enable it today, but Go does not currently apply it to LLM clients.

- [x] Apply model TLS settings in the Go runtime.
  `ModelConfig` TLS fields are translated, but Go does not currently honor them when creating model clients.

- [x] Honor `ollama.options` in the Go runtime.
  Python uses them; Go currently ignores them. A solution would be to use Ollama SDK if available.

- [ ] Decide whether context compaction is a must-fix or a documented Go limitation.

- [ ] Decide whether memory auto-save parity is required.
  Python auto-saves periodically; Go currently relies more on explicit `save_memory` tool usage.

- [x] Decide whether skills sandboxing parity is required.
  Python uses `srt`; Go currently executes skills shell commands directly. Might not be necessary if we are moving away from SRT

jsonmp-k8:
Regarding the skills sandboxing item (marked as resolved with the note "moving away from SRT"):

SRT is still actively used in both Python and Go runtimes:

- **Python**: `sandbox_code_executer.py`, `shell.py`, and `bash_tool.py` all invoke `srt` for command execution
- **Go**: `go/adk/pkg/skills/shell.go` runs commands via `srt bash -c` with settings from `KAGENT_SRT_SETTINGS_PATH`
- **Dockerfiles**: Both `python/Dockerfile` and `go/Dockerfile.full` install SRT from `anthropic-experimental/sandbox-runtime`

[...]

PR #1540, opened 2026-03-24
feat(kagent-adk): remove litellm as dependency from kagent-adk

jmhbh:
Removes `litellm` as a dependency from kagent. `litellm` is now replaced with provider specific sdks. 

### Testing
**ollama**
1. deploy ollama
```bash
kubectl apply -f - <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ollama
  namespace: kagent
spec:
  replicas: 1
  selector:
    matchLabels:
      app: ollama
  template:
    metadata:
      labels:
        app: ollama
    spec:
      containers:
      - name: ollama
        image: ollama/ollama:latest
        ports:
        - containerPort: 11434
        resources:
          requests:
            memory: "2Gi"
          limits:
            memory: "4Gi"
---
apiVersion: v1
kind: Service
metadata:
  name: ollama
  namespace: kagent
spec:
  selector:
    app: ollama
  ports:
  - port: 11434
    targetPort: 11434
EOF
```
2. Pull a small model into the ollama deployment - `kubectl -n kagent exec -it deploy/ollama -- ollama pull llama3.2:1b`
3. create model config an agent
```bash
kubectl apply -f - <<EOF
apiVersion: kagent.dev/v1alpha2
kind: ModelConfig
metadata:
  name: ollama-test-config
  namespace: kagent
spec:
  provider: Ollama
  model: llama3.2:1b
  ollama:
    host: "http://ollama:11434"
    options:
      num_ctx: "2048"
      temperature: "0.7"
      top_k: "40"
---
apiVersion: kagent.dev/v1alpha2
kind: Agent
metadata:
  name: ollama-test
  namespace: kagent
spec:
  type: Declarative
  description: "Ollama native SDK test"
  declarative:
    modelConfig: ollama-test-config
    systemMessage: "You are a helpful assistant. Answer concisely."
EOF
```
4. Port forward UI and test agent. `kubectl port-forward -n kagent svc/kagent-ui 3000:8080`
Screenshot 2026-03-25 at 7 21 20 PM
5. Test memory recall by pulling an embedding model `kubectl -n kagent exec -it deploy/ollama -- ollama pull nomic-embed-text`
6. Create the embedding model config an ollama memory test agent
```bash
kubectl apply -f - <<EOF
apiVersion: kagent.dev/v1alpha2
kind: ModelConfig
metadata:
  name: ollama-embedding-config
  namespace: kagent
spec:
  provider: Ollama
  model: nomic-embed-text
  ollama:
[...]

Issue #1473, opened 2026-03-10
fix(bedrock): sanitize tool names for Bedrock API constraints


frank-bee:
## Problem

When using kagent with Amazon Bedrock models, MCP servers (e.g. GitHub Copilot) may produce tool names containing dots, spaces, or other characters that violate Bedrock's API constraints.

Bedrock requires tool names to match `[a-zA-Z0-9_-]+` with a minimum length of 1. Any name outside this pattern causes API errors like:

```
ValidationException: A tool name is invalid. Tool names can only contain letters, numbers, underscores, and hyphens
```

## Root Cause

`KAgentLiteLlm` does not sanitize tool names before passing them to the Bedrock API, and does not sanitize tool names in the conversation history (function call parts) sent in subsequent requests.

## Fix

Override `generate_content_async` in `KAgentLiteLlm` to:
- Sanitize tool names in the outgoing `LlmRequest` history (function call parts in `contents`)
- Sanitize tool names in incoming `LlmResponse` parts

Invalid characters are replaced with `_`. Empty/completely invalid names fall back to `unknown_tool_<idx>`. A warning is logged whenever a name is sanitized, to aid debugging.

## Affected File

`python/packages/kagent-adk/src/kagent/adk/models/_litellm.py`

PR #1650, opened 2026-04-10
feat(go-adk): enhance model config and provider support

supreme-gg-gg:
## Motivation

This will allow most users to directly switch from `runtime: python` to `runtime: go` without needing to worry about existing LLM provider configs since everything will be supported on the Go side, facilitating adoption of the new go runtime

## Summary

Closes most of the gap between python and go identified in #1643 

- TLS and api key passthrough for LLM provider
- Support Ollama and Bedrock using client SDK as we've done earlier in #1540 
- Use Bedrock client instead of messages API for Anthropic on Bedrock to support all bedrock runtime models 
- Tightens tool config conversion for Anthropic + Bedrock and fixes issues like #1645, #1683
- Sanitize ToolName for bedrock LLMs #1473, see [bedrock API docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolSpecification.html)
- Refactor embedding models in python to be separate from memory service
- Strip approval confirmation synthetic tool calls from LLM requests, these messages are persisted in task / events store and used by ADK internally but sending them to the model will be wasting tokens and confuse the model. If the session is long and has many HITL events, these internal tool messages will be consuming many unnecessary token!

## Testing Plan

- [x] All new unit tests in go adk passes, all old unit tests in python passes
- [x] Test for no regression with OpenAI and Gemini models in go runtime
- [x] Validate with a wide range of use cases such as: builtin tools (ask user, save memory), ADK built-in tools (load memory) MCP tools, Remote A2A (subagent) tools, HITL tools (approvals)
- [x] Bedrock LLM and embedding model in Go runtime
- [x] OpenAI API key passthrough with A2A `--token` option
- [x] Ollama LLM and embedding in Go runtime (local models, Gemma 4 + embedding Gemma)
- [x] Ollama with TLS (local https server with self-signed certs)


python/packages/kagent-adk/src/kagent/adk/models/_embedding.py
```diff
@@ -0,0 +1,230 @@
...
+logger = logging.getLogger(__name__)
+
+
+class KAgentEmbedding:
+    """Client for generating embeddings using provider-specific SDKs.
+
+    This client is standalone and has no dependencies on the memory service.
+    It supports multiple embedding providers with dimension enforcement and
+    L2 normalization.
+    """
```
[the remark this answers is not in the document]
supreme-gg-gg: We intentionally NOT process embeddings if they are already in the correct dimension, we trust that the model provider return what it intends if that's the case. We only process then if they don't match, so that's why we have this statement here

python/packages/kagent-adk/src/kagent/adk/models/_embedding.py
```diff
@@ -0,0 +1,230 @@
...
+                processed_embedding = self._normalize_l2(processed_embedding).tolist()
+            elif dim < self.TARGET_DIMENSION:
+                logger.error(
+                    "Embedding dimension %d is smaller than required %d; rejecting embeddings batch",
+                    dim,
+                    self.TARGET_DIMENSION,
+                )
+                return []
+
+            processed.append(processed_embedding)
```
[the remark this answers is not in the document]
supreme-gg-gg: I responded to the above one

python/packages/kagent-adk/src/kagent/adk/models/_embedding.py
```diff
@@ -0,0 +1,230 @@
...
+        self,
+        texts: List[str],
+    ) -> List[List[float]]:
+        """Embed using the AWS Bedrock Titan Embedding API via boto3.
+
+        Uses the same credential chain (env vars, IRSA, instance profile) as
+        KAgentBedrockLlm.  Each text is embedded individually because the
+        Titan Embedding API accepts a single ``inputText`` per invocation.
+        """
+        import boto3
```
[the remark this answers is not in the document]
supreme-gg-gg: true

go/adk/pkg/agent/approval.go
```diff
@@ -3,10 +3,48 @@ package agent
...
+			}
+			if len(filtered) == 0 {
+				continue
+			}
+			c.Parts = filtered
+			out = append(out, c)
+		}
+		req.Contents = out
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
supreme-gg-gg: True

go/adk/pkg/embedding/embedding.go
```diff
@@ -210,6 +222,183 @@ func (c *Client) generateAzureOpenAI(ctx context.Context, texts []string) ([][]f
...
+	if region == "" {
+		region = "us-east-1"
+	}
+
+	awsCfg, err := awsconfig.LoadDefaultConfig(ctx, awsconfig.WithRegion(region))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
+	client := bedrockruntime.NewFromConfig(awsCfg)
```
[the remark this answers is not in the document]
supreme-gg-gg: Done that with both bedrock and gemini clients

go/adk/pkg/models/openai.go
```diff
@@ -134,17 +139,9 @@ func NewAzureOpenAIModelWithLogger(config *AzureOpenAIConfig, logger logr.Logger
 		option.WithHeader("Api-Key", apiKey),
 		option.WithMiddleware(azurePathRewriteMiddleware()),
 	}
```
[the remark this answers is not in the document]
supreme-gg-gg: This (and the following duplicated) comment is valid, I have made a change here by setting the api key from context at each SDK call with the options so it handles the appropriate api key header. An alternative is to do that at the transport layer, but that kind of couples api key header (which is owned by the SDK and can change) with the transport http client.

go/adk/pkg/embedding/embedding.go
EItanya, line 1: Why are these subfields on a `Client` and not their own structs?

go/adk/pkg/a2a/executor.go
```diff
@@ -114,6 +116,19 @@ func (e *KAgentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.RequestCont
 	}
 	sessionID := reqCtx.ContextID
 
+	// Extract Bearer token from incoming request for API key passthrough
+	if callCtx, ok := a2asrv.CallContextFrom(ctx); ok {
```
EItanya: Not sure if this would be better, but couldn't we just turn this into a function and reuse that for passthrough rather than adding a new context variable?

go/adk/pkg/models/tls.go
```diff
@@ -0,0 +1,88 @@
...
+	tlsConfig *tls.Config
+}
+
+func (t *tlsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// If the request's TLS config needs to be modified, we would need to
+	// create a new client for each request, which is inefficient.
+	// Instead, we rely on the base transport having the TLS config set.
+	// This wrapper is primarily for when base is not an *http.Transport.
+	return t.base.RoundTrip(req)
+}
```
EItanya: I don't quite understand this function

python/packages/kagent-adk/src/kagent/adk/models/_embedding.py
```diff
@@ -0,0 +1,230 @@
...
+from typing import Any, List, Union
+
+import numpy as np
+
+from kagent.adk.types import EmbeddingConfig
+
+logger = logging.getLogger(__name__)
+
+
+class KAgentEmbedding:
```
EItanya, line 24: Same comment about packing clients vs creating some sort of interface or using multiple structs

EItanya (commented):
A couple of structural things about the clients, but mostly looks good!

go/adk/pkg/embedding/embedding.go
supreme-gg-gg: Agreed, I've split them to providers

go/adk/pkg/models/tls.go
supreme-gg-gg: Removed it, I don't think it's necessary

go/adk/pkg/a2a/executor.go
supreme-gg-gg: We could, but I think context variable keeps models decoupled from the `a2asrv` package (A2A server), similar to how Python uses session state to pass the key rather than having models read the A2A request directly. 

An alternative works as well though, if this is what you're suggesting?

```go
func openAIPassthroughOpts(ctx context.Context, m *OpenAIModel) []option.RequestOption {
    // Read key from call context directly, avoid context var, we can make this a function
    callCtx, ok := a2asrv.CallContextFrom(ctx)                             
    if !ok { return nil }                                                  
    vals, _ := callCtx.RequestMeta().Get("authorization")                  
    // parse Bearer token...                                               
}                                                                          
```

python/packages/kagent-adk/src/kagent/adk/models/_embedding.py
supreme-gg-gg: For python I don't think it's too necessary, since each provider interface would just be a single method to generate embeddings so practically there would be no difference (this is because we don't need client caching for Python since asyncio is single threaded)

EItanya (approved):

PR #1694, opened 2026-04-18
fix(ui): poll agent list until all agents report deploymentReady

tjorourke:
## Bug

The agents page shows **Agent not Ready** permanently for agents that were still starting when the page loaded — even after the Kubernetes deployment becomes healthy.

## Steps to reproduce

1. Create a new agent via the wizard
2. Navigate to the agents list immediately after creation
3. The card shows **Agent not Ready**
4. Refresh the page — card now shows Ready (backend was fine all along)

## Root cause

`AgentsProvider` calls `fetchAgents()` once on mount with no re-fetch mechanism. The `deploymentReady` flag from the initial response is stale forever.

## Fix

Add a polling effect that activates only while any agent has `deploymentReady=false` and clears itself automatically once all agents are ready — no unnecessary API calls in steady state.


EItanya:
Hey there, in order to merge your PR you will need to sign your commits, you can look at the DCO job to see the details

PR #1696, opened 2026-04-18
feat: transparent zstd/gzip decompression for ConfigMap prompt data sources

tjorourke:
## Problem

Large knowledge bases often exceed what fits comfortably in a ConfigMap. Users must either truncate their prompt data or build separate MCP servers just to serve static content.

## Solution

Add transparent decompression for ConfigMap data sources. When a ConfigMap has the annotation `kagent.dev/compression` (value: `gzip` or `zstd`), `GetConfigMapData()` automatically base64-decodes and decompresses all values.

Completely transparent to the agent — `include()` works unchanged. Decompression happens at the ConfigMap read layer.

### Compression comparison (real-world 521KB knowledge base)

| Algorithm | Size | Ratio |
|-----------|------|-------|
| Raw | 521KB | 1:1 |
| gzip -9 | 163KB | 3.2:1 |
| zstd -19 | 130KB | 4:1 |

### Changes

- `config_map.go` — add CompressionAnnotation + decompress helper
- `config_map_compression_test.go` — unit tests for gzip, zstd, unsupported algo, invalid base64

### DIFF
--- a/go/core/internal/utils/config_map.go
+++ b/go/core/internal/utils/config_map.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/klauspost/compress/zstd"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


go/core/internal/utils/config_map.go
```diff
@@ -1,28 +1,116 @@
...
+	// transparently decompressed when read via GetConfigMapData.
+	CompressionAnnotation = "kagent.dev/compression"
+
+	// maxDecompressedSize is the upper bound on decompressed output (10 MB).
+	// This prevents a small compressed payload from expanding into an
+	// arbitrarily large allocation that could OOM the controller.
+	maxDecompressedSize = 10 << 20 // 10 MiB
+)
+
+// GetConfigMapData fetches all data from a ConfigMap. If the ConfigMap carries
```
EItanya, line 30: This data gets rewritten to a secret, so I don't think really makes sense unless we also compress that outgoing data as well

tjorourke: ah ok, i can fix that - but i think we should yes. Just compress it here too. Similar pattern to helm data in secret.

PR #1697, opened 2026-04-19
helm: Allow configuring agent runtime

marcofranssen:
This PR allows to configure the agent runtime as Go. In the agent.yaml templates I kept the default as python. 

In the kagent chart I set the runtime by default to Go. It needs much less memory and CPU which also allows better use on local kind clusters which are more resource constrainted.

This way people can more easily run and try kagent on their local system. The resources are still overprovisioned to leave some room for more heavy use. Real production use can still override the values to higher numbers.


helm/agents/cilium-policy/values.yaml
```diff
@@ -1,11 +1,14 @@
...
-
-resources:
-  requests:
-    cpu: 100m
-    memory: 256Mi
-  limits:
-    cpu: 1000m
-    memory: 1Gi
\ No newline at end of file
+# -- Configure the runtime for the agent. Defaults to python, but can also be configured as go.
```
marcofranssen, line 1: FYI: Windows line-endings got replaced by linux line endings

helm/kagent/values.yaml
```diff
@@ -339,11 +339,10 @@ kagent-tools:
   replicaCount: 1
   resources:
     requests:
-      cpu: 100m
-      memory: 256Mi
+      cpu: 50m
```
marcofranssen, line 347: Reduced resources for Go runtime. Still overprovisioned, allowing room for more advanced usecases. Helm chart users can still increase request and limits to fit their heavy usecases.

helm/kagent/Chart-template.yaml
```diff
@@ -15,51 +15,51 @@ dependencies:
   - name: grafana-mcp
     version: ${VERSION}
     repository: file://../tools/grafana-mcp
-    condition: tools.grafana-mcp.enabled, agents.observability-agent.enabled
+    condition: grafana-mcp.enabled, observability-agent.enabled
```
EItanya, line 18: Why did this get updated? IIRC this happened before and was incorrect

helm/kagent/values.yaml
```diff
@@ -369,112 +368,103 @@ proxy:
...
-  cilium-debug-agent:
-    enabled: true
-    resources:
-      requests:
-        cpu: 100m
-        memory: 256Mi
-      limits:
-        cpu: 1000m
-        memory: 1Gi
+k8s-agent:
```
EItanya, line 371: This is a breaking change

EItanya (changes requested):
I like the addition of runtime to the agent charts, but the changes to the parent chart are mostly breaking so I don't think we can make those

helm/kagent/Chart-template.yaml
marcofranssen: Because otherwise you can't pass in the values for the sub charts. 

If you keep the values aligned with the names of the subcharts, those values can be passed to the subcharts, so users can override e.g. resources etc.

helm/kagent/values.yaml
marcofranssen: It is a mandatory change as otherwise the values will not propagate to the subcharts. The current value structure is broken.

If we want to be able to override values of the subcharts we need name the values at the root.

marcofranssen:
Thanks for the questions in the review. I see the helm chart isn't following helm practices and has quite a lot more flaws in the design. Without changing value structure, none of the values would propagate to the subcharts.

E.g. I'm even thinking to drop the subcharts approach and make all part of the same chart. Then we don't have to change the value structure. We would end up with a single helm chart. Using Helm templates to loop over the agents.

I can also implement it that way. Simplifying the helm chart. Making the things more idiomatic etc.

Let me know. I'm planning to contribute more and to address a whole bunch of gaps. So we can do a couple more production grade configurations. Also would love to improve the Kagent cli to make it more useable, similar like the cilium cli e.g.

helm/agents/istio/templates/agent.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
 spec:
   description: An Istio Expert AI Agent specializing in Istio operations, troubleshooting, and maintenance.
   type: Declarative
+  runtime: {{ .Values.runtime | default "python" }}
```
jmhbh, line 11: should be `spec.declarative.runtime` instead.  Currently this doesn't work because `spec.runtime` does not exist.

marcofranssen: Woops, good catch.

helm/kagent/Chart-template.yaml
marcofranssen: Also see the values.yaml basically all values under 

grafana-mcp (the name of the dependency) will propagate to the nested charts.

helm/kagent/values.yaml
marcofranssen: Also see the Chart.yaml.

Values will only propagate to the nested/dependency charts if the keys in the yaml here match the name/alias of a chart dependency.

So if we have a dependency `-name: k8s-agent` then the value structure here must be `k8s-agent:` in order for the values to propagate. `agents.k8s-agent` will not propagate. This is simply how Helm works.

EItanya:
This PR is looking great, the only thing I'll say is that I don't think we should change the default runtime yet. We're planning to do that in `0.10.0`

helm/kagent/values.yaml
jmhbh: Tested on `main`, and subchart values are currently broken and not propagated properly when they are set here so these changes make sense to me

jmhbh (commented):
Tested this locally, works great!

jmhbh (dismissed):

marcofranssen:
@EItanya tell me what to change, or make suggestions. Too prevent more back and forth... Would love to get this in so my deployment start working and I can contribute in the next improvements.

EItanya:
> @EItanya tell me what to change, or make suggestions. Too prevent more back and forth... Would love to get this in so my deployment start working and I can contribute in the next improvements.

For example, you set the runtime to all agents in the `values.yaml` to `go`. We are planning on doing that in the next minor release, can you either unset those or set them to python for now
```
cilium-debug-agent:
  enabled: true
  runtime: go
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
-      requests:
-        cpu: 100m
-        memory: 256Mi
-      limits:
-        cpu: 1000m
-        memory: 1Gi
+
+k8s-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+kgateway-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+istio-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+promql-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+observability-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+argo-rollouts-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+helm-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+cilium-policy-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+cilium-manager-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

helm/kagent/values.yaml
```diff
@@ -411,109 +410,104 @@ proxy:
...
+  resources:
+    requests:
+      cpu: 50m
+      memory: 128Mi
+    limits:
+      memory: 256Mi
+cilium-debug-agent:
+  enabled: true
+  runtime: go
```
marcofranssen:
```suggestion
```

marcofranssen:
All feedback handled @EItanya, removed the runtime override for now.

jmhbh (approved):

EItanya (approved):

helm/kagent/Chart-template.yaml
onematchfox: FYI. Would be great if this sort of change could at least be highlighted in the release notes since this is ultimately a breaking change. I certainly didn't expect that my current Helm values setup would need to be rewritten in a patch release that doesn't even mention changes to Helm values - "Allow configuring agent runtime" doesn't really highlight this 😄

> Because otherwise you can't pass in the values for the sub charts.

To this point... This is incorrect - yes you can still pass values through to the sub-chart. The defaults here (e.g. `resources`) were not being passed through. But there was nothing stopping anyone from configuring their own values.yaml correctly with something like

``` yaml
sub-chart:
  runtime: "go" 
```

Also, FYI, the side-effect of moving `enabled` under `sub-chart:` is that now, that key/value pair `enabled: true` will always be passed through to the sub-chart as well.

PR #1713, opened 2026-04-20
fix(go-adk): use genai.Schema for sub-agent tool declaration

### LINKED ISSUE #1645: [BUG] Runtime: go | Agent fails to call sub-agents | Missing "request" field

mesutoezdil:
Fixes #1645.

Ran into this while testing a multi-agent setup with Bedrock. The orchestrator kept failing every time it tried to delegate to a sub-agent. Logs showed it first sent `$task` as the field name, got rejected, then tried empty args, got rejected again. Never got through.

Dug into it and found that `functiontool.New` puts the schema in `ParametersJsonSchema` on the function declaration. Turns out the Anthropic and Bedrock adapters in the ADK only read the `Parameters` field and just skip `ParametersJsonSchema`. So from the LLM side the tool shows up with no schema at all, and it has to guess what to send.

OpenAI and Gemini both handle both fields so they work fine, which is why it did not reproduce there.

The fix replaces `functiontool.New` with a small `remoteA2ATool` struct that sets `Parameters` directly using `genai.Schema`, the same way `agenttool` does it inside the ADK. It also wires up `Run` and `ProcessRequest` so the ADK runtime is happy without needing to reach into internal packages.

Builds clean, all existing tests pass.

### DIFF
--- a/go/adk/pkg/tools/remote_a2a_tool.go
+++ b/go/adk/pkg/tools/remote_a2a_tool.go
@@ -13,8 +13,9 @@ import (
 	"github.com/a2aproject/a2a-go/a2aclient/agentcard"
 	"github.com/kagent-dev/kagent/go/adk/pkg/a2a"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"google.golang.org/adk/model"
 	"google.golang.org/adk/tool"
-	"google.golang.org/adk/tool/functiontool"
+	"google.golang.org/genai"
 )
 
 // userIDContextKey is the context key for passing the session user_id to the subagent.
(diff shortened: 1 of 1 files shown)


supreme-gg-gg (commented):
@mesutoezdil this should be fixed already in #1650 the proper way is to use `functiontool.New` according to ADK docs and to avoid hand-rolling tool packing functions (and a lot of other boilerplate)

mesutoezdil:
Thanks for pointing that out, @supreme-gg-gg! Closing this in favour of #1650. Will pull the latest and look at how you approached it properly.

PR #1716, opened 2026-04-20
fix(cli): recognize all providers in install and improve missing key error

mesutoezdil:
Closes #1622.

`GetModelProvider` was missing Gemini, GeminiVertexAI, AnthropicVertexAI and Bedrock from its switch. Any of those values in `KAGENT_DEFAULT_MODEL_PROVIDER` fell through to the default and came back as OpenAI, which then required `OPENAI_API_KEY` even though the user had no intention of using OpenAI.

`GetProviderAPIKey` now also covers Gemini (`GOOGLE_API_KEY`). Providers that rely on cloud credentials rather than an API key (Ollama, Bedrock, VertexAI variants) already returned empty string and are left as is.

The error message now mentions `KAGENT_DEFAULT_MODEL_PROVIDER` so users know they can switch providers instead of being stuck.

### DIFF
--- a/go/core/cli/internal/cli/agent/const.go
+++ b/go/core/cli/internal/cli/agent/const.go
@@ -28,6 +28,14 @@ func GetModelProvider() v1alpha2.ModelProvider {
 		return v1alpha2.ModelProviderAnthropic
 	case GetModelProviderHelmValuesKey(v1alpha2.ModelProviderAzureOpenAI):
 		return v1alpha2.ModelProviderAzureOpenAI
+	case GetModelProviderHelmValuesKey(v1alpha2.ModelProviderGemini):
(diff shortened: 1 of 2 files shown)


jmhbh (commented):
LGTM, but the copilot feedback above regarding support for `GEMINI_API_KEY` makes sense to me and should be addressed

mesutoezdil:
Good point. @jmhbh Updated `GetProviderAPIKey` for Gemini to mirror the exact fallback logic in the runtime (`go/adk/pkg/agent/agent.go`): prefer `GOOGLE_API_KEY`, fall back to `GEMINI_API_KEY` if the former is not set. Branch also rebased onto current main.

EItanya (approved):

PR #1687, opened 2026-04-17
[FEATURE] helm: improve kagent chart podAnnotations value usage

TOMOFUMI-KONDO:
kagent helm template has "podAnnotations" value but lacks several features, documentation, and tests.

This PR fixes below:

- add placeholder values in values.yaml for podAnnotations of controller and ui for documentations
- add controller podAnnotations fallback to top-level podAnnotations for consistency with ui-deployment
- add helm unittests for podAnnotations values

fix https://github.com/kagent-dev/kagent/issues/1685

PR #1720, opened 2026-04-20
feat(helm): add podAnnotations defaults for controller and ui

### LINKED ISSUE #1685: [FEATURE] Helm: add missing per-component podAnnotations defaults to values.yaml
### 📋 Prerequisites

### 📝 Feature Summary

Add `controller.podAnnotations` and `ui.podAnnotations` to `values.yaml` with default `{}`, and align the controller template's fallback behavior with UI.

The Helm templates reference per-component `podAnnotations` values that have no corresponding defaults in `values.yaml`:

- `controller-deployment.yaml` references `.Values.controller.podAnnotations` (no fallback to global)
- `ui-deployment.yaml` references `.Values.ui.podAnnotations | default .Values.podAnnotations`

However, `values.yaml` only defines the top-level `.podAnnotations: {}`. Neither `.Values.controller.podAnnotations` nor `.Values.ui.podAnnotations` have default entries, making it unclear to users that these values are configurable.

Additionally, the controller template does **not** fall back to the global `.podAnnotations`, unlike the UI template — which is inconsistent.

### 💡 Proposed Solution

1. Add `podAnnotations: {}` under `controller:` and `ui:` sections in `helm/kagent/values.yaml`
2. Add `| default .Values.podAnnotations` fallback to `controller-deployment.yaml`, aligning with the UI template pattern

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

This inconsistency was introduced in PR #599 (Helm refactor, `4067b282`). The templates were newly created with per-component references, but the corresponding default values were never added to `values.yaml`.

mesutoezdil:
Closes #1685.

The controller and ui deployment templates reference .Values.controller.podAnnotations and .Values.ui.podAnnotations but neither had a default in values.yaml. Users had no way to discover these without reading through the templates.

Added podAnnotations: {} under both sections in values.yaml.

Also aligned the controller template with the ui template: the controller now falls back to .Values.podAnnotations when .Values.controller.podAnnotations is not set, same as the ui template already did.


EItanya (changes requested):
Can you please add helm unit tests. That will also help validate against Copilot's comments

mesutoezdil:
@EItanya thx for your feedback! Added helm unit tests for both components. Also fixed a duplicate podAnnotations key that appeared after rebasing on main (main already had podAnnotations: {} in both sections without a comment, my commit added another one with a comment.. now consolidated into a single entry with the doc comment). All 48 tests passing locally.

EItanya:
Was this fixed by this commit: https://github.com/kagent-dev/kagent/commit/eeac4f41d45b890d4463f1debf18b4853c92b7a3

mesutoezdil:
Yes, looks like #1687 already covered this. Closing in favor of that one. Thanks for checking @EItanya!

PR #1725, opened 2026-04-21
feat: add imagePullSecrets support for container-based skills

### LINKED ISSUE #1222: [FEATURE] Container-based skills image download authentication support
### 📋 Prerequisites

### 📝 Feature Summary

support of loading images for container-based skills from private registries 

Currently, an init-container cannot pull images for container-based skills from private registries due to the lack of authentication support.

### 💡 Proposed Solution

Add authentication support to the image pull/download function

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

ppeau:
Closes #1222

## Problem

Container-based skills using `krane` to pull OCI images had no way to authenticate against private registries (Artifactory, ACR, ECR, etc.). The `imagePullSecrets` defined on the agent deployment were not passed to the `skills-init` init container, causing authentication failures like:
No matching credentials were found for "docker.artifactory.dev.example.com"
Error: pulling ...: Authentication is required

## Solution

Follows the approach discussed in #1222 by @s10gopal:

1. Added an `imagePullSecrets` field under `spec.skills` accepting a list of `kubernetes.io/dockerconfigjson` secrets
2. When `imagePullSecrets` is set, a new `docker-auth-init` init container is prepended — it merges all referenced secrets into a single `config.json` using `jq`
3. The `skills-init` container reads that merged config via the `DOCKER_CONFIG` env var, which `krane` picks up automatically when pulling skill images

## Changes

- `go/api/v1alpha2/agent_types.go`: add `ImagePullSecrets []corev1.LocalObjectReference` to `SkillForAgent` struct
- `go/api/v1alpha2/zz_generated.deepcopy.go`: regenerated DeepCopy for new field
- `go/core/internal/controller/translator/agent/adk_api_translator.go`: `buildSkillsInitContainer` now returns `[]Container`, prepends `docker-auth-init` when `imagePullSecrets` are present
- `docker/skills-init/Dockerfile`: add `jq` to the Alpine base image
- `.gitattributes`: enforce LF line endings on `*.sh.tmpl` files (prevents shell script breakage on Windows contributors)

## Usage

```yaml
kind: Agent
spec:
  skills:
    refs:
      - private-registry.example.com/my-org/my-skill:v1
    imagePullSecrets:
      - name: my-registry-secret  # kubernetes.io/dockerconfigjson secret
```

## Testing
Validated end-to-end on a local Kubernetes cluster with a private registry protected by htpasswd authentication:

Skill image hosted on the private registry, inaccessible without credentials
Agent configured with imagePullSecrets referencing a dockerconfigjson secret
docker-auth-init merged the credentials, skills-init pulled the image successfully via krane
Skill was correctly loaded and executed by the agent


EItanya (commented):
Overall this makes sense, but before we go down the road of adding a new API, is there anyway we can re-use the ImagePullSecrets which already get used for image pulling, or do those remain on the node and never mounted into the pod themselves?

ppeau:
Hi @EItanya, great question. We actually explored reusing the existing deployment imagePullSecrets first before adding a new field.

On the technical side, imagePullSecrets defined on a pod spec are consumed exclusively by the kubelet to pull container images. They are never mounted into the pod or made accessible to running containers. This means krane, executing inside the skills-init init container, has no way to read those credentials. We hit this wall directly during testing.

Even if it were technically possible, there is a design reason why it would not be the right approach in enterprise environments. The registry used to deploy the kagent system and the registry hosting skill images are typically owned by completely different teams with different security boundaries. The platform/ops team manages the kagent deployment and its registry credentials, while the line-of-business team produces and owns the skill images, hosted in their own private registry (Artifactory, ACR, ECR, etc.). Reusing the deployment imagePullSecrets would couple these two security contexts together, violate the principle of least privilege, and make skills effectively unusable for any team whose registry differs from the one used to deploy kagent. That is the common case at scale.

The new imagePullSecrets field under spec.skills directly mirrors the standard Kubernetes pattern where a pod can reference multiple imagePullSecrets for different registries. It introduces no new concept, just applies the same model at the skill level.

Happy to discuss further if needed!

EItanya:
> Hi @EItanya, great question. We actually explored reusing the existing deployment imagePullSecrets first before adding a new field.
> 
> On the technical side, imagePullSecrets defined on a pod spec are consumed exclusively by the kubelet to pull container images. They are never mounted into the pod or made accessible to running containers. This means krane, executing inside the skills-init init container, has no way to read those credentials. We hit this wall directly during testing.
> 
> Even if it were technically possible, there is a design reason why it would not be the right approach in enterprise environments. The registry used to deploy the kagent system and the registry hosting skill images are typically owned by completely different teams with different security boundaries. The platform/ops team manages the kagent deployment and its registry credentials, while the line-of-business team produces and owns the skill images, hosted in their own private registry (Artifactory, ACR, ECR, etc.). Reusing the deployment imagePullSecrets would couple these two security contexts together, violate the principle of least privilege, and make skills effectively unusable for any team whose registry differs from the one used to deploy kagent. That is the common case at scale.
> 
> The new imagePullSecrets field under spec.skills directly mirrors the standard Kubernetes pattern where a pod can reference multiple imagePullSecrets for different registries. It introduces no new concept, just applies the same model at the skill level.
> 
> Happy to discuss further if needed!

Ok I buy that logic. What do you think about renaming the field to `PullSecrets` instead `ImagePullSecrets` since these aren't really images

ppeau:
> > Hi @EItanya, great question. We actually explored reusing the existing deployment imagePullSecrets first before adding a new field.
> > On the technical side, imagePullSecrets defined on a pod spec are consumed exclusively by the kubelet to pull container images. They are never mounted into the pod or made accessible to running containers. This means krane, executing inside the skills-init init container, has no way to read those credentials. We hit this wall directly during testing.
> > Even if it were technically possible, there is a design reason why it would not be the right approach in enterprise environments. The registry used to deploy the kagent system and the registry hosting skill images are typically owned by completely different teams with different security boundaries. The platform/ops team manages the kagent deployment and its registry credentials, while the line-of-business team produces and owns the skill images, hosted in their own private registry (Artifactory, ACR, ECR, etc.). Reusing the deployment imagePullSecrets would couple these two security contexts together, violate the principle of least privilege, and make skills effectively unusable for any team whose registry differs from the one used to deploy kagent. That is the common case at scale.
> > The new imagePullSecrets field under spec.skills directly mirrors the standard Kubernetes pattern where a pod can reference multiple imagePullSecrets for different registries. It introduces no new concept, just applies the same model at the skill level.
> > Happy to discuss further if needed!
> 
> Ok I buy that logic. What do you think about renaming the field to `PullSecrets` instead `ImagePullSecrets` since these aren't really images

Good point, I can see both sides here.

For keeping imagePullSecrets: Skills are stored and pulled as OCI artifacts using the exact same kubernetes.io/dockerconfigjson secrets as normal container images. The name imagePullSecrets is the standard Kubernetes convention, so it feels familiar right away. Anyone who’s used Kubernetes already knows what it means and how to set it up.

For renaming to pullSecrets: Skills aren’t executed as containers, they’re more like content or configuration that gets pulled. The original imagePullSecrets name is specifically tied to pulling runnable container images at the pod level, so using it in this context could feel a bit overloaded. pullSecrets is more neutral and probably more accurate here.

Both options are valid. 
I’m happy to go with pullSecrets if you prefer it. 

Just say the word and I’ll rename the field right away. :thumbsup:

EItanya:
> > > Hi @EItanya, great question. We actually explored reusing the existing deployment imagePullSecrets first before adding a new field.
> > > On the technical side, imagePullSecrets defined on a pod spec are consumed exclusively by the kubelet to pull container images. They are never mounted into the pod or made accessible to running containers. This means krane, executing inside the skills-init init container, has no way to read those credentials. We hit this wall directly during testing.
> > > Even if it were technically possible, there is a design reason why it would not be the right approach in enterprise environments. The registry used to deploy the kagent system and the registry hosting skill images are typically owned by completely different teams with different security boundaries. The platform/ops team manages the kagent deployment and its registry credentials, while the line-of-business team produces and owns the skill images, hosted in their own private registry (Artifactory, ACR, ECR, etc.). Reusing the deployment imagePullSecrets would couple these two security contexts together, violate the principle of least privilege, and make skills effectively unusable for any team whose registry differs from the one used to deploy kagent. That is the common case at scale.
> > > The new imagePullSecrets field under spec.skills directly mirrors the standard Kubernetes pattern where a pod can reference multiple imagePullSecrets for different registries. It introduces no new concept, just applies the same model at the skill level.
> > > Happy to discuss further if needed!
> > 
> > 
> > Ok I buy that logic. What do you think about renaming the field to `PullSecrets` instead `ImagePullSecrets` since these aren't really images
> 
> Good point, I can see both sides here.
> 
> For keeping imagePullSecrets: Skills are stored and pulled as OCI artifacts using the exact same kubernetes.io/dockerconfigjson secrets as normal container images. The name imagePullSecrets is the standard Kubernetes convention, so it feels familiar right away. Anyone who’s used Kubernetes already knows what it means and how to set it up.
> 
> For renaming to pullSecrets: Skills aren’t executed as containers, they’re more like content or configuration that gets pulled. The original imagePullSecrets name is specifically tied to pulling runnable container images at the pod level, so using it in this context could feel a bit overloaded. pullSecrets is more neutral and probably more accurate here.
> 
> Both options are valid. I’m happy to go with pullSecrets if you prefer it.
> 
> Just say the word and I’ll rename the field right away. 👍

Ok I buy that logic, let's stick with it for now. Just resolve merge conflicts and we'll get this merged

ppeau:
> > > > Hi @EItanya, great question. We actually explored reusing the existing deployment imagePullSecrets first before adding a new field.
> > > > On the technical side, imagePullSecrets defined on a pod spec are consumed exclusively by the kubelet to pull container images. They are never mounted into the pod or made accessible to running containers. This means krane, executing inside the skills-init init container, has no way to read those credentials. We hit this wall directly during testing.
> > > > Even if it were technically possible, there is a design reason why it would not be the right approach in enterprise environments. The registry used to deploy the kagent system and the registry hosting skill images are typically owned by completely different teams with different security boundaries. The platform/ops team manages the kagent deployment and its registry credentials, while the line-of-business team produces and owns the skill images, hosted in their own private registry (Artifactory, ACR, ECR, etc.). Reusing the deployment imagePullSecrets would couple these two security contexts together, violate the principle of least privilege, and make skills effectively unusable for any team whose registry differs from the one used to deploy kagent. That is the common case at scale.
> > > > The new imagePullSecrets field under spec.skills directly mirrors the standard Kubernetes pattern where a pod can reference multiple imagePullSecrets for different registries. It introduces no new concept, just applies the same model at the skill level.
> > > > Happy to discuss further if needed!
> > > 
> > > 
> > > Ok I buy that logic. What do you think about renaming the field to `PullSecrets` instead `ImagePullSecrets` since these aren't really images
> > 
> > 
> > Good point, I can see both sides here.
> > For keeping imagePullSecrets: Skills are stored and pulled as OCI artifacts using the exact same kubernetes.io/dockerconfigjson secrets as normal container images. The name imagePullSecrets is the standard Kubernetes convention, so it feels familiar right away. Anyone who’s used Kubernetes already knows what it means and how to set it up.
> > For renaming to pullSecrets: Skills aren’t executed as containers, they’re more like content or configuration that gets pulled. The original imagePullSecrets name is specifically tied to pulling runnable container images at the pod level, so using it in this context could feel a bit overloaded. pullSecrets is more neutral and probably more accurate here.
> > Both options are valid. I’m happy to go with pullSecrets if you prefer it.
> > Just say the word and I’ll rename the field right away. 👍
> 
> Ok I buy that logic, let's stick with it for now. Just resolve merge conflicts and we'll get this merged

Done! ✅
I’ve resolved the merge conflicts.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1377,7 +1431,36 @@ func buildSkillsInitContainer(
...
+`)
+	for _, secret := range imagePullSecrets {
+		sb.WriteString(`if [ -f /docker-secrets/` + secret.Name + `/.dockerconfigjson ]; then
+  merged="$(printf '%s\n%s\n' "$merged" "$(cat /docker-secrets/` + secret.Name + `/.dockerconfigjson)" | jq -s '.[0].auths * .[1].auths | {"auths": .}')"
+fi
+`)
+	}
+	sb.WriteString(`printf '%s' "$merged" > /docker-config-out/config.json
+`)
+	return sb.String()
```
EItanya: Sorry for the continue reviews, is there anyway you could put this into a tmpl file similar to [this existing one](https://github.com/kagent-dev/kagent/blob/cfcd66dfb71bd48307d2855d6722ac2fda53b8b4/go/core/internal/controller/translator/agent/skills-init.sh.tmpl#L4). In the future I want to move away from these scripts altogether, but I think they're a bit simpler to understand for now.

ppeau: Perfect, I'll check that out right away. I'll get back to you as soon as it's ready.

ppeau: Hi @Eltanya, done! The docker-auth-init script has been moved to a dedicated docker-auth-init.sh.tmpl file, following the same pattern as skills-init.sh.tmpl (//go:embed + template.Must + typed data struct). All existing tests pass including the imagePullSecrets ones.

I'll let you resolve the conversation if it looks good to you 👍

EItanya:
Hey there, I'm sorry for going back and forth about this PR, but I have some more questions. It's not clear to me why we need a new container for this, why can't we run this logic inside of the existing skills-init container?

ppeau:
> Hey there, I'm sorry for going back and forth about this PR, but I have some more questions. It's not clear to me why we need a new container for this, why can't we run this logic inside of the existing skills-init container?

Hi @EItanya, completely fair question. Technically yes, we could add the jq merge at the start of the skills-init script.

We went with a separate docker-auth-init container because the two-container approach was the design that came out of the discussion in #1222, and it felt like a clean separation between auth setup and skill pulling. skills-init.sh.tmpl is already handling git auth, SSH keys, and krane pulls, so adding credential merging on top would blur its responsibility. It also gives better failure isolation: if credentials fail to merge, Kubernetes reports the failed init container by name immediately, without having to parse skills-init logs.

Worth noting too that docker-auth-init only runs when imagePullSecrets is set, so existing deployments without private registries are completely unaffected.

That said, if you strongly prefer consolidating into skills-init, we are happy to do that. Just let us know!

EItanya:
> > Hey there, I'm sorry for going back and forth about this PR, but I have some more questions. It's not clear to me why we need a new container for this, why can't we run this logic inside of the existing skills-init container?
> 
> Hi @EItanya, completely fair question. Technically yes, we could add the jq merge at the start of the skills-init script.
> 
> We went with a separate docker-auth-init container because the two-container approach was the design that came out of the discussion in #1222, and it felt like a clean separation between auth setup and skill pulling. skills-init.sh.tmpl is already handling git auth, SSH keys, and krane pulls, so adding credential merging on top would blur its responsibility. It also gives better failure isolation: if credentials fail to merge, Kubernetes reports the failed init container by name immediately, without having to parse skills-init logs.
> 
> Worth noting too that docker-auth-init only runs when imagePullSecrets is set, so existing deployments without private registries are completely unaffected.
> 
> That said, if you strongly prefer consolidating into skills-init, we are happy to do that. Just let us know!

Although I agree with these, I also think that adding a new container comes with its own difficulties that I'd rather avoid. For example adding new `SecurityContext` and resources requirements options which clog up the `Agent` object. As I've mentioned before we're getting to the point where we really need to turn this `skills-init ` container into a golang program so it's clearer what it's actually doing, and I think it should be responsible for all pieces related to that.

ppeau:
> > > Hey there, I'm sorry for going back and forth about this PR, but I have some more questions. It's not clear to me why we need a new container for this, why can't we run this logic inside of the existing skills-init container?
> > 
> > 
> > Hi @EItanya, completely fair question. Technically yes, we could add the jq merge at the start of the skills-init script.
> > We went with a separate docker-auth-init container because the two-container approach was the design that came out of the discussion in #1222, and it felt like a clean separation between auth setup and skill pulling. skills-init.sh.tmpl is already handling git auth, SSH keys, and krane pulls, so adding credential merging on top would blur its responsibility. It also gives better failure isolation: if credentials fail to merge, Kubernetes reports the failed init container by name immediately, without having to parse skills-init logs.
> > Worth noting too that docker-auth-init only runs when imagePullSecrets is set, so existing deployments without private registries are completely unaffected.
> > That said, if you strongly prefer consolidating into skills-init, we are happy to do that. Just let us know!
> 
> Although I agree with these, I also think that adding a new container comes with its own difficulties that I'd rather avoid. For example adding new `SecurityContext` and resources requirements options which clog up the `Agent` object. As I've mentioned before we're getting to the point where we really need to turn this `skills-init ` container into a golang program so it's clearer what it's actually doing, and I think it should be responsible for all pieces related to that.

Totally makes sense, thanks for taking the time to explain your reasoning! 
I'll consolidate the credential merge logic directly into skills-init and drop the separate docker-auth-init container. 

I'll get back to you once it's ready! 👍

ppeau:
Hey @EItanya,

Pushed a new commit on PR #1725. Removed the separate docker-auth-init init container entirely - the credential merge logic is now embedded directly in the skills-init script. Each pull secret gets mounted under /docker-secrets/<name>, merged with jq, and DOCKER_CONFIG is exported before krane runs. One init container instead of two.

Tested locally on Kind with a real htpasswd-authenticated registry + a dummy secret in CI (TestE2ESkillImagePullSecrets) - both work fine.

Let me know if this looks good or if you'd like any changes!

EItanya (approved):

EItanya (approved):

PR #1730, opened 2026-04-23
fix(bedrock): sanitize tool names to satisfy Bedrock Converse API constraint

### LINKED ISSUE #1473: fix(bedrock): sanitize tool names for Bedrock API constraints

mesutoezdil:
Closes #1473.

MCP server tool names often contain dots, colons, or spaces (e.g. `fetch.get_url` or `filesystem:read_file`). Bedrock's Converse API rejects any tool name outside `[a-zA-Z0-9_-]+`, so agents wired up to MCP tools hit a hard validation error before the first call goes through.

Added `_sanitize_tool_name()` that replaces invalid characters with underscores and caches the mapping for the duration of a request, same pattern as the existing `_sanitize_tool_id()`. Wired it into `_convert_tools_to_converse()` so every `toolSpec.name` sent to Bedrock is clean. The same map is passed to `_convert_content_to_converse_messages()` so conversation history uses the sanitized names Bedrock expects. A reverse map is built from it so that when Bedrock echoes back a sanitized name in a `toolUse` response, we restore the original name before returning it in the `LlmResponse`, that way the ADK framework can still dispatch to the correctly registered tool.

Tests cover the sanitizer itself (dots, colons, spaces, hyphens, caching, empty input) and the end-to-end flow verifying that the tool spec reaches Bedrock with the sanitized name and the response comes back with the original.


supreme-gg-gg:
Hi @mesutoezdil thanks for the PR, can you also add it to Go runtime: https://github.com/kagent-dev/kagent/blob/main/go/adk/pkg/models/bedrock.go

Have you been able to test the fix with a MCP server that actually has invalid tool name for Bedrock? I'm pretty sure this issue is valid and this is the right fix, but just double checking since I haven't got the chance to reproduce it

mesutoezdil:
Go runtime is now updated. sanitizeBedrockToolName follows the same pattern as sanitizeBedrockToolID already in the file. The name map is built in convertGenaiToolsToBedrock, threaded through GenerateContent, applied to conversation history in convertGenaiContentsToBedrockMessages, and reversed in both generateNonStreaming and generateStreaming so the original tool name comes back to the ADK dispatcher.

On testing: I have not run this against a live MCP server with an invalid tool name. The fix mirrors the Python implementation exactly and is covered by unit tests (sanitizer, end-to-end tool spec conversion, name map contents). If you have a setup to test with fetch or filesystem MCP tools that expose names like fetch.get_url, that would give us the live confirmation.

EItanya:
Conflicts need resolving and then can approve

mesutoezdil:
@EItanya and @supreme-gg-gg  Conflicts resolved in 7c4e1b9c. The two PRs touched the same function signatures in bedrock.go: this branch added reverseNameMap for tool name sanitization, while #1732 added additionalFields for top_k and extended thinking. Both parameters are now in place. additionalFields is passed alongside reverseNameMap in generateStreaming and generateNonStreaming, and AdditionalModelRequestFields is wired into both Converse API calls.

supreme-gg-gg:
Seems like there's still a conflict in `bedrock.go`

mesutoezdil:
@supreme-gg-gg Rebased onto current main; no conflict markers remain. Both `additionalFields` and `reverseNameMap` are in the signatures of `generateStreaming` and `generateNonStreaming`, and both are passed through to the Bedrock API calls.

mesutoezdil:
The `build (app)` failure is a transient Alpine registry error (`ncurses-terminfo-base: temporary error (try again later)`): not related to this PR. Could someone with repo access re-run the failed jobs?

supreme-gg-gg (approved):

PR #1732, opened 2026-04-23
feat(bedrock): add top_k and extended thinking support for Claude models

### LINKED ISSUE #1721: [FEATURE] Bedrock modelconfig allow to configure Claude settings
### 📋 Prerequisites

### 📝 Feature Summary

Allow for Bedrock models to configure specific settings

When running kagent with Claude 4.6 on Bedrock token consumption is massive compared to using claude code. It would be great if we can configure settings like effort and such.

### 💡 Proposed Solution

It would be great if following could be configured.

**~/.claude/settings.json**
```json
{
  "awsAuthRefresh": "aws sso login --profile eks-testing-admin",
  "env": {
    "AWS_PROFILE": "bedrock",
    "AWS_REGION": "us-east-1",
    "CLAUDE_CODE_USE_BEDROCK": "true",
    "ANTHROPIC_DEFAULT_OPUS_MODEL": "us.anthropic.claude-opus-4-6-v1",
    "ANTHROPIC_DEFAULT_SONNET_MODEL": "us.anthropic.claude-sonnet-4-6",
    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "us.anthropic.claude-haiku-4-5-20251001-v1:0"
  },
  "model": "sonnet",
  "enabledPlugins": {
    "gopls-lsp@claude-plugins-official": true,
    "example-skills@anthropic-agent-skills": true
  },
  "effortLevel": "medium",
  "thinking": {
    "type": "adaptive",
    "display": "summarized"
  }
}
```

So users can define the model, effort and thinking parameters.

### 🔄 Alternatives Considered

N.A.

### 🎯 Affected Service(s)

Not Sure

### 📚 Additional Context

Normally my sonnet 4.6 cost is averaging at 15-20$ a day when doing all kind of development tasks.

With kagent this cost 🚀 went to 80-195$ a day for similar prompts like:

```
Figure out the root cause for crashing pods in namespace-x
```

In kagent I prompted way less then I do in my local claude-code.

mesutoezdil:
Closes #1721

Bedrock's Converse API keeps Claude-specific parameters separate from the standard InferenceConfiguration block. Fields like top_k and the extended thinking config have to go through additionalModelRequestFields. This PR wires those two fields end-to-end so operators can turn on Claude extended thinking via the ModelConfig CRD.

Changes:

- BedrockConfig CRD type gets optional topK and thinkingBudgetTokens fields. Deepcopy and CRD manifests regenerated with make manifests, Helm CRD template synced.
- Go ADK Bedrock struct and BedrockModel updated. buildAdditionalModelRequestFields() builds the payload only when at least one field is set.
- Python KAgentBedrockLlm gets matching top_k and thinking_budget_tokens fields and builds additionalModelRequestFields the same way.
- The translator passes both fields from ModelConfig through to the ADK config.

Both fields are fully optional so existing configs are not affected.

Signed-off-by: mesutoezdil <mesudozdil@gmail.com>


supreme-gg-gg (commented):
The intention of the change makes sense to me, but I want to discuss a potentially different way to structure bedrock config CRD.

In Bedrock there are two types of model parameters, one typed struct called [inferenceConfig](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_RequestSyntax) (which we use already) and an untyped JSON called [additionalModelRequestFields](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_RequestSyntax) which you are using here.

I think to stay close to Bedrock's API design intention we should make `inferenceConfig` values directly configurable on `BedrockConfig`, and other custom provider-specific headers a raw JSON. The immediate benefit I see is that this avoids adding new fields for all the (potentially conflicting) provider params and eventually having a massive config, since everyone has their different needs when using Bedrock.

I'm suggesting something similar to how `OllamaConfig` uses `map[string]string` options to allow for arbitrary model parameters, or you can perhaps use `*apiextensionsv1.JSON` here.

mesutoezdil:
cool point on the long-term maintainability concern. Typed fields would keep growing as new Claude-specific params appear. 1 thing worth noting: map[string]string would not work here bcs the thinking config is a nested object ({"type": "enabled", "budget_tokens": 16000}), not a flat string value. Would *apiextensionsv1.JSON be the right pick for additionalModelRequestFields, or do you have a different shape in mind? Happy to rework the PR along those lines once we agree on the type.

supreme-gg-gg:
`*apiextensionsv1.JSON` should work for nested objects since it just stores raw json bytes that can be unmarshalled. An alternative I think is `runtime.RawExtension`, you can pick the right one to use. You might want to add this validation rule: `// +kubebuilder:pruning:PreserveUnknownFields` to avoid these untyped fields from being pruned.

mesutoezdil:
Going with *apiextensionsv1.JSON then. Will add +kubebuilder:pruning:PreserveUnknownFields on the field and rework the PR to replace the two typed fields with a single additionalModelRequestFields raw JSON field. Will push shortly.

mesutoezdil:
Thx @supreme-gg-gg. Done. topK and thinkingBudgetTokens are gone. BedrockConfig now has a single additionalModelRequestFields *apiextensionsv1.JSON field with +kubebuilder:pruning:PreserveUnknownFields. The translator unmarshals the raw bytes into map[string]any and passes it straight to the Converse API. Python side gets the same field as a plain dict. CRD manifests and Helm template regenerated.

supreme-gg-gg (commented):
The changes look great, before we can merge this, have you been able to test this with cache point config / thinking config: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-get-started

It seems like the original issue discusses using prompt caching and it would be great if we can verify this does improve the cost usage with these thinking / caching config

mesutoezdil:
I do not have a Bedrock account to verify the cost reduction end-to-end against the live API.

What this PR is verified to do: the unit tests assert that whatever JSON is set on additionalModelRequestFields lands in the kwargs passed to boto3.converse / converse_stream verbatim. So a config with thinking enabled (e.g. {"thinking": {"type": "enabled", "budget_tokens": 16000}}) reaches Bedrock unchanged. That is the contract this PR adds.

Two things worth separating:

1. Thinking config: this PR covers it. The raw JSON pass-through is exactly how Bedrock expects extended thinking to be configured per the docs.

2. Prompt caching (cachePoint): this PR does NOT cover it. cachePoint blocks live inside the message array, not in additionalModelRequestFields, so they need a separate change in _convert_content_to_converse_messages. The original issue mentions cost concerns broadly but the proposed config (effort, thinking) is parameter-level rather than cache-point-level. Happy to follow up with a dedicated caching PR once this lands.

Would it be reasonable to merge this PR for the thinking/parameter use case and track caching separately, or do you want both in one PR? If the latter, would you have a way to test it on real Bedrock since I cannot?

supreme-gg-gg (approved):
I am able to test this with bedrock claude, seems to be working. Thanks @mesutoezdil

marcofranssen:
I have been testing this, and it doesn't seem to work.

```shell
LLM error: API_ERROR operation error Bedrock Runtime: Converse, https response error StatusCode: 400, RequestID: 7dbf6c3a-d526-43ef-a9b9-b48228b1821f, ValidationException: The value at messages.1.content.2.toolUse.input is empty.
```

https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-extended-thinking.html

According to Claude is thinks following:

```
Now I have the full picture. The Bedrock SDK has ContentBlockMemberReasoningContent with ReasoningTextBlock{Text, Signature}
  and RedactedContent []byte. The genai Part struct has Thought bool and ThoughtSignature []byte.

  The fix requires two things in bedrock.go:

  1. generateNonStreaming: When extracting response content, handle ContentBlockMemberReasoningContent and emit
  genai.Part{Thought: true, Text: ..., ThoughtSignature: ...}
  2. generateStreaming: Track reasoning content deltas (text + signature) and emit thinking parts
  3. convertGenaiContentsToBedrockMessages: When building messages from conversation history, handle part.Thought == true parts
  by emitting ContentBlockMemberReasoningContent
```

PR #1733, opened 2026-04-23
fix(go-adk): forward allowedHeaders from incoming request to MCP tool calls

### LINKED ISSUE #1679: [BUG] [Go Runtime] AllowedHeaders not implemented — JWT/headers not forwarded to MCP tool calls
### 📋 Prerequisites

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

The allowedHeaders field on `McpServer` (CRD `v1alpha2`) is documented and correctly translated into `HttpMcpServerConfig`/`SseMcpServerConfig`, but it is never used in the Go runtime. As a
result, headers from the incoming `A2A`/`MCP` request (e.g. Authorization: Bearer <jwt>) are never forwarded to downstream `MCP` server tool calls.

Using the `python` runtime, this feature works as expected.

### 🔄 Steps To Reproduce

1. Deploy a `RemoteMCPServer` that expect a `JWT` header passed from the agent
2. Deploy a declarative `kagent` with `go` as runtime
```yaml
# ...
    declarative:
      runtime: go
# ...
      tools:
      - type: McpServer
        mcpServer:
          allowedHeaders:
          - Authorization
          name: mcp-federation
          kind: RemoteMCPServer
          toolNames:
          - get_user_profile
```
3. Call agent via `MCP` or `A2A` and ask it to use the tool

### 🤔 Expected Behavior

Agent passes the Headers (`JWT` token in this example) to the `MCP` server.

### 📱 Actual Behavior

Headers are not forwarded.

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

mesutoezdil:
Closes #1679

The allowedHeaders field on McpServer was being populated by the translator and stored in HttpMcpServerConfig/SseMcpServerConfig, but the Go runtime never used it. Headers from the incoming A2A request were silently dropped even when the operator explicitly listed them in allowedHeaders. The Python runtime handled this correctly already.

What changed:

- allowedRequestHeaders helper added to registry.go. It reads the a2asrv.CallContext already present in the Go context and returns only the header values whose names appear in the configured allowedHeaders list (case-insensitive). No intermediate context key, no extra copy.
- headerRoundTripper gets an allowedHeaders field. On each outbound MCP HTTP request the transport calls allowedRequestHeaders and sets the matching headers. Static headers from the server spec are applied last so they always take precedence, mirroring the Python runtime behaviour.
- CreateToolsets now passes AllowedHeaders from each MCP server config into the transport params.
- executor.go no longer needs to touch headers at all. The A2A server already stores the CallContext in the Go context before Execute is called, so the round tripper picks it up automatically.

The change is fully backwards compatible. If allowedHeaders is empty, no extra headers are forwarded.

Tests use a2asrv.NewRequestMeta and a2asrv.WithCallContext to build the test context, matching what the real A2A server does. Coverage includes: normal forwarding, static-header-wins override, no A2A context, non-allowed header filter, and empty allowed list.

Signed-off-by: mesutoezdil <mesudozdil@gmail.com>


go/adk/pkg/mcp/registry.go
```diff
@@ -21,12 +22,32 @@ const (
...
 )
 
+// requestHeadersKey is the context key used to store incoming request headers
+// so MCP transports can forward them to downstream MCP servers.
+type requestHeadersKey struct{}
+
+// WithRequestHeaders stores the provided headers in ctx. The headers are
+// expected to come from the incoming A2A or MCP request and will be forwarded
+// to downstream MCP server calls according to each server's AllowedHeaders list.
+func WithRequestHeaders(ctx context.Context, headers map[string]string) context.Context {
```
jmhbh: nit: slightly more intuitive to just do `return context.WithValue(ctx, requestHeadersKey{}, headers)` in the `withRequestHeaders` function instead of creating a new function in the `mcp` package

go/adk/pkg/agent/agent.go
```diff
@@ -285,9 +285,11 @@ func CreateLLM(ctx context.Context, m adk.Model, log logr.Logger) (adkmodel.LLM,
...
 		// Use Bedrock Converse API for ALL models (including Anthropic)
 		cfg := &models.BedrockConfig{
-			TransportConfig: transportConfigFromBase(m.BaseModel, nil),
-			Model:           modelName,
-			Region:          region,
+			TransportConfig:      transportConfigFromBase(m.BaseModel, nil),
+			Model:                modelName,
+			Region:               region,
+			TopK:                 m.TopK,
+			ThinkingBudgetTokens: m.ThinkingBudgetTokens,
```
jmhbh: just curious about this addition since it doesn't seem relevant to this specific issue

jmhbh (commented):
header forwarding changes lgtm just curious about the bedrock changes in this PR since they aren't explicitly tied to the issue at hand, do we have a separate issue for that?

mesutoezdil:
Good catch on both counts. @jmhbh 

The Bedrock changes crept in because this branch was cut from the wrong base. Rebased onto main so only the allowedHeaders fix is here now, single clean commit.

On the nit: removed the private `withRequestHeaders` wrapper and inlined the header extraction directly in `Execute` alongside `withBearerToken`. The exported `mcp.WithRequestHeaders` is kept since `registry_test.go` needs it to set up context state for the round-tripper tests, but the extra layer in executor.go is gone.

go/adk/pkg/mcp/registry.go
EItanya: I'm actually not sure I agree, in general these functions need to be very easily discoverable because they are a bit magic.

go/adk/pkg/a2a/executor.go
```diff
@@ -117,6 +118,19 @@ func (e *KAgentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.RequestCont
...
+			for key, vals := range meta.List() {
+				if len(vals) > 0 && vals[0] != "" {
+					headers[key] = vals[0]
+				}
+			}
+			if len(headers) > 0 {
+				ctx = mcp.WithRequestHeaders(ctx, headers)
+			}
+		}
+	}
```
EItanya: Why are we moving these values from the original context value into a custom one, what value does that bring us? I'd rather just create a helper function in `registry.go` that pulls these values, what do you think?

mesutoezdil: Done. Replaced the WithRequestHeaders/requestHeadersFromContext approach with a single allowedRequestHeaders helper in registry.go that reads straight from the a2asrv.CallContext already present in the Go context. No intermediate copy, no exported helper needed. Tests use a2asrv.NewRequestMeta + a2asrv.WithCallContext to build the test context, matching what the real server does.

go/adk/pkg/mcp/registry.go
mesutoezdil: Agreed. WithRequestHeaders is gone in the latest commit. The helper now lives entirely inside registry.go and reads from the a2asrv.CallContext directly, so there is nothing magic to discover from outside the package.

go/adk/pkg/agent/agent.go
mesutoezdil: That code is gone. The branch was cut from the wrong base and picked up Bedrock commits by accident. It has been rebased onto main, so this PR now contains only the allowedHeaders change.

mesutoezdil:
I've now updated the top section as well. @EItanya @jmhbh

jsonmp-k8 (changes requested):
### Bug: `allowedRequestHeaders` reinvents `RequestMeta.Get()` poorly

The entire inner loop is unnecessary and introduces bugs:

```go
for _, name := range allowed {
    lower := strings.ToLower(name)
    for key, vals := range meta.List() {
        if strings.ToLower(key) == lower && len(vals) > 0 && vals[0] != "" {
            result[key] = vals[0]
            break
        }
    }
}
```

`RequestMeta.Get(key)` already does case-insensitive O(1) lookup (it lowercases keys at construction in `NewRequestMeta`, and `Get` lowercases the lookup key). This loop is O(n*m) for no reason. Worse, since `List()` yields already-lowercased keys, `result[key]` stores the lowercased form — so `req.Header.Set("authorization", v)` is what actually happens. Go's `Header.Set` canonicalizes it so it works by accident, but it's still wrong in intent.

Replace the entire function body with:

```go
func allowedRequestHeaders(ctx context.Context, allowed []string) map[string]string {
    if len(allowed) == 0 {
        return nil
    }
    callCtx, ok := a2asrv.CallContextFrom(ctx)
    if !ok {
        return nil
    }
    meta := callCtx.RequestMeta()
    if meta == nil {
        return nil
    }
    result := make(map[string]string)
    for _, name := range allowed {
        if vals, ok := meta.Get(name); ok && len(vals) > 0 && vals[0] != "" {
            result[name] = vals[0]
        }
    }
    if len(result) == 0 {
        return nil
    }
    return result
}
```

This preserves the original casing from the `allowedHeaders` config, does O(n) instead of O(n*m), and avoids returning an empty map (the callers range over it, so nil is cleaner).

### Missing: no test for case mismatch between config and request

The PR claims case-insensitive matching but none of the tests exercise it. Add a test where `allowedHeaders` says `"authorization"` but the A2A request sends `"Authorization"` (or vice versa). Without this, the case-insensitivity claim is untested.

### Missing: no test for multi-value headers

`RequestMeta` stores `[]string` per key. The PR takes `vals[0]` unconditionally, silently dropping all subsequent values. This is probably fine for headers like `Authorization`, but the PR should document this choice (first value wins) and test that a multi-value header only forwards the first value. If a header like `X-Forwarded-For` gets multiple values, dropping all but the first is data loss.

### Concern: context propagation is assumed, not verified

The entire approach relies on the A2A `CallContext` surviving from the executor through the Google ADK runner, through the agent invocation, through the MCP tool call, all the way down to `headerRoundTripper.RoundTrip(req)`. The PR doesn't verify this end-to-end. The unit tests construct the context manually with `a2asrv.WithCallContext` — they don't prove the real request path propagates it.

This is the kind of thing that works in tests and breaks in production when some middleware swaps the context. The PR description says "The A2A server already stores the CallContext in the Go context before Execute is called, so the round tripper picks it up automatically" — but "automatically" is doing a lot of work there. An integration test that actually sends an A2A request and verifies the MCP server receives the forwarded header would catch this.

### Nit: empty map allocation when no headers match

If all allowed header names are present in the config but none appear in the request metadata, the function allocates a `map[string]string` and returns it empty. Callers range over it harmlessly, but `nil` is more idiomatic and avoids the allocation.

### Tests: no `t.Parallel()`

All tests spin up real HTTP servers. They should call `t.Parallel()` to avoid unnecessarily serializing the test suite.

mesutoezdil:
Thanks for the thorough review @jsonmp-k8. Applied the fixes in 76e7c1a:

- allowedRequestHeaders now uses RequestMeta.Get for an O(1) case-insensitive lookup instead of iterating meta.List. Result keys preserve the casing from the allowedHeaders config rather than the lowercased form from List. Returns nil instead of an empty map when nothing matches.
- strings import removed since it is no longer needed.
- New test TestAllowedRequestHeaders_CaseInsensitiveLookup exercises both directions (config lowercase vs incoming capitalized, and the reverse). Multi-value test locks in the first-wins behaviour. Empty-match test verifies the nil return.
- t.Parallel added to every test function and subtest.

On the end-to-end context propagation concern: agreed that the unit tests do not prove the CallContext survives from executor through the ADK runner down to the round tripper. The current e2e suite does not have a hook that sends an A2A request and asserts headers on the MCP server side, so I did not add one here. Happy to follow up with an e2e test in a separate PR if the maintainers think the scope fits, otherwise this relies on the a2asrv runtime contract that the CallContext is stored in ctx before Execute is called.

EItanya (dismissed):

EItanya (approved):

EItanya (approved):

PR #1738, opened 2026-04-23
feat(ui): sort sessions by recent activity

officialasishkumar:
## Summary
- Sort session lists by `updated_at` descending so active long-running chats stay visible
- Touch session activity when events or tasks are stored
- Group and display sidebar sessions by last activity, with story coverage for older sessions updated today

Fixes #1556

## Testing
- `go test ./core/internal/database -run 'TestListSessionsOrdersByRecentActivity|TestStoreEventTouchesSessionActivity|TestStoreTaskTouchesSessionActivity' -count=1`\n- `npm run lint -- --quiet`

### DIFF
--- a/go/core/internal/database/client_postgres.go
+++ b/go/core/internal/database/client_postgres.go
@@ -148,28 +148,17 @@ func (c *postgresClient) DeleteSession(ctx context.Context, sessionID, userID st
 // ── Events ────────────────────────────────────────────────────────────────────
 
 func (c *postgresClient) StoreEvents(ctx context.Context, events ...*dbpkg.Event) error {
-	return c.withTx(ctx, func(q *dbgen.Queries) error {
-		for _, e := range events {
-			sessionID := strPtrIfNotEmpty(e.SessionID)
-			if err := q.InsertEvent(ctx, dbgen.InsertEventParams{
-				ID:        e.ID,
-				UserID:    e.UserID,
-				SessionID: sessionID,
-				Data:      e.Data,
-			}); err != nil {
-				return fmt.Errorf("failed to store event %s: %w", e.ID, err)
-			}
-			if sessionID != nil {
-				if err := q.TouchSessionForUser(ctx, dbgen.TouchSessionForUserParams{
-					ID:     *sessionID,
-					UserID: e.UserID,
-				}); err != nil {
-					return fmt.Errorf("failed to touch session %s: %w", *sessionID, err)
-				}
-			}
+	for _, e := range events {
+		if err := c.q.InsertEvent(ctx, dbgen.InsertEventParams{
+			ID:        e.ID,
+			UserID:    e.UserID,
+			SessionID: strPtrIfNotEmpty(e.SessionID),
+			Data:      e.Data,
+		}); err != nil {
+			return fmt.Errorf("failed to store event %s: %w", e.ID, err)
 		}
-		return nil
-	})
+	}
+	return nil
 }
 
 func (c *postgresClient) ListEventsForSession(ctx context.Context, sessionID, userID string, opts dbpkg.QueryOptions) ([]*dbpkg.Event, error) {
@@ -213,21 +202,10 @@ func (c *postgresClient) StoreTask(ctx context.Context, task *protocol.Task) err
 	if err != nil {
 		return fmt.Errorf("failed to serialize task: %w", err)
 	}
-	return c.withTx(ctx, func(q *dbgen.Queries) error {
-		sessionID := strPtrIfNotEmpty(task.ContextID)
-		if err := q.UpsertTask(ctx, dbgen.UpsertTaskParams{
-			ID:        task.ID,
-			Data:      string(data),
-			SessionID: sessionID,
-		}); err != nil {
-			return err
-		}
-		if sessionID != nil {
-			if err := q.TouchSession(ctx, *sessionID); err != nil {
-				return fmt.Errorf("failed to touch session %s: %w", *sessionID, err)
-			}
-		}
-		return nil
+	return c.q.UpsertTask(ctx, dbgen.UpsertTaskParams{
+		ID:        task.ID,
+		Data:      string(data),
+		SessionID: strPtrIfNotEmpty(task.ContextID),
 	})
 }
 
--- a/go/core/internal/database/client_postgres.go
+++ b/go/core/internal/database/client_postgres.go
@@ -148,17 +148,28 @@ func (c *postgresClient) DeleteSession(ctx context.Context, sessionID, userID st
 // ── Events ────────────────────────────────────────────────────────────────────
 
 func (c *postgresClient) StoreEvents(ctx context.Context, events ...*dbpkg.Event) error {
-	for _, e := range events {
-		if err := c.q.InsertEvent(ctx, dbgen.InsertEventParams{
-			ID:        e.ID,
-			UserID:    e.UserID,
-			SessionID: strPtrIfNotEmpty(e.SessionID),
-			Data:      e.Data,
-		}); err != nil {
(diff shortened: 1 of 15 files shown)


jsonmp-k8 (commented):
Good feature — sorting by activity is the right UX. A few issues to address:

**1. `StoreTask` uses `TouchSession` (no user_id check) while `StoreEvents` uses `TouchSessionForUser` (with user_id check)**

`TouchSession` updates *any* session matching the ID regardless of who owns it. `TouchSessionForUser` scopes to the user. Since `protocol.Task` doesn't carry a user_id, using `TouchSession` is understandable, but it means any task upsert can bump another user's session timestamp. Consider whether this matters for multi-tenant scenarios, or add a comment explaining the choice.

**2. Fake client `StoreTask` touch logic iterates all sessions**

```go
for _, session := range c.sessions {
    if session.ID == task.ContextID {
        session.UpdatedAt = now
    }
}
```

This scans every session to find one by ID. The sessions map is keyed by `sessionKey(id, userID)`, so you can't do a direct lookup without the user_id. This is fine for a fake/test client, but worth noting — a `break` after the match would be cleaner since session IDs are unique.

**3. `sortSessionsByActivity` can use `cmp.Compare` instead of manual if-chains**

The `time.Time.Compare` method returns -1/0/1 directly. You can simplify:
```go
func sortSessionsByActivity(sessions []database.Session) {
    slices.SortStableFunc(sessions, func(i, j database.Session) int {
        if c := j.UpdatedAt.Compare(i.UpdatedAt); c != 0 {
            return c
        }
        if c := j.CreatedAt.Compare(i.CreatedAt); c != 0 {
            return c
        }
        return strings.Compare(i.ID, j.ID)
    })
}
```
Note the `j` before `i` to get descending order.

**4. UI: `createdAt` prop name is now misleading**

In `SessionGroup.tsx`, the `createdAt` prop now receives `session.updated_at || session.created_at`. The prop name `createdAt` no longer reflects what it contains. Consider renaming it to `lastActivity` or `activityAt` to avoid confusion for future readers.

Overall the backend changes (SQL, transaction wrapping, tests) look solid. The test coverage is thorough.

jsonmp-k8 (commented):
.

officialasishkumar:
Addressed the review feedback in fe088386f363f2c279420ec8f32c525a1a12d122: added the task touch comment, broke the fake-client loop on match, simplified the sort comparator, and renamed the UI timestamp prop to `activityAt`.

go/core/internal/database/client_postgres.go
```diff
@@ -148,17 +148,28 @@ func (c *postgresClient) DeleteSession(ctx context.Context, sessionID, userID st
...
+				return fmt.Errorf("failed to store event %s: %w", e.ID, err)
+			}
+			if sessionID != nil {
+				if err := q.TouchSessionForUser(ctx, dbgen.TouchSessionForUserParams{
+					ID:     *sessionID,
+					UserID: e.UserID,
+				}); err != nil {
+					return fmt.Errorf("failed to touch session %s: %w", *sessionID, err)
+				}
+			}
```
EItanya: This feels ugly, maybe the `InsertEvent` query should also touch the session automatically?

go/core/internal/database/client_postgres.go
```diff
@@ -202,10 +213,23 @@ func (c *postgresClient) StoreTask(ctx context.Context, task *protocol.Task) err
 	if err != nil {
 		return fmt.Errorf("failed to serialize task: %w", err)
 	}
-	return c.q.UpsertTask(ctx, dbgen.UpsertTaskParams{
-		ID:        task.ID,
-		Data:      string(data),
-		SessionID: strPtrIfNotEmpty(task.ContextID),
+	return c.withTx(ctx, func(q *dbgen.Queries) error {
+		sessionID := strPtrIfNotEmpty(task.ContextID)
+		if err := q.UpsertTask(ctx, dbgen.UpsertTaskParams{
```
EItanya: Same comment here

go/core/internal/database/client_postgres.go
officialasishkumar: Moved the session activity update into the `InsertEvent` SQL path in 4699816.

go/core/internal/database/client_postgres.go
officialasishkumar: Handled the same way for task upserts in 4699816 by updating the session in the SQL CTE.

EItanya:
Hey there @officialasishkumar I deleted the fake db client in a recent PR, please resolve the conflicts and then I will merge

EItanya (dismissed):

officialasishkumar:
Rebased on current `main` and resolved the conflict by keeping the upstream fake-client deletion. `go test ./core/internal/database` passes from the `go/` directory.

go/core/internal/database/queries/tasks.sql
```diff
@@ -14,12 +14,20 @@ WHERE session_id = $1 AND deleted_at IS NULL
...
     data       = EXCLUDED.data,
     session_id = EXCLUDED.session_id,
-    updated_at = NOW();
+    updated_at = NOW()
+RETURNING session_id
+)
+UPDATE session
+SET updated_at = NOW()
+FROM upserted_task
+WHERE session.id = upserted_task.session_id
```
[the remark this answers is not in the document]
officialasishkumar: Fixed in b7a3418 by adding the `upserted_task.session_id IS NOT NULL` guard before touching the session.

go/core/internal/database/queries/events.sql
```diff
@@ -1,6 +1,15 @@
...
+WITH inserted_event AS (
 INSERT INTO event (id, user_id, session_id, data, created_at, updated_at)
-VALUES ($1, $2, $3, $4, NOW(), NOW());
+VALUES ($1, $2, $3, $4, NOW(), NOW())
+RETURNING user_id, session_id
+)
+UPDATE session
+SET updated_at = NOW()
+FROM inserted_event
+WHERE session.id = inserted_event.session_id
```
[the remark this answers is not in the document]
officialasishkumar: Fixed in b7a3418 by adding the `inserted_event.session_id IS NOT NULL` guard before touching the session.

ui/src/components/sidebars/GroupedChats.tsx
```diff
@@ -39,9 +39,11 @@ export default function GroupedChats({ agentName, agentNamespace, sessions, hide
       older: [],
     };
 
-    // Process each session and group by date
+    const getActivityDate = (session: Session) => new Date(session.updated_at || session.created_at);
+
```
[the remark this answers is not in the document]
officialasishkumar: Fixed in b7a3418 by computing one numeric activity timestamp per session and sorting on that cached value.

go/core/internal/database/client_test.go
```diff
@@ -233,6 +234,124 @@ func TestStoreSessionIdempotence(t *testing.T) {
...
+			ID:      s.id,
+			UserID:  userID,
+			AgentID: &agentID,
+		})
+		require.NoError(t, err)
+		_, err = db.Exec(ctx, `
+			UPDATE session
+			SET created_at = $1, updated_at = $2
+			WHERE id = $3 AND user_id = $4
+		`, s.createdAt, s.updatedAt, s.id, userID)
```
EItanya: In general tests like this shouldn't execute raw SQL. Rather we should use the new queries to prove the actual behavior

officialasishkumar: Updated the tests to avoid raw SQL. They now create sessions through the client and use event/task writes to advance activity, so the ordering assertion exercises the new queries directly.

EItanya (approved):

PR #1757, opened 2026-04-26
fix: propagate authorization header to subagents

### LINKED ISSUE #1745: [FEATURE]  _SubagentInterceptor should forward Authorization header from parent session to sub-agent A2A calls
### 📋 Prerequisites

### 📝 Feature Summary

When a supervisor agent dispatches to a sub-agent via `type: Agent`, the `_SubagentInterceptor` should forward the `Authorization` header from the parent agent's `session.state["headers"]`, so that the sub-agent's downstream MCP calls can propagate the caller's JWT end-to-end.

When using `allowedHeaders: [Authorization]` on Agent CRs, the user's JWT propagates correctly from the incoming A2A request to outbound MCP calls — but only for **direct** agent invocations. In multi-agent setups where a coordinator (supervisor) dispatches to sub-agents via `type: Agent`, the JWT is silently dropped at the A2A hop.

`_SubagentInterceptor` ([`_remote_a2a_tool.py:63-77`](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py#L63-L77)) currently only forwards two headers:
- `x-kagent-source: agent`
- `x-user-id` (from `context.state`)

The `Authorization` header stored in `context.state["headers"]` (populated by `_agent_executor.py` and consumed by `create_header_provider` in `types.py`) is **not forwarded** to the sub-agent.

This means:

| Path | JWT at MCP backend? |
|---|---|
| Browser → agent → MCP | ✅ Yes (`allowedHeaders` works) |
| Browser → coordinator → sub-agent → MCP | ❌ No (JWT lost at coordinator → sub-agent hop) |

### 💡 Proposed Solution

The fix requires two changes — the interceptor alone isn't enough because the headers are never passed into the `ClientCallContext` it receives.

### Change 1: Pass session headers into `ClientCallContext.state`

In `_handle_first_call()` ([line 242](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py#L242)) and `_handle_resume()` ([line 384](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py#L384)), the `ClientCallContext` is constructed with only `x-user-id`:

```python
# Current — headers from session.state are not included
call_context = ClientCallContext(state={_USER_ID_CONTEXT_KEY: tool_context.session.user_id})
```

The session headers (populated by `_agent_executor.py` from the incoming request) need to be forwarded:

```python
# Proposed — include session headers so the interceptor can forward them
ctx_state = {_USER_ID_CONTEXT_KEY: tool_context.session.user_id}
session_headers = tool_context.session.state.get("headers", {})
if session_headers:
    ctx_state["headers"] = session_headers
call_context = ClientCallContext(state=ctx_state)
```

This must be applied in both `_handle_first_call()` (line 242) and `_handle_resume()` (line 384).

### Change 2: Forward Authorization in `_SubagentInterceptor`

With headers now available in `context.state`, the interceptor can forward them:

```python
class _SubagentInterceptor(ClientCallInterceptor):
    async def intercept(self, method_name, request_payload, http_kwargs, agent_card, context):
        headers = dict(http_kwargs.get("headers", {}))
        headers[_SOURCE_HEADER] = _SOURCE_SUBAGENT
        if context:
            if _USER_ID_CONTEXT_KEY in context.state:
                headers["x-user-id"] = context.state[_USER_ID_CONTEXT_KEY]
            # Forward Authorization from the parent agent's incoming request
            # so sub-agents can propagate it to downstream MCP backends
            request_headers = context.state.get("headers", {})
            for key, value in request_headers.items():
                if key.lower() == "authorization":
                    headers[key] = value
                    break
        http_kwargs["headers"] = headers
        return request_payload, http_kwargs
```

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

## background of our setup: How coordinator → sub-agent dispatch works

In kagent's multi-agent setup, agents are deployed as individual pods behind **agentgateway**, which acts as both an A2A (agent-to-agent) and MCP (model-context-protocol) gateway:

```
┌─────────┐     A2A (8080)    ┌──────────────────┐    A2A (8080)     ┌─────────────┐    MCP (9090)    ┌─────────────┐
│ Browser  │ ──────────────►  │  agentgateway    │ ──────────────►  │  sub-agent   │ ──────────────►  │ MCP backend │
│ (+ JWT)  │                  │  (A2A listener)  │                  │  (kagent)    │                  │ (toolserver)│
└─────────┘                   └──────────────────┘                  └─────────────┘                  └─────────────┘
                                      │                                    │
                                      │ routes by agent name               │ MCP calls via agentgateway
                                      ▼                                    ▼
                              ┌──────────────────┐              ┌──────────────────┐
                              │  coordinator     │              │  agentgateway    │
                              │  agent (kagent)  │              │  (MCP listener)  │
                              └──────────────────┘              └──────────────────┘
```

**The coordinator → sub-agent hop** happens when:
1. A request arrives at a coordinator agent (via agentgateway A2A listener on port 8080)
2. The coordinator's LLM decides to invoke a tool of `type: Agent` (a sub-agent reference)
3. `_SubagentInterceptor` builds the outbound HTTP request to the sub-agent's A2A endpoint — **this is where the Authorization header is dropped**
4. The sub-agent receives the request without the JWT, processes it, and calls MCP backends via agentgateway's MCP listener (port 9090)
5. The MCP backend never sees the original user's JWT

When agentgateway is configured with JWT authentication on the MCP listener, it can validate/enforce identity on MCP calls — but only if the JWT survives the full chain. The coordinator → sub-agent A2A hop is the gap.

## Additional Context

- `create_header_provider()` in [`types.py:36`](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent/adk/types.py#L36) already reads from `session.state["headers"]` and forwards `Authorization` when listed in `allowedHeaders` — but this only applies to the agent → MCP path, not the coordinator → sub-agent A2A path.
- When STS is configured, the STS-provided token should still take precedence on the sub-agent side (same security model as `create_header_provider`).
- Verified on kagent v0.9.0 with agentgateway v1.1.0. Tested by sending a JWT via curl to a coordinator-agent → sub-agent → MCP chain. agentgateway logs confirm `jwt.sub` is present on direct calls and absent on coordinator-dispatched calls.

towsif-rahman:
## Summary
- Pass parent session headers into sub-agent A2A call contexts for both initial calls and HITL resume calls.
- Forward the parent `Authorization` header from `_SubagentInterceptor` while preserving existing `x-user-id` and `x-kagent-source` headers.
- Add focused unit coverage for context header propagation and interceptor filtering behavior.

## Test plan
- `UV_NATIVE_TLS=true UV_PROJECT_ENVIRONMENT=/home/ubuntu/oneai_devops/kagent/.venv ../.venv/bin/uv run --python /home/ubuntu/oneai_devops/kagent/.venv/bin/python pytest ./packages/kagent-adk/tests/unittests/test_remote_a2a_tool.py`

Closes #1745


supreme-gg-gg (commented):
hi @towsif-rahman, the changes look good here. Can you also do it for the go runtime: https://github.com/kagent-dev/kagent/blob/main/go/adk/pkg/tools/remote_a2a_tool.go

towsif-rahman:
@supreme-gg-gg Added same implementation to Go runtime

supreme-gg-gg (dismissed):

towsif-rahman:
@supreme-gg-gg Pushed a change for Go lint.. would you mind running again

go/adk/pkg/tools/remote_a2a_tool.go
```diff
@@ -20,18 +21,59 @@ import (
 // userIDContextKey is the context key for passing the session user_id to the subagent.
 type userIDContextKey struct{}
 
-// userIDForwardingInterceptor forwards the session user_id as an x-user-id header.
-type userIDForwardingInterceptor struct {
+// authorizationHeaderContextKey is the context key for passing the parent Authorization header to the subagent.
+type authorizationHeaderContextKey struct{}
```
EItanya, line 25: I really don't like that we're creating new context keys for all of our values.  Can we re-use something similar to this PR: https://github.com/kagent-dev/kagent/pull/1733

supreme-gg-gg:
> Pushed a change for Go lint.. would you mind running again

Doesn't seem like it passed, can you make sure to run it locally...

go/adk/pkg/tools/remote_a2a_tool.go
supreme-gg-gg: Agreed, we should also do the same for userID here so there's no need for context keys for the interceptor and it's self-contained (reads directly from A2A server and tool context)

supreme-gg-gg:
hi @towsif-rahman are you still working on this?

EItanya:
Superseded by #1858

PR #1763, opened 2026-04-27
fix: show error when required tool parameters are missing

### LINKED ISSUE #1722: [BUG] Values for MCP tool call parameters are unexpectedly size capped
### 📋 Prerequisites

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

I am using a custom RemoteMCPSserver CR for code-execution sandboxes. One tool is file-uploading into a sandbox, one parameter of the tool is the file content. The tool works fine with a content of about 10kB but fails with a file content of about 16kB. The problem is that this is unexpected and how it fails. It fails by setting the whole value (file content) to null, so that the tool call fails with something like "you did not provide a required parameter 'file_content'. ".

This is unexpected and hard to figure out, that this is done by kagent and not by anything downstream (custom MCP server or the connected application).

### 🔄 Steps To Reproduce

Run any remote MCP server and provide a value to a required parameter that is 20kB in size. Tool call should fail with "Required parameter not used" or similar.

### 🤔 Expected Behavior

I'd suggest to provide a clear feedback about _what_ is happening and _where_ it is happening. Example: 

"Error: MCP server 'my_server', tool 'my_tool', parameter 'my_parameter': Value size 32kB exceeds max value of 12kB for MCP tool parameter values in kagent."

This should be shown to the agent calling the tool so that it can react properly and also logged for the admin to understand and solve the problem.

### 📱 Actual Behavior

See above

### 💻 Environment

Latest version of kagent running on microk8s.

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

See above

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

areebahmeddd:
## summary

closes #1722

* if the llm skips a required tool parameter (like large file content hitting output token limits), the tool call now fails with a clear, actionable message instead of some vague downstream error
* catches missing or nil required params in `makebeforetoolcallback` before the tool even runs, for both mcp tools and native function tools

## ai model disclosure

used vscode copilot (claude sonnet 4.6) to understand the issue context and help generate a fix. changes were self-reviewed and verified via:

```bash
go build ./adk/...
go test ./adk/pkg/agent/...
make lint
```

### DIFF
--- a/go/adk/pkg/agent/agent.go
+++ b/go/adk/pkg/agent/agent.go
@@ -359,16 +359,16 @@ func makeBeforeToolCallback(logger logr.Logger) llmagent.BeforeToolCallback {
 			"args", truncateArgs(args),
 		)
 
-		// Short-circuit if required params are nil; the LLM may drop large values
+		// Short-circuit if required params are missing or nil; the LLM may drop large values
 		// due to output-token limits, producing an opaque downstream error otherwise.
-		if nilParams := findNilRequiredParams(t, args); len(nilParams) > 0 {
+		if missingParams := findNilRequiredParams(t, args); len(missingParams) > 0 {
 			msg := fmt.Sprintf(
-				"tool %q: required parameter(s) %v are nil; the LLM may have omitted large values due to output-token limits",
-				t.Name(), nilParams,
+				"tool %q: required parameter(s) %v are missing or nil; the LLM may have omitted large values due to output-token limits",
+				t.Name(), missingParams,
 			)
-			logger.Info("required parameters nil, skipping tool call",
+			logger.Info("required parameters missing or nil, skipping tool call",
 				"tool", t.Name(),
-				"nullParams", nilParams,
+				"missingParams", missingParams,
 			)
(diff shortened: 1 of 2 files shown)


areebahmeddd:
ping! @EItanya 🙂

supreme-gg-gg (commented):
This seems to me like an unnecessary change. This does exactly what ADK does, checking if the tool call is valid, only for the sake of adding a not-too-useful error message. The issue itself is not entirely valid IMO, since if the model is outputting too big of a file then there's little we can do in the callback to handle it / tell the model to retry, it's better to redesign the agent workflow or MCP server.

In general, a better place to put a guard is after the tool's execution, since if it fails after tool callback will trigger anyways. For resilience, I think a reflect and retry plugin is something that would be relevant (but not for this issue).

areebahmeddd:
thanks for the review. i didnt realize adk already validates args, so yeah this check is redundant. i'll close the pr. 🙂

PR #1773, opened 2026-04-28
fix(httpserver): replace stdlib log with structured logr in memory handler

mesutoezdil:
The memory handler was the only handler in the package still using the stdlib `log` package. Every other handler uses `go-logr/logr`. The logging middleware already injects a logr logger into the request context via `ctrllog.IntoContext`, so switching to `logr.FromContextOrDiscard(r.Context())` is a straight drop-in.

This makes memory handler logs consistent with the rest: same format, same log level control, same structured key/value fields.


EItanya (commented):
Can you please take a look at the copilot comments. Changes make sense overall though

mesutoezdil:
Fixed in 2c9ab6ae. Replaced logr.FromContextOrDiscard with ctrllog.FromContext in all four handler functions, and updated the log field keys to match the rest of the handlers (user_id -> userID, agent -> agentName) @EItanya .

EItanya (approved):

PR #1776, opened 2026-04-29
feat(ui): UI does not have ability to forward headers that are not authorization

jholt96:
Add an allowlist env for forwarding additional headers to the backend from oauth2-proxy. This allowed for setting additional headers from oauth2-proxy or other services such as X-Auth-Request-User, X-Auth-Request-Email, X-Auth-Request-Groups, X-Forwarded-User, X-Forwarded-Groups. 

We are working on an use case where we set a custom header -> goes through oauth2-proxy -> UI -> controller -> agent -> MCP server. The MCP server uses that header to validate the user before running any tool call as an additional layer of rbac


EItanya (commented):
Overall this PR makes a ton of sense to me, I do think the copilot comments are worth addressing first.

ui/src/app/a2a/[namespace]/[agentName]/route.ts
```diff
@@ -1,6 +1,14 @@
 import { NextRequest, NextResponse } from 'next/server';
 import { getBackendUrl } from '@/lib/utils';
-import { getAuthHeadersFromRequest } from '@/lib/auth';
+import { getAuthHeadersFromRequest, getAllowedForwardHeaderNames } from '@/lib/auth';
+
+function buildCorsAllowHeaders(): string {
```
Comment (peterj) [inline ui/src/app/a2a/[namespace]/[agentName]/route.ts]: this looks the same as `buildCorsAllowHeaders`

ui/src/lib/auth.ts
```diff
@@ -1,33 +1,91 @@
...
+  "keep-alive",
+  "transfer-encoding",
+  "upgrade",
+  "te",
+  "trailer",
+  "proxy-authenticate",
+  "proxy-authorization",
+  "content-length",
+  "content-encoding",
+]);
```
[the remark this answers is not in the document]
jholt96: personally i think this is not needed as at least for the oauth2-proxy cookie that should be encrypted. lmk what you all think and I can add it

jholt96:
implemented copilot suggested changes and added tests for this

ui/src/app/a2a/[namespace]/[agentName]/route.ts
Comment (jholt96) [reply ui/src/app/a2a/[namespace]/[agentName]/route.ts]: Not sure what you mean by this but can you re review? I am fairly rusty with Typescript so want to make sure this looks good to everyone

peterj:
LGTM!

peterj (approved):

PR #1782, opened 2026-04-30
test: move handler handler tests to testcontainers db

EItanya:
## Summary
- switch HTTP handler tests from the in-memory fake database client to the shared Postgres testcontainers setup
- add a shared database-backed test helper for the handlers test package
- delete the unused fake database implementation

## Testing
- go test ./core/internal/httpserver/handlers/...


go/core/internal/httpserver/handlers/database_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Skip("skipping database-backed handler test in short mode")
+	}
+
+	_, err := sharedDB.Exec(context.Background(), `
+		TRUNCATE TABLE
+			agent, session, event, task, push_notification, feedback,
+			tool, toolserver, lg_checkpoint, lg_checkpoint_write,
+			crewai_agent_memory, crewai_flow_state, memory
+		RESTART IDENTITY CASCADE
+	`)
```
iplay88keys: This will silently skip truncating tables if we add additional tables in the future and forget to add them to this list. Can we dynamically build them with something like:
```go
SELECT tablename FROM pg_tables WHERE schemaname = current_schema() AND tablename != 'schema_migrations'
```

iplay88keys (dismissed):

PR #1784, opened 2026-04-30
fix(bedrock): sanitize tool names to meet Bedrock API constraints

### LINKED ISSUE #1473: fix(bedrock): sanitize tool names for Bedrock API constraints

AnantKumar17:
## What this does

MCP servers can produce tool names containing dots, spaces, or other characters that Bedrock rejects. The Converse API requires tool names to match `[a-zA-Z0-9_-]+`, but the previous code passed names through unchanged, which caused `ValidationException` errors at runtime.

This PR adds `_sanitize_tool_name`, which replaces invalid characters with underscores and falls back to a generated name for empty inputs. A `name_map` is shared across `_convert_tools_to_converse` and `_convert_content_to_converse_messages` so the same original name always maps to the same sanitized name within a single request. A `reverse_name_map` is built from it and used to restore original names in Bedrock responses, so the rest of the agent runtime sees the names it registered.

A `WARNING` log is emitted whenever a name is sanitized to help with debugging.

Closes #1473

## Changes

- `python/packages/kagent-adk/src/kagent/adk/models/_bedrock.py`
  - Added `_BEDROCK_TOOL_NAME_VALID` and `_BEDROCK_TOOL_NAME_INVALID` regex constants
  - Added `_sanitize_tool_name` function (follows the same pattern as the existing `_sanitize_tool_id`)
  - Updated `_convert_content_to_converse_messages` to sanitize `function_call.name` in history
  - Updated `_convert_tools_to_converse` to sanitize tool declaration names
  - In `generate_content_async`, built a shared `name_map`/`name_counter` and a `reverse_name_map` to restore names in both streaming and non-streaming responses

- `python/packages/kagent-adk/tests/unittests/models/test_bedrock.py`
  - Added `TestSanitizeToolName`: unit tests for the new function
  - Added `TestConvertToolsToConverse`: verifies dotted names are sanitized in the tool spec
  - Added `TestBedrockToolNameRoundTrip`: integration tests confirming original names are restored in both non-streaming and streaming responses

## Testing

All 264 existing unit tests continue to pass. 12 new tests cover the added behaviour.

```
uv run pytest tests/unittests/ --ignore=tests/unittests/models/test_tls_e2e.py
264 passed, 27 warnings
```

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/models/_bedrock.py
+++ b/python/packages/kagent-adk/src/kagent/adk/models/_bedrock.py
@@ -29,6 +29,9 @@ logger = logging.getLogger(__name__)
 _BEDROCK_TOOL_ID_VALID = re.compile(r"^[a-zA-Z0-9_.:-]+$")
 _BEDROCK_TOOL_ID_INVALID = re.compile(r"[^a-zA-Z0-9_.:-]")
 
+_BEDROCK_TOOL_NAME_VALID = re.compile(r"^[a-zA-Z0-9_-]+$")
+_BEDROCK_TOOL_NAME_INVALID = re.compile(r"[^a-zA-Z0-9_-]")
+
 
 def _sanitize_tool_id(tool_id: str, id_map: dict[str, str], counter: list[int]) -> str:
     """Return a valid Bedrock toolUseId.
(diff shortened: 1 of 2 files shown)


supreme-gg-gg (commented):
Duplicate of #1730

AnantKumar17:
Closing this in favour of #1730, which I noticed is already in review and covers the same ground.

PR #1789, opened 2026-05-01
fix(app): add bearer token authentication support for Bedrock

kkkksu:
## summary

The Go controller already injects `AWS_BEARER_TOKEN_BEDROCK` as an env var into agent pods when the secret contains that key. However, the Python Bedrock runtime (`_bedrock.py`) only used boto3's standard IAM credential chain, ignoring the bearer token entirely — causing "Invalid API Key format" errors.

The AWS docs imply that boto3 should auto-detect AWS_BEARER_TOKEN_BEDROCK, but in practice it doesn’t. I got
"Invalid API Key format: Must start with pre-defined prefix"

That error happened because boto3 ignored the env var and tried its standard SigV4 credential chain (which failed since no IAM credentials were present).

This PR adds bearer token support to `_get_bedrock_client()`: when `AWS_BEARER_TOKEN_BEDROCK` is set, the client uses `UNSIGNED` signature and a `before-sign` event handler to inject the `Authorization: Bearer <token>` header. When the env var is not set, the existing IAM credential chain behavior is unchanged.

## changes

- Added `_inject_bearer_token()` event handler function
- Modified `_get_bedrock_client()` to detect bearer token and use UNSIGNED + event handler
- Added unit tests for bearer token path, event registration, header injection, and IAM fallback
- Updated module docstring to document both auth methods

## ai model disclosure

Used Claude (via Copilot CLI) to assist with implementation and test writing. Self-reviewed the approach against boto3/botocore docs and verified end-to-end on a GKE cluster with a real Bedrock bearer token. Tests verified via:

```bash
uv run pytest packages/kagent-adk/tests/unittests/models/test_bedrock.py -v  # 12 passed
```

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/models/_bedrock.py
+++ b/python/packages/kagent-adk/src/kagent/adk/models/_bedrock.py
@@ -2,7 +2,10 @@
 
 Uses boto3's Converse API which provides a consistent interface across all
 Bedrock-supported models (Anthropic, Meta, Mistral, Amazon, Cohere, etc.).
-Authenticates via the standard AWS credential chain (env vars, IAM role, etc.).
+
+Supports two authentication methods:
+- **Bearer token**: Set ``AWS_BEARER_TOKEN_BEDROCK`` env var (API key auth).
+- **IAM credentials**: Standard AWS credential chain (env vars, IAM role, etc.).
 """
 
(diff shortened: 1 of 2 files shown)


kkkksu:
No I might be wrong about this. It turns out I did not set the secret correctly for the first time.

PR #1798, opened 2026-05-04
fix: tolerate empty toolNames on RemoteMCPServer refs

### LINKED ISSUE #1797: [BUG] Agent pod CrashLoops with pydantic ValidationError when RemoteMCPServer ref omits toolNames (http_tools[*].tools=null)
### 📋 Prerequisites

### 🎯 Affected Service(s)

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

When an `Agent` (`kagent.dev/v1alpha2`) lists a tool of `type: McpServer`
referencing a `RemoteMCPServer` **without** an explicit `toolNames:` filter,
the controller materializes the agent's `config.json` Secret with
`http_tools[*].tools = null` (and `sse_tools[*].tools = null` for the SSE
protocol). The Python ADK runtime image
(`cr.kagent.dev/kagent-dev/kagent/app:0.9.1`) then crashes on startup with a
pydantic `ValidationError`, leaving the agent pod in `CrashLoopBackOff`.

Whether `toolNames` is intended to be required or optional, the current
behavior is a bug because omitting it silently produces a config that the
runtime cannot load, with no controller-side validation or admission error.

The mismatch is between two sides of the same wire format:

* **Controller (Go)** — `HttpMcpServerConfig.Tools` is `[]string` with the
  JSON tag `json:"tools"` (no `omitempty`). When `mcpServerTool.ToolNames`
  is `nil` in
  [`go/core/internal/controller/translator/agent/adk_api_translator.go`](https://github.com/kagent-dev/kagent/blob/main/go/core/internal/controller/translator/agent/adk_api_translator.go#L911-L937),
  `encoding/json` marshals the nil slice as JSON `null`.
* **Runtime (Python)** —
  [`HttpMcpServerConfig.tools`](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent/adk/types.py#L143-L154)
  is typed `list[str] = Field(default_factory=list)`, which rejects `None`
  with a `list_type` validator error.

The CRD schema for `McpServerTool.toolNames` does not appear to mark the
field as required, so users (and tooling) have no signal that omitting it
is unsafe.

### 🔄 Steps To Reproduce

1. Stand up kagent v0.9.1 against any working `RemoteMCPServer`
   (e.g. `cr.kagent.dev/kagent-dev/kagent/tools`).
2. Apply the following Agent (no `toolNames` on the tool ref):

   ```yaml
   kind: Agent
   metadata:
     name: repro
     namespace: kagent
   spec:
     type: Declarative
     description: repro
     declarative:
       modelConfig: <existing>
       systemMessage: "test"
       tools:
         - type: McpServer
           mcpServer:
             apiGroup: kagent.dev
             kind: RemoteMCPServer
             name: <existing-rms>
             # NOTE: no toolNames
   ```

3. Observe the agent pod enter `CrashLoopBackOff`.
4. `kubectl logs` shows the pydantic validation error below.
5. Inspect the rendered config:
   `kubectl get secret repro-config -o jsonpath='{.data.config\.json}' | base64 -d | jq '.http_tools[0].tools'`
   prints `null`.

Adding any non-empty `toolNames: [<one>]` to the tool ref makes the pod
start successfully — that's a viable workaround.

### 🤔 Expected Behavior

The agent should not CrashLoop on a config the controller itself produced.
Possible resolutions (deferring to maintainers on the intended semantics):

1. If `toolNames` is meant to be **required**, validate it in the CRD
   schema or in a controller-side admission check so users get a clear
   error at apply-time rather than a pydantic crash at pod-start.
2. If `toolNames` is meant to be **optional** (with omission meaning "all
   discovered tools"), the controller should emit an explicit list (or
   the runtime should tolerate `None`) rather than handing the runtime a
   `null` it can't parse.

### 📱 Actual Behavior

Agent pod CrashLoops on startup with:

```
ValidationError: 1 validation error for AgentConfig
http_tools.0.tools
  Input should be a valid list [type=list_type, input_value=None,
  input_type=NoneType]
```

The rejected blob is:

```json
{
  "http_tools": [
    {
      "params": {"url": "...", "headers": {}, "terminate_on_close": true},
      "tools": null
    }
  ]
}
```

### 💻 Environment

- Kubernetes: Talos Linux, k8s 1.34
- kagent: v0.9.1 (`cr.kagent.dev/kagent-dev/kagent/app:0.9.1` runtime,
  controller image `cr.kagent.dev/kagent-dev/kagent/controller:0.9.1`)
- Helm chart: `kagent` 0.9.1
- Topology: agents reference a single `RemoteMCPServer` named
  `agentgateway-mcp` that fronts several upstream MCP servers via
  [agentgateway](https://agentgateway.dev). The RMS reports ~50
  `discoveredTools` and is `Healthy`.

### 🔍 Additional Context

- Same field exists on `sse_tools[*].tools` and the same nil-slice → JSON
  `null` defect applies there.
- Workaround: pin explicit `toolNames` per agent.
- I'll open a PR shortly that addresses both sides defensively (controller
  expands discovered tools when `toolNames` is empty; runtime coerces
  `None` → `[]` so older controllers don't crash newer runtimes). Happy
  to rework toward whichever direction maintainers prefer if the intended
  semantics are different.

### 📋 Logs

```shell
$ kubectl logs -n kagent repro-<hash>
Traceback (most recent call last):
  ...
  File ".../kagent/adk/types.py", line ..., in to_agent
    config = AgentConfig.model_validate(...)
pydantic_core._pydantic_core.ValidationError: 1 validation error for AgentConfig
http_tools.0.tools
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.x/v/list_type
```

reilly3000:
## Summary

Fixes #1797. An Agent that references a `RemoteMCPServer` without an
explicit `toolNames` filter currently CrashLoops with a pydantic
`ValidationError` because the controller emits `http_tools[*].tools = null`
and the runtime rejects `None`. Whether `toolNames` is meant to be required
or optional is a call for maintainers — either way, the controller
shouldn't produce a config the runtime can't load.

This PR is intentionally defensive on both sides; happy to rework toward
a different direction (e.g. CRD validation making `toolNames` required, or
runtime-only fix) if maintainers prefer.

- **Controller (Go)**: when `toolNames` is empty, expand to
  `RemoteMCPServer.status.discoveredTools` so the rendered config is
  explicit and observable in the agent's `config.json` Secret. Explicit
  `toolNames` continue to take precedence. Falls back to an empty array
  rather than `null` if status has no discovered tools yet.
- **Runtime (Python)**: coerce `tools: None` → `[]` via a pydantic
  `BeforeValidator` on `HttpMcpServerConfig` / `SseMcpServerConfig`, so
  older controllers don't crash newer runtimes.

See #1797 for the full repro and root-cause analysis.

## Test plan

- [x] New Go translator regression test
      (`Test_AdkApiTranslator_RemoteMCPServer_NoToolNames`) asserts the
      rendered `AgentConfig.HttpTools[0].Tools` is non-nil, expanded from
      discovered tools, and serializes as a JSON array (not `null`).
- [x] New Python pydantic test
      (`test_http_tool_null_tools.py`) covers `tools: None`, missing
      `tools`, and explicit `tools: [...]` for both `Http` and `Sse`
      variants.
- [x] Full Go translator suite (`go test ./core/internal/controller/translator/agent/...`) — pass.
- [x] Full Python kagent-adk unit suite (`uv run pytest packages/kagent-adk/tests/unittests/`) — pass (9 unrelated TLS e2e failures from missing local fixtures).
- [x] **Runtime-level e2e**: invoked the actual `kagent-adk static --filepath ...` startup path (the same command the runtime container runs) against a `config.json` containing `http_tools[*].tools = null`.
      - Pre-fix: reproduces the exact `ValidationError ... type=list_type, input_value=None` and the process exits.
      - Post-fix: `Application startup complete. / Uvicorn running on http://127.0.0.1:...`.
- [x] **In-cluster e2e against a real `RemoteMCPServer`**: built an overlay image of the v0.9.1 runtime with only the patched `types.py`, swapped it in via `controller.agentImage`, and deployed against a live `RemoteMCPServer` that fronts ~50 federated MCP tools.
      - Pre-fix: a fresh Agent with no `toolNames` CrashLoops with the exact `ValidationError ... input_value=None` from the issue.
      - Post-fix: same Agent reaches `Application startup complete` / `Uvicorn running on http://0.0.0.0:8080` and serves `200` on `/.well-known/agent-card.json`.
- [x] **Controller-side e2e**: built an overlay image of the v0.9.1 controller with only the patched binary, swapped it in via `controller.image`, and applied a fresh Agent referencing the same `RemoteMCPServer` (23 `discoveredTools`) with no `toolNames` specified.
      - Resulting agent pod reaches `Application startup complete` / `Uvicorn running` cleanly.

## Open questions for reviewers

1. Is `toolNames` intended to be required, or optional with omission
   meaning "all discovered tools"? If the former, this PR can be reduced
   to CRD validation + a clear admission error.
2. Is the controller-side expansion to `discoveredTools` desirable for
   observability, or would you prefer the controller emit `[]` (or
   omit the field entirely) and have the runtime handle the "all tools"
   semantics on its own?

## Notes

- Same nil-slice → JSON `null` defect existed for `sse_tools[*].tools`;
  fixed via the same `resolveToolNames` helper.
- Commit is DCO-signed.


reilly3000:
Hi folks, I hate to be the ai slop PR guy, but ran into this issue and thought others may make the same assumption as well. I have a couple tweaks inbound before I raise it for review.

PR #1807, opened 2026-05-06
Confine kind setup and e2e tests to the kind context

### LINKED ISSUE #1806: [BUG] Leaking kind-only resources onto the active kubectl context
## 📋 Prerequisites

## 🎯 Affected Service(s)
Multiple services / System-wide issue (developer tooling: `Makefile`, `go/Makefile`, `scripts/kind/`, plus the e2e test code under `go/core/test/e2e/`)

## 🚦 Impact/Severity
Blocker — these are normal developer commands; nothing in their names hints at "may mutate whichever cluster your kubeconfig happens to point at right now." Anyone running them with a non-kind context active (intentionally or by accident) modifies that cluster instead.

## 🐛 Bug Description
Several developer-facing entry points run `kubectl` and `istioctl` without `--context`, so the commands target whichever context is currently active in the kubeconfig instead of the kind cluster they describe. The `kube-public/local-registry-hosting` ConfigMap that landed on a real cluster recently is the smallest possible reproducer of this; the same code paths are how a developer ends up:

- creating a `kagent` namespace on a remote cluster (`make use-kind-cluster`);
- mutating the local kubeconfig to set that remote context's default namespace to `kagent` (also `use-kind-cluster`, via `kubectl config set-context --current --namespace kagent`);
- installing the entire MetalLB stack on a remote cluster (`scripts/kind/setup-metallb.sh`, called by `make create-kind-cluster`);
- installing Istio with the demo profile on a remote cluster (`make kagent-addon-install`, the `istioctl install` line);
- port-forwarding services from a remote cluster to local ports (`make kagent-cli-port-forward` / `kagent-ui-port-forward`);
- creating test Agents, MCPServers, ModelConfigs and other CRs in a remote cluster's `kagent` namespace (`make -C go e2e`, whose Go test code calls `controller-runtime`'s `config.GetConfig()` and bare `exec.Command("kubectl", ...)` — both honor whatever the active context is).

These are all everyday developer flows: standing up a kind cluster, running e2e, port-forwarding the dashboard, exercising the addon stack. Each of them silently follows the active kubeconfig context unless the operator has manually exported `KUBECONFIG` to a kind-only file beforehand.

## 🔄 Steps To Reproduce
1. Set `kubectl config current-context` to a non-kind cluster (e.g., a GKE cluster).
2. Run `make create-kind-cluster`.
3. Observe a `local-registry-hosting` ConfigMap in the remote cluster's `kube-public` namespace, with `host: "localhost:5001"` pointing at the local kind registry. The same run also tries to install MetalLB onto the remote cluster.

The same shape reproduces with `make use-kind-cluster` (creates the `kagent` namespace remotely and rewrites the kubeconfig), `make kagent-addon-install` (installs Istio remotely), and `make -C go e2e` (creates `test-model-config-*`, `everything-mcp-server-*`, and dynamically-named test Agents in the remote `kagent` namespace).

## 🤔 Expected Behavior
All `kubectl`/`istioctl` invocations issued by these scripts and Make targets should pin to the kind cluster they're operating on (via `--context "kind-${KIND_CLUSTER_NAME}"`), and `make e2e` should isolate the test process with a kind-only `KUBECONFIG`, so a stray current-context cannot cause any of these flows to mutate another cluster.

## 📱 Actual Behavior
The unscoped invocations follow the user's current-context. Confirmed instance: a `kube-public/local-registry-hosting` ConfigMap was applied to a real GKE cluster after `make create-kind-cluster` was run while that context was active. The same run would have installed MetalLB there too if the cluster had matching node networking; on a hardened cluster it would have been a noisier failure but still a write attempt.

The same `--kube-context` / `--context kind-$(KIND_CLUSTER_NAME)` discipline is already enforced in `helm-install*`, `helm-uninstall`, `push-test-agent`, and the `kubectl apply` lines inside `kagent-addon-install`. The scripts and remaining Make targets just hadn't been brought to the same baseline.

## 💻 Environment
- OS: macOS / Linux (any host running the Make targets)
- Application version: `main` at or after 526107f0
- Kubernetes provider: kind locally, plus whatever remote cluster the developer happens to be authenticated to

## 🔍 Additional Context
This has been the case since the affected scripts and targets were introduced; not a regression. Workaround for users today: explicitly export `KUBECONFIG=$(kind get kubeconfig --name kagent)` before running these targets, or `kubectl config use-context kind-kagent` first. The repair is to pin every `kubectl` / `istioctl` call to `kind-${KIND_CLUSTER_NAME}` and have `make -C go e2e` materialize a kind-only `KUBECONFIG` before invoking `go test`.

## 🙋 Are you willing to contribute?

danielorbach:
A stray `kubectl config current-context` was enough to make `make create-kind-cluster`, `make use-kind-cluster`, `make kagent-addon-install`, and `make -C go e2e` mutate whichever cluster was active. None of those targets read like "may rewrite an arbitrary cluster," yet that is what they did. Resolves kagent-dev/kagent#1806.

`scripts/kind/setup-{kind,metallb}.sh` now pin every `kubectl` invocation to `--context "kind-${KIND_CLUSTER_NAME}"`. The Makefile targets `use-kind-cluster`, `kagent-cli-port-forward`, `kagent-ui-port-forward`, and the `istioctl install` line inside `kagent-addon-install` get the same treatment; `use-kind-cluster` no longer touches `--current` and instead sets the namespace on the kind context by name, so any other context the user has is left untouched. `go/Makefile`'s `e2e` target materializes the kind cluster's kubeconfig to a temp file and runs `go test` with `KUBECONFIG` forced to it — both `controller-runtime`'s `config.GetConfig()` and the bare `exec.Command("kubectl", ...)` calls inside the test code now have no path to a remote cluster. The target aborts early with a clear message if the kind cluster is missing.

The same `--kube-context kind-$(KIND_CLUSTER_NAME)` discipline already applied to `helm-install*`, `helm-uninstall`, `push-test-agent`, and the `kubectl apply` lines inside `kagent-addon-install`; this brings the rest of the developer tooling to the same baseline.

### Verification

`bash -n` parses both kind scripts. `make -n use-kind-cluster`, `make -n kagent-addon-install`, `make -n kagent-cli-port-forward`, `make -n kagent-ui-port-forward`, and `make -C go -n e2e` all expand with `--context kind-kagent` (or `KUBECONFIG=/tmp/kind-config-e2e`) on every state-mutating line.

### Out of scope

The Go test code itself still calls `kubectl` and `controller-runtime` without an explicit context; the safety here is provided by the wrapping `make e2e` target setting `KUBECONFIG`. Anyone running `go test ./go/core/test/e2e` directly is still responsible for setting `KUBECONFIG` themselves.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,8 +194,8 @@ create-kind-cluster:
 use-kind-cluster:
 	kind get kubeconfig --name $(KIND_CLUSTER_NAME) > /tmp/kind-config
 	KUBECONFIG=~/.kube/config:/tmp/kind-config kubectl config view --merge --flatten > ~/.kube/config.tmp && mv ~/.kube/config.tmp ~/.kube/config && chmod $(KUBECONFIG_PERM) ~/.kube/config
-	kubectl create namespace kagent || true
-	kubectl config set-context --current --namespace kagent || true
+	kubectl --context kind-$(KIND_CLUSTER_NAME) create namespace kagent || true
+	kubectl config set-context kind-$(KIND_CLUSTER_NAME) --namespace kagent || true
 
 .PHONY: delete-kind-cluster
 delete-kind-cluster:
@@ -425,17 +425,17 @@ kagent-cli-install: use-kind-cluster build-cli-local helm-version helm-install-p
 .PHONY: kagent-cli-port-forward
 kagent-cli-port-forward: use-kind-cluster
 	@echo "Port forwarding to kagent CLI..."
-	kubectl port-forward -n kagent service/kagent-controller 8083:8083
+	kubectl --context kind-$(KIND_CLUSTER_NAME) port-forward -n kagent service/kagent-controller 8083:8083
 
 .PHONY: kagent-ui-port-forward
 kagent-ui-port-forward: use-kind-cluster
 	open http://localhost:8082/
-	kubectl port-forward -n kagent service/kagent-ui 8082:8080
+	kubectl --context kind-$(KIND_CLUSTER_NAME) port-forward -n kagent service/kagent-ui 8082:8080
 
 .PHONY: kagent-addon-install
 kagent-addon-install: use-kind-cluster
 	# to test the kagent addons - installing istio, grafana, prometheus, metrics-server
-	istioctl install --set profile=demo -y
+	istioctl install --context kind-$(KIND_CLUSTER_NAME) --set profile=demo -y
 	kubectl apply --context kind-$(KIND_CLUSTER_NAME) -f contrib/addons/grafana.yaml
 	kubectl apply --context kind-$(KIND_CLUSTER_NAME) -f contrib/addons/prometheus.yaml
 	kubectl apply --context kind-$(KIND_CLUSTER_NAME) -f contrib/addons/metrics-server.yaml
--- a/go/Makefile
+++ b/go/Makefile
@@ -124,9 +124,19 @@ run: fmt vet ## Run a controller from your host.
 test: ## Run all unit tests.
 	UPDATE_GOLDEN=$(UPDATE_GOLDEN) go test -race -skip 'TestE2E.*' -v ./...
 
+# Pin e2e tests to the kind cluster so a stray current-context (or an
+# ambient KUBECONFIG) can't leak resources onto a remote cluster. Override
+# KIND_CLUSTER_NAME to target a differently-named kind cluster.
+KIND_CLUSTER_NAME ?= kagent
+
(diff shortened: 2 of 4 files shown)


EItanya:
This change definitely makes sense, you'll need to get DCO to pass before I can approve.

danielorbach:
> This change definitely makes sense, you'll need to get DCO to pass before I can approve.

Apologies for having missed that, will not happen again 💪🏿 Fixed and pulled latest `main`.

EItanya (approved):

Issue #1812, opened 2026-05-07
[FEATURE] Namespace filtering support


maazghani:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add namespace-based filtering to the Agents UI so users can quickly narrow the agent list to a specific Kubernetes namespace.

### ❓ Problem Statement / Motivation

The current Agents page shows agents from all namespaces in one list.
This is harder to use in multi-namespace or multi-team clusters where many agents exist.
Users who manage agents per namespace need a faster way to find relevant agents without scanning the full list.
Adding a namespace filter improves usability, reduces clutter, and makes agent management easier.

### 💡 Proposed Solution

Add a namespace filter to the Agents page in the kagent UI.

Add a namespace dropdown or selector near the top of the agent list.
Default to showing agents from all namespaces, with the option to narrow the list to a specific namespace.
Keep the selected namespace in the page URL/query params so the filtered view is shareable and persists on refresh.
Apply the filter consistently in both card and list views.
Show a clear empty state when no agents exist in the selected namespace.
This approach keeps the change simple, improves usability in multi-namespace environments, and aligns with how other kagent UI pages already scope resources by namespace.

### 🎯 Affected Service(s)

UI Service

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR for this feature

maazghani:
PR part 1 ready for review to implement server-side logic for namespace filtering in UI.

EItanya:
Can we close this since #1878 went in?

maazghani:
@EItanya I meant to create this PR earlier to complete the UI side of the namespace filtering. Let me know if this makes sense #1923

maazghani:
Closing this out now that both the API and UI namespace filtering PRs are merged.

PR #1813, opened 2026-05-07
implement filtering AgentList by namespace

maazghani:
addresses #1812 

The agents page showed a single unscoped list across namespaces, which made the view noisy in multi-tenant clusters and harder to navigate. This change adds namespace-aware filtering to the existing agents page without changing the underlying data flow.

- **Namespace-scoped list state**
  - Adds a namespace selector to the agents page header.
  - Sources the active filter from the `namespace` query param so the view persists across refresh/navigation.
  - Includes an explicit **All namespaces** option.

- **Consistent filtering across views**
  - Applies the namespace filter before rendering, so both `AgentGrid` and `AgentListView` stay in sync.
  - Derives available namespace options from the loaded agent set rather than introducing a separate fetch path.

- **Filtered empty state**
  - Preserves the existing global empty state when no agents exist at all.
  - Adds a namespace-specific empty state when the selected namespace has no matching agents, with a one-click reset to clear the filter.

- **Coverage**
  - Adds component tests for:
    - URL-driven namespace filtering
    - namespace-filtered empty state behavior

```tsx
const namespaceFilter = searchParams.get("namespace") || ALL_NAMESPACES;

const filteredAgents =
  namespaceFilter === ALL_NAMESPACES
    ? agents || []
    : (agents || []).filter(
        (item) => (item.agent.metadata.namespace || "") === namespaceFilter,
      );
```


maazghani:
@copilot thoughts on solutions for if the cluster has a large number of namespaces?

ui/src/components/AgentList.tsx
```diff
@@ -116,10 +181,21 @@ export default function AgentList() {
...
           </Button>
         </div>
+      ) : filteredAgents.length === 0 ? (
+        <div className="rounded-xl border border-border/60 bg-card/30 py-12 text-center shadow-sm">
+          <KagentLogo className="mx-auto mb-4 h-16 w-16" />
+          <h2 className="mb-2 text-lg font-medium tracking-tight">No agents in this namespace</h2>
+          <p className="mb-6 text-pretty text-sm text-muted-foreground">
+            No agents match the <span className="font-mono">{namespaceFilter}</span> namespace filter.
+          </p>
+          <Button type="button" variant="outline" onClick={() => onNamespaceChange(ALL_NAMESPACES)}>
```
peterj, line 191: this doesn't work -- e.g. `http://localhost:8001/agents?namespace=aknjdskjdsgjknds`, clicking the Show all namespaces button doesn't do anything.

peterj (changes requested):
clicking show all namepaces doesn't reload the page/change the url

EItanya (commented):
Thanks for the PR! I think this sort of filtering behavior is usually best done on the backend so that the frontend doesn't have to handle ALL of the data. What do you think?

maazghani:
> Thanks for the PR! I think this sort of filtering behavior is usually best done on the backend so that the frontend doesn't have to handle ALL of the data. What do you think?

Got it, thanks for the comments @peterj @EItanya. That makes a lot more sense, I'll move the namespace filtering to the agents handler in the API, and then use that for this. 

I was thinking it will also make sense to have the filtered namespace state live in the URL/query state. Will also try to address cases of large number of namespaces. Like using limits or typeahead.

maazghani:
> Thanks for the PR! I think this sort of filtering behavior is usually best done on the backend so that the frontend doesn't have to handle ALL of the data. What do you think?

@EItanya I implemented the filtering on the backend instead here in #1878, closing this. Want to get that reviewed before having the UI use it.

PR #1843, opened 2026-05-11
feat(helm): add extraGetterRules and extraWriterRules to rbac values

### LINKED ISSUE #1826: [FEATURE] Expose RBAC knobs in kagent-tools chart
### 📋 Prerequisites

### 📝 Feature Summary

_No response_

The `kagent-tools` chart ships with a `ClusterRole` that grants wildcard permissions across all resources and verbs:

```yaml
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - '*'
  verbs:
  - '*'
```

This is effectively cluster-admin level access. There is currently no way to narrow these permissions without either disabling the chart entirely or applying a post-install patch (which gets overwritten on every `helm upgrade`).

### 💡 Proposed Solution

A two-part fix:
**1. In the `kagent-tools` chart** — expose the ClusterRole rules as a configurable value
**2. In the main `kagent` chart** — pass those knobs through so users can configure them from the top-level `values.yaml`

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

mesutoezdil:
Closes #1826

The getter and writer ClusterRoles have no way to add custom rules without editing chart templates directly. 
This adds two new values that append rules to each role.
Both default to empty, so existing installs are unchanged.

### DIFF
--- a/helm/kagent/templates/rbac/getter-role.yaml
+++ b/helm/kagent/templates/rbac/getter-role.yaml
@@ -93,6 +93,9 @@
   - get
   - list
   - watch
+{{- with .Values.rbac.extraGetterRules }}
+{{ toYaml . }}
+{{- end -}}
 {{- end -}}
 
 {{- include "kagent.rbac.validate" . -}}
--- a/helm/kagent/templates/rbac/writer-role.yaml
+++ b/helm/kagent/templates/rbac/writer-role.yaml
@@ -75,6 +75,9 @@
   - update
   - patch
   - delete
+{{- with .Values.rbac.extraWriterRules }}
+{{ toYaml . }}
+{{- end -}}
 {{- end -}}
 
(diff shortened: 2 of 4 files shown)


EItanya (commented):
A couple of comments on this PR:
1. I don't think that adding this sort of thing actually makes sense in the core helm chart, rather if a user wants additional RBAC they can add the `ClusterRole/Role` and `Binding` themselves, otherwise it can potentially lead to scope creep in the chart. 
2. Based on the issue I don't think this solution actually fixes it, I commented on the issue.

mesutoezdil:
Thanks for the review and the context.
Closing this PR accordingly.

PR #1849, opened 2026-05-11
feat(helm): expose context compaction in all agent subcharts

### LINKED ISSUE #1827: k8s-agent: oversized k8s_get_resources tool_result causes Anthropic 400 (request body exceeds per-request limit)
## Summary

When a user prompt asks `k8s-agent` (v0.9.2, Anthropic provider) to look up resources in a busy namespace, the resulting `k8s_get_resources` tool_result can be hundreds of kilobytes. After 2–3 LLM iterations of accumulated history, the next `messages.create()` call to the Anthropic API returns **HTTP 400 Bad Request** because the total request body exceeds the per-request size limit. The runner then closes and retries from the same persisted session, producing the same 400 again.

Net effect: any agent that delegates to `k8s-agent` via A2A receives an error instead of the lookup result, with no graceful degradation.

## Reproduction

1. Bring up `kagent` 0.9.2 with the default `k8s-agent` Agent CR (no `spec.declarative.context.compaction` set — the subchart values.yaml at `helm/agents/k8s/values.yaml` does not expose this field).
2. Provider configured as Anthropic-compatible (Claude Haiku class model).
3. From any A2A caller, send `k8s-agent` a prompt that asks it to look up several resource kinds in a single namespace where one of them returns a large list (e.g. all `pods` in a tenant namespace running a non-trivial application — observed with a namespace running ~30 pods × deep `.spec`/`.status`).

## Observed evidence

Captured kagent-controller session events (`/api/sessions/<id>/events?...&limit=-1`):

| Event | Author      | Role  | Size      | Content                                                     |
|-------|-------------|-------|-----------|-------------------------------------------------------------|
| 1     | user        | user  | 769 B     | A2A request prompt                                          |
| 2     | k8s_agent   | model | 373 B     | text + 3 parallel `k8s_get_resources` tool_uses             |
| 3     | k8s_agent   | user  | 8.3 KB    | 3 small tool_results                                        |
| 4     | k8s_agent   | model | 304 B     | 3 more parallel `k8s_get_resources` tool_uses               |
| 5     | k8s_agent   | user  | **569 KB**| tool_results 145 / 133 / **568 843 B**                      |
| 6–7   | k8s_agent   | —     | 0         | runner-close events after the 400                           |

Cumulative content sent on the third LLM iteration: ~579 KB in one HTTP request. Anthropic returned HTTP 400; httpx logged only the status, not the body, but the size correlates exactly.

The tool_use/tool_result pairs are well-formed (each `tool_use_id` has its matching `tool_result` in the immediately following user-role event). This is **not** the orphan-tool_use case that some other LLM stacks guard against — the failure is purely a single tool_result blowing the per-request size budget.

## Why this is not user-fixable today

- `kagent-tools-server` does not size-cap tool results; `k8s_get_resources` returns the full list-response JSON verbatim.
- The k8s-agent subchart at `helm/agents/k8s/values.yaml` does not expose `compaction`, so users cannot enable history compaction without forking the chart or replacing the Agent CR.
- `KAgentSessionService.get_session()` uses `limit=-1`, so all prior events are reloaded on every retry; the 400 is deterministic from the persisted history.
- `_rearrange_events_for_latest_function_response()` in `google/adk/flows/llm_flows/contents.py` only rejects mismatches on the *last* event; nothing rejects oversize content blocks.

## Suggested fixes (any one of these would close the gap; combining is best)

1. **Size cap at the tools-server layer.** Truncate any single `k8s_get_resources` (and similar list-style) result above a configurable threshold (e.g. 64 KB), returning the truncated payload plus a `"…[truncated, N items omitted]"` marker. Analogous to size-bounded MCP tool wrappers in other ecosystems. Ship as a default-on knob, configurable via Helm values.
2. **Size cap at the kagent-adk layer.** Before posting `messages.create()`, walk the assembled `messages` list and either truncate any single content block above a configurable cap or summarize and replace it with a synthetic note. Catches misbehaving tools generally, not just `k8s_get_resources`.
3. **Expose `compaction` on the agent subcharts.** At minimum surface `context.compaction.compactionInterval` and `context.compaction.overlapSize` as values in `helm/agents/k8s/values.yaml` (and the other agent subcharts) so users can opt into history compaction without replacing the chart-managed Agent CR.

## Workaround we've adopted

Replaced the chart-managed `k8s-agent` Agent CR with a locally-managed one that adds `spec.declarative.context.compaction.{compactionInterval: 5, overlapSize: 2}`. This is defence-in-depth — compaction may summarize a 569 KB tool_result before it gets re-sent, but the underlying fix is tool-result size capping (suggestion 1 above).

## Environment

- kagent 0.9.2
- Provider: Anthropic-compatible gateway, model class Claude Haiku 4.5
- Runtime: kagent-adk Python (kagent_adk package)
- Observed in two separate sessions with different tenant namespaces; behaviour is reliably reproducible against any namespace where `k8s_get_resources` for a list kind returns ≥~500 KB.

mesutoezdil:
Fixes #1827

All agent subcharts now expose a `compaction` value so users can configure context compaction without replacing the chart-managed Agent CR.

By default `compaction` is empty and no context block is rendered.
Users can set it in their values file:

```yaml
compaction:
  compactionInterval: 5
  overlapSize: 2
```

This adds the same change to all 10 agent subcharts: argo-rollouts, cilium-debug, cilium-manager, cilium-policy, helm, istio, k8s, kgateway, observability, and promql.

### DIFF
--- a/helm/agents/argo-rollouts/templates/agent.yaml
+++ b/helm/agents/argo-rollouts/templates/agent.yaml
@@ -182,6 +182,10 @@ spec:
             - "Should I convert my Deployment directly or use workloadRef for my 'user-service'?"
             - "How can I check if the Argo Rollouts controller is installed and running in my cluster before I start migrating?"
             - "My Deployment YAML is complex. Can you help me get its YAML so I can plan the conversion to a Rollout?"
+    {{- if .Values.compaction }}
+    context:
+      compaction: {{- toYaml .Values.compaction | nindent 8 }}
+    {{- end }}
     deployment:
       {{- with coalesce (empty .Values.imagePullSecrets | ternary nil .Values.imagePullSecrets) .Values.global.imagePullSecrets }}
       imagePullSecrets:
--- a/helm/agents/argo-rollouts/values.yaml
+++ b/helm/agents/argo-rollouts/values.yaml
@@ -14,3 +14,5 @@ resources:
     cpu: 1000m
     memory: 1Gi
 
+
+compaction: {}
--- a/helm/agents/cilium-debug/templates/agent.yaml
+++ b/helm/agents/cilium-debug/templates/agent.yaml
@@ -155,6 +155,10 @@ spec:
             - "Monitor real-time traffic using cilium-dbg monitor"
             - "Check the health status of Cilium endpoints"
             - "Verify encryption status between pods"
+    {{- if .Values.compaction }}
+    context:
+      compaction: {{- toYaml .Values.compaction | nindent 8 }}
+    {{- end }}
     deployment:
       {{- with coalesce (empty .Values.imagePullSecrets | ternary nil .Values.imagePullSecrets) .Values.global.imagePullSecrets }}
       imagePullSecrets:
--- a/helm/agents/cilium-debug/values.yaml
+++ b/helm/agents/cilium-debug/values.yaml
@@ -13,3 +13,5 @@ resources:
   limits:
     cpu: 1000m
     memory: 1Gi
+
+compaction: {}
--- a/helm/agents/cilium-manager/templates/agent.yaml
+++ b/helm/agents/cilium-manager/templates/agent.yaml
@@ -427,6 +427,10 @@ spec:
             - "I need a CiliumNetworkPolicy for my application."
             - "How do I write a policy to allow only specific DNS queries?"
             - "What's the syntax for creating a Cilium egress policy?"
+    {{- if .Values.compaction }}
+    context:
+      compaction: {{- toYaml .Values.compaction | nindent 8 }}
+    {{- end }}
     deployment:
       {{- with coalesce (empty .Values.imagePullSecrets | ternary nil .Values.imagePullSecrets) .Values.global.imagePullSecrets }}
       imagePullSecrets:
--- a/helm/agents/cilium-manager/values.yaml
+++ b/helm/agents/cilium-manager/values.yaml
@@ -13,3 +13,5 @@ resources:
   limits:
     cpu: 1000m
     memory: 1Gi
+
+compaction: {}
--- a/helm/agents/cilium-policy/templates/agent.yaml
+++ b/helm/agents/cilium-policy/templates/agent.yaml
@@ -519,6 +519,10 @@ spec:
             - "I want to restrict traffic between pods in the same namespace, which policy should I use?"
             - "Can you create a Kubernetes NetworkPolicy for me?"
             - "What's the difference between CiliumNetworkPolicy and CiliumClusterwideNetworkPolicy for securing my nodes?"
+    {{- if .Values.compaction }}
+    context:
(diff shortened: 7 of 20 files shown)


EItanya (commented):
Rather than needing to add this to every helm chart for every feature. Is there a way we can figure out to do a generic passthrough based on the agent spec, or come up with a better system?

mesutoezdil:
> Rather than needing to add this to every helm chart for every feature. Is there a way we can figure out to do a generic passthrough based on the agent spec, or come up with a better system?

Hm. A generic additionalSpec: {} passthrough in values that merges into the declarative block would cover context, memory, and any future fields without touching each subchart again. Want me to rework it that way, or do you have a preferred shape in mind?

marcofranssen:
This is the helm native way though. Normally global is used to set defaults. And then use the current pattern to pass value to subcharts and allowing to override per agent.

mesutoezdil:
@EItanya the per-subchart values pattern is intentional, it keeps each chart usable standalone, which @marcofranssen also argued in #1869. That said, I agree the template duplication is a real maintenance concern. A shared partial in _helpers.tpl would centralize the rendering logic so adding a new field only requires touching one file, not ten. Happy to go that route if that addresses your concern.

marcofranssen:
> @EItanya the per-subchart values pattern is intentional, it keeps each chart usable standalone, which @marcofranssen also argued in #1869. That said, I agree the template duplication is a real maintenance concern. A shared partial in _helpers.tpl would centralize the rendering logic so adding a new field only requires touching one file, not ten. Happy to go that route if that addresses your concern.

Yes, you could create a separate library chart with some helpers. Meaning each agent chart could depend on the helpers chart to render the agent. However the values.yaml part will always need the duplication. However that means you will have to update and release the library chart. Then update the agent charts to bump to the new version and align their values.yaml. Not sure if it is worth the effort as it would complicate things for a single CR definition. If there would be more objects in the agents charts it would probably make sense, but for now I think this is best you can do.

mesutoezdil:
> > @EItanya the per-subchart values pattern is intentional, it keeps each chart usable standalone, which @marcofranssen also argued in #1869. That said, I agree the template duplication is a real maintenance concern. A shared partial in _helpers.tpl would centralize the rendering logic so adding a new field only requires touching one file, not ten. Happy to go that route if that addresses your concern.
> 
> Yes, you could create a separate library chart with some helpers. Meaning each agent chart could depend on the helpers chart to render the agent. However the values.yaml part will always need the duplication. However that means you will have to update and release the library chart. Then update the agent charts to bump to the new version and align their values.yaml. Not sure if it is worth the effort as it would complicate things for a single CR definition. If there would be more objects in the agents charts it would probably make sense, but for now I think this is best you can do.

Thx that confirms my thinking too. @EItanya given that values.yaml duplication is unavoidable either way and the current pattern keeps each chart usable standalone, are you okay with merging as-is?

EItanya (approved):

PR #1854, opened 2026-05-12
feat(controller): make MCP stateless mode configurable via `KAGENT_MCP_STATELESS`

onematchfox:
The `StreamableHTTPHandler` stores sessions in memory and returns `404` for any `Mcp-Session-Id` it doesn't recognise. With multiple controller replicas, a request may land on a different replica than the one that created the session.

`KAGENT_MCP_STATELESS=true` enables stateless mode, which skips session-ID validation and creates a temporary session per request, making any replica capable of handling any request. This enables use of the MCP when the network path does/can not provide sticky session routing based on the `Mcp-Session-Id` header.

Server->client initiated streaming is disabled in stateless mode, but that is not needed here — both tools are request/response and A2A conversation continuity is carried in the `context_id` field of the `invoke_agent` input.

### DIFF
--- a/go/core/internal/mcp/mcp_handler.go
+++ b/go/core/internal/mcp/mcp_handler.go
@@ -13,6 +13,7 @@ import (
 	authimpl "github.com/kagent-dev/kagent/go/core/internal/httpserver/auth"
 	"github.com/kagent-dev/kagent/go/core/internal/version"
 	"github.com/kagent-dev/kagent/go/core/pkg/auth"
+	"github.com/kagent-dev/kagent/go/core/pkg/env"
 	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


supreme-gg-gg (commented):
this lgtm overall. I agree with copilot we should add test coverage though. Perhaps not e2e test since it could be flaky and also make CI run even longer, I think integration test that spins up two handlers (simulating two replicas) and verifying stateful fails while stateless succeeds would be good

EItanya (commented):
I wonder if we should actually just make it stateless by default always. Realistically this will not be stateful without a bunch more work on our side

onematchfox:
> I wonder if we should actually just make it stateless by default always. Realistically this will not be stateful without a bunch more work on our side

Yeah, initially I was going to just make this the default. However, as mentioned in the comment, stateless mode does disable server-side notifications which, I am introducing in #1856 😅

onematchfox:
> this lgtm overall. I agree with copilot we should add test coverage though. Perhaps not e2e test since it could be flaky and also make CI run even longer, I think integration test that spins up two handlers (simulating two replicas) and verifying stateful fails while stateless succeeds would be good

Honestly I don't think this is worth it - we'd ultimately just be validating the SDK behaviour rather than kagent's code. The behaviours that the tests would exercise are implemented entirely within the SDK which has it's [own test suite](https://github.com/modelcontextprotocol/go-sdk/blob/v1.6.0/mcp/streamable_test.go). This also gets a little complex because the feature (within the context of Kagent) is enabled/disabled via an env var - which, to then test the 3 lines of configuration glue in KAgent would require setting up tests with that environment variable set/not-set. If you really want to, I can setup a test that ensures that the `httpOpts` is correctly set base on the presence of the env var?

supreme-gg-gg (approved):
> Honestly I don't think this is worth it - we'd ultimately just be validating the SDK behaviour rather than kagent's code. The behaviours that the tests would exercise are implemented entirely within the SDK which has it's [own test suite](https://github.com/modelcontextprotocol/go-sdk/blob/v1.6.0/mcp/streamable_test.go).

Sure that's reasonable. This lgtm

EItanya (approved):

PR #1855, opened 2026-05-12
refactor(controller): invoke agents directly in MCP handler

onematchfox:
Replace the HTTP round-trip through the controller's own A2A listener with direct invocation via a new `AgentClientRegistry`. The registry is owned by `A2ARegistrar`, which already maintains an `A2AClient` per agent for its HTTP mux — the registry gives the MCP handler access to those same clients without an extra network hop.

The old approach routed through the controller's public A2A endpoint, meaning requests could traverse the external network (and any ingress or load-balancer in front of it) unnecessarily. The new path stays in-process.

The old handler also cached its own `A2AClient` per agent in a `sync.Map` with no eviction, so clients for deleted agents would remain indefinitely. The registry is kept consistent by the registrar's add/update/delete lifecycle, eliminating that staleness.

`A2ARegistrar.upsertAgentHandler` writes to both the HTTP mux (for inbound /api/a2a/<ns>/<name>/ routing) and the registry (for direct invocation). The registry is exposed via `ClientRegistry()` and passed to `NewMCPHandler` in app.go.

~Note:~
~- currently includes changes from #1853 as well since this needs to adjust the tests added in that PR.~


onematchfox:
Now rebased and adjusted to account for #1853 being closed

go/core/internal/a2a/a2a_registrar.go
```diff
@@ -60,6 +62,12 @@ func NewA2ARegistrar(
 	return reg
 }
 
+// ClientRegistry returns the registry of A2A clients for direct agent
+// invocation, populated as agents are registered and deregistered.
+func (a *A2ARegistrar) ClientRegistry() *AgentClientRegistry {
+	return a.clientRegistry
+}
```
EItanya: I think this can be avoided if we just construct the client registry outside of this component and pass it into here AND the mcp handler, what do you think? I personally prefer injecting dependencies that way.

onematchfox: It's funny you say that because that was the way I implemented it initially. However, I ended up refactoring because it felt like the `Registrar` should wholly own the `Registry`; rather than viewing this as a dependency, maintaining the registry is a core function of the registrar's job. But, this isn't something I'm too firmly opinionated on so I'm happy to change tact again if you prefer?

EItanya: I would prefer that if it's ok

onematchfox: Addressed in https://github.com/kagent-dev/kagent/pull/1855/commits/1cafc0ad652bc512d057eea09535226e08bc570b

EItanya:
Hey there @onematchfox, there was a merge conflict created by the most recent PR, should be a simple test thing and I will merge after

EItanya (dismissed):

onematchfox:
> merge after

Done.

EItanya (dismissed):

onematchfox:
@EItanya Sorry, missed that CI had been failing after I addressed the merge conflicts. Should all be good now.

PR #1856, opened 2026-05-12
feat(mcp): expose agents as MCP resource with change notifications

onematchfox:
Adds a `kagent://agents` [MCP resource](https://modelcontextprotocol.io/specification/2025-06-18/server/resources) that lists  agents as JSON. Clients can read the resource directly or subscribe to receive notifications when the agent list changes. I've left the existing `list_agents` tool in place for now to not break existing consumers although this can potentially be removed prior to the next breaking version?

Notes: 
- Contains changes from both #1853 and #1855 at present. Diff will clean up as those get reviewed.
- Post this change we can also adjust the implementation of list agents to use the registry introduced in #1855 to avoid needing to query Kube API every time someone calls the endpoint.


EItanya (commented):
My initial review for this is that it looks great. I think a dynamic list makes a lot of sense here. My one nit pick is the following:
```
// SetAgentChangeCallback registers a function called whenever an agent is
// added, updated, or removed. The callback receives the context from the
// informer event so it can propagate cancellation.
func (a *A2ARegistrar) SetAgentChangeCallback(fn func(ctx context.Context)) {
	a.onAgentChange = fn
}

func (a *A2ARegistrar) notifyAgentChange(ctx context.Context) {
	if a.onAgentChange != nil {
		a.onAgentChange(ctx)
	}
}
```

Those 2 functions and the way the interact with the mcp_handlers. Specifically I Have a really hard time tracking down callbacks, maybe we could use a small interface for `NotifyAgentsChanged` or something? What do you think?

As far as removing the existing `list_agents`. I'm not sure about that. Resources aren't as widely used yet so I'd rather wait until there's more wide usage of them

onematchfox:
> Specifically I Have a really hard time tracking down callbacks, maybe we could use a small interface for `NotifyAgentsChanged` or something? What do you think?

Not 100% sure I'm following you... Were you thinking something like https://github.com/kagent-dev/kagent/pull/1856/commits/e3b33c9d45b6833cb684934726d93b4be3dde16b instead? 

> As far as removing the existing list_agents. I'm not sure about that. Resources aren't as widely used yet so I'd rather wait until there's more wide usage of them

Fine by me 👍

go/core/internal/a2a/a2a_registrar.go
```diff
@@ -88,6 +93,18 @@ func (a *A2ARegistrar) Start(ctx context.Context) error {
 	return nil
 }
 
+// ClientRegistry returns the registry of A2A clients for direct agent
+// invocation, populated as agents are registered and deregistered.
+func (a *A2ARegistrar) ClientRegistry() *AgentClientRegistry {
+	return a.clientRegistry
+}
```
EItanya: I think this can be removed as we pass it in now

go/core/internal/a2a/a2a_registrar.go
```diff
@@ -88,6 +93,18 @@ func (a *A2ARegistrar) Start(ctx context.Context) error {
...
+
+// SetAgentObserver registers an observer notified whenever an agent is added,
+// updated, or removed.
+func (a *A2ARegistrar) SetAgentObserver(o AgentObserver) {
+	a.agentObserver = o
+}
```
EItanya: Same as my comment about the client registry I'd much rather pass this in at construct time rather than do this and have to check for `nil` later.

EItanya (commented):
Just a couple small nits.

What do you think about adding an e2e style test for this. I know it's also testing the mcp library, but just a quick smoke test could make sure the functionality doesn't randomly stop working

go/core/internal/a2a/a2a_registrar.go
onematchfox: Yeah, my bad... Missed when dealing with merge conflicts. Will fix in the am.

onematchfox:
> What do you think about adding an e2e style test for this. I know it's also testing the mcp library, but just a quick smoke test could make sure the functionality doesn't randomly stop working

I'm good with that. Will put something together.

onematchfox:
> > What do you think about adding an e2e style test for this. I know it's also testing the mcp library, but just a quick smoke test could make sure the functionality doesn't randomly stop working
> 
> I'm good with that. Will put something together.

Done

EItanya (approved):
Looks great :)

PR #1857, opened 2026-05-12
fix(openclaw): bind gateway to loopback interface

felipevicens:
Add Bind: loopback to the openclaw gateway bootstrap configuration
   so the local gateway only listens on the loopback interface rather
   than all network interfaces.

Without this setting, OpenClaw defaults to 0.0.0.0. However, the sandbox iptables rules block all non-proxy traffic, so the TUI can only reach the gateway through loopback inside the sandbox network namespace. As a result, binding to 0.0.0.0 makes the gateway unreachable from the TUI (not connected to gateway).

Setting "bind": "loopback" fixes the issue at the source in the generated config, instead of requiring a manual post-creation patch to openclaw.json.


peterj:
thanks for the PR! Could you also just update the unit test [here](https://github.com/kagent-dev/kagent/blob/e35d1a04c72f5256a6b07daa2ac686e5a997f49c/go/core/pkg/sandboxbackend/openshell/openclaw/bootstrap_test.go#L59), so it also check the bind value being set correctly?

peterj (approved):

EItanya (approved):

PR #1858, opened 2026-05-12
fix(adk): Propagate access token to subagents and mcp tools

### LINKED ISSUE #1745: [FEATURE]  _SubagentInterceptor should forward Authorization header from parent session to sub-agent A2A calls

jmhbh:
**Overview**
- Fixes an issue where setting the `KAGENT_PROPAGATE_TOKEN` env var for `Agent` CR's using the `python` runtime would only propagate the access token to mcp tools and not to sub agents. 
- Fixes an issue where the setting the `KAGENT_PROPAGATE_TOKEN` env var for the `go` run would not propagate the token to mcp or subagents as it was unimplented.

**Testing**
- Manually tested by deploying a coordinator agent and a subagent that uses the `kagent-tool-server` and making an a2a request to the coordinator agent then inspecting the logs. I added logs [here](https://github.com/kagent-dev/kagent/pull/1858/changes#diff-2c6ac47a132ff8be9df40e1902380810f3d2c8ead904caaf53d22161440903c5R56) and [here](https://github.com/kagent-dev/kagent/pull/1858/changes#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfR267) for the go runtime and [here](https://github.com/kagent-dev/kagent/pull/1858/changes#diff-e2f83de8fb21d0c7da42da4ab5f1ad96d03fe03abbb40432b80ac2adaf483ba8R90) for the python runtime to validate the allowed header was being added by the interceptor

closes https://github.com/kagent-dev/kagent/issues/1745


supreme-gg-gg (commented):
lgtm overall, I agree using `KAGENT_PROPAGATE_TOKEN` is a better and more consistent approach than the previous PR / #1757

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
```diff
@@ -58,21 +59,34 @@
 _USER_ID_CONTEXT_KEY = "x-user-id"
 _SOURCE_HEADER = "x-kagent-source"
 _SOURCE_SUBAGENT = "agent"
+_HEADERS_STATE_KEY = "headers"
+_AUTHORIZATION_CONTEXT_KEY = "authorization"
+_PROPAGATE_TOKEN = bool(os.getenv("KAGENT_PROPAGATE_TOKEN"))
```
supreme-gg-gg: nit: this requires `KAGENT_PROPAGATE_TOKEN` to be `True` instead of just check presence, which seems different from the behaviour in `cli.py` which would work for any values like "yes"

go/adk/pkg/tools/remote_a2a_tool.go
```diff
@@ -32,6 +34,28 @@ func (u *userIDForwardingInterceptor) Before(ctx context.Context, req *a2aclient
...
+	callCtx, ok := a2asrv.CallContextFrom(ctx)
+	if !ok {
+		return ctx, nil
+	}
+	meta := callCtx.RequestMeta()
+	if meta == nil {
+		return ctx, nil
+	}
+	if vals, ok := meta.Get(constants.AuthorizationHeader); ok && len(vals) > 0 && vals[0] != "" {
+		req.Meta.Append(constants.AuthorizationHeader, vals[0])
```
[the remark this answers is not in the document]
supreme-gg-gg: this might be valid

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
iplay88keys: I think we should fix that in `cli.py`. I've had an issue locally which I have been meaning to create where:

Setting `KAGENT_PROPAGATE_TOKEN: "false"` on the k8s agent (or any ADK agent) does not disable token propagation. The user's Bearer token is still sent to the tool server (e.g. kagent-tools) when the agent makes MCP calls.

## Root cause

In `python/packages/kagent-adk/src/kagent/adk/cli.py`, the env var is read as a raw string and used in a boolean context:

```python
propagate_token = os.getenv("KAGENT_PROPAGATE_TOKEN")

def create_sts_integration() -> Optional[ADKTokenPropagationPlugin]:
    if sts_well_known_uri or propagate_token:  # bug: "false" is truthy in Python
        ...
        return ADKTokenPropagationPlugin(sts_integration)
```

In Python, any non-empty string is truthy. So when the Agent deployment has:

```yaml
env:
  - name: KAGENT_PROPAGATE_TOKEN
    value: "false"
```

`propagate_token` is the string `"false"`, and `if propagate_token` is `True`. The token propagation plugin is therefore always added whenever the env var is **set**, regardless of value. Only leaving the env var **unset** currently disables propagation.

go/adk/pkg/tools/remote_a2a_tool.go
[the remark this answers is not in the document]
jmhbh: fixed so existing headers take precedence.

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
jmhbh: makes sense, I fixed this in cli.py and made sure the behavior is consistent in `_remote_a2a_tool.py`

go/adk/pkg/mcp/registry.go
```diff
@@ -230,30 +239,41 @@ func createTransport(ctx context.Context, params mcpServerParams) (mcpsdk.Transp
...
+	// A2A request independently of allowedHeaders.
+	if rt.propagateToken {
+		if callCtx, ok := a2asrv.CallContextFrom(req.Context()); ok {
+			if meta := callCtx.RequestMeta(); meta != nil {
+				if vals, ok := meta.Get(constants.AuthorizationHeader); ok && len(vals) > 0 && vals[0] != "" {
+					req.Header.Set(constants.AuthorizationHeader, vals[0])
+				}
+			}
+		}
+	}
```
iplay88keys, line 269: I haven't worked with golang context much. Is this the idiomatic way to store/retrieve something like this? It seems like a lot of nesting for this value.

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
```diff
@@ -381,7 +404,14 @@ async def _handle_resume(self, tool_context: ToolContext) -> Any:
...
         client = await self._ensure_client()
-        call_context = ClientCallContext(state={_USER_ID_CONTEXT_KEY: tool_context.session.user_id})
+        call_context_state: dict[str, Any] = {_USER_ID_CONTEXT_KEY: tool_context.session.user_id}
+        if self._propagate_token:
+            incoming = tool_context.session.state.get(_HEADERS_STATE_KEY) or {}
+            if isinstance(incoming, dict):
+                auth = incoming.get("authorization") or incoming.get("Authorization")
+                if auth:
+                    call_context_state[_AUTHORIZATION_CONTEXT_KEY] = auth
+        call_context = ClientCallContext(state=call_context_state)
```
iplay88keys: This looks to be duplicated from above. Would it make sense to pull it out to a new function?

iplay88keys (dismissed):
Looks good overall. Just a few small comments.

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
jmhbh: fixed

go/adk/pkg/mcp/registry.go
jmhbh: request metadata is generally considered acceptable. Excessive non-request scoped values are considered context stuffing

go/adk/pkg/agent/agent.go
```diff
@@ -48,7 +48,8 @@ func CreateGoogleADKAgentWithSubagentSessionIDs(ctx context.Context, agentConfig
 		return nil, nil, fmt.Errorf("agent config is required")
 	}
 
-	toolsets := mcp.CreateToolsets(ctx, agentConfig.HttpTools, agentConfig.SseTools)
+	propagateToken := os.Getenv("KAGENT_PROPAGATE_TOKEN") != ""
```
iplay88keys: It looks like you fixed this in the python side to check for `== "true"`. Let's fix it here as well.

jmhbh: good catch fixed.

go/adk/pkg/tools/remote_a2a_tool.go
```diff
@@ -32,6 +34,31 @@ func (u *userIDForwardingInterceptor) Before(ctx context.Context, req *a2aclient
 	return ctx, nil
 }
 
+// authzForwardingInterceptor forwards the Authorization header from the
+// incoming A2A request context to outbound sub-agent A2A calls.
+type authzForwardingInterceptor struct {
+	a2aclient.PassthroughInterceptor
+	name string
```
EItanya: nit: what is this used for?

EItanya (commented):
Just one small comment

go/adk/pkg/tools/remote_a2a_tool.go
jmhbh: I used this for debugging to enrich my logging when i was testing. I'll remove it.

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
```diff
@@ -58,21 +59,34 @@
 _USER_ID_CONTEXT_KEY = "x-user-id"
 _SOURCE_HEADER = "x-kagent-source"
 _SOURCE_SUBAGENT = "agent"
+_HEADERS_STATE_KEY = "headers"
+_AUTHORIZATION_CONTEXT_KEY = "authorization"
+_PROPAGATE_TOKEN = os.getenv("KAGENT_PROPAGATE_TOKEN", "").lower() == "true"
```
iplay88keys: It feels weird to me to read an env var here -- especially since we are reading it in `cli.py` already. Is this something that can be passed through instead, or is it needed here?

jmhbh: yeah we can prob pass this as an arg

EItanya: The issue is that the MCP code uses a different header provider. Can we re-use that for this somehow?

iplay88keys (dismissed):

python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
EItanya: Bump on this, passing `propogate_token` through is confusing because it's already handled for MCP. Can we pass `HeaderProviders` to `KAgentRemoteA2ATool`?

jmhbh: makes sense. fixed

EItanya (approved):

EItanya (approved):

supreme-gg-gg (approved):

PR #1869, opened 2026-05-14
helm: Allow configuring agent memory

marcofranssen:


EItanya:
@mesutoezdil here's another example.

https://github.com/kagent-dev/kagent/pull/1849#issuecomment-4445455863

marcofranssen:
> @mesutoezdil here's another example.
> 
> 
> 
> https://github.com/kagent-dev/kagent/pull/1849#issuecomment-4445455863

https://github.com/kagent-dev/kagent/pull/1849#issuecomment-4458628018

I think we are about to break helm best practices again if we go that route. Similar like I recently fixed to be able to pass through per agent settings in my earlier PR.

This is really the way to pass through values to subcharts. 

One example of a chart that does this correctly is the spire helm chart(s). Please let's not break the pattern again.

global is another pattern that allows global defaults. Usually used for e.g. image registry. And usually that is used combined with post component ways of overriding.

Even though there is duplication like this it makes the helm charts usable as individual components. Which is actually great as it allows people to make their own composite charts for their own specific usecases.

marcofranssen:
Fixed the DCO just now.

helm/agents/argo-rollouts/templates/agent.yaml
```diff
@@ -120,6 +120,13 @@ spec:
           name: kagent-builtin-prompts
           alias: builtin
     modelConfig: {{ .Values.modelConfigRef | default (printf "%s" (include "kagent.defaultModelConfigName" .)) }}
+    {{- with .Values.memory }}
+    {{- if .enabled }}
+    memory:
+      modelConfig: {{ .modelConfigRef | default (printf "%s" (include "kagent.defaultModelConfigName" .)) }}
```
supreme-gg-gg: Default model config uses a LLM model name not embedding model name so this will break if it falls back to default. I think we will need to either create a default embedding model or just don't use a default value and require the user to create a separate embedding modelconfig themselves

marcofranssen: I'm not sure if I understand your feedback.

`uses a LLM model name not embedding model name`? 🤔 can you rephrase?

I have tested it, it seems to work, as I see e.g. my k8s-agent now trying to look up memories. I'm not defining a separate Model for this and just use the default modeConfigRef.

mesutoezdil: > I'm not sure if I understand your feedback.
> 
> `uses a LLM model name not embedding model name`? 🤔 can you rephrase?
> 
> I have tested it, it seems to work, as I see e.g. my k8s-agent now trying to look up memories. I'm not defining a separate Model for this and just use the default modeConfigRef.

Memory uses vector embeddings. A separate model category (e.g. text-embedding-3-small for OpenAI, not gpt-4). The embedding client in _embedding.py supports openai, azure_openai, ollama, gemini/vertex_ai, and bedrock. There is no anthropic provider, bcs Anthropic has no embedding API.

Falling back to the default LLM modelConfig silently breaks this: the embedding call will either fail or produce garbage, and the memory service just logs a warning and continues. So it can look like it's working while vector search isn't actually happening.

The fix should be to remove the default fallback and use Helm's required function instead: modelConfig: {{ required "memory.modelConfigRef must be set when memory.enabled is true".Values.memory.modelConfigRef }}

This way users are forced to create a dedicated embedding ModelConfig (pointing to an actual embedding model) rather than accidentally reusing their LLM config.

marcofranssen: Ok, what should be the model to use in combination with e.g. Sonnet 4.6 on Bedrock? Can it be text-embedding-3-small by OpenAI? or can it be Amazon Titan Text Embeddings?

supreme-gg-gg: > I have tested it, it seems to work, as I see e.g. my k8s-agent now trying to look up memories.

This is interesting lol, does it actually look up memories or does it log an error? I meant default model config points to a language model like `gpt-4.1-mini` and you need an embedding model like `text-embedding-3-small`, so you can't just share the two configs directly even though they're both OpenAI providers.

In the case of bedrock I use `amazon.titan-embed-text-v2:0`

marcofranssen: Updated using your suggestion

marcofranssen: > > I have tested it, it seems to work, as I see e.g. my k8s-agent now trying to look up memories.
> 
> This is interesting lol, does it actually look up memories or does it log an error? I meant default model config points to a language model like `gpt-4.1-mini` and you need an embedding model like `text-embedding-3-small`, so you can't just share the two configs directly even though they're both OpenAI providers.
> 
> In the case of bedrock I use `amazon.titan-embed-text-v2:0`

No memories, just noticed in UI it was trying to get one. Also no error logs as far I could see.

marcofranssen: Now tried with `titan-embed-text-v2:0`

```
{
  "content": "Kagent namespace workload improvement findings (analyzed 2026-05-15):\n\nHIGH PRIORITY:\n1. Missing CPU Limits: helm-agent, k8s-agent, kagent-controller, kagent-grafana-mcp, kagent-querydoc, kagent-kmcp-controller-manager, and kagent-ui all lack CPU limits. Only kagent-tools (1 CPU) and kagent-oauth2-proxy (100m) have CPU limits set.\n2. kagent-grafana-mcp uses image tag 'latest' (mcp/grafana:latest) with imagePullPolicy: Always — should be pinned to a specific version tag and use imagePullPolicy: IfNotPresent.\n\nMEDIUM PRIORITY:\n3. Missing Liveness Probes: helm-agent, k8s-agent, kagent-grafana-mcp, kagent-querydoc, kagent-tools, kagent-controller, and kagent-ui all lack liveness probes. Only kagent-kmcp-controller-manager (/healthz) and kagent-oauth2-proxy (/ping) have liveness probes.\n4. Missing Security Context hardening (runAsNonRoot, readOnlyRootFilesystem, allowPrivilegeEscalation: false, seccompProfile) on helm-agent and k8s-agent. kagent-oauth2-proxy and kagent-kmcp-controller-manager are the best hardened. kagent-controller and kagent-ui have partial hardening (runAsNonRoot: true, readOnlyRootFilesystem: true).\n\nAll 9 pods were healthy and running at time of analysis. Deployments are Helm-managed (release: kagent, namespace: kagent)."
}
{"error":"failed to generate embedding: failed to invoke Bedrock model for text 0: operation error Bedrock Runtime: InvokeModel, https response error StatusCode: 400, RequestID: 2599be3c-793c-43b6-a930-d39fa16b67bd, ValidationException: messages: Field required"}
```

supreme-gg-gg: Are you using a custom build of kagent since I just verified that this works. Perhaps it has to do with your local changes in the other bedrock PRs you put up?

marcofranssen: yeah maybe.

Anything required to get this one merged? I'll retry the embeddings parts once this gets in without the other changes.

.editorconfig
EItanya, line 1: Can we please remove this from the PR. If we want an `.editorconfig` let's leave that in a separate PR where we can discuss it

EItanya (changes requested):

mesutoezdil (changes requested):
2 issues need to be fixed.

`required` argument order is wrong in all 10 templates

The Helm `required` function signature is `required "message" .value`, but every template has the arguments swapped:

```
# current (broken)
modelConfig: {{ required .modelConfigRef "A compatible modelConfig must be provided when memory is enabled" }}

# should be
modelConfig: {{ required "A compatible modelConfig must be provided when memory is enabled" .modelConfigRef }}
```

With the arguments flipped, `required` treats the literal string as the value and `.modelConfigRef` as the error message. 

An empty `modelConfigRef` will silently render the error string as the field value instead of failing at template render time. 

This affects all 10 agent templates.

Trailing blank lines in argo-rollouts/templates/agent.yaml

The diff adds 6 empty lines at the end of the file. 

Also agree with EItanya that `.editorconfig` belongs in a separate PR.

.editorconfig
marcofranssen: Done and created https://github.com/kagent-dev/kagent/pull/1899

helm/agents/promql/templates/agent.yaml
```diff
@@ -215,4 +222,4 @@ spec:
       {{- end }}
       resources:
         {{- toYaml .Values.resources | nindent 8 }}
-      
\ No newline at end of file
+
```
marcofranssen: These kind of inconsistencies are autoresolved by a editorconfig for every user that has it in their IDE and any AI agent generating stuff.

https://github.com/kagent-dev/kagent/pull/1899

helm/agents/argo-rollouts/values.yaml
```diff
@@ -1,6 +1,12 @@
 # -- Optional reference to a ModelConfig resource for this agent. If not specified, the default model config will be used.
 modelConfigRef: ""
 
+memory:
+  enabled: false
```
supreme-gg-gg, line 5: I think we should follow the CRD and gate if memory config should be added in template with the presence of memory block instead of an explicit enabled for consistency

helm/agents/argo-rollouts/values.yaml
```diff
@@ -1,6 +1,12 @@
...
+  ttlDays: 15
+  # Optional reference to a ModelConfig resource for the agent's memory. If not specified, the default model config will be used.
```
supreme-gg-gg: Also need to update comment there isn't a default as per prev discussion

helm/agents/argo-rollouts/values.yaml
marcofranssen: That will complicate the helm chart.

1. Most succesful large charts use this pattern.
2. This allows us to define defaults and document the value structure as we can pre-define all values.
3. The advantage of that is that the charts nicely show up in a documented form at https://artifacthub.io/ _kagent should integrate there as well IMHO (happy to help with that as I did this before for couple projects)_

marcofranssen:
Rebased to fix conflicts with main.

marcofranssen:
Rebased again to resolve conflicts with main.

EItanya (approved):

PR #1873, opened 2026-05-15
fix(bedrock): preserve thinking blocks in multi-turn tool use

### LINKED ISSUE #1870: [BUG]: ModelConfig Bedrock Thinking not working.
I have been testing this, and it doesn't seem to work.

```shell
LLM error: API_ERROR operation error Bedrock Runtime: Converse, https response error StatusCode: 400, RequestID: 7dbf6c3a-d526-43ef-a9b9-b48228b1821f, ValidationException: The value at messages.1.content.2.toolUse.input is empty.
```

https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-extended-thinking.html

According to Claude it thinks following:

```
Now I have the full picture. The Bedrock SDK has ContentBlockMemberReasoningContent with ReasoningTextBlock{Text, Signature}
  and RedactedContent []byte. The genai Part struct has Thought bool and ThoughtSignature []byte.

  The fix requires two things in bedrock.go:

  1. generateNonStreaming: When extracting response content, handle ContentBlockMemberReasoningContent and emit
  genai.Part{Thought: true, Text: ..., ThoughtSignature: ...}
  2. generateStreaming: Track reasoning content deltas (text + signature) and emit thinking parts
  3. convertGenaiContentsToBedrockMessages: When building messages from conversation history, handle part.Thought == true parts
  by emitting ContentBlockMemberReasoningContent
```

_Originally posted by @marcofranssen in https://github.com/kagent-dev/kagent/issues/1732#issuecomment-4460321507_

mesutoezdil:
Fixes #1870

When extended thinking is active, Bedrock returns thinking content blocks together with tool use blocks. 

The API requires these blocks to be sent back unmodified in the next request. 

Without them, Bedrock returns a ValidationException with toolUse.input is empty.

See the API requirement: https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-extended-thinking.html

Tests added for thinking block round-trip and temperature/top_p exclusion.

cc @marcofranssen @supreme-gg-gg @EItanya

### DIFF
--- a/go/adk/pkg/models/bedrock.go
+++ b/go/adk/pkg/models/bedrock.go
@@ -171,20 +171,9 @@ func (m *BedrockModel) GenerateContent(ctx context.Context, req *model.LLMReques
 		// is written with the sanitized name Bedrock already knows about.
 		messages, systemInstruction := convertGenaiContentsToBedrockMessages(req.Contents, nameMap)
 
-		// Build inference config
-		var inferenceConfig *types.InferenceConfiguration
-		if m.Config.MaxTokens != nil || m.Config.Temperature != nil || m.Config.TopP != nil {
-			inferenceConfig = &types.InferenceConfiguration{}
-			if m.Config.MaxTokens != nil {
-				inferenceConfig.MaxTokens = aws.Int32(int32(*m.Config.MaxTokens))
-			}
-			if m.Config.Temperature != nil {
-				inferenceConfig.Temperature = aws.Float32(float32(*m.Config.Temperature))
-			}
-			if m.Config.TopP != nil {
-				inferenceConfig.TopP = aws.Float32(float32(*m.Config.TopP))
-			}
-		}
+		// temperature/top_p must not be sent when thinking is active. https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-extended-thinking.html
+		_, thinkingEnabled := m.Config.AdditionalModelRequestFields["thinking"]
+		inferenceConfig := buildInferenceConfig(m.Config, thinkingEnabled)
 
 		// Build system prompt
 		var systemPrompt []types.SystemContentBlock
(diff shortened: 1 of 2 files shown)


mesutoezdil:
@supreme-gg-gg could you test this with extended thinking and tool use? That is the scenario that was broken.

marcofranssen:
I was doing an atempt here as well, but couldn't get it to work just yet, allthough there might be parts useful for this PR.

https://github.com/kagent-dev/kagent/pull/1872

mesutoezdil:
> I was doing an atempt here as well, but couldn't get it to work just yet, allthough there might be parts useful for this PR.
> 
> #1872

Thx!  I took a look at #1872 and incorporated the token explosion fix from there.

EItanya:
> I was doing an atempt here as well, but couldn't get it to work just yet, allthough there might be parts useful for this PR.
> 
> #1872

@marcofranssen any chance you can give this PR a smoke test before I review/approve?

EItanya (commented):
Can you please separate the tracing changes into a different PR. The changes make sense but they are unrelated. Separately I think it should be optional so as to not break current flows

mesutoezdil:
Addressing the three Copilot review comments:

1. TestThinkingDropsTemperatureAndTopP
replaced by `TestBuildInferenceConfig` which calls `buildInferenceConfig()` directly and asserts actual `InferenceConfiguration` field values (Temperature, TopP, MaxTokens) for four scenarios. The original map-key-only check is gone.

2. Helm metrics / ServiceMonitor bearer token
the Helm metrics changes have been removed from this PR entirely. They will be submitted separately.

3. PR scope
tracing changes also removed. This PR now only touches `bedrock.go` and `bedrock_test.go`.

mesutoezdil:
Done. Tracing changes have been removed from this PR.

EItanya (approved):

EItanya (dismissed):

PR #1875, opened 2026-05-15
ui: Display the authenticated user by configured userIdClaim

marcofranssen:
## Summary

- Adds `KAGENT_USER_ID_CLAIM` environment variable to the UI pod, sourced from `controller.auth.userIdClaim` in `values.yaml`
- Exposes a `getUserIdClaim()` server action that reads `KAGENT_USER_ID_CLAIM` (defaults to `"sub"`)
- Extends `AuthContext` to fetch the configured claim name alongside the JWT and expose it as `userIdClaim`
- Updates `AppInitializer` to sync the authenticated user's identity into `userStore` using the configured claim (e.g. `email`, `preferred_username`, `sub`) so the correct value is displayed in the UI

## Motivation

When deploying with `auth.mode: trusted-proxy` and `controller.auth.userIdClaim: email`, the UI was still showing the raw `sub` value (typically an opaque ID) instead of the user's email. This was because `userStore` was seeded from `localStorage` and never synchronized with the authenticated JWT claims.

## How it works

```
values.yaml
  controller.auth.userIdClaim: email
        │
        ▼
ui-deployment.yaml
  KAGENT_USER_ID_CLAIM=email
        │
        ▼
auth.ts: getUserIdClaim()   +   getCurrentUser() (JWT decode)
        │                               │
        └───────────┬───────────────────┘
                    ▼
           AuthContext: { user, userIdClaim }
                    │
                    ▼
          AppInitializer: setUserId(user[userIdClaim])
                    │
                    ▼
             userStore (displayed in UI)
```

## Test plan

- [ ] Deploy with `controller.auth.userIdClaim: email` and `auth.mode: trusted-proxy` — UI should display the email from the JWT
- [ ] Deploy with `controller.auth.userIdClaim: preferred_username` — UI should display the preferred username
- [ ] Deploy without setting `userIdClaim` (default) — UI should fall back to `sub`
- [ ] Deploy in `unsecure` mode (no JWT) — UI should fall back to the default `admin@kagent.dev`

### DIFF
--- a/helm/kagent/templates/ui-deployment.yaml
+++ b/helm/kagent/templates/ui-deployment.yaml
@@ -60,13 +60,17 @@ spec:
               value: {{ .Values.ui.publicBackendUrl | quote }}
             - name: BACKEND_INTERNAL_URL
               value: {{ .Values.ui.backendInternalUrl | default (include "kagent.controllerInternalHttpApiBase" .) | quote }}
-            {{- if .Values.ui.auth }}
+            {{- with .Values.ui.auth }}
             - name: SSO_REDIRECT_PATH
-              value: {{ .Values.ui.auth.ssoRedirectPath | default "/oauth2/start" | quote }}
+              value: {{ .ssoRedirectPath | default "/oauth2/start" | quote }}
             {{- end }}
             {{- if .Values.ui.streamTimeoutSeconds }}
             - name: KAGENT_STREAM_TIMEOUT_MS
               value: {{ mul (int .Values.ui.streamTimeoutSeconds) 1000 | quote }}
+            {{- end }}
+            {{- with .Values.controller.auth.userIdClaim }}
+            - name: KAGENT_USER_ID_CLAIM
+              value: {{ . | quote }}
             {{- end }}
             {{- with .Values.ui.additionalForwardedHeaders }}
             - name: KAGENT_ADDITIONAL_FORWARDED_HEADERS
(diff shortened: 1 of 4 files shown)


EItanya (commented):
We already have the [current user handler](https://github.com/kagent-dev/kagent/blob/ae3c873406ff712ced5c5c2c81c247711504eb85/go/core/internal/httpserver/handlers/current_user.go). Should we instead expand that to include more information? I favor backend solutions usually

marcofranssen:
I think we need both. The backend seems to map the id to sub, and claims is never used. Mapping ID to sub, imho is wrong as well. It should just forward the JWT and then client side use that to render a attribute of that token.

EItanya:
> I think we need both. The backend seems to map the id to sub, and claims is never used. Mapping ID to sub, imho is wrong as well. It should just forward the JWT and then client side use that to render a attribute of that token.

Should we maybe edit the backend endpoint rather than just masking this in the front-end?

EItanya:
> > I think we need both. The backend seems to map the id to sub, and claims is never used. Mapping ID to sub, imho is wrong as well. It should just forward the JWT and then client side use that to render a attribute of that token.
> 
> Should we maybe edit the backend endpoint rather than just masking this in the front-end?

bump on this comment @marcofranssen

peterj:
ping @marcofranssen  in case you're still interested in getting this PR in.

PR #1880, opened 2026-05-15
feat: go runtime sts client and adk integration

supreme-gg-gg:


go/adk/pkg/runner/adapter.go
```diff
@@ -61,11 +72,57 @@ func CreateRunnerConfig(
...
+		nil,   // getSubjectToken
+		0,     // default timeout
+		true,  // default verifySSL
+		false, // default useIssuerHost
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize STS integration: %w", err)
+	}
+
+	log.Info("Enabling STS token propagation plugin",
```
jmhbh: nit: could we format this log as one line?

go/adk/pkg/runner/adapter.go
```diff
@@ -61,11 +72,57 @@ func CreateRunnerConfig(
...
+		log.Info("Enabling token propagation plugin without STS exchange")
+		return sts.NewTokenPropagationPlugin(nil, log), nil
+	}
+
+	integration, err := sts.NewSTSIntegration(
+		stsWellKnownURI,
+		"",
+		nil,   // fetchActorToken
+		nil,   // getSubjectToken
+		0,     // default timeout
```
jmhbh: isn't the default time out 5s?

go/adk/pkg/sts/client.go
```diff
@@ -0,0 +1,275 @@
...
+
+// NewSTSClient creates a new STS client with the given configuration.
+func NewSTSClient(config STSConfig) *STSClient {
+	config = normalizeSTSConfig(config)
+	return &STSClient{
+		config: config,
+	}
+}
+
+func normalizeSTSConfig(config STSConfig) STSConfig {
```
jmhbh: what is the rationale behind this function?

go/adk/pkg/sts/client.go
```diff
@@ -0,0 +1,275 @@
...
+	errorDescription := ""
+	if ed, ok := responseData["error_description"].(string); ok {
+		errorDescription = ed
+	}
+
+	return nil, NewTokenExchangeError(errorCode, errorDescription, resp.StatusCode)
+}
+
+// Impersonate performs an impersonation token exchange (no actor token).
+func (c *STSClient) Impersonate(
```
jmhbh, line 216: nit: the following functions are only used in tests but I do like them since they're very clear. I think it would be easier to follow if you used a switch statement in `integration.go`'s `ExchangeTokenWithActorToken` function to invoke these two functions based on the presence of `actorToken` and `actorTokenType`

go/adk/pkg/mcp/registry_test.go
```diff
@@ -319,3 +319,80 @@ func TestAllowedRequestHeaders_ReturnsNilWhenNoMatches(t *testing.T) {
...
+
+// TestDynamicHeaders_OverridePropagatedAuthorization verifies dynamic headers
+// take precedence over propagated and allowed request headers.
+func TestDynamicHeaders_OverridePropagatedAuthorization(t *testing.T) {
+	t.Parallel()
+	var capturedAuth, capturedTrace string
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		capturedAuth = r.Header.Get("Authorization")
+		capturedTrace = r.Header.Get("X-Trace-Id")
```
jmhbh: nit: `x-trace-id` here seems to be a bit of a misrepresentation given that our implementation only overwrites the `Authorization` header.

go/adk/pkg/sts/client.go
supreme-gg-gg: This has to do with the defaulting path originally, I've removed this after centralizing the defaulting behaviour to only happen in `client.go`

go/adk/pkg/runner/adapter.go
supreme-gg-gg: Yes but a non-positive (<=0) value will default to 5, I've rewritten these default logic to be more centralized

go/adk/pkg/mcp/registry_test.go
supreme-gg-gg: The reason why I included that is because `Authorization` only tests overriding the forwarded auth header which uses the `propagateToken` path, but `x-trace-id` (I've renamed to `x-custom` to not imply anything about tracing) tests that the dynamic header provider also overrides the normal `allowedHeaders` path

jmhbh (approved):

EItanya (approved):

PR #1886, opened 2026-05-18
feat: add `ANTHROPIC_BASE_URL` support

dongjiang1989:
## Summary                                                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  - Add `ANTHROPIC_BASE_URL` environment variable support across Go controller and Python ADK, mirroring the existing `OPENAI_API_BASE` pattern                                                                                     
  - Inject `ANTHROPIC_BASE_URL` into agent pods when Anthropic provider has a custom BaseURL configured                                                                                                                             
  - Add unittest test for Anthropic


EItanya (commented):
Thanks so much for the contribution. Can we leave the new sample agent out of this PR. Frankly the existing sample agents are already causing issues with dependencies and other issues. I think a separate repo with sample agents might be a better long-term solution, but for now can you please remove it.

dongjiang1989:
> Thanks so much for the contribution. Can we leave the new sample agent out of this PR. Frankly the existing sample agents are already causing issues with dependencies and other issues. I think a separate repo with sample agents might be a better long-term solution, but for now can you please remove it.

Thanks @EItanya 
I totally agree. I will remove the sample agent and add corresponding unit test cases. Please review again.

dongjiang1989:
cc @EItanya  Please re-check it

python/packages/kagent-adk/src/kagent/adk/models/_anthropic.py
```diff
@@ -44,11 +44,12 @@ def _create_http_client(self):
     @cached_property
     def _anthropic_client(self) -> AsyncAnthropic:
         api_key = self._api_key or os.environ.get("ANTHROPIC_API_KEY")
+        base_url = self.base_url or os.environ.get("ANTHROPIC_BASE_URL")
         kwargs = {}
         if api_key:
             kwargs["api_key"] = api_key
-        if self.base_url:
-            kwargs["base_url"] = self.base_url
+        if base_url:
```
supreme-gg-gg, line 51: What would be a use case to inject the `ANTHROPIC_BASE_URL` env var? As shown here we already set the `base_url` kwarg on the anthropic client (which has a higher order of precedence than the env var), so if you specify a base url in the modelconfig for anthropic models it already works.

dongjiang1989: Thanks @supreme-gg-gg 

Totally agree. The explicit `base_url` passed into the Anthropic client has higher precedence and fully covers all custom endpoint requirements defined in ModelConfig.

We keep supporting the `ANTHROPIC_BASE_URL` environment variable mainly for native SDK convention compliance, convenient local debugging, unified global proxy configuration, and compatibility with external tooling that only respects official Anthropic environment variables.

Additionally, it provides a smooth out-of-the-box experience for new users: when first installing `kagent`, they can simply set `ANTHROPIC_BASE_URL`(with a proxy URL) and `ANTHROPIC_API_KEY` at the system level to get a working default setup immediately, without needing to configure ModelConfig upfront.

EItanya: > Additionally, it provides a smooth out-of-the-box experience for new users: when first installing `kagent`, they can simply set `ANTHROPIC_BASE_URL`(with a proxy URL) and `ANTHROPIC_API_KEY` at the system level to get a working default setup immediately, without needing to configure ModelConfig upfront.

Specifically here you mean when running the agent locally? If so why do we need this code in the controller. I think this is a lot of added code with complexity to optimize for a small local use case.

dongjiang1989: Thanks @EItanya                                                                                                                                  

To clarify the use case: in certain regions (e.g. China), the Anthropic API is not directly accessible. Users need to configure third-party providers that implement the Anthropic Messages API protocol, such as https://help.aliyun.com/zh/model-studio/anthropic-api-messages#ac2a3b4c5d6f7. They set this via ModelConfig.Spec.Anthropic.BaseURL. 
                                                                                                                                                                               
However, since the controller already passes base_url from ModelConfig to the Python ADK explicitly (via KAgentAnthropicLlm(base_url=...)), injecting ANTHROPIC_BASE_URL as an env var into the pod is indeed redundant for the K8s deployment path. The env var fallback on the Python side is still valuable for local development, but that doesn't require controller involvement.                                                                                                                                              
                                                                                                                                                                               
I'll remove the controller-side changes (adk_api_translator.go, providers.go, and the corresponding Go test) and keep only the Python ADK changes. Will push shortly.

dongjiang1989: @EItanya  revert the controller-side changes and keep only the Python ADK changes.

python/packages/kagent-adk/src/kagent/adk/_a2a.py
```diff
@@ -52,6 +52,19 @@ def thread_dump(request: Request) -> PlainTextResponse:
 kagent_url_override = os.getenv("KAGENT_URL")
 
 
+def _configure_anthropic_client() -> None:
```
supreme-gg-gg, line 55: Is there a reason why this function is needed? Since ANTHROPIC_BASE_URL is only used in BYO cases, perhaps you can just do the logging elsewhere in your custom code?

python/packages/kagent-adk/src/kagent/adk/models/_anthropic.py
supreme-gg-gg: Makes sense to me as well 👍

PR #1905, opened 2026-05-21
RemoteMCPServer TLS + companion-Secret API + plaintext egress rewrite

iplay88keys:
# Description

Adds TLS configuration to `RemoteMCPServer` so declarative agents can connect to MCP servers behind private CAs, self-signed certs, or corporate internal CAs. Mirrors the existing `ModelConfig.spec.tls` shape so the operator-facing surface is consistent across resource types.

```yaml
kind: RemoteMCPServer
metadata:
  name: corp-mcp
  namespace: kagent
spec:
  description: Corporate-CA MCP server
  url: https://mcp.corp.internal/mcp
  tls:
    caCertSecretRef: corp-ca
    caCertSecretKey: ca.crt
```

`https://` URLs that the agent's system trust already covers don't need `spec.tls` — omit the block and TLS origination uses Go's default trust store. The CRD only rejects the contradictory `http://` + `spec.tls` combo.

```yaml
spec:
  url: https://api.example.com/mcp   # public CA, no spec.tls block needed
```

The companion egress URL rewrite (details in the `core/pkg/egress` bullet below) is opt-in via a new helm value; off by default. Enable when MCP traffic should egress to a TLS-terminating proxy:

```yaml
controller:
  mcpEgressPlaintext: true
```

## Egress URL-rewrite contract

| # | `spec.url` | `spec.tls` | CRD | `EffectiveScheme` | `EffectivePort` | `RewriteDialURL` (egress on) |
|---|---|---|---|---|---|---|
| 1 | `host/path` (scheme-less) | nil | ✅ | `http` | 80 | `http://host:80/path` |
| 2 | `host:1234/path` | nil | ✅ | `http` | 1234 | `http://host:1234/path` |
| 3 | `host/path` | `{}` | ✅ | `https` | 443 | `http://host:443/path` |
| 4 | `host/path` | non-empty | ✅ | `https` | 443 | `http://host:443/path` |
| 5 | `host:1234/path` | `{}` or non-empty | ✅ | `https` | 1234 | `http://host:1234/path` |
| 6 | `http://host/path` | nil | ✅ | `http` | 80 | `http://host:80/path` |
| 7 | `http://host:8080/path` | nil | ✅ | `http` | 8080 | `http://host:8080/path` |
| 8 | `http://host/path` | `{}` or non-empty | 🛑 | — | — | (CRD rejects) |
| 9 | `https://host/path` | nil | ✅ | `https` | 443 | `http://host:443/path` |
| 10 | `https://host/path` | `{}` | ✅ | `https` | 443 | `http://host:443/path` |
| 11 | `https://host/path` | non-empty | ✅ | `https` | 443 | `http://host:443/path` |
| 12 | `https://host:8443/path` | `{}` or non-empty | ✅ | `https` | 8443 | `http://host:8443/path` |

Invariants the table pins:
- `spec.tls != nil` is the TLS opt-in signal (even an empty `{}` counts); an explicit `https://` URL is a parallel signal. Either one is sufficient.
- Only one combination is admission-rejected: `http://` URL + non-nil `spec.tls` (row 8). `https://` with no `spec.tls` (row 9) is admitted and defaults to system trust.
- `RewriteDialURL` resolves every `spec.url` to the same plaintext URL — `http://host:<effective-port>/<path>` — using the tls-aware effective port. The agent's tool URL (via `RewriteConfigForAgent`) and the controller's tool-discovery probe dial an identical endpoint.

Things to note:

- `RemoteMCPServer.spec.tls *TLSConfig` reuses the `TLSConfig` from `ModelConfig`. The `caCertSecretRef`/`caCertSecretKey`/`disableSystemCAs` CEL validation rules were lifted onto the `TLSConfig` type itself so every consumer (ModelConfig today, RemoteMCPServer in this PR, future resources) gets identical admission validation without re-declaring rules per spec.
- CRD admission on `RemoteMCPServerSpec`: `spec.tls` must be unset when `spec.url` carries the `http://` scheme (a TLS opinion contradicts a plaintext URL). `https://` with no `spec.tls` is admitted and defaults to system trust.
- The controller reads `RemoteMCPServer.spec.tls` so that listing tools with a custom CA cert works. The tool-discovery dial uses the same TLS config the agent will use at runtime.
- `RemoteMCPServer.status.secretHash` captures the bytes of the referenced CA Secret so the agent translator can fold it into the agent config hash — in-place rotations (cert-manager updating bytes under the same Secret name) flip the agent config hash and roll the pod.
- The RemoteMCPServer controller watches Secrets (mirrors the existing ModelConfig controller pattern): an in-place CA rotation triggers an immediate RMS reconcile via reverse-lookup, not the 60s `RequeueAfter` tick. `GenerationChangedPredicate` on `.For()` filters out the controller's own Status writes.
- Added `core/pkg/egress` package, gated by the `--mcp-egress-plaintext` flag. When enabled, two entry points resolve a `RemoteMCPServer.spec.url` to the same plaintext URL: `RewriteConfigForAgent` (called in-line by the agent translator before the config Secret is serialized, so the existing config-hash captures the rewrite and rolls the pod on toggle) and `RewriteDialURL` (used by the controller's tool-discovery probe). Both use the RMS's tls-aware effective port so the agent and the controller dial an identical endpoint.
- `addTLSConfiguration` mounts a per-Secret read-only volume at `/etc/ssl/certs/custom/<id>/<key>`, where `<id>` is the Secret name verbatim when it fits the DNS-1123-label grammar and length, or a short SHA-256 prefix otherwise (cert-manager-style names with dots, long names). Previously a single fixed-name volume silently collided when multiple TLS sources referenced different Secrets on the same agent.
- `POST /api/toolservers` accepts an optional `secrets []SecretMaterial` array so operators can create an RMS or MCPServer and its companion Secrets in a single round-trip. Mirrors the existing `ModelConfig` inline-Secret pattern.
- Companion-Secret writes use generic `createOrUpdateCompanionSecrets` + `isOwnedBy` helpers (no longer ModelConfig-specific). OwnerRef matching now requires non-empty UID — safer than the old wildcard match against a stale-UID Secret left over from a delete-recreate.
- All three Create paths (`HandleCreateModelConfig`, `handleCreateRemoteMCPServer`, `handleCreateMCPServer`) close their partial-failure window: a companion-Secret failure rolls back the owner CR via `rollbackOwnerOnCompanionSecretFailure`, so the operator's retry sees a clean slate instead of an `AlreadyExists` on a half-created owner.
- `HandleUpdateModelConfig` similarly closes its partial-failure window with a reorder (companion Secrets written before the Spec Update) plus a sweep pass: after a successful Spec Update, companion Secrets that the prior Spec referenced but the new Spec doesn't get deleted **only if** they carry an OwnerRef back to this ModelConfig. Externally-managed Secrets are left untouched. Cleans up the rename case (`Spec.TLS.CACertSecretRef: ca-v1 → ca-v2`) so old Secrets don't accumulate.

# Out of scope

- mTLS (client certs to upstream).
- ConfigMap-sourced CA bundles.
- UI form fields for `spec.tls` — the TypeScript type binding is in this PR; rendering the form is a UI follow-up.
- Automatic cleanup of companion Secrets created via earlier PUT requests that the UI never surfaces (no GET endpoint exposes existing companion Secrets, so the sweep is intentionally scoped to "Secrets the prior Spec referenced and the new Spec doesn't" — full PUT-replace semantics on `req.Secrets` would silently delete Secrets the UI didn't know about).


go/api/config/crd/bases/kagent.dev_remotemcpservers.yaml
```diff
@@ -179,6 +179,63 @@ spec:
...
+                      This field follows the same pattern as APIKeySecretKey.
+                      Required when CACertSecretRef is set (unless DisableVerify is true).
+                    type: string
+                  caCertSecretRef:
+                    description: |-
+                      CACertSecretRef is a reference to a Kubernetes Secret containing
+                      CA certificate(s) in PEM format. The Secret must be in the same
+                      namespace as the ModelConfig.
+                      When set, the certificate will be used to verify the provider's SSL certificate.
+                      This field follows the same pattern as APIKeySecret.
```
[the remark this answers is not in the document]
iplay88keys: Fixed

helm/kagent-crds/templates/kagent.dev_remotemcpservers.yaml
```diff
@@ -179,6 +179,63 @@ spec:
...
+                    description: |-
+                      CACertSecretKey is the key within the Secret that contains the CA certificate data.
+                      This field follows the same pattern as APIKeySecretKey.
+                      Required when CACertSecretRef is set (unless DisableVerify is true).
+                    type: string
+                  caCertSecretRef:
+                    description: |-
+                      CACertSecretRef is a reference to a Kubernetes Secret containing
+                      CA certificate(s) in PEM format. The Secret must be in the same
+                      namespace as the ModelConfig.
```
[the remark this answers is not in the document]
iplay88keys: Fixed

go/core/internal/httpserver/handlers/toolservers.go
```diff
@@ -150,13 +176,19 @@ func (h *ToolServersHandler) handleCreateRemoteMCPServer(w ErrorResponseWriter,
 		return
 	}
 
+	if err := createOrUpdateCompanionSecrets(r.Context(), h.KubeClient, toolServerRequest, remoteMCPServerGVK, secrets); err != nil {
```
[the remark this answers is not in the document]
iplay88keys: Fixed

go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -72,7 +71,17 @@ func (a *adkApiTranslator) BuildManifest(
...
+	// hash is computed and stamped after plugins run (see below): a
+	// plugin that mutates the config Secret needs the hash to reflect
+	// the mutation so pods roll on the next reconcile. buildPodTemplate
+	// initializes the Annotations map by reference; the post-plugin
+	// stamp writes into that same map, which workloadObjects share via
+	// Go's map-reference semantics — so the stamp propagates to
+	// whichever workload type buildWorkloadObjects produced (Deployment
+	// in normal mode, Sandbox in sandbox mode) without the translator
+	// needing to know about the workload's concrete type.
+	podTemplate := buildPodTemplate(manifestCtx, podRuntime, 0)
```
EItanya: Why do we even need to pass this in if it's always zero, should we just return it. Also what's the value of any pre-plugin hashing if the plugin can change anything it wants

go/core/pkg/translator/adk_api_translator_types.go
```diff
@@ -20,3 +20,21 @@ type TranslatorPlugin interface {
...
+//
+// The hook controls the URL only. TLS configuration is built from
+// spec.tls independently and applied to the underlying http.Client.
+// Implementations that change scheme (e.g. https:// → http:// to route
+// through a destination that handles TLS elsewhere) must ensure
+// spec.tls remains consistent with the new dial target. The rewriter
+// currently cannot supply TLS config of its own.
+type RemoteMCPServerURLRewriter interface {
+	RewriteRemoteMCPServerURL(ctx context.Context, remoteMCPServer *v1alpha2.RemoteMCPServer) (string, error)
+}
```
EItanya: Can we just add this to the above plugin instead? Or is this a reconciler plugin?

go/core/pkg/sandboxbackend/backend.go
```diff
@@ -25,4 +25,11 @@ type Backend interface {
 
 	// ComputeReady reflects implementation-specific status into condition pieces for Agent.status.
 	ComputeReady(ctx context.Context, cl client.Client, nn types.NamespacedName) (status metav1.ConditionStatus, reason, message string)
+
+	// StampPodTemplateAnnotation sets the given annotation on the
+	// backend's workload pod template, returning true if obj is a kind
+	// the backend produces. Used by the agent translator to stamp the
+	// post-plugin config-hash without needing to import the backend's
+	// concrete CRD types.
+	StampPodTemplateAnnotation(obj client.Object, key, value string) bool
```
EItanya: This seems to leak implementation details, and related to my other comment about doing all of the hashing after the translators/plugins

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -289,15 +294,16 @@ type TLSConfig struct {
...
+	// When set, the certificate will be used to verify the upstream's
+	// SSL certificate.
 	// +optional
 	CACertSecretRef string `json:"caCertSecretRef,omitempty"`
 
-	// CACertSecretKey is the key within the Secret that contains the CA certificate data.
-	// This field follows the same pattern as APIKeySecretKey.
-	// Required when CACertSecretRef is set (unless DisableVerify is true).
+	// CACertSecretKey is the key within the Secret that contains the
+	// CA certificate data (PEM-encoded). Required when CACertSecretRef
```
EItanya, line 305: Why is this required, can't we infer a single key if there is one? I see it's required today but not enforced via `cel`, so I suppose this is just a cleaner UX given the current constraint.

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -274,9 +274,14 @@ type SAPAICoreConfig struct {
...
-// that use self-signed certificates or custom certificate authorities.
+// TLSConfig contains TLS/SSL configuration options for outbound HTTPS
+// connections from the agent (model provider, RemoteMCPServer). The
+// XValidation rules below apply at admission to every CRD field that
+// uses TLSConfig, so callers don't need to re-declare them per spec.
+//
+// +kubebuilder:validation:XValidation:message="caCertSecretKey requires caCertSecretRef",rule="!(has(self.caCertSecretKey) && size(self.caCertSecretKey) > 0 && (!has(self.caCertSecretRef) || size(self.caCertSecretRef) == 0))"
+// +kubebuilder:validation:XValidation:message="caCertSecretRef requires caCertSecretKey",rule="!(has(self.caCertSecretRef) && size(self.caCertSecretRef) > 0 && (!has(self.caCertSecretKey) || size(self.caCertSecretKey) == 0))"
+// +kubebuilder:validation:XValidation:message="disableSystemCAs requires caCertSecretRef or disableVerify (trust-nothing config rejects every upstream)",rule="!(has(self.disableSystemCAs) && self.disableSystemCAs && (!has(self.disableVerify) || !self.disableVerify) && (!has(self.caCertSecretRef) || size(self.caCertSecretRef) == 0))"
```
jmhbh, line 284: if we have tests around this it would be good to validate that these changes are still consistent with the validation rules they replace in line 328-330

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -1028,35 +1092,125 @@ func (a *kagentReconciler) createMcpTransport(ctx context.Context, s *v1alpha2.R
...
+// disableSystemCAs trust-only-the-named-bundle — so tool discovery
+// trusts the same upstream chain the agent will trust at runtime.
+func (a *kagentReconciler) buildRemoteMCPServerTLSConfig(ctx context.Context, s *v1alpha2.RemoteMCPServer) (*tls.Config, error) {
+	tlsSpec := s.Spec.TLS
+	if tlsSpec.IsEmpty() {
+		return nil, nil
+	}
+
+	cfg := &tls.Config{
+		InsecureSkipVerify: tlsSpec.DisableVerify, //nolint:gosec // operator-authored test-fixture escape hatch
```
EItanya: What does this comment mean?

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -90,23 +93,30 @@ type kagentReconciler struct {
 	watchedNamespaces []string
 
 	sandboxBackend sandboxbackend.Backend
+
+	// rmsURLRewriter optionally transforms the URL the controller dials
+	// when discovering tools on a RemoteMCPServer. Nil means dial
+	// s.Spec.URL verbatim. See pkgtranslator.RemoteMCPServerURLRewriter.
+	rmsURLRewriter pkgtranslator.RemoteMCPServerURLRewriter
```
jmhbh: nit: prefer `remoteMCPServerURLRewriter` here instead. We use the fully qualified name elsewhere in our codebase so keeping that consistent would be good.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -933,7 +986,33 @@ func (a *adkApiTranslator) translateRemoteMCPServerTarget(ctx context.Context, a
...
 		})
 	}
-	return nil
+	// Mount the CA Secret on the agent pod when the RemoteMCPServer pins a TLS bundle.
+	// Converters that synthesize RMSs from in-cluster MCPServer/Service
+	// references don't set Spec.TLS, so this is a no-op for those. Returns
+	// the controller-resolved TLS Secret hash so callers can mix it into
+	// the agent's config hash — that's the signal that drives a rollout
+	// when the CA Secret rotates in place (same Secret name, new PEM).
+	if mdd != nil {
```
EItanya: How would mdd be nil?

go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -72,7 +71,12 @@ func (a *adkApiTranslator) BuildManifest(
 		return nil, err
 	}
 
-	podTemplate := buildPodTemplate(manifestCtx, podRuntime, configSecret.configHash)
+	// Build the pod template with a placeholder config-hash. The real
+	// hash is computed and stamped onto each workload's pod template
+	// after plugins run (see below): a plugin that mutates the config
+	// Secret needs the hash to reflect the mutation so pods roll on the
+	// next reconcile.
+	podTemplate := buildPodTemplate(manifestCtx, podRuntime, 0)
```
EItanya: it feels wrong to me that this takes in zero rather than potentially returning the hash if it's creating it, does that make sense? Unless of course there are other callers I'm missing.

go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -89,7 +93,44 @@ func (a *adkApiTranslator) BuildManifest(
...
+	return outputs, nil
+}
+
+// stampConfigHashOnWorkloads walks the translated manifest and sets the
+// kagent.dev/config-hash annotation on every workload object's pod
+// template. Deployments are stamped here directly; sandbox-backend kinds
+// delegate to the backend so this file stays independent of the
+// agent-sandbox CRD type.
+func (a *adkApiTranslator) stampConfigHashOnWorkloads(manifest []client.Object, configHash uint64) {
+	const key = "kagent.dev/config-hash"
```
EItanya: Can we make this annotation a const for clarity

go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -89,7 +93,44 @@ func (a *adkApiTranslator) BuildManifest(
...
+	value := fmt.Sprintf("%d", configHash)
+	for _, obj := range manifest {
+		if d, ok := obj.(*appsv1.Deployment); ok {
+			if d.Spec.Template.Annotations == nil {
+				d.Spec.Template.Annotations = map[string]string{}
+			}
+			d.Spec.Template.Annotations[key] = value
+			continue
+		}
+		if a.sandboxBackend != nil && a.sandboxBackend.StampPodTemplateAnnotation(obj, key, value) {
```
EItanya: This code is seriously smelling to me right now and I can't quite put my finger on why. I think maybe it's because there's an interaction here at all. It seems to me that the translator should have zero knowledge of a "SandboxBackend". Rather the translator should be a pure function which may be called by either the deployment translator or the sandbox translator. Does that make sense?

EItanya (commented):
Initial review. functionality makes complete sense, but there are a few code smells I'd like to try and work out

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -1028,35 +1092,125 @@ func (a *kagentReconciler) createMcpTransport(ctx context.Context, s *v1alpha2.R
...
+// trusts the same upstream chain the agent will trust at runtime.
+func (a *kagentReconciler) buildRemoteMCPServerTLSConfig(ctx context.Context, s *v1alpha2.RemoteMCPServer) (*tls.Config, error) {
+	tlsSpec := s.Spec.TLS
+	if tlsSpec.IsEmpty() {
+		return nil, nil
+	}
+
+	cfg := &tls.Config{
+		InsecureSkipVerify: tlsSpec.DisableVerify, //nolint:gosec // operator-authored test-fixture escape hatch
+	}
```
jmhbh, line 1151: configuring this here seems like a potential code smell. disableVerify should be configured explictly by the user

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -1028,35 +1092,125 @@ func (a *kagentReconciler) createMcpTransport(ctx context.Context, s *v1alpha2.R
...
+		// TLS config. If the default transport isn't *http.Transport
+		// (extremely unusual — only happens if something earlier swapped
+		// it out), fall back to a fresh transport so TLS still applies.
+		if t, ok := http.DefaultTransport.(*http.Transport); ok {
+			clone := t.Clone()
+			clone.TLSClientConfig = tlsConfig
+			base = clone
+		} else {
+			base = &http.Transport{TLSClientConfig: tlsConfig}
+		}
```
jmhbh: this seems to be dead code. `http.DefaultTransport.(*http.Transport)` will always evaluate to true

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -1028,35 +1092,125 @@ func (a *kagentReconciler) createMcpTransport(ctx context.Context, s *v1alpha2.R
...
+	if tlsConfig != nil {
+		// Clone the default transport to preserve its dial/keepalive
+		// settings (proxies, dual-stack, HTTP/2) and override only the
+		// TLS config. If the default transport isn't *http.Transport
+		// (extremely unusual — only happens if something earlier swapped
+		// it out), fall back to a fresh transport so TLS still applies.
+		if t, ok := http.DefaultTransport.(*http.Transport); ok {
+			clone := t.Clone()
+			clone.TLSClientConfig = tlsConfig
+			base = clone
```
jmhbh: this cloning and type assertion seems unnecessary to me. We don't ingest a transport in our arguments so we are deep cloning a default from the http library. We can just init a base transport here with the desired settings instead.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -217,56 +217,108 @@ func (r *adkApiTranslator) GetOwnedResourceTypes() []client.Object {
...
+	}
+	insecureSkipVerify = &tlsConfig.DisableVerify
+	disableSystemCAs = &tlsConfig.DisableSystemCAs
 	if tlsConfig.CACertSecretRef != "" && tlsConfig.CACertSecretKey != "" {
-		certPath := fmt.Sprintf("%s/%s", tlsCACertMountPath, tlsConfig.CACertSecretKey)
-		baseModel.TLSCACertPath = &certPath
+		_, _, p := tlsCAPaths(tlsConfig.CACertSecretRef, tlsConfig.CACertSecretKey)
+		caCertPath = &p
 	}
+	return
```
jmhbh: nit: prefer explictly reteturning `insecureSkipVerify, disableSystemCAs, caCertPath` instead of doing named returns

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -217,56 +217,108 @@ func (r *adkApiTranslator) GetOwnedResourceTypes() []client.Object {
...
+		}
+
 		modelDeploymentData.Volumes = append(modelDeploymentData.Volumes, corev1.Volume{
-			Name: tlsCACertVolumeName,
+			Name: volumeName,
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
 					SecretName:  tlsConfig.CACertSecretRef,
-					DefaultMode: new(int32(0444)), // Read-only for all users
+					DefaultMode: new(int32(0444)),
```
jmhbh: nit: would be good to keep the comment

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -749,7 +800,7 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
 			AuthUrl:       model.Spec.SAPAICore.AuthURL,
 		}
 
-		populateTLSFields(&sapAICore.BaseModel, model.Spec.TLS)
+		sapAICore.TLSInsecureSkipVerify, sapAICore.TLSCACertPath, sapAICore.TLSDisableSystemCAs = deriveTLSFields(model.Spec.TLS)
```
jmhbh: nit: given that all baseModels use the same fields `TLSInsecureSkipVerify` , `TLSCACertPath` etc. feels like we are being overly verbose in our approach to replace `populateTLSFields`. We may be able to keep `populateTLSFields` instead and wrap any additional logic into that function.

go/core/internal/controller/translator/agent/manifest_builder.go
iplay88keys: Removed as there is no post-plugin recompute anymore. This has been reworked so that the plugin handles adding a separate annotation which can trigger pod rollout.

go/core/pkg/sandboxbackend/backend.go
iplay88keys: This no longer exists as there is no longer any post-plugin hashing done. The plugin is in charge of changing annotations to result in pod rollout.

go/core/internal/controller/translator/agent/manifest_builder.go
iplay88keys: See the [other comment](https://github.com/kagent-dev/kagent/pull/1905/changes/BASE..c7707a2d8cce2993a3f50c22af3a269a50ab9bdf#r3307105430), no longer takes in 0 as we no longer re-compute the hash after processing plugins.

go/core/internal/controller/translator/agent/manifest_builder.go
iplay88keys: The annotation aspect of this is gone, but the sandbox backend smell predates this work. I believe that unwinding that should be tackled separately.

go/core/pkg/translator/adk_api_translator_types.go
iplay88keys: No longer using a plugin for this

go/api/v1alpha2/modelconfig_types.go
iplay88keys: We can definitely infer if there is only one key. This just keeps with the existing modelconfig approach for now.

Also, this is more extensible. We can make sure that there's a default and it's not required, but I think it's just beneficial to have the ability to specify it here. For example, cert-manager-generated TLS secrets have `tls.crt`, `tls.key`, and `ca.crt`, so there is precedent for a secret with multiple keys and the right one would need to be fetched.

go/core/internal/controller/reconciler/reconciler.go
iplay88keys: Updated

go/core/internal/controller/reconciler/reconciler.go
iplay88keys: No longer a plugin

go/core/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Resolved, this guard is no longer there. The call site now invokes `addTLSConfiguration(mdd, remoteMcpServer.Spec.TLS)` directly. `addTLSConfiguration`  short-circuits on `tlsConfig == nil`, and all callers of `translateRemoteMCPServerTarget` pass a non-nil `mdd`

go/api/v1alpha2/modelconfig_types.go
iplay88keys: This should be semantically equivalent to what was there before. There are no admission-level tests before, so I will add some envtest assertions around this.

go/core/internal/controller/reconciler/reconciler.go
iplay88keys: The text here has been updated to be more clear:
```
InsecureSkipVerify: tlsSpec.DisableVerify, //nolint:gosec // G402: explicit user opt-in via spec.tls.disableVerify
```

go/core/internal/controller/reconciler/reconciler.go
iplay88keys: The clone preserves the configured defaults on http.DefaultTransport. We'd lose them if we built a fresh `*http.Transport` with just `TLSClientConfig` set:
- Proxy: `ProxyFromEnvironment` (honors HTTP_PROXY/HTTPS_PROXY)
- 30s dial timeout + 30s keepalive via `DialContext`
- `ForceAttemptHTTP2: true`
- `MaxIdleConns: 100`, `IdleConnTimeout: 90s` (connection pooling)
- `TLSHandshakeTimeout: 10s`: the safety-relevant one: a misconfigured upstream that accepts TCP but never completes the TLS handshake would hang this reconcile forever without it.

`http.DefaultTransport` is declared as `http.RoundTripper` (interface) but `.Clone()` is a method on `*http.Transport`. The assertion is required by the type system to call `.Clone()`.

go/core/internal/controller/reconciler/reconciler.go
iplay88keys: This has been removed.

go/core/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Updating -- there are two places where it has to be different (`StreamableHTTPConnectionParams`, `SseConnectionParams`) as they don't embed `BaseModel`, but the others can use `populateTLSFields`.

go/core/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Fixed

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -289,15 +294,16 @@ type TLSConfig struct {
...
+	// SSL certificate.
 	// +optional
 	CACertSecretRef string `json:"caCertSecretRef,omitempty"`
 
-	// CACertSecretKey is the key within the Secret that contains the CA certificate data.
-	// This field follows the same pattern as APIKeySecretKey.
-	// Required when CACertSecretRef is set (unless DisableVerify is true).
+	// CACertSecretKey is the key within the Secret that contains the
+	// CA certificate data (PEM-encoded). Required when CACertSecretRef
+	// is set (unless DisableVerify is true).
```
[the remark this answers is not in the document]
iplay88keys: Fixed

go/core/pkg/egress/egress_test.go
```diff
@@ -0,0 +1,375 @@
...
+		// Scheme-less is rewritten just like the agent path.
+		{"scheme-less with explicit port rewritten", rms("host.docker.internal:13443/mcp"), "http://host.docker.internal:13443/mcp"},
+		{"scheme-less no port no tls defaults to 80", rms("svc.ns/mcp"), "http://svc.ns:80/mcp"},
+		// Per CRD-validated contract, spec.tls != nil signals TLS opt-in (even
+		// the empty struct {}); the previously-divergent shape (scheme-less,
+		// port-less, TLS-backed) resolves to :443 on both paths. The combos
+		// that would be contradictory (http://+tls, https://+empty/nil-tls)
+		// are rejected at admission; the runtime cases below cover both the
+		// well-formed shapes that reach it and the would-be-rejected ones
+		// (kept as defensive coverage if a webhook is bypassed).
```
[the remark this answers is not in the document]
iplay88keys: Fixed

helm/kagent-crds/templates/kagent.dev_remotemcpservers.yaml
```diff
@@ -179,13 +179,81 @@ spec:
...
             required:
             - description
             - url
             type: object
+            x-kubernetes-validations:
+            - message: 'spec.tls must be unset when spec.url has http:// scheme: a
+                TLS opinion contradicts a plaintext URL. Either drop spec.tls, or
+                use https:// / a scheme-less URL.'
+              rule: '!self.url.startsWith(''http://'') || !has(self.tls)'
           status:
```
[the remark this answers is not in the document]
iplay88keys: Fixed

go/core/pkg/egress/egress_test.go
```diff
@@ -0,0 +1,375 @@
...
+		{"http url no tls", rms("http://svc.ns/mcp"), "http"},
+		{"scheme-less no tls", rms("svc.ns/mcp"), "http"},
+		{"non-empty tls + http url", rmsWith("http://svc/mcp", nonEmptyTLS), "https"},
+		{"non-empty tls + scheme-less", rmsWith("svc/mcp", nonEmptyTLS), "https"},
+		{"non-empty tls + DisableVerify only", rmsWith("svc/mcp", &v1alpha2.TLSConfig{DisableVerify: true}), "https"},
+		// Per CRD-validated contract, spec.tls != nil ⇒ TLS opt-in, even when
+		// the struct has no fields set. The combos CRD rejects (http://+tls,
+		// https://+nil-tls, https://+empty-tls) are kept as defensive runtime
+		// coverage; the runtime's safer answer is "https" whenever either
+		// signal expresses TLS intent.
```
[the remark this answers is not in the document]
iplay88keys: Fixed

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -609,12 +642,16 @@ func (a *kagentReconciler) ReconcileKagentRemoteMCPServer(ctx context.Context, r
 	} else {
 		l.Info("successfully registered remote MCP server", "url", server.Spec.URL, "toolCount", len(tools), "duration", time.Since(start))
 	}
+	if secretErr != nil {
+		err = multierror.Append(err, secretErr)
+	}
```
EItanya, line 652: nit: why is this added after the upsert and not right when the error occurs?

go/core/pkg/egress/egress.go
```diff
@@ -0,0 +1,282 @@
...
+// share a hostname but differ in path or scheme don't collide. The only way to
+// reach a collision is to author two distinct RMS objects with byte-identical
+// spec.url strings, which is itself a misconfiguration of the RMS resources
+// rather than something the rewrite needs to police.
+//
+// Refs that fail to resolve (NotFound) are skipped — RMS presence is validated
+// upstream, so reaching this code with a missing RMS means a transient mismatch
+// that reconverges on the next reconcile. Other client errors propagate so the
+// caller re-tries.
+func CollectRMSHosts(ctx context.Context, c client.Client, agent v1alpha2.AgentObject) (map[string]string, error) {
```
EItanya: Why is this function public, seems to only be used here?

go/core/pkg/egress/egress.go
```diff
@@ -0,0 +1,282 @@
...
+// rejecting spec.tls on an http:// URL. Declared so any future Validate
+// entry point can return the same sentinel the admission rule would surface.
+var ErrTLSConflictsWithHTTPScheme = errors.New("spec.tls must be unset when spec.url has http:// scheme")
+
+// RewriteConfigForAgent rewrites every RemoteMCPServer-backed tool URL in cfg
+// to its plaintext form. Called inline by the agent translator's
+// config-phase (before the config Secret is serialized and the config-hash is
+// computed) when the OSS egress gate is enabled — registration is the gate,
+// so this function itself applies unconditionally. Skill-only / MCPServer-only
+// / Service-only agents resolve to an empty RMS set and are a no-op.
```
EItanya: Why do we do this in a loop rather than applying it a single RMS at a time?

go/core/internal/controller/reconciler/reconciler.go
iplay88keys: This is deliberate — secretErr must not short-circuit tool discovery (which has its own previously-discovered-tools fallback), and the status update needs both error streams folded together, so the merge happens at the one point where they converge. Added a comment making that explicit.

go/core/pkg/egress/egress.go
iplay88keys: Good catch, I have unexported it, along with `RewriteConfig`/`rewriteHTTPSForEgress` which had the same issue, and dropped the unused error sentinel. The package now exports just the two entry points plus the `spec.url-resolution` helpers.

go/core/pkg/egress/egress.go
iplay88keys: The rewrite is deliberately an allowlist with exact-match misses, and the misses carry meaning. The key case: when `--proxy-url` is set, `applyProxyURL` has already redirected in-cluster RMS tool URLs to the global proxy host earlier in `CompileAgent`. Those entries no longer equal any RMS `spec.url`, so the map lookup misses and they're intentionally left alone. Proxy routing takes precedence for internal URLs, and only external upstreams (which the proxy never touches) still match and get the plaintext-egress rewrite.

The shape follows the input — the agent side asks "of these N flattened URLs, which are RMS-backed?" (set-membership → build the set once), while the dial side asks "rewrite this one RMS's URL" (no set needed).

iplay88keys: Refactoring this code

EItanya (approved):

PR #1907, opened 2026-05-22
fix(compaction): preserve namespaced tool names in LLM summarizer prompt

### LINKED ISSUE #1550: [BUG] Agents with "compaction" | Tools name error
### 📋 Prerequisites

- [ ] I am using the latest version of the software

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

As our user are likely to have longer session with Kagent, I wanted to enable the [compaction](https://kagent.dev/docs/kagent/resources/api-ref#contextcompressionconfig) feature. For testing purposes used the following configs:
```yaml
    context:
      compaction:
        compactionInterval: 5
        overlapSize: 2
        summarizer:
          modelConfig: bedrock-model-config
```
This is the _orchestrator_ agent in multi-agent architecture.
Howerver, after the "compaction" of the context is triggered, the Agent seems to forget the sub-agent available, to use as tools and throws an error:
```text
Tool 'agent_name_aa' not found.\nAvailable tools: datetime_get_current_time, kagent__NS__agent_name_aa, kagent__NS__agent_name_bb ....
```

### 🔄 Steps To Reproduce

1. create an Agent
2. enable the compaction config 
3. trigger the compaction, e.g. based on the configs above, asking >5 promts
4. when propt a task that requires delegating to other agents, error is thrown

### 🤔 Expected Behavior

The `compaction` is effectively reducing the context size, by summarizing it.
The Agent acting as orchestrator, should continue its work normally, even after the compaction. 
It should not "forget" the available sub-agent, used for delegeting tasks.

### 📱 Actual Behavior

An error is thrown, and the task is not delegated to appropriate sub-agent. 
Seems like the names of the sub-agent are different from what's expected.

### 💻 Environment

- AWS EKS - eks.29 
- Kubernetes - 1.33
- Kagent: 0.7.23
- Model: Bedrock - Anthropic Claude Sonnet 4.5

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

yashrajshuklaaa:
after context compaction the LlmEventSummarizer rewrote conversation history using an LLM. 
also the default ADK prompt had no instruction to preserve exact tool names so the LLM abbreviated namespaced tool names like 'kagent__default__agent_name_aa' to just 'agent_name_aa'
when the orchestrator agent resumed after compaction it could not dispatch to the sub-agent because the short name does not match the registered tool name

Fix: inject a kagent-specific default prompt_template that explicitly instructs the summarizer LLM to preserve exact tool/agent names including namespace prefixes.users who supply a custom prompt_template are unaffected (their value takes precedence via 'or')

Fixes #1550


EItanya (commented):
Thanks so much for the fix. What happens if the user does provide a prompt template, do we always want to append the bit about tool name correctness?

yashrajshuklaaa:
@EItanya great point
**two ways to handle this -**
Keep it as is , custom prompt overrides everything. User is responsible for preserving tool names in their prompt
or 
always append the tool name warning to whatever prompt is used , even custom ones

i prefer 2nd one cuz appending the warning is a small change but saves users from a confusing failure

EItanya:
> @EItanya great point **two ways to handle this -** Keep it as is , custom prompt overrides everything. User is responsible for preserving tool names in their prompt or always append the tool name warning to whatever prompt is used , even custom ones
> 
> i prefer 2nd one cuz appending the warning is a small change but saves users from a confusing failure

sounds good let's do it

yashrajshuklaaa:
@EItanya done 
PTAL

python/packages/kagent-adk/src/kagent/adk/types.py
```diff
@@ -615,6 +615,27 @@ def _create_llm_from_model_config(model_config: ModelUnion):
...
+    "The following is a conversation history between a user and an AI"
+    " agent. Please summarize the conversation, focusing on key"
+    " information and decisions made, as well as any unresolved"
+    " questions or tasks. The summary should be concise and capture the"
+    " essence of the interaction.\n\n"
+    "IMPORTANT: When referencing any tools or agents that were used,"
+    " you MUST preserve their exact registered names including any"
+    " namespace prefixes (for example: 'kagent__default__agent_name')."
+    " Never shorten, abbreviate, or rephrase tool or agent names.\n\n"
+    "{conversation_history}"
```
EItanya: nit: why is this doubled up, can't you just append the above string

yashrajshuklaaa:
fixed @EItanya 
sorry for the mistake

EItanya (approved):

yashrajshuklaaa:
@EItanya is it okay to merge ?

PR #1912, opened 2026-05-25
fix: address PR review comments

RustyCoderX:
Addressed the review feedback by wiring `clusterDomain` into the actual runtime flow instead of leaving it unused, tightening internal `.svc` URL detection to avoid matching external domains incorrectly, fixing the related comments/documentation, restoring the Grafana MCP default service URL, and adding regression tests for cluster-domain-aware URL handling. Also propagated the config through the HTTP server and translator paths so custom cluster domains are consistently supported end-to-end

### DIFF
--- a/go/core/internal/controller/translator/agent/adk_api_translator.go
+++ b/go/core/internal/controller/translator/agent/adk_api_translator.go
@@ -33,6 +33,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -130,6 +131,7 @@ type AdkApiTranslator interface {
 		inputs *AgentManifestInputs,
 	) (*AgentOutputs, error)
 	GetOwnedResourceTypes() []client.Object
+	WithClusterDomain(clusterDomain string) AdkApiTranslator
 }
 
 // probeConfig holds readiness probe timing configuration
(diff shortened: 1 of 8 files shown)


go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -177,16 +179,30 @@ func NewAdkApiTranslatorWithWatchedNamespaces(kube client.Client, watchedNamespa
...
+func (a *adkApiTranslator) WithClusterDomain(clusterDomain string) AdkApiTranslator {
+	if a == nil {
+		return a
+	}
+	clusterDomain = strings.TrimSpace(clusterDomain)
+	if clusterDomain == "" {
+		clusterDomain = "cluster.local"
+	}
+	a.clusterDomain = clusterDomain
+	return a
```
jmhbh, line 196: nit: prefer using `a != nil` instead for slightly better readability
```suggestion
	if a != nil {
	   	clusterDomain = strings.TrimSpace(clusterDomain)
	    if clusterDomain == "" {
		  clusterDomain = "cluster.local"
	    }
	    a.clusterDomain = clusterDomain
	}
	return a
```

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -953,34 +969,68 @@ func (a *adkApiTranslator) isInternalK8sURL(ctx context.Context, urlStr, namespa
...
+			return false
+		}
+		if len(validation.IsDNS1123Label(parts[0])) > 0 || len(validation.IsDNS1123Label(parts[1])) > 0 {
+			return false
+		}
 		return true
 	}
 
-	// Extract namespace from hostname pattern: {name}.{namespace}
-	// Examples: test-mcp-server.kagent -> namespace is "kagent"
```
jmhbh, line 962: nit: can we keep this comment? this seems to still be relevant so it would be good context.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -953,34 +969,68 @@ func (a *adkApiTranslator) isInternalK8sURL(ctx context.Context, urlStr, namespa
...
+		if len(validation.IsDNS1123Label(service)) > 0 || len(validation.IsDNS1123Label(potentialNamespace)) > 0 {
+			return false
 		}
-		// Controller is using namespaced RBAC, so check if the namespace is watched
-		if (apierrors.IsForbidden(err) || apierrors.IsUnauthorized(err)) && len(a.watchedNamespaces) > 0 {
-			return slices.Contains(a.watchedNamespaces, potentialNamespace)
+		return a.namespaceExistsOrWatched(ctx, potentialNamespace)
+	}
+
+	if len(parts) == 3 && parts[2] == "svc" {
```
jmhbh, line 1009: nit: would be helpful to add a comment that this conditional specifically addressesing a hostname using the `grafana.kagent.svc` construction

PR #1921, opened 2026-05-25
Migrate from A2A v0 to v1 (part 1)

supreme-gg-gg:
The 1st of 3 PRs for A2A v0 to v1 zero downtime migration. This will go into release 0.10.0

Makes the controller compatible with A2A v0 (trpc-a2a-go) and v1 (a2a-go) data and both v0 and v1 JSON-RPC routes / agent card routes depending on `A2A-Version` header (designed explicitly by the protocol for version negotiations). Replaces **all** internal usages of `trpc-a2a-go` with `a2a-go` except in data conversion code and CLI client code.

This ensure that the controller is compatible with old / existing agents and UI in A2A v0 and existing v0 data in db, while adding the capability to communicate with v1 agents and UI during upgrade as they are added in the next part.

Follows https://a2a-protocol.org/latest/whats-new-v1/#migration-strategy-recommendations and db conversion reuses as much of existing a2a backward compatible code as possible.


go/core/internal/a2a/trace.go
supreme-gg-gg, line 1: For reviewer, tracing context injection and associated tests have been moved to `client_interceptors.go` and `client_interceptor_test.go` (also where auth headers injection happens)

go/core/internal/a2a/passthrough_handler.go
supreme-gg-gg, line 1: For reviewer, this is a modified version of existing https://github.com/kagent-dev/kagent/blob/main/go/core/internal/a2a/manager.go (removed) for very similar functionality 

This new passthrough handler is simpler and forwards RPC methods directly to the client and avoids having to go through things like task lifecycle, task status, etc. that the old `AgentExecutor` would do, and fits better since the controller is just a proxy to upstream agent runtime that handles the real executor.

supreme-gg-gg:
@copilot review

go/core/internal/a2a/a2a_registrar.go
```diff
@@ -42,9 +42,9 @@ func NewA2ARegistrar(
 	a2aBaseUrl string,
 	sandboxA2ABaseURL string,
 	authenticator auth.AuthProvider,
-	streamingMaxBuf int,
-	streamingInitialBuf int,
-	streamingTimeout time.Duration,
+	_ int,
+	_ int,
+	_ time.Duration,
```
EItanya: Does the new library not expose these? I remember these were very important to ensure larger streams/payloads weren't rejected

go/core/internal/a2a/client_interceptors.go
EItanya, line 1: Can you please add comments to this file to explain what these types are for

go/core/internal/controller/translator/agent/utils.go
```diff
@@ -2,38 +2,58 @@ package agent
...
+		Skills:             []a2atype.AgentSkill{},
 		DefaultInputModes:  []string{"text"},
 		DefaultOutputModes: []string{"text"},
 	}
 	if spec.Type == v1alpha2.AgentType_Declarative && spec.Declarative != nil && spec.Declarative.A2AConfig != nil {
-		decl := spec.Declarative
-		card.Skills = slices.Collect(utils.Map(slices.Values(decl.A2AConfig.Skills), func(skill v1alpha2.AgentSkill) server.AgentSkill {
-			return server.AgentSkill(skill)
-		}))
+		card.Skills = make([]a2atype.AgentSkill, 0, len(spec.Declarative.A2AConfig.Skills))
```
EItanya, line 38: Is there a TODO or work item to move our k8s API type away from trpc go? I guess for now we can just mirror it to make sure we don't break the marshalling

EItanya (commented):
These changes look awesome overall!

Can you potentially create a unified `TODO` marker for all changes which need to go in as a part of this migration so that it's clear when a change will be handled by a future PR, or needs to be. That way we can just do `ctrl + shift + f` and find all of them.

go/core/internal/controller/translator/agent/utils.go
supreme-gg-gg: Yes I've done that in #1936 (the 0.12.0 one), I'll move it here. I've removed the dependency of agents CRD on the a2a library

go/core/internal/a2a/a2a_registrar.go
supreme-gg-gg: The new library doesn't have direct options to set these, I think we can just use the default ones from the library, I found these:

```go
const defaultRequestTimeout = 3 * time.Minute

// MaxSSETokenSize is the maximum size for SSE data lines (10MB).
// The default bufio.Scanner buffer of 64KB is insufficient for large payloads
MaxSSETokenSize = 10 * 1024 * 1024 // 10MB
```

I'll remove these parameters

jmhbh (dismissed):

jmhbh (approved):

EItanya (approved):

PR #1925, opened 2026-05-26
feat(adk): propagate parent + root context_id headers on outbound A2A calls

### LINKED ISSUE #1924: [FEATURE] Propagate parent + root context_id headers on outbound A2A calls
### 📋 Prerequisites

### 📝 Feature Summary

Propagate two conversation-lineage HTTP headers (`x-kagent-parent-context-id` + `x-kagent-root-context-id`) on outbound `KAgentRemoteA2ATool` calls so a peer agent can correlate a turn with the originating chat conversation across a chain of A2A hops.

Today `KAgentRemoteA2ATool.run_async` ships the outbound A2A message with `context_id = self._last_context_id`, which is a `uuid4()` minted once per tool instance at construction (`_remote_a2a_tool.py` line 159). The receiving peer sees that opaque uuid as its own A2A `context_id` and has no way to correlate the turn back to the originating chat conversation.

`_SubagentInterceptor` only forwards two headers today: `x-user-id` and `x-kagent-source: agent` (`_remote_a2a_tool.py` lines 73-84). The receive side already copies all inbound headers into `session.state["headers"]` in `_agent_executor.py` lines 541-545 — so the *plumbing* to consume new headers exists; it is only the outbound-stamping side that lacks lineage data.

Concrete failure mode that motivated this issue: a chat-tier Declarative kagent Agent delegates to a router agent over A2A; the router spawns a per-conversation worker pod (e.g. a kubernetes-sigs/agent-sandbox `SandboxClaim` running `claude --print`) and wants the same pod to serve every subsequent turn of the same chat thread. Without a stable cross-hop identifier the router has to re-create the worker on every turn — because the tool's self-generated `context_id` is opaque to the chat-tier and changes the moment the chat-tier's tool list is rebuilt. The user-visible symptom is "my coding agent loses memory between turns" even though the kagent UI thread is the same.

Anyone building multi-hop kagent fleets that key state on a conversation id today either has to hand-roll a `header_provider` callback on every `KAgentRemoteA2ATool` instance or accept losing per-conversation continuity at every A2A hop. Both options are friction that should not exist for the obvious correlation case.

### 💡 Proposed Solution

Add two outbound HTTP headers, set automatically inside `KAgentRemoteA2ATool._build_call_context`:

```
x-kagent-parent-context-id  — immediate caller's session id (= the agent
                              that just ran this tool). Changes with every
                              hop.

x-kagent-root-context-id    — top-of-chain context_id. Forwarded unchanged
                              when an inbound `x-kagent-root-context-id`
                              header is present in this agent's session
                              state; otherwise set to the caller's own
                              session id (this agent is the chain root).
                              Stays stable across hops and across turns of
                              the same conversation.
```

Derivation logic (self-healing, no per-agent configuration needed):

```python
parent_context_id = tool_context.session.id
inbound_headers   = tool_context.session.state.get("headers", {}) or {}
root_context_id   = (
    inbound_headers.get("x-kagent-root-context-id")     # forwarded
    or inbound_headers.get("x-kagent-parent-context-id") # legacy fallback
    or parent_context_id                                  # we are root
)
```

Receiving agents already see every inbound HTTP header in `session.state["headers"]` (set by `A2aAgentExecutor.execute`), so no server-side plumbing change is required — a peer that wants to use the root context_id just reads `state["headers"]["x-kagent-root-context-id"]`.

Header names are exported as module-level constants (`PARENT_CONTEXT_ID_HEADER`, `ROOT_CONTEXT_ID_HEADER`) so downstream BYO agents can consume them by reference rather than by string literal.

The patch preserves backward compatibility:
- Existing `x-user-id` and `x-kagent-source` propagation is unchanged.
- Constructor-supplied `header_provider` callbacks still win — lineage headers are layered underneath the provider so a custom provider can override them when it really needs to.
- When the caller cannot resolve a session id (older `tool_context` shapes used in tests / unusual integration paths), no lineage headers are emitted — outbound request matches pre-feature behavior.

A draft PR with the implementation, unit-test coverage (root, mid-chain, legacy-inbound, missing session id, provider-override, end-to-end through `_SubagentInterceptor`) and a DCO sign-off is attached to this issue.

### 🔄 Alternatives Considered

1. **Per-tool `header_provider` callback** — already exists; works but every consumer has to write the same lineage-extraction code, and most consumers don't know they should. Friction tax on the common case.

2. **Reuse `context_id` field on `A2AMessage` itself** to carry root identity — would mis-use a field the a2a-sdk uses for per-call session correlation, breaks the abstraction.

3. **Add the data to `tool_context.session.state` (a session-state key)** — would only round-trip if the receiving agent runs in the same kagent process, which is not the case for cross-pod A2A calls.

4. **OTEL trace context propagation** — already there (`gen_ai.conversation.id` span attribute set in `_agent_executor.py`) but trace IDs are not exposed to agent-level code and are not a stable per-conversation identifier across UI turns.

The explicit-header approach is the smallest viable change that solves the correlation problem at the right layer, mirrors how `x-user-id` is already propagated, and stays inside the HTTP plane where downstream code already reads it.

### 🎯 Affected Service(s)

App Service

### 📚 Additional Context

- Receiving-side helper code already copies all inbound HTTP headers into `session.state["headers"]`, so this PR only changes the outbound stamping logic plus tests.
- Draft PR with full diff + unit tests: linked above.

Prefix:
## Summary

Stamps two conversation-lineage HTTP headers on every outbound `KAgentRemoteA2ATool` call so a remote peer can correlate the turn with the originating chat conversation across a chain of A2A hops.

- `x-kagent-parent-context-id` — immediate caller's session id (changes per hop)
- `x-kagent-root-context-id` — top-of-chain context_id (stable across hops + across turns of the same conversation)

Closes #1924.

## Why

`KAgentRemoteA2ATool.run_async` ships the outbound A2A message with `context_id = self._last_context_id`, a `uuid4()` minted once per tool instance. The receiving peer has no way to correlate the turn with the chat that started it. Anyone building multi-hop kagent fleets that key per-conversation state (sessions, worker pods, idempotency tokens, cache entries) on a stable identifier today either hand-rolls a `header_provider` on every tool or accepts losing continuity at every hop.

Motivating concrete failure mode: a chat-tier Declarative agent delegates to a router that spawns a per-conversation worker pod (e.g. a `kubernetes-sigs/agent-sandbox` `SandboxClaim` running `claude --print`); the router has no stable identifier to key the claim on, so a fresh pod spawns every turn and the user-visible symptom is "the agent loses memory between turns" even though the kagent UI thread is the same.

The receive side already copies all inbound HTTP headers into `session.state["headers"]` (set by `A2aAgentExecutor.execute`, lines 541-545), so no server-side plumbing change is needed — a peer that wants the root context_id just reads it from session state.

## What changed

`python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py`

- Two new module-level header constants: `PARENT_CONTEXT_ID_HEADER` and `ROOT_CONTEXT_ID_HEADER`.
- `_build_call_context` now also derives a lineage-headers dict and merges it into the outbound `ClientCallContext` state. The existing `header_provider` callback still wins on overlap, so callers can override lineage when they need to.
- Helper `_build_lineage_headers` derives `parent_context_id` from `tool_context.session.id` and `root_context_id` from the inbound `state["headers"]` (forwarded unchanged when present, falls back to legacy `parent` header for older callers, falls back to caller's own session id when this agent is the chain root). Returns `{}` when no session id is resolvable so the outbound request matches pre-feature behavior on stub tool contexts.

`python/packages/kagent-adk/tests/unittests/test_remote_a2a_tool.py`

- Extends `_MockSession`/`MockToolContext` to expose `id` and `state` so lineage logic is testable.
- Extends `_make_tool` helper to forward an optional `header_provider`.
- New `TestLineageHeaderPropagation` class covers six cases:
  - Root agent stamps own id as parent + root.
  - Mid-chain agent forwards root unchanged + overrides parent with own id.
  - Legacy inbound with only `parent` header promotes it to root.
  - No session id → no lineage headers (non-breaking).
  - `header_provider` overrides lineage when callers need to.
  - End-to-end through `_SubagentInterceptor` onto outbound `http_kwargs.headers`.

## Test plan

- [x] `uv run ruff format --diff` → clean
- [x] `uv run ruff check` → "All checks passed!"
- [x] `uv run pytest packages/kagent-adk/tests/unittests/test_remote_a2a_tool.py -q` → **28 passed**
- [ ] E2E coverage — happy to add if a maintainer points at the right place; the change is non-breaking by construction (additive headers + no-op when no session id), so I scoped this PR to unit coverage initially.

## Notes for reviewers

- Marked **draft** to gather design feedback first per CONTRIBUTING.md guidance. Happy to iterate on header naming, derivation rules, or scope.
- Backward-compatible: callers that ignore the new headers see no behavior change; callers that opt in just read `session.state["headers"][ROOT_CONTEXT_ID_HEADER]`.
- The export of header-name constants is intentional so BYO consumers (`from kagent.adk._remote_a2a_tool import ROOT_CONTEXT_ID_HEADER`) avoid string-literal drift.

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_remote_a2a_tool.py
@@ -61,6 +61,24 @@ _SOURCE_SUBAGENT = "agent"
 _HEADERS_STATE_KEY = "headers"
 _EXTRA_HEADERS_CONTEXT_KEY = "_a2a_extra_headers"
 
+# Conversation-lineage headers propagated on outbound A2A calls so a remote
+# agent can correlate this turn with the originating chat conversation —
+# useful when downstream code keys per-conversation state (sessions, sandbox
+# pods, cache entries) on a stable identifier across A2A hops.
+#
+# `x-kagent-parent-context-id` is the immediate caller's A2A context_id
+# (the session id of the agent that ran this tool). It changes with every
+# hop in a chain of A2A calls.
+#
+# `x-kagent-root-context-id` is the top-of-chain context_id — the agent at
+# the start of the chain (typically the user-facing chat agent). It stays
+# stable across every hop and across every turn of the same conversation,
+# so downstream agents can use it to key state that should outlive a single
+# A2A call (e.g. claim a per-conversation worker pod that survives between
+# turns).
+PARENT_CONTEXT_ID_HEADER = "x-kagent-parent-context-id"
+ROOT_CONTEXT_ID_HEADER = "x-kagent-root-context-id"
+
 
 class _SubagentInterceptor(ClientCallInterceptor):
     """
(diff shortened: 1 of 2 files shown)


Prefix:
Closing — will reopen clean under a fresh fork. Sorry for the churn.

PR #1928, opened 2026-05-26
fix(security): replace shell-templated skills-init with Go binary (#1842)

### LINKED ISSUE #1842: [FEATURE] Transition skills-init container to golang
### 📋 Prerequisites

### 📝 Feature Summary

_No response_

The skills-init container originally used bash because we, (wrongly), assumed that the scope and complexity of it would remain small. It is still relatively simple, but the templated bash script is making it very very difficult to understand. We should just use a simple go program to do this work instead and build it alongside everything else.

### 💡 Proposed Solution

Build a go version. I suppose I'm open to other languages but it has to be fast. I much prefer go because I don't want to use `Python`, and I don't want to introduce any new languages.

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

EItanya:
## Summary

Fixes #1842 — arbitrary command execution in the `skills-init` container via Agent CRD fields.

The previous implementation rendered a Bash script from a Go `text/template`, interpolating user-controlled fields (Git URL/Ref/Path/Name, OCI image, secret names) into `cat <<'ENDVAL' ... ENDVAL` heredocs. A low-privileged user with `create`/`update` on Agent could embed `ENDVAL` to escape the heredoc and execute arbitrary commands inside the init container — reaching node IMDS, other in-cluster services, or the shared `/skills` volume.

This PR eliminates the shell entirely:

- **New `skills-init` Go binary** (`go/core/cmd/skills-init` + `go/core/internal/skillsinit/`) consumes a structured JSON config and invokes `git` / `ssh-keyscan` via `exec.Command` with argv vectors. User strings never reach a shell.
- **OCI fetch** moves to in-process `go-containerregistry` (no more `krane` subprocess, no more `jq` for docker-config merging). Tar extraction rejects absolute paths, `..` traversal, and symlinks whose targets escape the destination.
- **Controller** emits a per-Agent ConfigMap with the JSON config; the pod template carries a `kagent.dev/skills-init-hash` annotation so config changes still trigger pod rollout.
- **Dockerfile** rewritten as a multi-stage build of the Go binary; `krane` and `jq` are dropped from the runtime image (smaller surface area, fewer CVE sources).

## Test plan

- [x] Unit tests updated to assert ConfigMap JSON content instead of script text — all green
- [x] Golden translator outputs regenerated (`UPDATE_GOLDEN=true`)
- [x] `go vet ./...` clean
- [x] End-to-end on a kind cluster:
  - Agent with `gitRefs[].path` → repo cloned via argv, subPath applied, agent container starts with files at `/skills/<name>/`
  - Agent with OCI `refs` → pull + TLS + token auth succeed; tar extraction works for benign content
  - Path-escape symlink in a tar entry → correctly rejected with a clear error
- [ ] CI green
- [ ] Maintainer review

🤖 Comment / PR created by Claude on behalf of @EItanya


go/core/cmd/skills-init/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+	cfg, err := skillsinit.LoadConfig()
+	if err != nil {
+		log.Fatalf("skills-init: %v", err)
+	}
+
+	home, _ := os.UserHomeDir()
+	if home == "" {
+		home = "/root"
+	}
```
iplay88keys: What does this give us? Is it better to run with relative paths from where the binary is in the container? I just don't like how it can be dependent on the `$HOME` env var.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1115,53 +1114,75 @@ func gitSkillName(ref v1alpha2.GitRepo) string {
...
+var validSkillNamePattern = regexp.MustCompile(`^[A-Za-z0-9._-]+$`)
+
+// validateSkillName rejects names that would escape /skills/ or look like
+// dotfiles that hide the skill.
+func validateSkillName(name string) error {
+	if name == "" {
+		return fmt.Errorf("skill name is empty")
+	}
+	if name == "." || name == ".." {
+		return fmt.Errorf("skill name %q is reserved", name)
```
peterj: nit: are `.` and `..` a name? :)

peterj (dismissed):

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1115,53 +1114,75 @@ func gitSkillName(ref v1alpha2.GitRepo) string {
...
 
-	return sshHostData{Host: host}, true
+	return skillsinit.SSHHost{Host: host}, true
+}
+
+// validSkillNamePattern restricts skill directory names to a safe alphabet.
+// The name becomes the final path segment under /skills/, so anything beyond
+// [a-zA-Z0-9._-] (notably "/" and "..") could escape the skills volume.
+var validSkillNamePattern = regexp.MustCompile(`^[A-Za-z0-9._-]+$`)
+
```
iplay88keys, line 1175: I think it would be better to use `root, err := os.OpenRoot(dir)`  and `root.Open(path)` to prevent path traversal rather than depend on a regex. This is fine for a first layer of validation, but enforcing that would be good as well.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1275,95 +1253,135 @@ func prepareSkillsInitData(
 		ref.Ref = gitRef
 
 		name := gitSkillName(ref)
+		if err := validateSkillName(name); err != nil {
+			return skillsinit.Config{}, fmt.Errorf("git skill %q: %w", ref.URL, err)
+		}
 		if seen[name] {
-			return skillsInitData{}, fmt.Errorf("duplicate skill directory name %q", name)
+			return skillsinit.Config{}, fmt.Errorf("duplicate skill directory name %q", name)
```
iplay88keys, line 1262: Is it better to return an error or just dedup?

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1275,95 +1253,135 @@ func prepareSkillsInitData(
...
 	}
 
 	for _, imageRef := range ociRefs {
 		name := ociSkillName(imageRef)
+		if err := validateSkillName(name); err != nil {
+			return skillsinit.Config{}, fmt.Errorf("oci skill %q: %w", imageRef, err)
+		}
 		if seen[name] {
-			return skillsInitData{}, fmt.Errorf("duplicate skill directory name %q", name)
+			return skillsinit.Config{}, fmt.Errorf("duplicate skill directory name %q", name)
```
iplay88keys, line 1293: Same as above, should we just dedup?

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1275,95 +1253,135 @@ func prepareSkillsInitData(
...
-// under /docker-secrets/<name> and the script merges them into a single config.json in /tmp;
-// krane reads the credentials via the DOCKER_CONFIG env var exported by the script.
+// SkillsInitConfigMapSuffix is appended to the Agent name to form the
+// ConfigMap that carries the skills-init container's JSON config.
+const SkillsInitConfigMapSuffix = "-skills-init"
+
+// SkillsInitConfigMapName returns the name of the skills-init ConfigMap for
+// the given Agent.
+func SkillsInitConfigMapName(agentName string) string {
+	return agentName + SkillsInitConfigMapSuffix
```
iplay88keys, line 1320: We should probably have validation on this for length. There also might be differences in what K8s accepts for configmaps vs pod names.

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1394,23 +1410,23 @@ func buildSkillsInitContainer(
...
+			MountPath: skillsinit.DockerSecretsDir + "/" + secret.Name,
 			ReadOnly:  true,
 		})
 	}
 
 	skillsInitContainer := corev1.Container{
 		Name:            "skills-init",
 		Image:           DefaultSkillsInitImageConfig.Image(),
-		Command:         []string{"/bin/sh", "-c", script},
+		Command:         []string{"/usr/local/bin/skills-init"},
```
iplay88keys: Would it be better to just use the existing [docker container entrypoint](https://github.com/kagent-dev/kagent/pull/1928/changes#diff-89bc839540d2ecb21673065742d39d26c31b8caf49b0eb4b8cc7aedf4afcf24dR47)?

go/core/internal/skillsinit/config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// GitRef describes a single git clone operation.
+type GitRef struct {
+	URL string `json:"url"`
+	// Ref is a branch name, tag, or commit SHA. When IsCommit is true a
+	// full clone + checkout is used; otherwise a shallow --branch clone.
+	Ref      string `json:"ref"`
+	Dest     string `json:"dest"`
+	IsCommit bool   `json:"isCommit,omitempty"`
```
iplay88keys: Is there a better name for this? Maybe `shallow` or `full` depending on what you want the default to be?

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -1275,95 +1253,135 @@ func prepareSkillsInitData(
...
 				}
 			}
 		}
 
-		data.GitRefs = append(data.GitRefs, gitRefData{
+		cfg.GitRefs = append(cfg.GitRefs, skillsinit.GitRef{
 			URL:      ref.URL,
 			Ref:      gitRef,
-			Dest:     "/skills/" + name,
+			Dest:     skillsinit.SkillsDir + "/" + name,
```
iplay88keys: Should this dir be separate so that git skills named the same as other types of skills don't collide as easily? Or would you just provide a different name in that case?

go/core/internal/skillsinit/git.go
```diff
@@ -0,0 +1,135 @@
...
+	return cmd.Run()
+}
+
+// applySubPath replaces dest with the contents of dest/subPath. The result is
+// that dest contains only the requested subdirectory. We materialize the
+// content into a sibling tmp dir under the same parent so the final rename is
+// atomic on the same filesystem.
+func applySubPath(dest, subPath string) error {
+	// Defense in depth: refuse traversal even if upstream validation slipped.
+	clean := filepath.Clean(subPath)
```
iplay88keys: Mentioned above, but I think with Go 1.24+, the recommended approach is to use `os.OpenRoot()`

jmhbh (dismissed):

iplay88keys (commented):
Not a full review, but here's my comments so far.

go/core/internal/controller/translator/agent/adk_api_translator.go
EItanya: I think we were erroring before, so I think let's start with maintaining the same behavior

go/core/internal/controller/translator/agent/adk_api_translator.go
EItanya: It has to be a different name, the skills systems use the top-level folder as the name deliniation

iplay88keys: Ok, makes sense!

go/core/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Covered previously

go/core/internal/controller/translator/agent/adk_api_translator.go
iplay88keys: Covered with `validateSkillsInitConfigMapName` below

iplay88keys (approved):
LGTM

PR #1948, opened 2026-05-31
fix: preserve Ollama streaming tool calls

### LINKED ISSUE #1922: [BUG] Ollama streaming adapter drops tool_calls emitted before the done chunk
### 📋 Prerequisites
- [x] Searched existing issues
- [x] Reproducible

### 🐛 Bug Description
The `KAgentOllamaLlm` streaming path in `kagent-adk/src/kagent/adk/models/_ollama.py` only reads `tool_calls` from the chunk where `chunk.done == True`. However, Ollama's `/api/chat` streaming protocol emits `tool_calls` in an **earlier** chunk and then sends a separate final chunk with `done=True, tool_calls=None, content=""`. As a result, when an Agent has `spec.declarative.stream: true` (the default), every tool call the model makes is silently discarded. The agent yields an `LlmResponse` with empty `content.parts: []`, no event is enqueued, and the A2A request hangs in a `dequeue_event` poll loop until the client times out.

### 🔄 Steps to Reproduce
1. Apply a `ModelConfig` pointing at any Ollama-hosted model with native tool calling (`llama3.2:3b`, `qwen2.5:3b`, etc.).
2. Apply a declarative `Agent` with `stream: true` and at least one MCP tool (e.g. the default `my-first-k8s-agent` with `k8s_get_resources`).
3. Send a prompt that should trigger a tool call ("any exciting events in my cluster recently?").
4. Observe: no reply is ever returned to the UI; Phoenix shows an `LlmResponse` with `parts: []` despite non-zero `eval_count`.

### 🔬 Direct evidence

Streaming Ollama with the same tool, hitting the upstream directly:
```
$ curl -s POST /api/chat -d '{"model":"llama3.2:3b","stream":true,"tools":[...],"messages":[...]}'
done=False content=''  tool_calls=[{'function': {'name': 'k8s_get_resources', 'arguments': {'resource_type': 'events'}}}]
done=True  content=''  tool_calls=None
```

The tool call arrives in the **non-final** chunk.

### 🩹 Code location

[`kagent-adk/src/kagent/adk/models/_ollama.py`](https://github.com/kagent-dev/kagent/blob/main/python/packages/kagent-adk/src/kagent/adk/models/_ollama.py) (streaming branch in `generate_content_async`):

```python
async for chunk in response:
    if chunk.message.content:
        aggregated_text += chunk.message.content
        yield LlmResponse(..., partial=True, ...)
    if chunk.done:
        final_parts = []
        if aggregated_text:
            final_parts.append(types.Part.from_text(text=aggregated_text))
        for tc in chunk.message.tool_calls or []:   # ← only the done chunk
            ...
```

Should accumulate `tool_calls` across all chunks:

```python
aggregated_tool_calls: list = []
async for chunk in response:
    if chunk.message.content:
        ...
    if chunk.message.tool_calls:
        aggregated_tool_calls.extend(chunk.message.tool_calls)
    if chunk.done:
        ...
        for tc in aggregated_tool_calls:
            ...
```

The non-streaming branch in the same function handles this correctly — it's only the streaming path that's broken.

### 🩺 Workaround
Set `spec.declarative.stream: false` on the Agent CR. The non-streaming path correctly emits `function_call` parts.

### 💻 Environment
- Chart: `kagent-0.9.4`
- App image: `cr.kagent.dev/kagent-dev/kagent/app:0.9.4`
- `kagent-adk`: 0.3.0
- Ollama backend: tested against `llama3.2:3b` and `gemma3n:e4b` aliases; reproducible with any tool-capable Ollama model
- Kubernetes: kind in devcontainer

andreivince:
Closes #1922.

Ollama can emit `tool_calls` before the final `done=true` chunk when streaming. The adapter was only reading `chunk.message.tool_calls` from the final chunk, so a real streamed tool call could turn into an empty final ADK response.

I changed the streaming path to accumulate tool calls across chunks, then include them when building the final `LlmResponse`. I also moved the Ollama tool-call to ADK part conversion into one helper so streaming and non-streaming stay consistent.

I verified this two ways:

- Added a regression test where the tool call arrives on a non-final chunk and the done chunk has no tool calls. This failed before the fix and passes after it.
- Reproduced against a live local Ollama server with `qwen3:0.6b`. Raw Ollama emitted `get_temperature({"city": "Tokyo"})` on `done=false`, then a final `done=true` chunk with no tool calls. Before the fix, kagent returned `function_calls=0`; after the fix, kagent returned the expected `get_temperature` function call.

Checks run:

- `uv run pytest packages/kagent-adk/tests/unittests/models/test_ollama.py -q`
- `uv run pytest packages/kagent-adk/tests/unittests/models --ignore=packages/kagent-adk/tests/unittests/models/test_tls_e2e.py -q`
- `uv run ruff check packages/kagent-adk/src/kagent/adk/models/_ollama.py packages/kagent-adk/tests/unittests/models/test_ollama.py`

One note: the full model test directory has unrelated local TLS E2E failures because the cert fixture files are missing in my checkout. The non-TLS model tests passed.

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/models/_ollama.py
+++ b/python/packages/kagent-adk/src/kagent/adk/models/_ollama.py
@@ -133,6 +133,13 @@ def _convert_tools_to_ollama(tools: list[types.Tool]) -> list[ollama_sdk.Tool]:
     return ollama_tools
 
 
+def _convert_tool_call_to_part(tc: OllamaMessage.ToolCall) -> types.Part:
+    part = types.Part.from_function_call(name=tc.function.name, args=dict(tc.function.arguments))
+    if part.function_call:
+        part.function_call.id = str(uuid.uuid4())
+    return part
+
+
 class KAgentOllamaLlm(KAgentTLSMixin, BaseLlm):
     """Ollama model via the native Ollama SDK.
 
@@ -190,6 +197,7 @@ class KAgentOllamaLlm(KAgentTLSMixin, BaseLlm):
         try:
             if stream:
                 aggregated_text = ""
+                tool_calls = []
                 response: AsyncIterator[ollama_sdk.ChatResponse] = await self._client.chat(
                     model=llm_request.model or self.model,
                     messages=messages,
@@ -198,6 +206,7 @@ class KAgentOllamaLlm(KAgentTLSMixin, BaseLlm):
                     stream=True,
                 )
                 async for chunk in response:
+                    tool_calls.extend(chunk.message.tool_calls or [])
                     if chunk.message.content:
                         aggregated_text += chunk.message.content
                         yield LlmResponse(
@@ -211,13 +220,7 @@ class KAgentOllamaLlm(KAgentTLSMixin, BaseLlm):
                         final_parts = []
                         if aggregated_text:
                             final_parts.append(types.Part.from_text(text=aggregated_text))
-                        for tc in chunk.message.tool_calls or []:
-                            part = types.Part.from_function_call(
-                                name=tc.function.name, args=dict(tc.function.arguments)
-                            )
-                            if part.function_call:
-                                part.function_call.id = str(uuid.uuid4())
-                            final_parts.append(part)
+                        final_parts.extend(_convert_tool_call_to_part(tc) for tc in tool_calls)
                         finish_reason = _done_reason_to_finish_reason(chunk.done_reason) if chunk.done_reason else None
                         usage_metadata = None
                         if chunk.prompt_eval_count is not None or chunk.eval_count is not None:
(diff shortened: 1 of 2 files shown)


EItanya:
Thanks so much for the contribution, you will need to sign your commits to pass DCO before we can merge

jmhbh (commented):
lgtm after commits are signed.

andreivince:
Thanks @EItanya, signed the commit and force-pushed the clean branch. @jmhbh DCO is green now too.

andreivince:
Looks like the remaining `build (app)` failure is a CI flake. It fails in the arm64 Docker base layer with a QEMU illegal-instruction error before kagent source is copied in.

This PR only touches the Ollama adapter and its unit test, and the Python lint/tests are green. Could someone rerun the failed build jobs?

jmhbh (approved):

PR #1950, opened 2026-06-01
Link fixes

kristin-kronstain-brown:
## fix: update stale and broken URLs in Go API type comments

Three sets of comment URLs in the API type definitions were either redirecting or pointing to renamed anchors, causing failures in downstream docs link checking.

## Changes

- **`github.com/google/A2A` → `github.com/a2aproject/A2A`** (`v1alpha1` and `v1alpha2` `agent_types.go`): The A2A project moved to a new GitHub org. Updated the link in the `a2aConfig` field description.

- **`#cross-namespace-routing` → `#cross-namespace-route-attachment`** (`agent_types.go`, `remotemcpserver_types.go`, `common_types.go`): The Gateway API docs renamed this anchor. The old fragment no longer exists on the `guides/multiple-ns/` page.

- **`adk-python/issues/3921)`** (`agent_types.go`): Added a space before the closing parenthesis so URL parsers don't include it as part of the link.

These comments flow into the auto-generated API reference docs via `crd-ref-docs`, so fixing them here eliminates the link checker warnings in downstream doc sites.

### DIFF
--- a/go/api/v1alpha2/agent_types.go
+++ b/go/api/v1alpha2/agent_types.go
@@ -80,7 +80,7 @@ type AgentSpec struct {
 	// This follows the Gateway API pattern for cross-namespace route attachments.
 	// If not specified, only Agents in the same namespace can reference this Agent as a tool.
 	// This field only applies when this Agent is used as a tool by another Agent.
-	// See: https://gateway-api.sigs.k8s.io/guides/multiple-ns/#cross-namespace-routing
+	// See: https://gateway-api.sigs.k8s.io/guides/multiple-ns/#cross-namespace-route-attachment
 	// +optional
 	AllowedNamespaces *AllowedNamespaces `json:"allowedNamespaces,omitempty"`
 }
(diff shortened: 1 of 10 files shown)


go/api/v1alpha2/agent_types.go
```diff
@@ -212,7 +212,7 @@ type DeclarativeAgentSpec struct {
 	// If true, the agent will automatically execute python code blocks in the LLM responses.
 	// Code will be executed in a sandboxed environment.
 	// +optional
-	// due to a bug in adk (https://github.com/google/adk-python/issues/3921), this field is ignored for now.
+	// due to a bug in adk (https://github.com/google/adk-python/issues/3921 ), this field is ignored for now.
```
peterj, line 215: this space probably added by mistake

peterj:
thanks for the PR! Can you also make sure you run `make controller-manifests` and include the files that get generated?

go/api/v1alpha2/agent_types.go
kristin-kronstain-brown: @peterj The link was getting interpreted as `https://github.com/google/adk-python/issues/3921%29`, so adding the space was Claude's way of getting rid of the parentheses. I can investigate other options if you think it's the wrong way to go about it.

kristin-kronstain-brown: Actually, it's not just lychee. It is in the markdown output too.
https://docs.solo.io/kagent-enterprise/docs/latest/reference/api/kagent/#declarativeagentspec
Screenshot 2026-06-01 at 1 52 26 PM

peterj: oh, looks like it's taking the `)` as part of the link

peterj: nah, it's fine, I thought it was a typo :)

peterj (approved):

EItanya (approved):

PR #1962, opened 2026-06-04
fix(translator): forward Anthropic config fields to model config

mesutoezdil:
MaxTokens, Temperature, TopP, and TopK were defined in AnthropicConfig but never copied to the adk.Anthropic struct during translation, so they had no effect at runtime.

This applies them the same way OpenAI does in the same function.


mesutoezdil:
Re Copilot comments: (1) ParseStringToFloat64 returns nil for empty strings, no zero value is set when fields are omitted. (2) >0 guard matches the OpenAI pattern in this same function; 0 is not valid for max_tokens or top_k. (3) Golden test covers all four fields with real values.

EItanya (approved):

Issue #1637, opened 2026-04-07
Add upgrade compatibility tests for database migrations


iplay88keys:
## Context

The migration PR (#1561) replaces GORM with golang-migrate + sqlc. Manual testing confirmed upgrade safety, but we need automated tests to prevent regressions as the schema evolves.

### 1. Rolling upgrade compatibility: previous release code against new schema

Verify that the previous release's code can read and write against the current version's schema. This simulates a rolling deploy where old pods are still running while the new schema has already been applied.

### 2. Data-preserving upgrade: previous release with data to current version

Verify that an upgrade from the previous release preserves all existing data and passes e2e tests.

## Notes

- These tests should run in CI on PRs that touch migration files or database queries

PR #1963, opened 2026-06-04
fix(controller): skip DB migration when DB schema is ahead of binary

### LINKED ISSUE #1881: [BUG] Rollback of controller fails when older binary lacks DOWN files for newer schema versions
### 📋 Prerequisites

- [ ] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

When the kagent controller is downgraded to a previous version, the older controller fails to start if the newer version applied schema migrations that are unknown to the older version. This was observed when attempting to rollback after a failed upgrade from `0.9.0` to `0.9.3`´. The .down.sql files for those migrations are embedded only in the newer container. The older container can neither roll the schema back nor start against the newer schema, resulting in a crash loop.

### 🔄 Steps To Reproduce

1. Deploy application `v0.9.0` — embeds migrations 000001–000003. Schema reaches version 3.
2. Upgrade to `v0.9.3` — embeds migrations 000001–000004. Schema reaches version 4.
3. `v0.9.3` is found to be broken; downgrade back to v0.9.0.
4. `v0.9.0` starts against a DB at schema version 4. Its embedded source only knows versions 1–3.

### 🤔 Expected Behavior

Rolling back to a previous version is a standard recovery operation (e.g. when a new release introduces a regression) and should work reliably.

### 📱 Actual Behavior

_No response_

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

onematchfox:
When an older binary starts against a database that a newer binary has already migrated, calling `mg.Up()` fails because golang-migrate tries to read the down file for the current DB version, which does not exist in the older binary's embedded FS. This results in the controller crash looping (issue #1881).

Instead, detect when the database version exceeds the binary's highest known migration and return early. This relies on the expand-then-contract policy documented in [database-migrations.md](https://github.com/kagent-dev/kagent/blob/main/.claude/skills/kagent-dev/references/database-migrations.md#backward-compatible-schema-changes), which guarantees each release's code is compatible with the schema applied by the previous release.

Adds TestApplyDir_SucceedsWhenDBVersionAhead to cover this path.


go/core/pkg/migrations/runner.go
```diff
@@ -56,12 +57,37 @@ func applyDir(url string, migrationsFS fs.FS, dir, migrationsTable string) (prev
...
+	// attempt failed and must be resolved, not silently accepted.
+	if maxVer, scanErr := maxEmbeddedVersion(migrationsFS, dir); scanErr != nil {
+		log.Error(scanErr, "could not determine max embedded migration version; proceeding with Up", "track", dir)
+	} else if prevVersion > maxVer {
+		if dirty {
+			// DB is both dirty and ahead of this binary. Attempting Up/rollback would
+			// fail (the migration files for prevVersion don't exist), producing noisy
+			// and misleading logs. Return a clear error so operators act on the real
+			// problem rather than chasing rollback noise.
+			return prevVersion, fmt.Errorf("database is dirty at version %d and ahead of this binary's max known version %d for track %s: manual operator intervention required: %w",
```
iplay88keys, line 83: I just wonder if this might not be enough of an indicator that the db is in a state that requires manual intervention. Having the binary fail definitely notifies the user, but has the obvious downsides.

onematchfox: How would you propose handling it? Log a warning and skip migrations so the binary doesn't fail (as we do with "compatability" mode)?

onematchfox: I'm good with either approach - this should really be an edge case.

iplay88keys: Actually, I got this mixed up with the other case below where the db is not dirty. I think this part is fine. The  other case where the binary is running in compatibility mode is less visible, but failing on it there as well would crashloop the process, so that seems to be the best option.

iplay88keys (dismissed):
This might be changed when we update the db migration/rollback logic, but this is a good incremental improvement in the meantime.

iplay88keys:
Sorry for un-approving this PR. I want to make sure the implications of this change are well thought through.

Let's start with aligning on a few things:
- We only support n+1 upgrades and n-1 downgrades.
- We also don't currently have a guarantee that a downgrade doesn't require manual steps, but we try to avoid them on upgrades.

The current db migrations are set up such that the application fails to start if there are more migrations applied to the db than the binary knows about. This is by design to prevent data corruption.

There is no way for the binary to know what migrations are in n+1 vs n+2 without going back and editing the old artifacts or making this incompatible with airgapped environments.

This PR removes the aforementioned check (comparing known migrations vs what's migrated), effectively introducing a requirement that Kagent stays compatible with future database schema shape changes and queries. It also leaves a subtle footgun as the information regarding the "compatibility mode" is only surfaced in a log line.

Even with this change, there is still a manual database rollback requirement in order to not be in an inconsistent state. So I'm not sure that it's worth taking on these additional support burdens in that case.

onematchfox:
> We also don't currently have a guarantee that a downgrade doesn't require manual steps, but we try to avoid them on upgrades.

That's a bit of a understatement 😅; the only way to downgrade is through manual intervention at present. Even though [the docs](https://github.com/kagent-dev/kagent/blob/main/.claude/skills/kagent-dev/references/database-migrations.md#down-migrations) seem to indicate that this shouldn't ever be the case? 

> ... and by automatic rollback on migration failure


> The current db migrations are set up such that the application fails to start if there are more migrations applied to the db than the binary knows about. This is by design to prevent data corruption.

This is the issue. Any database schema update ultimately prevents any form of automated rollback. This design choice will also cause the existing version of the controller to crash-loop if it scales out during the rollout of a new version - this despite the fact that it [is documented](https://github.com/kagent-dev/kagent/blob/main/.claude/skills/kagent-dev/references/database-migrations.md#backward-compatible-schema-changes) that:

> During a rolling deploy, old pods will be reading and writing a schema that has already been upgraded. Every migration must be backward-compatible with the previous version's code.

> There is no way for the binary to know what migrations are in n+1 vs n+2

Yes, unfortunately there is no alignment between the DB schema versions and the application version. Hence, I did not try to implement a check on whether a user and rather opted for a warning that would be visible if the controller crashed due to an incompatible schema. Worth noting here, is that if the documented process is followed then this would only ever happen when rolling back N+2 minor versions. E.g. rolling back `0.10.x`->`0.8.x`. And even then, incompatible/breaking schema changes (dropping tables, columns, etc) are far less likely to happen than compatible/additive ones surely? Whilst there's isn't muchhistory in the current set of migrations, I don't see any backwards incompatible changes so far so we could theoretically be running `0.7.x` against the schema in `0.9.x`. Whereas in reality we can't even run `0.9.2` against the schema introduced by `0.9.3` (a patch release).

> effectively introducing a requirement that Kagent stays compatible with future database schema shape changes and queries.

I don't expect KAgent to be compatible with all future database schema changes. But I do expect to be able to rollback a patch release, or a single minor version without manual intervention. 

> It also leaves a subtle footgun as the information regarding the "compatibility mode" is only surfaced in a log line.

IMO, this is a lot better than the current behaviour. If the controller crash loops, an operator will most likely look at logs and see the warning 🤞 .

> Even with this change, there is still a manual database rollback requirement in order to not be in an inconsistent state

I disagree. If the expand/contract pattern is followed as documented there should not need to be a manual database rollback requirement when rolling back from N+1->N.

iplay88keys:
>by automatic rollback on migration failure

At this point, the design around this is focused on temporary migration upgrade failures, with the goal of retrying db migrations. Full rollback of the application after a successful upgrade was not in scope for that design.

>Yes, unfortunately there is no alignment between the DB schema versions and the application version.

This, coupled by the lack of preventing multiple version downgrades is still my biggest concern around this at the moment.

>If the expand/contract pattern is followed as documented there should not need to be a manual database rollback requirement when rolling back from N+1->N.

Whether or not the application runs does not remove the fact that the db is no longer in the state expected by the migration system for the version of Kagent running. That may or may not be an issue. Another risk I see here is data being orphaned as a result of the upgrade process. The expand and contract model should cover this if implemented correctly, though.

> IMO, this is a lot better than the current behaviour

I definitely see your point on this. I am actively looking into what db rollbacks looks like for Kagent and am including this as an option to fill in that feature gap.

iplay88keys (approved):
After further consideration, I think that with our compatibility goals of `n-1` application works with `n` db, it makes sense to allow the application to start up against db with additional migrations when you upgrade within a supported window (any patch in one minor to any patch in the next minor).

We can document this, but a few things come to mind:
- Application rollbacks are back to the exact version you had run previously
- Before upgrading or rolling back more, the db should be reset to that version's set of migrations
  - For now, this can be done via golang-migrate directly against the migration files
- There does not have to be a mechanism for preventing rolling back further. The documentation around this covers that by saying the db needs to be on the appropriate version before a rollback of the code.

iplay88keys:
It's also worth noting that [this issue](https://github.com/kagent-dev/kagent/issues/1637) will help us continually test this rollback path.

Issue #1958, opened 2026-06-03
AgentHarness: phaseToCondition should fall back to status.conditions when phase=UNSPECIFIED


nloke:
## Summary

`phaseToCondition` in `translate.go` switches on `sb.GetPhase()` to derive the AgentHarness `Ready` condition. When the OpenShell gateway returns `phase=UNSPECIFIED` (or omits the field entirely — see https://github.com/NVIDIA/OpenShell/issues/1710), the controller permanently sets `Ready=Unknown / reason=SandboxPhaseUnknown` and requeues every 10 s, even though `status.conditions[Ready=True]` is present in the same response.

## Affected code

`go/core/pkg/sandboxbackend/openshell/translate.go`

```go
case openshellv1.SandboxPhase_SANDBOX_PHASE_UNKNOWN, openshellv1.SandboxPhase_SANDBOX_PHASE_UNSPECIFIED:
    return metav1.ConditionUnknown, "SandboxPhaseUnknown", msg
```

## Observed behaviour

- Gateway returns full `status.conditions[{type:Ready, status:True}]` but no `phase` field
- Controller logs: `reason=SandboxPhaseUnknown` on every reconcile loop (every 10 s via `agentHarnessNotReadyRequeue`)
- AgentHarness never reaches `Ready=True` despite sandbox pod running and supervisor connected

## Expected behaviour

When `phase` is `UNSPECIFIED` (or missing from the wire response), the controller should inspect `status.conditions` as a fallback rather than immediately giving up.

## Proposed fix

```go
case openshellv1.SandboxPhase_SANDBOX_PHASE_UNKNOWN, openshellv1.SandboxPhase_SANDBOX_PHASE_UNSPECIFIED:
    // Gateway may omit the phase field (see NVIDIA/OpenShell#1710).
    // Fall back to status.conditions so an older gateway binary does not
    // permanently block AgentHarness readiness.
    for _, c := range sb.GetStatus().GetConditions() {
        if c.GetType() == "Ready" && c.GetStatus() == "True" {
            return metav1.ConditionTrue, "SandboxReady", msg
        }
    }
    return metav1.ConditionUnknown, "SandboxPhaseUnknown", msg
```

## Evidence

```
# grpcurl GetSandbox — phase field absent, conditions show Ready=True
fields returned: ['metadata', 'spec', 'status']
phase present: False
status.conditions[0]: {type: Ready, status: True, reason: DependenciesReady, message: Pod is Ready; Service Exists}
```

[...]

PR #1964, opened 2026-06-04
fix(agentharness): upsert and attach inference provider at AgentHarness sandbox creation

nloke:
## Problem

When deploying an `AgentHarness` using the `openclaw` backend, the agent starts and Slack connects successfully, but all LLM calls fail with a `Connection error`.

## Root Cause

openclaw uses an HTTP proxy (`HTTPS_PROXY=http://10.200.0.1:3128`) to resolve credentials at request time. The `openclaw.json` config stores an unresolved placeholder:

```json
"apiKey": "openshell:resolve:env:OPENAI_API_KEY"
```

When openclaw (running as UID 998) makes an LLM call, the proxy intercepts the `Bearer openshell:resolve:env:OPENAI_API_KEY` header, looks up `OPENAI_API_KEY` from the **sandbox's attached inference provider**, and replaces it with the real key before forwarding to the upstream LLM gateway.

`ClawBackend.EnsureAgentHarness` calls `attachMessagingProviders` to attach Slack providers but **never creates or attaches an inference provider**. The sandbox is created without one — the proxy has nothing to resolve and closes the connection.

## Fix

Added `upsertInferenceProviderForHarness` wired into `EnsureAgentHarness`:

1. Reads the `ModelConfig` referenced by the `AgentHarness`
2. Resolves the API key from the referenced k8s secret
3. Calls `UpsertInferenceProvider` on the OpenShell gRPC API with `{OPENAI_API_KEY, OPENAI_BASE_URL}`
4. Passes the provider name into `attachMessagingProviders` so the sandbox is created with the provider attached

The sandbox process never holds the real key — credential resolution happens in the proxy at request time, preserving the OpenShell security model.

## Files Changed

- `openclaw.go` — `EnsureAgentHarness` wired to call `upsertInferenceProviderForHarness` and pass provider into sandbox spec
- `providers.go` — new `UpsertInferenceProvider` method
- `openclaw/modelconfig.go` — `InferenceProviderName` helper
- `openclaw/provider.go` — exported `BootstrapProviderBaseURL` wrapper
- `translate.go` — phase fallback fix (related: #1958)

## Testing

Validated end-to-end on a local kind cluster with a Slack bot (`socket mode`) backed by an OpenAI-compatible LiteLLM gateway. Bot receives messages and returns LLM responses after this fix.

## Notes

- This is a draft PR to share the code alongside the bug report — happy to iterate on the approach
- The `translate.go` phase fix is bundled here but tracked separately in #1958

### DIFF
--- a/go/core/pkg/sandboxbackend/openshell/openclaw.go
+++ b/go/core/pkg/sandboxbackend/openshell/openclaw.go
@@ -50,7 +50,46 @@ func (b *ClawBackend) EnsureAgentHarness(ctx context.Context, ah *v1alpha2.Agent
 	if res, found, err := b.findExistingSandbox(ctx, ah); err != nil || found {
 		return res, err
 	}
-	return b.ensureAgentHarnessSandbox(ctx, ah, buildClawCreateRequest)
+
+	// Upsert the inference provider so the OpenShell proxy can resolve
+	// openshell:resolve:env:<VAR> placeholders in LLM Authorization headers.
+	// The real API key lives only in the OpenShell provider record; the sandbox
+	// process env never receives it directly.
+	inferenceProviderName, err := b.upsertInferenceProviderForHarness(ctx, ah)
+	if err != nil {
+		ctrllog.FromContext(ctx).Error(err, "failed to upsert inference provider; LLM credential resolution may fail",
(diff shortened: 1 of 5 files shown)


nloke:
Fixes #1965

PR #1971, opened 2026-06-06
feat(go-adk): add compaction support to Go ADK runtime

### LINKED ISSUE #1783: [FEATURE] Add compaction to golang ADK
### 📋 Prerequisites

### 📝 Feature Summary

Add auto compaction support to golang declarative and BYO agents

Compaction is a key part of long-running agentic workflows

### 💡 Proposed Solution

Add it to our libraries if possible, if not contribute upstream

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

None

### 📚 Additional Context

_No response_

yashrajshuklaaa:
Implements post-invocation event history compaction for the **Go ADK runtime** 
The upstream **Go ADK v1.1.0** does not ship this so it is implemented within **kagent** via a new compaction package 
Closes #1783


yashrajshuklaaa:
@iplay88keys @jmhbh @supreme-gg-gg PTAL 
addressed the comments of copilot cuz they are genuine 
if u need any changes pls let me know

EItanya (commented):
Thanks for taking this on. Adding Go-runtime compaction is a useful direction, and the current PR is a helpful starting point. I took a closer look against the Python ADK compaction path, and I think there is one important design gap to resolve before merging.

The main issue is that this does not yet match Python ADK compaction semantics. Python ADK does not durably delete compacted raw events. It appends a special compaction marker event with actions.compaction containing the compacted time range and compacted content, then prompt construction later uses that marker to replace covered raw events with the summary. In this PR, the Go runtime appends a normal model-looking summary event and trims only the in-memory local session. After a fresh GetSession, the persisted store still has the raw events plus the summary, and Go ADK has no marker-aware filtering step, so the next model request can see both raw history and the summary.

That means I would treat the durable-store Copilot concern as real, but I would not solve it by deleting old backend rows unless we intentionally want different semantics from Python. A closer match would be to persist some compaction marker/range and add a Go-side prompt-time filtering layer, or otherwise wrap/filter session events before Go ADK builds LLM contents.

A few related differences from the Python implementation also seem worth addressing:

- compaction_interval in Python means number of new invocations since the last compaction. The current Go code checks total grouped invocations, which can repeatedly compact the same logical history after reload or behave differently depending on in-memory truncation.
- summarizer_model appears to be parsed but not used to construct SummarizerLLM. Python builds an LlmEventSummarizer from the configured model, and otherwise initializes one from the agent model.
- token_threshold and event_retention_size are validated as a pair in Python. The Go implementation currently accepts partial token config and applies event retention to the kept events after interval compaction too, while Python uses retention for token-threshold compaction.
- Python takes care not to compact pending function calls or split function call/function response pairs. The Go implementation serializes calls/responses into text and can drop the structured pairing from the retained context.

The smaller Copilot comments about defensive copying, sentinels, and test cleanup are either already addressed or easy to handle. The extra GetSession call is a performance concern, but I think it is secondary to getting the compaction model aligned with Python first.

Suggested acceptance test: create enough persisted events to trigger compaction, run compaction, reload the session from the backing service, then verify that the next model request sees summary plus retained events rather than raw compacted history plus summary. That would catch the main semantic issue here.

Again, thank you for pushing this forward. I think the feature is valuable; it just needs the durable marker/filtering piece before it will behave like the existing Python ADK path.

yashrajshuklaaa:
@EItanya thanks for the detailed explanation , it really helped me understand what the correct implementation should look like
**I** went and read through the **Python ADK** source (  **apps/compaction.py  ,  flows/llm_flows/contents.py  ,  events/event_actions.py**  ) to understand the marker based approach you described
that makes total sense now - the compaction event carries **start_timestamp , end_timestamp and compacted_content** in **actions.compaction** and then **contents.py** filters out raw events that fall within that range at prompt-build time
The raw events stay in the backing store untouched.

So i checked the **Go ADK** module to see if the same primitives exist there. Went through every released version from **v0.1.0 all the way to v1.4.0** , searching for **EventCompaction** , **CompactedContent** , **StartTimestamp** and any other naming variation I could think of. Nothing. The **EventActions struct** in Go ADK doesn't have a Compaction field at all and **contents_processor.go** has no filtering step for compaction markers either

yashrajshuklaaa:
so i think the right move is to contribute to **EventCompaction** support upstream to  **google.golang.org/adk first** by adding the struct to  **EventActions**  and the filtering logic in **contents_processor.go** to mirror what Python ADK does
once that's done coming back and implementing proper marker-based compaction in kagent would be pretty straightforward

yashrajshuklaaa:
@EItanya opened an issue at https://github.com/google/adk-go after that we can continue with this

PR #1979, opened 2026-06-08
fix(ui): fix model name editing on the model config form

### LINKED ISSUE #1947: Model edit form: pencil button triggers a save API call
While editing an existing ModelConfig (/models/new?edit=true), there are two related problems with the model name field and its pencil ("Edit Auto-Generated Name") button.

  ---
  Issue 1 — Clicking the pencil icon fires an API call (saves the form)

  Clicking the pencil icon to edit the name immediately triggers a server-side update call (updateModelConfig) instead of just enabling name editing. The form gets submitted, the model is saved, and the page navigates away — so you can't actually edit anything.

  Captured request on pencil click:
  POST /models/new?edit=true&name=ollama-mistral&namespace=kagent
  payload: ["kagent/ollama-mistral",{"apiKey":null,"spec":{"model":"mistral","provider":"Ollama","ollama":{}}}]

https://github.com/user-attachments/assets/53b5d8e5-03f7-4d1d-b743-6f9c590a1d28

  ---
  Issue 2 — After fixing Issue 1, the name is editable but the change is useless

Once the pencil no longer submits and you can actually type a new name and save, the new name still doesn't take
effect — the model keeps its original name.

Cause: a ModelConfig is a Kubernetes resource, and its metadata.name is immutable — Kubernetes does not allow renaming a resource via update. The update only changes the spec, never the name.

  ---

Valyrian-Code:
## Summary

Two related problems with the model name field and its "Edit Auto-Generated Name" pencil button on the model form (`/models/new`, including `?edit=true`):

1. **Pencil submitted the form.** The button had no explicit `type`, so as a child of the `<form>` it defaulted to `type="submit"` and fired the `updateModelConfig` action on click — the form saved and navigated away instead of just toggling name editing. Fixed by setting `type="button"`, matching the sibling "fetch models" button.

2. **Renaming in edit mode silently did nothing.** A `ModelConfig`'s `metadata.name` is immutable in Kubernetes, so an update can never rename the resource. The pencil is now disabled in edit mode — consistent with the namespace, provider, and model fields, which are already disabled when editing — and a hint clarifies that the name can't be changed after creation.

Fixes #1947

## Changes

- `ui/src/components/models/new/BasicInfoSection.tsx`:
  - add `type="button"` to the edit-name pencil so it no longer submits the form
  - disable the pencil in edit mode and show a "name can't be changed after creation" hint

## Testing

- On `/models/new?edit=true`, clicking the pencil no longer triggers a network call or form submission.
- In edit mode the name field is read-only with the hint; create mode is unchanged (the auto-generated name is still editable).


peterj (approved):

Valyrian-Code:
The failing `build (app)` check is unrelated to this change, it's a QEMU/binfmt emulation failure while building the `sandbox-runtime` image in the `build-kagent-adk` step:

```
installing: amd64 cannot register "/usr/bin/qemu-x86_64" to /proc/sys/fs/binfmt_misc/register: no such file or directory
#36 ERROR: process "... git clone ... sandbox-runtime ... npm run build ..." did not complete successfully: exit code: 132
make: *** [Makefile:283: build-kagent-adk] Error 1
```

This PR only adds `type="button"` and an `isEditMode` guard to a single UI component, so it can't affect the Go app image build. Every other check (go-unit-tests, go-lint, ui build, python, helm, snyk) is green. @peterj would you be able to re-run the failed `build (app)` job so this can merge? Thanks for the review!

Valyrian-Code:
Thanks for approving @peterj! The only failing check is `build (controller)`, a flaky build that isn't related to this UI change. Could you run it again so it can merge?

PR #1980, opened 2026-06-08
fix(agentharness): fall back to status.conditions when phase is UNSPECIFIED

### LINKED ISSUE #1958: AgentHarness: phaseToCondition should fall back to status.conditions when phase=UNSPECIFIED

mesutoezdil:
When the OpenShell gateway returns `phase=UNSPECIFIED` (or omits the field), `phaseToCondition` was setting `Ready=Unknown` permanently and requeuing every 10s, even when `status.conditions[Ready=True]` was present in the same response.

Adds a fallback: when phase is `UNSPECIFIED` or `UNKNOWN`, check `status.conditions` for `Ready=True` before returning unknown.

Fixes #1958

### DIFF
--- a/go/core/pkg/sandboxbackend/openshell/translate.go
+++ b/go/core/pkg/sandboxbackend/openshell/translate.go
@@ -78,6 +78,11 @@ func phaseToCondition(sb *openshellv1.Sandbox) (metav1.ConditionStatus, string, 
 	case openshellv1.SandboxPhase_SANDBOX_PHASE_DELETING:
 		return metav1.ConditionFalse, "SandboxDeleting", msg
 	case openshellv1.SandboxPhase_SANDBOX_PHASE_UNKNOWN, openshellv1.SandboxPhase_SANDBOX_PHASE_UNSPECIFIED:
+		for _, c := range sb.GetStatus().GetConditions() {
+			if c.GetType() == "Ready" && c.GetStatus() == "True" {
+				return metav1.ConditionTrue, "SandboxReady", msg
+			}
+		}
 		return metav1.ConditionUnknown, "SandboxPhaseUnknown", msg
 	default:
 		return metav1.ConditionUnknown, "SandboxPhaseUnrecognized", fmt.Sprintf("unrecognized phase %s", sb.GetPhase())
--- a/go/core/pkg/sandboxbackend/openshell/translate_test.go
+++ b/go/core/pkg/sandboxbackend/openshell/translate_test.go
@@ -3,12 +3,91 @@ package openshell
 import (
 	"testing"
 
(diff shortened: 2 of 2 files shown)


mesutoezdil:
Closing as duplicate of #1966.

PR #1985, opened 2026-06-09
chore(linter): add new kube-api-linter and fix CRD lint issues 

dongjiang1989:
Add the `kube-api-linter` custom golangci-lint plugin to enforce Kubernetes API conventions across CRD type definitions. The following sub-linters are now enabled:                                     
                                                                                                                                                                                                          
  - `nodurations` — disallows `time.Duration` fields in API types (use string-based durations)                                                                                                            
  - `noreferences` — prevents use of pointer/reference types in API fields                                                                                                                                
  - `notimestamp` — disallows `time.Time` fields in API types                                                                                                                                             
  - `jsontags` — ensures JSON struct tags follow Go naming conventions (lowercase first letter)                                                                                                           
  - `statusoptional` — enforces `// +optional` marker on status subfields                                                                                                                                 
                                                                                                                                                                                                          
  All resulting lint violations across `v1alpha1` and `v1alpha2` CRD types have been resolved. Most fixes involve lowercasing the first letter of JSON struct tags for compliance.


dongjiang1989:
@EItanya  PTAL, thanks

iplay88keys:
I think that it would be nice to move in this direction, but I'd prefer if we had a plan for doing so (especially around breaking changes) rather than just enabling linting and skipping checks on specific fields that would fail otherwise. It could be useful to see which of these checks are currently passing with our current CRDs and which require changes.

dongjiang1989:
> I think that it would be nice to move in this direction, but I'd prefer if we had a plan for doing so (especially around breaking changes) rather than just enabling linting and skipping checks on specific fields that would fail otherwise. It could be useful to see which of these checks are currently passing with our current CRDs and which require changes.

Thanks @iplay88keys 

Add https://github.com/kagent-dev/kagent/issues/2026 for TODO tasks.

iplay88keys:
Thinking more about this, I'm not sure I really see the value in this at this point.

dongjiang1989:
> Thinking more about this, I'm not sure I really see the value in this at this point.

Although it's a breaking change, it makes sure we follow k8s api convention before moving the API to beta.

@iplay88keys WDYT?

PR #1987, opened 2026-06-09
add RemoteMCPServerPlugin reconcile-phase extension point

iplay88keys:
# Description

Adds a reconcile-phase extension point, `RemoteMCPServerPlugin`, so an extension can author Kubernetes resources for a `RemoteMCPServer` without standing up a second controller that watches the same resource. It mirrors the existing agent-side `TranslatorPlugin`: the plugin appends desired objects to an outputs struct, and the `RemoteMCPServer` reconciler owns them to the RMS and reconciles them (apply + prune).

```go
// translator.RemoteMCPServerPlugin
type RemoteMCPServerPlugin interface {
    // Called on each reconcile of server; the plugin appends the objects it
    // wants to exist to outputs.Manifest.
    ProcessRemoteMCPServer(ctx context.Context, server *v1alpha2.RemoteMCPServer, outputs *RemoteMCPServerOutputs) error
    // The object types this plugin may author, so the reconciler can set up
    // owner indexes and prune stale objects.
    GetOwnedResourceTypes() []client.Object
}
```

Plugins are registered through the existing extension-config seam (`ExtensionConfig.RemoteMCPServerPlugins`), alongside `AgentPlugins` and `MCPServerPlugins`. With no plugin registered the behavior is unchanged — this is plumbing for out-of-tree extensions.

## Reconcile flow

On each `RemoteMCPServer` reconcile, registered plugins run **before** tool discovery so any resources they author (routing, etc.) exist before the discovery dial traverses them. For every object a plugin returns, the reconciler:

- validates the object is non-nil (a plugin returning `nil` fails the reconcile with a clear error rather than panicking in `SetControllerReference`);
- sets the controller reference to the `RemoteMCPServer` (so it is garbage-collected on RMS delete and found by the owner index for pruning — this also rejects cross-namespace objects, since `SetControllerReference` requires same namespace);
- applies it and prunes any previously-owned object of the plugin's `GetOwnedResourceTypes()` that the plugin no longer returns.

Things to note:

- `RemoteMCPServerPlugin` / `RemoteMCPServerOutputs` live in `core/pkg/translator` next to the agent-side `TranslatorPlugin`, so both extension surfaces sit in one place and share the same shape (`Manifest []client.Object` + `GetOwnedResourceTypes()`).
- Registered plugins are invoked **unconditionally** — any feature gating is the plugin's own decision, matching how `AgentPlugins` work.
- A plugin failure fails **only** the `RemoteMCPServer` reconcile (logged, then returned so the controller requeues); it does not block tool discovery or the status update, and it never affects agent translation. The status is still updated from the discovery result before the plugin error is surfaced.
- `NewKagentReconciler` takes the registered plugins; `app.Start` threads `ExtensionConfig.RemoteMCPServerPlugins` through. `GetOwnedResourceTypes()` returns the union of the agent translator's owned types and every registered plugin's owned types, so the owner index `FindOwnedObjects` relies on (for pruning) is registered for extension-authored types too.
- Owned-type **deduplication moved to `SetupOwnerIndexes`** (the indexer layer that actually has the constraint). A type owned by more than one source — e.g. the agent translator and a plugin both owning `ConfigMap` — would otherwise make `IndexField` error on a duplicate field registration. `SetupOwnerIndexes` now skips duplicate Go types via a small `distinctByType` helper, so callers need not pre-deduplicate and `GetOwnedResourceTypes` is a plain union.
- The owner-index key function (`ownerIndexValue`) now keys on the controller-owner **UID** for any owner, dropping the prior `Agent`/`SandboxAgent` Kind allow-list. The allow-list was an index-size optimization, but it silently excluded `RemoteMCPServer`-owned objects — so `FindOwnedObjects` never returned them and the prune step above no-op'd (stale objects leaking). Keying on UID alone is correct because UIDs are unique cluster-wide: an owner's children match by UID regardless of Kind, and foreign-owned objects of the same type are indexed but never match our queries. This removes the per-kind list (and its drift risk) entirely.

# Out of scope

- Any concrete plugin implementation — this PR is the interface and reconcile wiring only.
- Plugin support for resources outside the `RemoteMCPServer`'s own namespace (controller-ref ownership requires same namespace by construction).
- Ordering/dependencies between multiple registered plugins (they run in registration order; no inter-plugin contract).

### DIFF
--- a/go/core/internal/controller/reconciler/mcp_server_reconciler_test.go
+++ b/go/core/internal/controller/reconciler/mcp_server_reconciler_test.go
@@ -117,6 +117,7 @@ func TestReconcileKagentMCPServer_ErrorPropagation(t *testing.T) {
 				[]string{}, // No namespace restrictions for tests
 				nil,
 				false,
+				nil,
 			)
 
 			// Call ReconcileKagentMCPServer
--- a/go/core/internal/controller/reconciler/owned_resource_types_test.go
+++ b/go/core/internal/controller/reconciler/owned_resource_types_test.go
@@ -1,0 +1,69 @@
+package reconciler
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/kagent-dev/kagent/go/api/v1alpha2"
+	agenttranslator "github.com/kagent-dev/kagent/go/core/internal/controller/translator/agent"
+	pkgtranslator "github.com/kagent-dev/kagent/go/core/pkg/translator"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// stubTranslator implements agenttranslator.AdkApiTranslator; only
+// GetOwnedResourceTypes is exercised by these tests.
+type stubTranslator struct{ owned []client.Object }
+
+func (s *stubTranslator) CompileAgent(context.Context, v1alpha2.AgentObject) (*agenttranslator.AgentManifestInputs, error) {
+	return nil, nil
+}
+
+func (s *stubTranslator) BuildManifest(context.Context, v1alpha2.AgentObject, *agenttranslator.AgentManifestInputs) (*agenttranslator.AgentOutputs, error) {
+	return nil, nil
+}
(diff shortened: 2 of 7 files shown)


iplay88keys:
Closing for now until there is a use case for this raised by the community.

PR #1988, opened 2026-06-09
Add CEL validation rule for RemoteMCPServer with CA Cert and AllowedNamespaces

iplay88keys:
# Description

Adds a CEL admission rule on `RemoteMCPServer` that rejects setting both `spec.allowedNamespaces` and `spec.tls.caCertSecretRef`. A `RemoteMCPServer` that pins a CA Secret must be same-namespace-only; one that allows cross-namespace references must not pin a CA Secret.

This closes a silent footgun in the cross-namespace + TLS combination without changing any working behavior: the `AllowedNamespaces` field stays, the reconciler is unchanged, and non-CA cross-namespace sharing keeps working exactly as before.

# Why

When an agent consumes a `RemoteMCPServer` that pins a CA bundle, the translator mounts that Secret onto the **agent's** pod by bare name (`addTLSConfiguration`). A `SecretVolumeSource` has no namespace field — Kubernetes always resolves it in the pod's own namespace, not the `RemoteMCPServer`'s. So if the RMS is referenced cross-namespace, the CA mount either:

- **dangles** — the Secret doesn't exist in the agent's namespace and the pod fails to start; or
- **silently mounts the wrong CA** — an unrelated Secret of the same name that happens to exist in the agent's namespace is trusted for the upstream, with no error anywhere.

Meanwhile the controller validates and hashes the Secret in the `RemoteMCPServer`'s namespace, so it reports `Accepted=true` while the pod mount looks elsewhere. The cert reference itself is structurally same-namespace (`caCertSecretRef` is a bare name with no namespace field, resolved in the owner's namespace by design); the only thing that broke co-location was opening the RMS object up cross-namespace via `allowedNamespaces`. Making the two mutually exclusive removes the unsound combination at admission time rather than letting it surface as a runtime mount failure on the consuming agent.

# Validation rule

Spec-level `XValidation` on `RemoteMCPServerSpec`:

```
rule:    !(has(self.allowedNamespaces) && has(self.allowedNamespaces.from)
            && (self.allowedNamespaces.from == 'All' || self.allowedNamespaces.from == 'Selector')
            && has(self.tls) && has(self.tls.caCertSecretRef) && size(self.tls.caCertSecretRef) > 0)
message: spec.allowedNamespaces must not permit cross-namespace access (from: All or from: Selector)
         when spec.tls.caCertSecretRef is set: a pinned CA Secret is mounted onto the consuming agent's
         pod and Kubernetes resolves it in the agent's namespace, not this RemoteMCPServer's. Use
         from: Same (the default), or remove the CA Secret reference.
```

The rule only rejects the genuinely-hazardous case — a CA Secret combined with an `allowedNamespaces` that actually permits other namespaces (`from: All` or `from: Selector`). `from: Same` (or omitted/empty, which defaults to same-namespace) is always allowed alongside a CA Secret, since it carries no cross-namespace mount hazard.

Regenerated `kagent.dev_remotemcpservers.yaml` and synced the helm CRD copy under `helm/kagent-crds/templates/`.

# Tests

Added cases to the existing envtest CEL suite (`tlsconfig_cel_test.go`), exercised against a real kube-apiserver:

- `allowedNamespaces` `from: All` + `caCertSecretRef` → rejected.
- `allowedNamespaces` `from: Selector` + `caCertSecretRef` → rejected.
- `allowedNamespaces` `from: Same` + `caCertSecretRef` → accepted (same-namespace, no hazard).
- `allowedNamespaces` `from: All`, no TLS → accepted (cross-namespace sharing still works).
- `caCertSecretRef`, no `allowedNamespaces` → accepted (same-namespace CA pinning still works).

# Compatibility

Not a breaking API change — no field is removed and no working configuration changes behavior. The only new effect is at admission: a manifest that combines a CA Secret with a *cross-namespace-permitting* `allowedNamespaces` (`from: All` / `from: Selector`) is now rejected on create/update. That combination was already non-functional (dangling or wrong-CA mount), so this turns a silent runtime failure into an explicit, actionable admission error. Same-namespace configurations (`from: Same` or omitted) are unaffected, and objects already stored are not re-validated until their next write.

### DIFF
--- a/go/api/config/crd/bases/kagent.dev_remotemcpservers.yaml
+++ b/go/api/config/crd/bases/kagent.dev_remotemcpservers.yaml
@@ -59,6 +59,12 @@ spec:
                   This follows the Gateway API pattern for cross-namespace route attachments.
                   If not specified, only Agents in the same namespace can reference this RemoteMCPServer.
                   See: https://gateway-api.sigs.k8s.io/guides/multiple-ns/#cross-namespace-route-attachment
+
+                  Mutually exclusive with spec.tls.caCertSecretRef (enforced by a spec-level
+                  XValidation rule): a pinned CA Secret is mounted onto the consuming agent's
+                  pod by bare name and Kubernetes resolves it in the agent's namespace, not
+                  this RemoteMCPServer's, so a CA-pinning RemoteMCPServer cannot be referenced
+                  cross-namespace and must omit allowedNamespaces.
                 properties:
                   from:
                     default: Same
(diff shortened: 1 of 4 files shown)


EItanya (dismissed):
We can't do this, it's a breaking API change

iplay88keys:
>We can't do this, it's a breaking API change

Makes sense. I have reworked this to be a CEL validation rule instead so you cannot create a RemoteMCPServer with both `spec.allowedNamespaces` and `spec.tls.caCertSecretRef ` set.

EItanya (approved):

PR #1989, opened 2026-06-09
fix(bedrock): stream events incrementally instead of buffering

### LINKED ISSUE #1986: [BUG] Bedrock streaming is buffered - users wait for full response before seeing any text
### 📋 Prerequisites

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

We're using a Bedrock-backed agent with stream: true and noticed our users have to wait 4-5 seconds before seeing
  anything, even for a simple "Hi" message. After some digging through the code, I found the culprit in
  kagent/adk/models/_bedrock.py around line 320:

  ```
def _run_converse_stream(**kw):
      resp = client.converse_stream(**kw)
      return list(resp.get("stream", []))

```
  That list() call collects the entire Bedrock response into memory before anything happens downstream. The rest of the
  streaming chain actually works great (partial LlmResponse events, A2A SSE, the whole thing) but it's all waiting on this
  one function to finish collecting everything first.

### 🔄 Steps To Reproduce

 1. Deploy an agent with stream: true and a Bedrock model config (we use eu.anthropic.claude-sonnet-4-6 in eu-central-1)
  2. Send a simple message like "Hi" through the A2A protocol
  3. Notice nothing appears for 4-5 seconds, then the full answer shows up all at once
  4. For comparison, calling Bedrock's converse_stream API directly gives you the first token in about 500ms

### 🤔 Expected Behavior

Text should start appearing within roughly 500ms (which is Bedrock's actual time-to-first-token), with the rest streaming
  in progressively.

### 📱 Actual Behavior

  The UI is blank for 4-5 seconds, then the complete response appears in one shot. Looking at the A2A events, the
  kagent_adk_partial: true events do get emitted but they all arrive as a burst after buffering, not incrementally as
  Bedrock produces them.

### 💻 Environment

- kagent-adk: 0.3.0
- google-adk: 1.31.1
- a2a-sdk: 0.3.23
- Model: eu.anthropic.claude-sonnet-4-6 via Bedrock (eu-central-1)
- Running on EKS, agent configured with stream: true

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

I think the fix could be fairly contained. Replace the list() buffering with something that bridges boto3's synchronous
  iterator to the async world incrementally. Something along these lines:

  ```
async def _iter_converse_stream(client, **kw):
      queue = asyncio.Queue()

      def _produce():
          resp = client.converse_stream(**kw)
          for event in resp.get("stream", []):
              queue.put_nowait(event)
          queue.put_nowait(None)

      loop = asyncio.get_event_loop()
      loop.run_in_executor(None, _produce)
      while (event := await queue.get()) is not None:
          yield event
```

  Or whatever pattern fits the project's conventions better. The important thing is just not materializing the whole stream
  upfront.

  All the downstream plumbing (event converter, A2A event queue, SSE to client) already handles partial events correctly.
  It's really just this one spot that's holding things up.

  Happy to test a fix on our cluster if you want to point me at a branch.

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

mesutoezdil:
Fixes #1986

`_run_converse_stream` used `list()` to collect all stream events before returning. This blocked the caller until the full response was ready.

Replaced with an `asyncio.Queue` bridge so events are forwarded as they arrive from boto3.


mesutoezdil:
Hey @paerts, can you test this on your cluster? Want to confirm the first token arrives within ~500ms with this change.

paerts:
Hey @mesutoezdil, tested this on our cluster. The fix works.

  Before (unpatched, June 9):
  - Simple "Hi" message: 5-7s of nothing, then full response appears at once
  - Bedrock stream was fully buffered before any SSE events reached the client

  After (patched, June 10):
  - HTTP 200 + stream opens within ~70-80ms
  - Text starts flowing to the user as Bedrock generates it
  - Total response time is the same, but perceived latency is way better

  The remaining delay our users see (2-4s before first visible text) is due to the agent calling tools (memory service lookups) before generating its text response.
  That's expected behavior, not a streaming issue.

  Confirmed on:
  - kagent-adk 0.3.0 with your patch applied
  - Bedrock eu-central-1, claude-sonnet-4-6
  - Streamed via A2A SSE through OpenWebUI pipe

  Nice fix, thanks for the quick turnaround!

mesutoezdil:
> Hey @mesutoezdil, tested this on our cluster. The fix works.
> 
> Before (unpatched, June 9):
> 
> * Simple "Hi" message: 5-7s of nothing, then full response appears at once
> * Bedrock stream was fully buffered before any SSE events reached the client
> 
> After (patched, June 10):
> 
> * HTTP 200 + stream opens within ~70-80ms
> * Text starts flowing to the user as Bedrock generates it
> * Total response time is the same, but perceived latency is way better
> 
> The remaining delay our users see (2-4s before first visible text) is due to the agent calling tools (memory service lookups) before generating its text response. That's expected behavior, not a streaming issue.
> 
> Confirmed on:
> 
> * kagent-adk 0.3.0 with your patch applied
> * Bedrock eu-central-1, claude-sonnet-4-6
> * Streamed via A2A SSE through OpenWebUI pipe
> 
> Nice fix, thanks for the quick turnaround!

thx, good to hear!

jmhbh (approved):
lgtm thanks @mesutoezdil!

PR #1998, opened 2026-06-11
fix(controller): decouple skills from privileged sandbox

### LINKED ISSUE #1997: Skills path sets privileged: true unconditionally, incompatible with restricted-PSS clusters
## Problem

On clusters with restricted Pod Security Standards, declarative agents with `spec.skills` set never reach `Accepted: True`. The controller sets `privileged: true` on the main container when skills are present, and restricted-PSS rejects it.

## Why

The `skills-init` init container handles git clone and OCI image pull — the main container doesn't need elevated privileges for skills loading. `privileged: true` is only used for the in-pod BashTool srt/bubblewrap sandbox.

Tying the two together means you can't use skills on any restricted-PSS cluster, even if you don't use the BashTool or don't need the in-pod sandbox (e.g. you call tools via MCP instead).

## Current workaround

Set `allowPrivilegeEscalation: false` on the Agent CR. The controller already has a guard (`allowPrivilegeEscalationExplicitlyFalse`) that suppresses `privileged: true` when this is set:

```yaml
spec:
  declarative:
    deployment:
      securityContext:
        allowPrivilegeEscalation: false
```

This works but requires every skills-using Agent CR to carry the override, and it silently disables the BashTool sandbox without any indication to the user.

## Requested change

Decouple skills from the sandbox: don't set `privileged: true` unless the sandbox is explicitly requested. The existing guard already implements the opt-out direction — the default just needs to flip so restricted-PSS clusters work out of the box without silently losing the sandbox.

mesutoezdil:
Skills are loaded by the init container. The main container does not
need privileged=true for skill loading.

Before this change, any agent with skills got privileged=true on the
main container, which breaks restricted PSS clusters.

Now only BashTool sandbox (cfg.GetExecuteCode) sets privileged=true.

Fixes #1997

### DIFF
--- a/go/core/internal/controller/translator/agent/manifest_builder.go
+++ b/go/core/internal/controller/translator/agent/manifest_builder.go
@@ -302,8 +302,11 @@ func buildPodRuntime(
 	volumeMounts := append([]corev1.VolumeMount{}, secretMounts...)
 	volumeMounts = append(volumeMounts, manifestCtx.deployment.VolumeMounts...)
 
-	needCodeExecIsolation := cfg != nil && cfg.GetExecuteCode()
-	initContainers, skillsInitCM, err := buildSkillsRuntime(manifestCtx, &sharedEnv, &volumes, &volumeMounts, &needCodeExecIsolation)
+	// privileged is required whenever srt (the sandbox runtime) is present in the image:
+	// skills use BashTool which calls srt, and code execution also calls srt.
+	// needsSRTSettings captures exactly those cases.
+	needCodeExecIsolation := needsSRTSettings(manifestCtx.agent, sandboxCfg)
+	initContainers, skillsInitCM, err := buildSkillsRuntime(manifestCtx, &sharedEnv, &volumes, &volumeMounts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


jmhbh (commented):
LGTM! makes sense to me

jmhbh (dismissed):

EItanya (changes requested):
The issue here is more nuanced. There are 2 features in our API, code execution and skills. Skills may execute code even if code_execution is not set. code_execution on the API specifically refers to executing code generated by the LLM, not code within skills themselves. We should probably deprecate the "code_execution" API feature altogether because it's very confusing

mesutoezdil:
> The issue here is more nuanced. There are 2 features in our API, code execution and skills. Skills may execute code even if code_execution is not set. code_execution on the API specifically refers to executing code generated by the LLM, not code within skills themselves. We should probably deprecate the "code_execution" API feature altogether because it's very confusing

skills run entirely in the init container, so whatever code they execute happens there, not in the main container. i try here only removes privileged from the main container, the init container is unaffected. is not correct?

EItanya:
> > The issue here is more nuanced. There are 2 features in our API, code execution and skills. Skills may execute code even if code_execution is not set. code_execution on the API specifically refers to executing code generated by the LLM, not code within skills themselves. We should probably deprecate the "code_execution" API feature altogether because it's very confusing
> 
> skills run entirely in the init container, so whatever code they execute happens there, not in the main container. i try here only removes privileged from the main container, the init container is unaffected. is not correct?

No, skills are downloaded in the init container, but RUN in the main container

EItanya:
Closing as per our conversation

PR #2001, opened 2026-06-11
implement agent reliability features from ADK

peterj:
Adds reliability features toagents and ModelConfig. 

Agent CRD has the `reliability` field with the following config:

- **`toolRetries`** (1–10): reflect-and-retry on failed tool calls — the runtime injects structured reflection guidance into the model context so the agent can self-correct instead of repeating the same failing call
- **`maxLLMCalls`** (≥1): cost safety rail capping model calls per request; the run stops with a clear error instead of looping (runtime default 500)
- **`debugLogging`**: verbose logging of every LLM request/response and tool call to the agent pod logs (off by default)

ModelConfig `retry` field:

- **`attempts`** (0–20): automatic retries of failed LLM HTTP requests (429/408/5xx) with exponential backoff via the provider SDK. Supported for OpenAI, Azure OpenAI, Anthropic, and Gemini; other providers log a warning

UI is updated to show the settings in the advanced section:

Screenshot 2026-06-11 at 4 11 52 PM


peterj:
@copilot resolve the merge conflicts in this pull request

PR #2005, opened 2026-06-12
test(translator): tighten skills security context assertions

mesutoezdil:
Skills mounting files via an init container do not require a privileged container. That requirement belongs to ExecuteCode (the code sandbox). The previous behavior set needCodeExecIsolation = true inside buildSkillsRuntime as a side effect, causing any agent with skills to get Privileged: true even when code execution was disabled. This change decouples the two: only cfg.GetExecuteCode() drives needCodeExecIsolation, and skills set up their own volumes without touching that flag.

Changes:

- Remove the needCodeExecIsolation pointer parameter from buildSkillsRuntime so skills no longer set the privileged flag as a side effect.
- Update golden outputs to remove the privileged security context from skills-based agent manifests.
- Replace the weak if containerSecurityContext != nil guard with assert.Nil so tests fail when a security context is unexpectedly present.
- Simplify multi-line comment blocks to single lines in TestSecurityContext_SkillsNoPrivileged and TestSecurityContext_SkillsPSSRestricted.


go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -297,7 +297,7 @@ func buildPodRuntime(
 	volumeMounts = append(volumeMounts, manifestCtx.deployment.VolumeMounts...)
 
 	needCodeExecIsolation := cfg != nil && cfg.GetExecuteCode()
-	initContainers, skillsInitCM, err := buildSkillsRuntime(manifestCtx, &sharedEnv, &volumes, &volumeMounts, &needCodeExecIsolation)
+	initContainers, skillsInitCM, err := buildSkillsRuntime(manifestCtx, &sharedEnv, &volumes, &volumeMounts)
```
EItanya, line 300: Why was the behavior here changed, I don't see mention of it in the PR description

mesutoezdil: done

EItanya:
> Skills mounting files via an init container do not require a privileged container. That requirement belongs to ExecuteCode (the code sandbox). The previous behavior set needCodeExecIsolation = true inside buildSkillsRuntime as a side effect, causing any agent with skills to get Privileged: true even when code execution was disabled. This change decouples the two: only cfg.GetExecuteCode() drives needCodeExecIsolation, and skills set up their own volumes without touching that flag.

Skills can execute code though, that's part of the point

mesutoezdil:
> > Skills mounting files via an init container do not require a privileged container. That requirement belongs to ExecuteCode (the code sandbox). The previous behavior set needCodeExecIsolation = true inside buildSkillsRuntime as a side effect, causing any agent with skills to get Privileged: true even when code execution was disabled. This change decouples the two: only cfg.GetExecuteCode() drives needCodeExecIsolation, and skills set up their own volumes without touching that flag.
> 
> Skills can execute code though, that's part of the point

hm, skills run trusted pythn packs installed via the init container, they execute inside the normal python runtime without needing a sandbox. and needcodeexecisolation drives the gvisor sandbox, which is for untrusted arbitrary code execution (executedode). the 2 are separate concerns?

EItanya:
> > > Skills mounting files via an init container do not require a privileged container. That requirement belongs to ExecuteCode (the code sandbox). The previous behavior set needCodeExecIsolation = true inside buildSkillsRuntime as a side effect, causing any agent with skills to get Privileged: true even when code execution was disabled. This change decouples the two: only cfg.GetExecuteCode() drives needCodeExecIsolation, and skills set up their own volumes without touching that flag.
> > 
> > 
> > Skills can execute code though, that's part of the point
> 
> hm, skills run trusted pythn packs installed via the init container, they execute inside the normal python runtime without needing a sandbox. and needcodeexecisolation drives the gvisor sandbox, which is for untrusted arbitrary code execution (executedode). the 2 are separate concerns?

Why don't they need a sandbox? Skills very often allow the agents to generate code files to run. They also use `srt` today. This will change with substrate, but that's how it works today.

mesutoezdil:
> > > > Skills mounting files via an init container do not require a privileged container. That requirement belongs to ExecuteCode (the code sandbox). The previous behavior set needCodeExecIsolation = true inside buildSkillsRuntime as a side effect, causing any agent with skills to get Privileged: true even when code execution was disabled. This change decouples the two: only cfg.GetExecuteCode() drives needCodeExecIsolation, and skills set up their own volumes without touching that flag.
> > > 
> > > 
> > > Skills can execute code though, that's part of the point
> > 
> > 
> > hm, skills run trusted pythn packs installed via the init container, they execute inside the normal python runtime without needing a sandbox. and needcodeexecisolation drives the gvisor sandbox, which is for untrusted arbitrary code execution (executedode). the 2 are separate concerns?
> 
> Why don't they need a sandbox? Skills very often allow the agents to generate code files to run. They also use `srt` today. This will change with substrate, but that's how it works today.

y r right, missed the srt dependency. should i close this and reopen after substrate replaces srt?

PR #2021, opened 2026-06-15
feat(controller): make Go (ADK) runtime agent image independently configurable

### LINKED ISSUE #2018: [FEATURE] Make the Go runtime agent image independently configurable
### 📋 Prerequisites

### 📝 Feature Summary

Allow the Go (ADK) runtime agent image to be configured independently, instead of deriving its repository from the Python agent image repository.

When a declarative `Agent` uses `spec.declarative.runtime: go`, the controller does **not** read a dedicated image setting for the Go runtime. Instead it *derives* the Go repository from the configured Python repository (`controller.agentImage.repository`, surfaced as `IMAGE_REPOSITORY`) by replacing the last path segment with `golang-adk`.

This derivation breaks for any registry layout where the Python image repository is a **flat name** (org/namespace carried in the registry, not in the repository path) — a very common pattern for mirrored/proxy registries.

We would like to mirror the images in our registry, using a consistent name prefix `kagent-`. This is currently not configurable.

### 💡 Proposed Solution

Add a dedicated, independent image configuration for the Go runtime — mirroring how the Python agent image and skills-init image are already configured — instead of deriving it.

Concretely:

- New Helm values, e.g. `controller.goAgentImage.{registry,repository,tag,pullPolicy}`, defaulting to today's derived behavior so existing setups don't change.
- New controller config keys surfaced from those values, e.g. `GO_IMAGE_REGISTRY` / `GO_IMAGE_REPOSITORY` / `GO_IMAGE_TAG` (parallel to the existing `IMAGE_*` and `SKILLS_INIT_IMAGE_*` keys in the `kagent-controller` ConfigMap).
- `getRuntimeImageRepository` uses the explicit Go repository when set, and only falls back to the current derive-from-Python logic when it is not — preserving backward compatibility.

This makes the Go runtime image a first-class, configurable input rather than a value inferred from an unrelated repository's path structure.

### 🔄 Alternatives Considered

- **Mirror the Go image as plain `golang-adk`** to match what the controller derives. Works as an immediate unblock, but forces an inconsistent naming exception for this one image and doesn't fix the underlying assumption for other registry layouts.
- **Restructure `controller.agentImage.repository` into a path form** (e.g. `kagent/app`) so the last-segment replacement yields `kagent/golang-adk`. Still cannot produce a hyphen-prefixed flat name like `kagent-golang-adk`, and changes the Python image path as a side effect.
- **Make the derivation prefix-aware** (preserve a `kagent-`-style prefix when replacing the suffix). Fragile — it bakes one naming convention into the controller. Explicit configuration is more general and predictable.

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

- Source: `go/core/internal/controller/translator/agent/deployments.go` (`getRuntimeImageRepository`).
- Existing comparable configuration that this would follow: `controller.agentImage.*` and `controller.skillsInitImage.*` → `IMAGE_*` / `SKILLS_INIT_IMAGE_*` keys in `helm/kagent/templates/controller-configmap.yaml`.
- Observed on kagent `0.9.6`.

QuentinBisson:
Fixes #2018.

## What

Adds `DefaultGoImageConfig` (registry, repository, tag, pullPolicy) so the Go (ADK) runtime agent image is configured independently instead of being derived from the Python repository by string surgery. Exposed end-to-end as:

- `--go-image-{registry,repository,tag,pull-policy}` controller flags (auto-mapped to `GO_IMAGE_*` env vars via `LoadFromEnv`)
- `controller.goAgentImage.{registry,repository,tag,pullPolicy}` Helm values
- `GO_IMAGE_*` ConfigMap keys, following the same defaulting as `agentImage`/`skillsInitImage` (registry/pullPolicy default to the globals, tag coalesces with the global tag then the Chart version)

Defaults are encoded in the variable (`ghcr.io/kagent-dev/kagent/golang-adk`), matching `DefaultImageConfig` and `DefaultSkillsInitImageConfig`. Since #2242, regular declarative agents reference the image by tag; sandbox agents still pin by digest via `GoADKImageDigest` / `GoADKFullImageDigest`. The `-full` variant selection is unchanged for both paths.

## Why

When `spec.declarative.runtime: go` is used, the controller derived the Go runtime repository from the Python repository by replacing the last path segment with `golang-adk`. This breaks for flat-name registry layouts (e.g. a mirrored registry where the image is `my-registry.io/kagent-golang-adk`, which the derivation cannot produce).

## Breaking change

Operators who mirror images and only set `agentImage` no longer get a derived Go image location; they must set `controller.goAgentImage` as well. The controller logs a startup warning when the two registries differ so a misconfigured mirror is visible before a Go agent fails to pull. As discussed in the review thread, this targets a minor release (0.10.x).

## Files changed

| File | Change |
|------|--------|
| `go/core/internal/controller/translator/agent/adk_api_translator.go` | `DefaultGoImageConfig` with encoded defaults |
| `go/core/internal/controller/translator/agent/deployments.go` | Go image resolution via `DefaultGoImageConfig`; tag threaded through `resolveRuntimeImage` |
| `go/core/pkg/app/app.go` | four `--go-image-*` flags + registry-mismatch startup warning |
| `helm/kagent/values.yaml` | `controller.goAgentImage` block |
| `helm/kagent/templates/controller-configmap.yaml` | `GO_IMAGE_*` keys |
| `go/core/internal/controller/translator/agent/runtime_test.go`, `imageconfig_test.go` | tests updated to tag-based expectations |


go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -122,6 +122,13 @@ var PythonADKImageDigest string
 var GoADKImageDigest string
 var GoADKFullImageDigest string
 
+// DefaultGoImageConfig overrides the image used for the Go (ADK) runtime agent.
+// Any field left empty falls back to DefaultImageConfig for registry/pullPolicy, or to
+// the derived repository (last segment of DefaultImageConfig.Repository replaced with
+// "golang-adk") when Repository is empty. Tag is unused — Go runtime images are
+// digest-pinned at controller link time via GoADKImageDigest / GoADKFullImageDigest.
+var DefaultGoImageConfig = ImageConfig{}
```
iplay88keys: Why don't we just encode the defaults into this variable like we do for the other image configs?

QuentinBisson: Hey @iplay88keys I wanted to preserve the issue's framing.

The empty default was there to avoid a breaking change: the controller currently derives the Go image from agentImage.repository, so anyone mirroring images only sets that one value and the Go image follows. If we hardcode the default and always emit `GO_IMAGE_REPOSITORY`, their next helm upgrade silently repoints the Go image back to `cr.kagent.dev` and pods fail to pull. 

I do agree that this is not the better change though and I will replace this PR with the way it is done for the other images.

iplay88keys: I see, is there another workaround we can build in? Maybe something like: if they are overriding the image for the python image, but not for the go image, throw an error/warning?

QuentinBisson: I think we should be able to add a log warning if the 2 registries are different yes

iplay88keys: @EItanya thoughts on the breaking change aspect? I'm thinking it's fine if we document it and we'd likely want that to go in on a `minor` release to draw attention to the breaking nature of that change.

supreme-gg-gg: I agree with @iplay88keys suggestion, this would likely go into the next minor release where we will make go runtime by default and it would be helpful to be able to configure go runtime images. So I think the breaking change would be fine as it's more clear in the long run

EItanya: Agreed, we can put it in `0.10.x`

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -124,6 +124,15 @@ var PythonADKFullImageDigest string
 var GoADKImageDigest string
 var GoADKFullImageDigest string
 
+// DefaultGoImageConfig is the image config for the Go (ADK) runtime agent.
+// Tag is unused: Go runtime images are digest-pinned at controller link time
+// via GoADKImageDigest / GoADKFullImageDigest.
+var DefaultGoImageConfig = ImageConfig{
+	Registry:   "cr.kagent.dev",
```
iplay88keys: This should be changed to `ghcr.io`

iplay88keys (commented):
There are a few changes that will need to be made to work with `main`. The biggest is that regular declarative agents are now referenced by tag and digests are only used for sandbox agents.

go/core/internal/controller/translator/agent/adk_api_translator.go
QuentinBisson: Done. Rebased on main and moved the defaults to ghcr.io/kagent-dev/kagent/golang-adk. Also adapted to #2242: regular Go agents are now tag-referenced (new tag field on goAgentImage, --go-image-tag, GO_IMAGE_TAG), sandbox agents still pin by digest.

go/core/pkg/app/app.go
```diff
@@ -218,6 +218,9 @@ func (cfg *Config) SetFlags(commandLine *flag.FlagSet) {
...
 	commandLine.StringVar(&agent_translator.DefaultSkillsInitImageConfig.PullPolicy, "skills-init-image-pull-policy", agent_translator.DefaultSkillsInitImageConfig.PullPolicy, "The pull policy to use for the skills init image.")
 	commandLine.StringVar(&agent_translator.DefaultSkillsInitImageConfig.Repository, "skills-init-image-repository", agent_translator.DefaultSkillsInitImageConfig.Repository, "The repository to use for the skills init image.")
+	commandLine.StringVar(&agent_translator.DefaultGoImageConfig.Registry, "go-image-registry", "", "The registry to use for the Go (ADK) runtime agent image. When empty, falls back to --image-registry.")
+	commandLine.StringVar(&agent_translator.DefaultGoImageConfig.Repository, "go-image-repository", "", "The repository to use for the Go (ADK) runtime agent image. When empty, derived from --image-repository by replacing the last path segment with \"golang-adk\".")
+	commandLine.StringVar(&agent_translator.DefaultGoImageConfig.PullPolicy, "go-image-pull-policy", "", "The pull policy to use for the Go (ADK) runtime agent image. When empty, falls back to --image-pull-policy.")
```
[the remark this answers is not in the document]
QuentinBisson: Fixed: the flags now use the DefaultGoImageConfig fields as StringVar defaults, same as the skills-init-image-* flags. Defaults are encoded in the config since the derivation fallback was dropped.

go/core/internal/controller/translator/agent/runtime_test.go
```diff
@@ -487,3 +487,215 @@ func TestRuntime_CustomRepositoryPath_WithSkillsUsesFullTag(t *testing.T) {
...
+			deployment = dep
+			break
+		}
+	}
+	require.NotNil(t, deployment)
+
+	img := deployment.Spec.Template.Spec.Containers[0].Image
+	assert.Contains(t, img, "/kagent-golang-adk@", "Image should use the explicit flat repository")
+	assert.NotContains(t, img, "/golang-adk@", "Derived path must not appear when repository is explicitly set")
+	assert.Contains(t, img, "@sha256:test-go-base")
```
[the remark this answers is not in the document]
QuentinBisson: Fixed: the tests now assert equality on the full image reference (e.g. my-registry.io/kagent-golang-adk:v0.0.0-test) instead of substring matching.

go/core/internal/controller/translator/agent/deployments.go
```diff
@@ -175,8 +171,21 @@ func resolveInlineDeployment(agent v1alpha2.AgentObject, mdd *modelDeploymentDat
...
+	baseRegistry := DefaultImageConfig.Registry
+	basePullPolicy := DefaultImageConfig.PullPolicy
+	if runtime == v1alpha2.DeclarativeRuntime_Go {
+		if DefaultGoImageConfig.Registry != "" {
+			baseRegistry = DefaultGoImageConfig.Registry
+		}
+		if DefaultGoImageConfig.PullPolicy != "" {
+			basePullPolicy = DefaultGoImageConfig.PullPolicy
+		}
+	}
```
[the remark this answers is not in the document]
QuentinBisson: Added TestRuntime_GoImageConfig_ExplicitPullPolicy: asserts the Go deployment uses the explicit pull policy and the Python deployment does not inherit it.

EItanya (approved):

PR #2034, opened 2026-06-17
fix(ui): avoid false stale chat send guard

### LINKED ISSUE #2033: UI chat: "New messages loaded" notification can incorrectly block sending the next message
When using the kagent UI chat, attempting to send a new message can show this notification instead of sending the message:

> New messages loaded — please review before sending

This happens even when the chat appears up to date and there are no obvious new messages that need review. From the user's perspective, the conversation is ready, but the first send attempt is blocked. Retrying the same send action usually succeeds.

This has been observed during normal same-tab usage, after receiving a completed response and then trying to send the next message.

## Screenshot

Image

## Steps to reproduce

1. Open the kagent UI in Chrome on macOS.
2. Start or open a chat session with an agent.
3. Send a message and wait for the agent response to complete.
4. Type another message.
5. Try to send it using either:
   - `Cmd+Enter`, or
   - the **Send** button.
6. Observe that the message may not send. Instead, a toast appears:

   > New messages loaded — please review before sending

7. Try sending the same message again.
8. The second attempt usually succeeds.

## Expected behavior

If the user is already viewing the latest messages in the current chat, pressing **Send** or using `Cmd+Enter` should send the message immediately.

The "New messages loaded" warning should only appear when there are actually unseen messages or session changes that the user needs to review before sending.

## Actual behavior

The UI sometimes blocks the first send attempt and shows:

> New messages loaded — please review before sending

The user then has to retry the send action for the message to go through.

## Impact

This makes the chat input feel unreliable because a normal send action can appear to fail even though the chat is ready. It affects both keyboard submission and clicking the **Send** button.

## Workaround

Retry the send action after the notification appears. The second attempt usually succeeds.

## Environment

- kagent version / commit: latest `main` as of 2026-06-16, observed around `67431ef6`
- Deployment method: kagent UI from current upstream/main build
- Browser: Chrome
- OS: macOS
- Chat path: regular kagent UI chat session
- Repro context: most noticeable after an agent response completes and the user sends the next message in the same chat

## Additional notes

The warning is useful when another tab or client updates the session and the user genuinely needs to review new content before sending. The problem is that it can also appear during normal same-tab usage where the latest response is already visible.

erauner12:
Fixes a false-positive chat send guard that can show `New messages loaded — please review before sending` during normal same-tab usage after an agent response has already completed and is visible.

The send guard now compares backend-derived messages against all locally visible comparable messages instead of only persisted/stored messages. This preserves the intended cross-tab stale-message warning while allowing the next same-tab send to proceed.

Fixes #2033.

## Changes

- Count all locally visible comparable chat messages when deciding whether the backend has unseen messages.
- Apply the same comparable-message filtering to backend-derived messages before comparing counts.
- Add mounted regression tests covering:
  - same-tab completed stream does not block the next send
  - true cross-tab/unseen messages still block
  - cross-tab/unseen messages still block after a same-tab stream

## Validation

Automated:

```bash
cd ui
npm test -- --runTestsByPath src/components/chat/__tests__/ChatInterface.sendGuard.test.tsx --runInBand
```

Result:

```text
Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
```

Previously also run successfully during branch validation:

```text
src/lib/__tests__/messageHandlers.test.ts: 27 passed
Full UI Jest suite: 268 passed
```

Manual:

- Deployed/tested locally against `kind-kagent`.
- Verified the next same-tab message sends without the false `New messages loaded` toast after a completed agent response.


mesutoezdil:
1 small thing worth adding: a 1-line comment on why `ToolCallSummaryMessage` is in the excluded set. It is the only newly added type and not obvious from the name alone whether it can indicate real unseen context.

ui/src/lib/messageHandlers.ts
```diff
@@ -448,6 +448,88 @@ export interface ADKMetadata {
...
+  "AskUserRequest", // ask_user UI duplicates the pending backend task state.
+  "ToolCallSummaryMessage", // UI-only marker that closes tool calls; not a backend chat turn.
+]);
+
+function isSendGuardComparableMessage(message: Message): boolean {
+  const meta = message.metadata as ADKMetadata | undefined;
+  return !meta?.originalType || !SEND_GUARD_EXCLUDED_ORIGINAL_TYPES.has(meta.originalType);
+}
+
+export function countSendGuardComparableMessages(messages: Message[]): number {
```
peterj, line 464: nit: you could do this with a oneliner :) (typescript can sometimes feel like python :D)

```
messages.filter(isSendGuardComparableMessage).length;
```

ui/src/lib/messageHandlers.ts
```diff
@@ -448,6 +448,88 @@ export interface ADKMetadata {
...
+  for (const message of messages) {
+    if (isSendGuardComparableMessage(message)) {
+      count += 1;
+    }
+  }
+  return count;
+}
+
+function getSendGuardMessageContentSignature(message: Message): string {
+  const meta = message.metadata as ADKMetadata | undefined;
```
peterj, line 469: nit -- can you pull out the `\u0001` and `\u0000` as consts within the function, so it doesn't look like random strings.

peterj: actually, pull them out of the function, since they are using in other functions too

erauner12: addressed

ui/src/lib/messageHandlers.ts
erauner12: addressed :)

peterj (approved):

PR #2039, opened 2026-06-17
fix(helm): rename $defaultProfider to $defaultProvider

mesutoezdil:
Fixes a typo in `helm/kagent/templates/modelconfig.yaml`. The variable was named `$defaultProfider` instead of `$defaultProvider`.


jmhbh (approved):
nice catch

mesutoezdil:
> nice catch

stumbled upon it by chance :)

EItanya (approved):

PR #2041, opened 2026-06-17
fix(substrate): return 200 with empty result when ate.dev CRDs are not installed

### LINKED ISSUE #2024: GET /api/substrate/status returns 500 when ate.dev CRDs are not installed
### Description

On clusters where the `agent-substrate` project is not deployed (CRDs `ate.dev/v1alpha1`
WorkerPool / ActorTemplate are absent), `GET /api/substrate/status` returns HTTP 500.

### Root cause

`listSubstrateCRs` (substrate.go) calls `KubeClient.List` on `WorkerPoolList` and
`ActorTemplateList` unconditionally. When the CRDs are missing the REST mapper returns
a `*meta.NoKindMatchError` which propagates to an HTTP 500.

### Expected behaviour

The endpoint should return 200 with `Enabled: false` and empty resource lists. Substrate
support is optional and `Enabled` is already controlled by `AteClient != nil`.

Fix: treat `meta.IsNoMatchError` as an empty result in `listSubstrateCRs`.

Reproducer: deploy kagent without `agent-substrate`; open the Chat UI (it calls
`/api/substrate/status` on load). The "New Chat" input fails to render ("Failed to
fetch servers").

mesutoezdil:
Closes #2024

`listSubstrateCRs` calls `KubeClient.List` on `WorkerPoolList` and `ActorTemplateList` unconditionally. When `ate.dev` CRDs are not installed, the REST mapper returns a `NoKindMatchError` which propagates as HTTP 500.

Treat `meta.IsNoMatchError` as an empty result so the endpoint returns 200 with `Enabled: false` on clusters without agent-substrate.

### DIFF
--- a/go/core/internal/httpserver/handlers/substrate.go
+++ b/go/core/internal/httpserver/handlers/substrate.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kagent-dev/kagent/go/core/internal/httpserver/errors"
 	"github.com/kagent-dev/kagent/go/core/pkg/auth"
 	"github.com/kagent-dev/kagent/go/core/pkg/sandboxbackend/substrate"
+	"k8s.io/apimachinery/pkg/api/meta"
 	utilvalidation "k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


mesutoezdil:
duplicate of #2023

PR #2042, opened 2026-06-17
fix(translator): honor APIKeyPassthrough for Gemini provider

mesutoezdil:
The Gemini provider in the model translator was not checking `APIKeyPassthrough`. It always tried to mount the Google API key secret, even when passthrough mode was enabled. It also never set `APIKeyPassthrough` on the ADK model struct. All other providers (OpenAI, Anthropic, AzureOpenAI, Ollama) already handle this correctly.

The CRD validation also rejected `apiKeyPassthrough: true` for Gemini, so the flag could never be enabled at all. Gemini uses plain API key auth like OpenAI and Anthropic, so passthrough is valid for it. The rule now only blocks the Vertex providers, which authenticate with service accounts. CRD manifests regenerated.

On the runtime side, `KAgentGeminiLlm` now accepts the Bearer token from the incoming request via `set_passthrough_key` and uses it as the Gemini API key, falling back to `GOOGLE_API_KEY` or `GEMINI_API_KEY` env vars as before. Setting a new token invalidates the cached API clients. Covered by new unit tests.


EItanya (commented):
Makes sense, is passthrough implemented for Gemini in our adk + adk go?

mesutoezdil:
> Makes sense, is passthrough implemented for Gemini in our adk + adk go?

hm not fully. the python adk has the api_key_passthrough field on kagentgeminillm but is missing set_passthrough_key(), so llmassthroughplugin skips gemini at runtime. go adk has no gemini model implementation at all. my pr fixes the translator side only. 
a follow-up is needed in the python adk to add set_passthrough_key() to kagentgeminillm.

EItanya:
> > Makes sense, is passthrough implemented for Gemini in our adk + adk go?
> 
> hm not fully. the python adk has the api_key_passthrough field on kagentgeminillm but is missing set_passthrough_key(), so llmassthroughplugin skips gemini at runtime. go adk has no gemini model implementation at all. my pr fixes the translator side only. a follow-up is needed in the python adk to add set_passthrough_key() to kagentgeminillm.

If this doesn't fully work then can we add those fixes?

mesutoezdil:
> > > Makes sense, is passthrough implemented for Gemini in our adk + adk go?
> > 
> > 
> > hm not fully. the python adk has the api_key_passthrough field on kagentgeminillm but is missing set_passthrough_key(), so llmassthroughplugin skips gemini at runtime. go adk has no gemini model implementation at all. my pr fixes the translator side only. a follow-up is needed in the python adk to add set_passthrough_key() to kagentgeminillm.
> 
> If this doesn't fully work then can we add those fixes?

done

EItanya:
> go adk has no gemini model implementation at all

How can this be true, seems like a bigger issue than passthrough in python

mesutoezdil:
> > go adk has no gemini model implementation at all
> 
> How can this be true, seems like a bigger issue than passthrough in python

correction: Go ADK does have Gemini (agent.go:215). What's missing is passthrough support there. it always requires env vars regardless of ApiKeyPassthrough. This PR covers translator + Python ADK; Go ADK passthrough is a separate follow-up.

mesutoezdil:
closing. go/core/internal/controller/translator/agent is gone on main, the whole v1 translator moved to v2. the fix this covers is already there: model.go sets gemini.APIKeyPassthrough from model.Spec.APIKeyPassthrough same as every other provider.

PR #2043, opened 2026-06-17
feat: substrate support for BYO and python runtimes for SandboxAgent CR

jmhbh:
- Implements substrate support for BYO typed agents and the python runtime for declarative agents. 
- Fixes an existing issue where on config changes the previous golden snapshot was still being served. We now correctly serve the newest golden snapshot with the updated config.
- Fixes an issue where golden snapshot building could use stale config. Substrate caches config values based on the `KAGENT_CONFIG_JSON` secret key ref name which is static. This can result in stale config being fetched. The secret key ref name now contains the config hash as a suffix so different configs now have a unique cache key.

**Testing**
- Tested locally in a kind cluster. Below is an example conversation with a declarative sandbox agent using the python runtime where I switch between model providers to test config rollout.
Screenshot 2026-06-22 at 4 27 29 PM


Makefile
```diff
@@ -60,14 +60,18 @@ SKILLS_INIT_IMAGE_NAME ?= skills-init
 CONTROLLER_IMAGE_TAG ?= $(VERSION)
 UI_IMAGE_TAG ?= $(VERSION)
 APP_IMAGE_TAG ?= $(VERSION)
+APP_FULL_IMAGE_TAG ?= $(VERSION)-full
```
jmhbh, line 66: as a part of moving away from the base chainguard images since we currently cannot pin the images we are introducing a `-full` suffixed tag which uses node:20-bookworm-slim as the base image to include shell, apt, and the sandbox runtime. This mirrors the existing pattern for the golang images. The existing unsuffixed flavor now uses a distroless base image instead.

go/api/v1alpha2/agent_types.go
```diff
@@ -282,16 +282,10 @@ func EffectiveDeclarativeRuntime(spec *AgentSpec) DeclarativeRuntime {
 }
 
 // EffectiveDeclarativeRuntimeForAgent returns the runtime for a reconciled agent object.
-// Substrate SandboxAgents always use Go; regular Agents honor spec.declarative.runtime.
+// All agents (including substrate SandboxAgents) honor spec.declarative.runtime, defaulting
+// to Python when unset.
 func EffectiveDeclarativeRuntimeForAgent(agent AgentObject) DeclarativeRuntime {
```
EItanya: This feels like an unnecessary wrapper now?

python/Dockerfile
```diff
@@ -1,140 +1,101 @@
...
+# materialized into /config at startup (the env-injected path); distroless runs as nonroot and
+# cannot create top-level dirs, so it must exist with the right owner ahead of time. On the
+# normal Deployment path /config is overlaid by the mounted Secret volume, so this is harmless.
+RUN mkdir -p /staging/config
+
+### STAGE 3: final (distroless)
+# distroless/cc provides glibc + libstdc++ (required by the standalone CPython build) but no
+# shell or package manager. Agents that need in-container code execution / bash tools use the
+# "full" image (python/Dockerfile.full) instead.
+FROM gcr.io/distroless/cc-debian12:nonroot
```
EItanya: Can we pick a tag/digest for this so we don't randomly get upgraded?

go/core/internal/controller/sandboxagent_substrate.go
```diff
@@ -33,9 +33,45 @@ func (r *SandboxAgentController) reconcileSubstrateSandboxAgent(ctx context.Cont
...
 	return ctrl.Result{}, nil
 }
 
+// reconcileSubstrateBlueGreen drives the cleanup half of a config-change rollout: it retires
+// ActorTemplates superseded by a newer Ready one (deleting each old template with its now-Suspended
+// golden), and best-effort reaps stale per-session actors. The new template keeps serving the old
+// golden until its own golden is Ready (see ResolveCurrentActorTemplate), so this never causes
+// downtime. Returns true when nothing more remains to retire. Errors are logged, not surfaced, so a
+// transient ate-api failure doesn't wedge reconciliation.
+func (r *SandboxAgentController) reconcileSubstrateBlueGreen(ctx context.Context, sa *v1alpha2.SandboxAgent) bool {
```
EItanya: So this is very interesting, I think we need to think through this whole use-case. I wonder if we can use some sort of name based on the hash, similar to a `GenerateName` so that everything is eventually consistent, rather than blocking

go/core/internal/controller/sandboxagent_substrate.go
```diff
@@ -33,9 +33,45 @@ func (r *SandboxAgentController) reconcileSubstrateSandboxAgent(ctx context.Cont
...
+	retireDone, err := r.SubstrateLifecycle.RetireSupersededTemplates(ctx, sa)
+	if err != nil {
+		sandboxAgentControllerLog.Info("retiring superseded substrate templates failed (will retry)",
+			"sandboxagent", sa.Namespace+"/"+sa.Name, "err", err.Error())
+		return true
+	}
+
+	// Best-effort reap of stale session actors keyed to a previous config. Not required for
+	// correctness (config-hashed ids mean they're never reused), so failures don't requeue.
+	if r.SubstrateActorBackend != nil {
```
EItanya: These constant nil checks are super ugly, we gotta figure that out.

go/core/pkg/sandboxbackend/substrate/agent_actor.go
```diff
@@ -116,7 +170,64 @@ func (b *SandboxAgentActorBackend) DeleteSandboxAgentSessionActor(ctx context.Co
...
+}
+
+// ReapStaleSessionActors deletes this agent's per-session actors that were created from a
+// superseded ActorTemplate (before a config change). It only deletes SUSPENDED actors: a RUNNING
+// one may be mid-request, and kagent's transport suspends session actors after each request, so a
+// stale actor converges to SUSPENDED on its own — force-suspending it could cut a live response.
+// With config-hashed actor ids these actors are never addressed again, so this is storage hygiene,
+// not correctness, and is best-effort. (Superseded GOLDEN actors are handled by
+// Lifecycle.RetireSupersededTemplates, which removes them with their template once a newer
+// template is Ready.) Returns true when no stale suspended session actors remain.
```
EItanya: I don't think we should do this. Just because the template is gone doesn't mean the actor is invalid. What do you think?

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
```diff
@@ -15,11 +16,34 @@ import (
...
+	pythonRuntimeLibPath = "/usr/lib/kagent-libs"
+	pythonVenvPath       = "/.kagent/.venv"
+
+	// SandboxConfigHashAnnotation carries the rendered-config hash on the generated
+	// ActorTemplate. It mirrors the translator's "kagent.dev/config-hash" pod-template
+	// annotation (keep in sync). A golden snapshot is an immutable memory image, so a
+	// config change must produce a NEW ActorTemplate (substrate snapshots once and
+	// no-ops in PhaseReady); folding the hash into the template name does that, and the
+	// annotation lets the chat path/reaper key session actors to the active template.
+	SandboxConfigHashAnnotation = "kagent.dev/config-hash"
```
EItanya: Can we define this somewhere globally, or use the generated name like I mentioned above

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
```diff
@@ -76,31 +114,83 @@ func findKagentContainer(containers []corev1.Container) *corev1.Container {
...
+// pythonRuntimeImageEnv returns the runtime-critical ENV directives baked into the Python
+// ADK image (python/Dockerfile). Substrate builds the OCI Process.Env from a hardcoded PATH
+// plus the ActorTemplate env only — it does NOT apply the image's ENV directives (the same
+// way it ignores the image entrypoint). Without LD_LIBRARY_PATH the standalone interpreter
+// cannot locate its bundled shared libraries (libz, libsqlite3, ...) and crashes on import
+// (e.g. numpy: "ImportError: libz.so.1: cannot open shared object file"); the failed startup
+// then surfaces as a gVisor "inconsistent private memory files on restore" error because the
+// golden snapshot captures only the pause container. The Go static binary needs none of this.
+// Keep in sync with the final-stage ENV block of python/Dockerfile.
+func pythonRuntimeImageEnv() []corev1.EnvVar {
```
EItanya, line 189: This feels hacky, is there someway we can fix substrate, or the gvisor impl to properly do this?

EItanya (commented):
Awesome job overall. I have some design questions here about getting rid of old templates and how that should work.

go/api/v1alpha2/agent_types.go
jmhbh: agreed. removing this function

go/core/internal/controller/sandboxagent_substrate.go
jmhbh: agreed, cleaned this up.

go/core/internal/controller/sandboxagent_substrate.go
jmhbh: yeah this is definitely worth discussing in more detail. These changes are intended to addres the current gap where a config change would not be reflected in the agent (ie: if changing the model provider in the sandbox agent). So, to your point of using a name based on the hash, the actor template name is now a combination of its truncated name + truncated config hash. New golden snapshots are created on creation of the new template, and we write the config hash as an annotation to the actor template which is embedded in the actor id (we retire the old template as a part of this function). When the new actor is being resolved on the request path with the new template's hash any misses will trigger actor creation and the request will be held until the actor is reachable. So right now on the request path, if you rollout a config change and message the agent while the config is being rolled out the if the new golden isn't ready the request will be served by the previous actor (there's some nuance here depending on the available workers) but if it is ready then a new actor will be created and the request will be held until the new actor can serve the request.

go/core/pkg/sandboxbackend/substrate/agent_actor.go
jmhbh: I agree that they aren't invalid but these actors would essentially be unreachable given our current implementation since we now use the config hash on the actor template to resolve actors. I'll look into what the cost is on the substrate side for leaving these actors around but given the current implementation I think it makes more sense to just remove them.

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
jmhbh: yup, this can be fixed in substrate. I can open a fix for this upstream.

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
jmhbh: moved to a separate consts package.

go/core/internal/controller/sandboxagent_controller.go
```diff
@@ -109,7 +109,7 @@ func (r *SandboxAgentController) SetupWithManager(mgr ctrl.Manager) error {
 	if err != nil {
 		return err
 	}
-	if r.SubstrateLifecycle != nil {
+	if r.substrateConfigured() {
```
EItanya, line 111: Can we get rid of this `r.substrateConfigured()` in a follow-up. If substrate is not configured we shouldn't even run this reconciler at all, nothing will work. Right now it's scattered everywhere

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
EItanya: Can you please link to that issue

go/core/internal/controller/sandboxagent_controller.go
jmhbh: I think it makes sense to separate this into a separate substrate controller, I can work on that as a follow up.

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
jmhbh: https://github.com/agent-substrate/substrate/issues/300

azabris1:
Is there a strong reason for this MR to proceed without skill support? It seems to me like this is one of the biggest advantages of the python substrate runtime. Particularly with this already trailblazer by initial Google ax, not apposed but I want to understand what the blocker is to a workaround for the init container.

jmhbh:
> Is there a strong reason for this MR to proceed without skill support? It seems to me like this is one of the biggest advantages of the python substrate runtime. Particularly with this already trailblazer by initial Google ax, not apposed but I want to understand what the blocker is to a workaround for the init container.

hey @azabris1 thanks for your comment. We are planning to add skills but will do so in a follow up PR

EItanya (approved):

PR #2049, opened 2026-06-18
add ACP shim for communication with agents in substrate

peterj:
Adds an [ACP shim](https://agentclientprotocol.com/) that runs inside the base images for openclaw/hermes on substrate. this allows us to re-use the websocket connection to the substrate actor (shim) and the shim does the translation from ws -> stdio. Both opencalw and hermes support the ACP protocol. 

I've also removed the openshell and agent-sandbox from the templates/code as substrate is the thing we'll go with.


go/core/pkg/sandboxbackend/substrate/constants.go
```diff
@@ -0,0 +1,15 @@
...
+
+// Default Substrate workload images for the acp-shim agent targets
+// (docker/acp-sandbox/Dockerfile). Substrate admission requires digest-pinned
+// refs. Both backend images are kept together here.
+const (
+	// AcpSandboxOpenClawImage is the default Substrate workload image for
+	// OpenClaw harnesses: the kagent acp-sandbox openclaw target, which layers
+	// the acp-shim and the restore-proof gateway-ensure scripts onto an
+	// OpenClaw install.
+	AcpSandboxOpenClawImage = "ttl.sh/kagent-acp-openclaw@sha256:f8c7b73253dd00098d3f2cb2c3a3d7585fa549daadeefdacd563362e4d40c7e6"
```
peterj: we'll need to build/push these images to kagent repo (cc @EItanya as I am not sure I have access to it)

go/api/v1alpha2/agentharness_types.go
```diff
@@ -18,31 +18,29 @@ import (
...
 		return true
 	default:
 		return false
 	}
 }
 
 // AgentHarnessRuntime selects which control plane provisions the harness VM.
-// +kubebuilder:validation:Enum=openshell;substrate
+// +kubebuilder:validation:Enum=substrate
 type AgentHarnessRuntime string
```
EItanya: I think we can remove this entirely. Since we only have one

go/core/cmd/acp-shim/sandbox/hermes-gateway-ensure.sh
EItanya, line 1: Can we add `.sh` extensions to these shell files

go/core/internal/database/gen/models.go
```diff
@@ -116,14 +116,15 @@ type PushNotification struct {
...
-	Source    *string
+	ID           string
+	UserID       string
+	Name         *string
+	CreatedAt    *time.Time
+	UpdatedAt    *time.Time
+	DeletedAt    *time.Time
+	AgentID      *string
+	Source       *string
+	AcpSessionID *string
```
EItanya: Any reason we can't re-use the ID here for the session itself?

EItanya (commented):
A couple of questions for a first pass

go/core/internal/database/gen/models.go
peterj: I can try to rework this; but the issue is the ordering -- we first create the session in the DB when the user clicks New Chat, only then we go and establish the connection with the actor. We can't then just take the session ID we created and pass it over to the openclaw/hermes and tell it to use that. I'll try to change the ordering, so we first create the session on the actor, then get that ID and use it in the session.id on our side

EItanya: And the actor generates the ID? I’m not following 100%

go/core/cmd/acp-shim/sandbox/hermes-gateway-ensure.sh
```diff
@@ -0,0 +1,41 @@
...
+# live actor incarnation. The gateway is intentionally NOT pre-warmed into the
+# golden snapshot (see buildAcpStartupScript): a snapshot-baked gateway would be
+# restored with a dead Telegram long-poll TCP connection in a fresh network
+# namespace, and this guard would then wrongly skip relaunching it. With no
+# prewarm, the PID file is absent on first post-restore connection, so we launch
+# a fresh gateway that dials Telegram in the live network.
+if [ -f "${PIDFILE}" ] && kill -0 "$(cat "${PIDFILE}" 2>/dev/null)" 2>/dev/null; then
+  log "gateway already running (pid $(cat "${PIDFILE}" 2>/dev/null)); nothing to do"
+  exit 0
+fi
```
jmhbh, line 29: this makes sense, but feels a bit hacky. Ideally we could perform the necessary cleanup right before a suspend operation but I understand we might want to iterate on this.

go/core/cmd/acp-shim/main.go
```diff
@@ -0,0 +1,87 @@
...
+	if cfg.TokenFile == "" {
+		cfg.TokenFile = os.Getenv("ACP_SHIM_TOKEN_FILE")
+	}
+	// Substrate ActorTemplate containers support env (incl. secretKeyRef)
+	// but not volume mounts, so allow passing the token directly. A literal
+	// token wins over the token file (the base image bakes in a default
+	// ACP_SHIM_TOKEN_FILE that only exists when a Secret is mounted).
+	if cfg.Token == "" {
+		cfg.Token = os.Getenv("ACP_SHIM_TOKEN")
+	}
```
jmhbh: nit: might be a bit cleaner to wrap this in a `acpshim.LoadConfig(cfg, policy)` func

go/core/internal/controller/agentharness_substrate_controller.go
```diff
@@ -124,83 +125,21 @@ func (r *SubstrateAgentHarnessController) Reconcile(ctx context.Context, req ctr
...
-		setAgentHarnessCondition(ah, v1alpha2.AgentHarnessConditionTypeActorReady, st, reason, msg)
-		if pending {
-			setAgentHarnessCondition(ah, v1alpha2.AgentHarnessConditionTypeBootstrapReady, metav1.ConditionFalse,
-				"ActorNotReady", "waiting for actor before post-ready bootstrap")
-		}
-		setAgentHarnessCondition(ah, v1alpha2.AgentHarnessConditionTypeReady, st, reason, msg)
-	}
+	// The AgentHarness is a template: once its generated ActorTemplate golden
+	// snapshot is Ready, the harness is Ready. We do NOT create a persistent
+	// actor here. Each chat session gets its own actor, created on demand by the
```
jmhbh: nit: this seems to conflict with the implementation in go/core/pkg/sandboxbackend/substrate/agentharness_actor.go given that currently, the actorTemplate chosen is just based on the harness and not the session id

design/EP-XXXX-acp-integration.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Open Questions
+
+* **Substrate + stdio-only ACP agents (Hermes, Codex, …)**: all backends use the in-sandbox `acp-shim` (see Implementation Details). Remaining sub-questions: shim delivery (baked into workload images vs. injected via init-container + shared volume, if Substrate workload templates allow it), and whether to wait for upstream Streamable HTTP adoption instead for some backends. The shim, once built, covers OpenClaw, Hermes, Codex, and any other stdio ACP agent uniformly.
+* **Shim child policy**: one child per WS connection (clean lifecycle) vs. one long-lived child with exactly-one-client semantics (preserves in-memory ACP sessions — required for Hermes). Proposed default: long-lived, configurable per backend.
```
jmhbh: I think this would also depend on what our suspension model looks if the default implementation is to keep long running actors then this makes sense.

jmhbh (commented):
look good to me so far (I'm around hafway through). Posting intial comments and will continue reviewing

go/core/cmd/acp-shim/sandbox/hermes-gateway-ensure.sh
peterj: yeah, in general we'll have to figure out how do we deal with the openclaw/hermes gateways -- they are launched and there mostly for the Telegram/Slack connections to work.

go/core/internal/controller/agentharness_substrate_controller.go
peterj: yep; fixing (we went back and forth on whether we have one actor for each session or single actor to host all sessions  :))

design/EP-XXXX-acp-integration.md
peterj: added a note

go/core/pkg/sandboxbackend/channels/channels.go
```diff
@@ -0,0 +1,98 @@
+// Package channels holds backend-agnostic helpers for translating AgentHarness
```
jmhbh: nit: wdyt about naming this file channel_helpers instead?

go/core/pkg/sandboxbackend/substrate/agentharness_actor.go
```diff
@@ -0,0 +1,194 @@
...
+		client:          client,
+		atenetRouterURL: atenetRouterURL,
+	}
+}
+
+// EnsureSessionActor creates (if needed) and resumes the harness's single shared
+// actor, then waits for it to be reachable via atenet-router. The sessionID
+// identifies the chat for logging only; chats are multiplexed as ACP sessions
+// inside the one actor, so they all resolve to the same ActorID(ah).
+func (b *AgentHarnessSessionActorBackend) EnsureSessionActor(ctx context.Context, ah *v1alpha2.AgentHarness, sessionID string) (sandboxbackend.EnsureResult, error) {
```
jmhbh, line 43: curious, is the acp shim's child process stateful? AFAICT we are only managing the gateway connection to the backend type so this may not be relevant but the reason I'm asking is because if the child process maintains state that pertains to a specific session multiplexing on the actor like this could leak that state to potentially irrelevant sessions.

peterj: the session specific state shouldn't leak across other sessions; however, the things that are per-actor -- e.g. skills or tools -- those will be (and should be) available across sessions.

go/core/cmd/acp-shim/sandbox/hermes-gateway-ensure.sh
EItanya: Long term I think we'll have to take over connectors for all agents in our system FWIW

go/core/pkg/acpshim/server.go
```diff
@@ -0,0 +1,333 @@
...
+		return true // auth disabled (prototype/testing only)
+	}
+	presented := ""
+	if h := r.Header.Get("Authorization"); len(h) > 7 && h[:7] == "Bearer " {
+		presented = h[7:]
+	} else if q := r.URL.Query().Get("access_token"); q != "" {
+		presented = q
+	}
+	return subtle.ConstantTimeCompare([]byte(presented), []byte(s.cfg.Token)) == 1
+}
```
EItanya: Why do we have this? Checking a static token seems fairly useless? Is this supposed to be the Gateway token equivalent?

go/core/pkg/acpshim/server.go
EItanya, line 1: There's a lot going on in this file, and it's quite hard to read, so I just want to check in on the functionality here before we merge it. Are there multiple modes because openclaw and hermes each require different behavior, or is it for some future state? Also, can you potentially try and do a clean-up run on this file to make it more readable, the constant locking and channels is very difficult to grok.

go/core/pkg/sandboxbackend/substrate/constants.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// AcpSandboxOpenClawImageRef and AcpSandboxHermesImageRef are the digest-pinned
+// workload image refs (registry/repo/name@sha256:...), set at controller link
+// time via -X ...substrate.AcpSandbox*ImageRef=... They are empty in source and
+// in unit tests, in which case resolution returns an error rather than an
+// unpinned (or wrong-registry) ref.
+var (
+	AcpSandboxOpenClawImageRef string
+	AcpSandboxHermesImageRef   string
```
EItanya: https://github.com/kagent-dev/kagent/issues/2055

We need to take into account image registry/repository somehow

scripts/controller-digest-ldflags.sh
```diff
@@ -84,17 +89,41 @@ image_digest() {
...
-append_digest_ldflag "PythonADKImageDigest" "${APP_IMG}"
-append_digest_ldflag "GoADKImageDigest" "${GOLANG_ADK_IMG}"
-append_digest_ldflag "GoADKFullImageDigest" "${GOLANG_ADK_FULL_IMG}"
+# append_pinned_ref_ldflag emits a -X ldflag setting go_var to the fully
+# resolved, digest-pinned ref (registry/repo/name@sha256:...) for image_ref.
+# Unlike append_digest_ldflag (digest only, with the registry composed in Go),
+# this bakes in the registry/repo from the image that was actually built/pushed,
+# so the controller points at the right registry (ghcr.io for releases,
+# localhost:5001 for local kind/`make helm-install`).
+append_pinned_ref_ldflag() {
```
EItanya: See my other comment about registry/repository

go/core/pkg/acpshim/server.go
peterj: leftover from previous implementation; will clean it up. it was for the gateway in openclaw

go/core/internal/httpserver/handlers/agentharness_session.go
EItanya, line 1: TODO: unify session handlers

EItanya (approved):

Issue #2045, opened 2026-06-18
SSO: re-run OIDC flow on expired session instead of rendering logged-out UI


dimetron:
## Summary

When kagent runs behind oauth2-proxy, the oauth2-proxy session cookie can still be valid while the forwarded `id_token` has expired. Today the UI treats an expired/missing token the same as "no user" and silently renders a logged-out UI, forcing a manual reload.

## Proposal

Distinguish three auth states in the UI instead of a single `CurrentUser | null`:

- `authenticated` — valid, non-expired token forwarded by oauth2-proxy
- `expired` — oauth2-proxy session valid but forwarded `id_token` is missing/expired → UI should re-run the OIDC flow
- `unsecured` — no `Authorization` header at all (no oauth2-proxy in front) → UI must NOT redirect, or it would loop with no `/oauth2` endpoint

On `expired`, redirect to the oauth2-proxy re-auth endpoint (`/oauth2/start?rd=<current>`) to mint a fresh token, with a short-window guard in `sessionStorage` to prevent redirect loops if re-auth keeps returning a stale token.

## Scope

- `ui/src/app/actions/auth.ts` — `AuthStatus` / `AuthResult`
- `ui/src/contexts/AuthContext.tsx` — re-auth redirect + loop guard
- `ui/src/components/UserMenu.tsx`, `ui/src/app/login/page.tsx`
- `docs/OIDC_PROXY_AUTH_ARCHITECTURE.md`, helm oauth2-proxy values

A design EP (`design/EP-<this issue>-sso-session-expiry-redirect.md`) will accompany the implementation PR.

PR #2051, opened 2026-06-18
feat(sso) expired token session redirect

dimetron:
Improve SSO expired token redirect (#2045) 

This pull request implements robust handling of SSO session expiry in the UI when deployed behind an OIDC proxy (such as oauth2-proxy). The main improvements are the introduction of a clear authentication status model, transparent re-authentication when the session expires, and safeguards against redirect loops. The changes ensure users are not unexpectedly logged out when their forwarded token expires, while avoiding infinite redirects in unsecured (no proxy) environments. Comprehensive unit tests and documentation updates are included.

**Authentication state handling and re-authentication flow:**

- Introduced an explicit `AuthStatus` model (`"authenticated"`, `"expired"`, `"unsecured"`) and updated `getCurrentUser` to return a structured `AuthResult` reflecting these states. This allows the UI to distinguish between a missing proxy, an expired token, and a valid session. [[1]](diffhunk://#diff-810f2f7cdb894e7c5a0daec17b6864a866d561bc4cf0c37b61873cae792c8504L14-R41) [[2]](diffhunk://#diff-a85c35b1f2858b3c4d567a9b413ab3efe379383afc6b3f9f7a713c6908607bf5R1-R122)
- Updated `AuthProvider` in `AuthContext.tsx` to trigger a redirect to the OIDC re-auth endpoint (`/oauth2/start`) only when the session is `"expired"`, never in `"unsecured"` mode. Added a `sessionStorage`-based guard to prevent redirect loops if re-authentication fails to refresh the token. The guard is cleared on successful authentication. [[1]](diffhunk://#diff-6f392efee8540a2e02c4f02b3b4cb7e10aba056ca88b1bbc0cd3d5bfd478ba32L4-R15) [[2]](diffhunk://#diff-6f392efee8540a2e02c4f02b3b4cb7e10aba056ca88b1bbc0cd3d5bfd478ba32R25-R35) [[3]](diffhunk://#diff-6f392efee8540a2e02c4f02b3b4cb7e10aba056ca88b1bbc0cd3d5bfd478ba32R47-R73)

**Testing and reliability:**

- Added comprehensive unit tests for `getCurrentUser` to cover all authentication states, ensuring correct status assignment and token handling.
- Added tests for `AuthProvider` to verify redirect and guard logic, including prevention of redirect loops and correct guard clearing on authentication.

**Documentation and UI updates:**

- Updated the architecture documentation and added a new design doc to explain the three authentication states, the re-authentication flow, and the rationale for the changes. The flowchart and rationale in `OIDC_PROXY_AUTH_ARCHITECTURE.md` were updated for clarity. [[1]](diffhunk://#diff-c05806e7bead589fd90ba06d6ed6d060af7c46653bcba4a22fbb80442a3e8936L132-R144) [[2]](diffhunk://#diff-a85c35b1f2858b3c4d567a9b413ab3efe379383afc6b3f9f7a713c6908607bf5R1-R122)
- Minor UI alignment: adjusted the login page to match the new SSO redirect path and improved layout consistency.

These changes collectively ensure a seamless and predictable authentication experience for users behind SSO, while maintaining safe behavior in unsecured deployments.


ui/src/contexts/AuthContext.tsx
```diff
@@ -1,10 +1,18 @@
 "use client";
 
 import React, { createContext, useContext, useEffect, useState, ReactNode } from "react";
-import { getCurrentUser, CurrentUser } from "@/app/actions/auth";
+import { getCurrentUser, CurrentUser, AuthStatus } from "@/app/actions/auth";
```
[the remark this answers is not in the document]
dimetron: Fixed in 37d7d387 — `ReactNode`, `CurrentUser`, and `AuthStatus` are now `type`-only imports.

ui/src/contexts/AuthContext.tsx
```diff
@@ -1,10 +1,18 @@
...
+
+// oauth2-proxy endpoint that (re)starts the OIDC flow. Mirrors the branded
+// login page's SSO_REDIRECT_PATH.
+const SSO_REAUTH_PATH = process.env.NEXT_PUBLIC_SSO_REAUTH_PATH || "/oauth2/start";
+// Guards against redirect loops if re-auth keeps returning a stale token.
+const REAUTH_GUARD_KEY = "kagent_reauth_attempt";
```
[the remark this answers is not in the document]
dimetron: Fixed in 37d7d387 — `AuthContext` now reads `NEXT_PUBLIC_SSO_REDIRECT_PATH`, and the Helm chart (`ui-deployment.yaml`) injects it from `ui.auth.ssoRedirectPath` alongside `SSO_REDIRECT_PATH`, so client-side overrides take effect.

ui/src/contexts/AuthContext.tsx
```diff
@@ -34,8 +44,33 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+    const lastAttempt = Number(sessionStorage.getItem(REAUTH_GUARD_KEY) || "0");
+    if (Date.now() - lastAttempt < REAUTH_GUARD_WINDOW_MS) {
+      setError(
+        new Error("Authentication expired and re-authentication did not refresh the session.")
+      );
+      return;
+    }
+    sessionStorage.setItem(REAUTH_GUARD_KEY, String(Date.now()));
+    const rd = encodeURIComponent(window.location.pathname + window.location.search);
+    window.location.href = `${SSO_REAUTH_PATH}?rd=${rd}`;
```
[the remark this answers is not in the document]
dimetron: Fixed in 37d7d387 — the re-auth redirect now uses `window.location.assign(...)`.

ui/src/contexts/__tests__/AuthContext.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+// jsdom hardens window.location (non-configurable, href setter throws
+// "navigation not implemented"), so rather than asserting the redirect URL we
+// assert the observable contract: the sessionStorage re-auth guard that gates
+// the redirect. The guard is written immediately before window.location.href
+// is set, so "guard written" == "redirect attempted".
+describe("AuthProvider re-auth behavior", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    sessionStorage.clear();
+  });
```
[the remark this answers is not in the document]
dimetron: Addressed in 37d7d387. jsdom in this repo fully hardens `window.location` (non-configurable; `assign` is read-only), so it can't be spied/replaced here. The test instead asserts the observable contract that gates the redirect — the `sessionStorage` re-auth guard written immediately before `assign()` — and silences jsdom's expected "navigation not implemented" console noise. Rationale documented inline.

docs/OIDC_PROXY_AUTH_ARCHITECTURE.md
```diff
@@ -129,14 +129,19 @@ flowchart TD
...
+    style I fill:#ff9,stroke:#333
 ```
 
-**Design rationale**: The UI does not redirect on auth failure. If `getCurrentUser()` fails, it indicates a misconfiguration (oauth2-proxy should have intercepted the request) rather than a normal session expiry. The error state surfaces this for debugging rather than masking it with a redirect loop.
+**Design rationale**: oauth2-proxy gates access using its session cookie (valid up to `cookie-expire`, default 168h), while the UI derives the user from the forwarded id_token. These lifetimes are decoupled, so the id_token can go stale while the session cookie is still valid. To keep them aligned, oauth2-proxy is configured with `cookie-refresh` (+ `offline_access` scope) to refresh the id_token. As a safety net, `getCurrentUser()` distinguishes three states:
+
+- **authenticated** — valid token → set user state.
+- **expired** — `Authorization` header present but token missing/expired → the UI re-runs the OIDC flow (`/oauth2/start`) to mint a fresh token, guarded against redirect loops.
+- **unsecured** — no `Authorization` header (no oauth2-proxy in front) → no user and no redirect (there is no `/oauth2` endpoint to redirect to).
```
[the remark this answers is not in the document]
dimetron: Fixed in 37d7d387 — the doc now states `cookie-refresh` + `offline_access` is opt-in and NOT enabled by the chart defaults (default scope is `openid profile email groups`).

design/EP-2045-sso-session-expiry-redirect.md
```diff
@@ -0,0 +1,122 @@
...
+surfaces an error instead of redirecting again. The guard is cleared on a
+successful `authenticated` result.
+
+### UI surface
+
+- `ui/src/components/UserMenu.tsx` — reflects the new status (e.g. distinguishes a
+  logged-out/unsecured menu from an authenticated one).
+- `ui/src/app/login/page.tsx` — aligns the branded login page's SSO redirect path
+  with `SSO_REAUTH_PATH`.
+- `docs/OIDC_PROXY_AUTH_ARCHITECTURE.md` — documents the three states and the
```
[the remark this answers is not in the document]
dimetron: Fixed in 37d7d387 — the EP now clarifies the login page (server component) reads `SSO_REDIRECT_PATH` while `AuthContext` (client) reads `NEXT_PUBLIC_SSO_REDIRECT_PATH`, both injected by the chart.

mesutoezdil:
3 small questions

nr1// auth.ts : when `decodeJWT` returns null (malformed token), the function returns `status: "expired"`. but a malformed token is not really expired, it is invalid. would it make more sense to have a third path here so the ui does not try to reauth on a broken token?

nr2//  jwt.ts:`isTokenExpired` returns false when `claims.exp` is missing, so a token with no expiry is treated as valid forever. is that intentional? feels like it could be a risk if a token without exp somehow ends up here.

nr3//  AuthContext.tsx :`REAUTH_GUARD_WINDOW_MS` is 10 seconds. if the oidc flow at the idp takes longer than that (slow network, slow idp), the guard expires and the user hits the error path instead of completing reauth. would 60 seconds be more appropriate here?

dimetron:
Thanks @mesutoezdil — all three are fair points. Addressed in the latest commit:

**nr2 (jwt.ts — missing `exp`)**: Agreed, valid-forever was a risk. `isTokenExpired` now treats a missing `exp` as **expired**. OIDC id_tokens are required to carry `exp`, so a token without one is non-compliant/garbled and should force re-auth rather than be trusted indefinitely.

**nr3 (AuthContext.tsx — guard window)**: Bumped `REAUTH_GUARD_WINDOW_MS` from 10s → **60s** to cover a slow IdP round-trip, so an in-flight re-auth isn't misread as a failed loop while still catching a fast redirect loop.

**nr1 (auth.ts — malformed token → `expired`)**: Kept malformed-token handling on the `expired` path intentionally. Behind oauth2-proxy a malformed forwarded `id_token` almost always means the proxy needs to re-mint one, so re-running the OIDC flow is the correct recovery, and the `sessionStorage` loop-guard (now 60s) prevents thrashing if the proxy keeps forwarding a broken token (the user lands on the error path after one failed cycle instead of looping). A separate `"invalid"` status would behave identically to `"expired"` here, so I left it as one path to avoid added surface area — happy to split it out if you'd prefer the explicit state.

Also added `ui/src/lib/__tests__/jwt.test.ts` covering decode (valid/malformed) and `isTokenExpired` (future/past/missing `exp`).

ui/src/app/actions/auth.ts
```diff
@@ -11,20 +11,32 @@ export interface CurrentUser extends Record<string, unknown> {
   groups?: string[];
 }
 
-export async function getCurrentUser(): Promise<CurrentUser | null> {
+// authenticated → valid, non-expired token forwarded by oauth2-proxy
+// expired       → oauth2-proxy session is still valid but the forwarded
+//                 id_token is missing/expired → the UI should re-run OIDC
+// unsecured     → no Authorization header at all (no oauth2-proxy in front);
+//                 the UI must NOT redirect or it would loop with no /oauth2 endpoint
+export type AuthStatus = "authenticated" | "expired" | "unsecured";
```
peterj, line 19: this is a nit, but should "unsecured" be "unauthenticated"? or would that be a state where aouth2-proxy is confgiured, but the user is unauthenticated?

peterj: I think I answered my question by looking at getCurrentUser :D

ui/src/app/login/page.tsx
```diff
@@ -12,7 +12,7 @@ export default function LoginPage() {
       {/* Preload background image for faster rendering */}
       <link rel="preload" href="/login-bg.webp" as="image" type="image/webp" fetchPriority="high" />
 
-      <div className="login-page relative fixed inset-0 z-50 overflow-hidden bg-[#0B0B15] text-white">
```
peterj, line 15: any reason for removing the `relative` class herE?

ui/src/app/actions/auth.ts
```diff
@@ -11,20 +11,32 @@ export interface CurrentUser extends Record<string, unknown> {
...
+
+export interface AuthResult {
+  status: AuthStatus;
+  user: CurrentUser | null;
+}
+
+export async function getCurrentUser(): Promise<AuthResult> {
```
peterj: since we're not returning just the user anymore, can we rename this to `authenticate` or `getAuthResult`?

ui/src/contexts/AuthContext.tsx
```diff
@@ -14,15 +26,17 @@ const AuthContext = createContext<AuthContextValue | undefined>(undefined);
...
 
   const fetchUser = async () => {
     setIsLoading(true);
     setError(null);
     try {
-      const currentUser = await getCurrentUser();
-      setUser(currentUser);
+      const result = await getCurrentUser();
+      setStatus(result.status);
+      setUser(result.user);
```
peterj: do we need to set the user separately like this, or should we just use the AuthResult as a state variable?

ui/src/contexts/AuthContext.tsx
```diff
@@ -34,8 +48,33 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+    const lastAttempt = Number(sessionStorage.getItem(REAUTH_GUARD_KEY) || "0");
+    if (Date.now() - lastAttempt < REAUTH_GUARD_WINDOW_MS) {
+      setError(
+        new Error("Authentication expired and re-authentication did not refresh the session.")
+      );
+      return;
+    }
+    sessionStorage.setItem(REAUTH_GUARD_KEY, String(Date.now()));
+    const rd = encodeURIComponent(window.location.pathname + window.location.search);
+    window.location.assign(`${SSO_REDIRECT_PATH}?rd=${rd}`);
```
peterj: `.replace` instead of assign

ui/src/contexts/AuthContext.tsx
```diff
@@ -34,8 +48,33 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+  }, [isLoading, status]);
+
+  useEffect(() => {
+    if (status === "authenticated" && typeof window !== "undefined") {
+      sessionStorage.removeItem(REAUTH_GUARD_KEY);
```
peterj, line 69: also clear the error here (setError(null))

peterj (changes requested):
added a couple of comments

ui/src/app/login/page.tsx
dimetron: Fixed in 388628cb — restored `relative`; it was removed unintentionally during layout cleanup.

ui/src/app/actions/auth.ts
dimetron: Fixed in 388628cb — renamed to `getAuthResult()`.

ui/src/contexts/AuthContext.tsx
dimetron: Fixed in 388628cb — `AuthProvider` now keeps a single `authResult` state and exposes `user`/`status` from it.

ui/src/contexts/AuthContext.tsx
dimetron: Fixed in 388628cb — re-auth redirect now uses `window.location.replace(...)`.

ui/src/contexts/AuthContext.tsx
dimetron: Fixed in 388628cb — `setError(null)` runs when auth status becomes `authenticated`.

peterj (approved):

PR #2052, opened 2026-06-18
feat(MCP UI) chat widgets

dimetron:
This pull request adds support for rendering interactive MCP UI widgets (MCP Apps) directly within the chat interface. It introduces backend and frontend changes to detect, handle, and render tools that provide UI resources, ensuring a richer and more interactive user experience. The backend now distinguishes between regular tools and those with UI widgets, compacts tool responses sent to the model to prevent unnecessary repeated calls, and exposes new endpoints for MCP-app interactions. The UI is updated to render these widgets inline and broker interactions between the app and the chat.

**Backend: MCP App tool detection and handling**
- The `CreateToolsets` function in `registry.go` now returns both the toolsets and a set of tool names that support MCP Apps, enabling the agent to treat these tools differently. [[1]](diffhunk://#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfL97-R100) [[2]](diffhunk://#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfL116-R124) [[3]](diffhunk://#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfL138-R156) [[4]](diffhunk://#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfL165-R178) [[5]](diffhunk://#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfL313-R422) [[6]](diffhunk://#diff-aa78af986baa8c0d72522820c58c6a60eb1d656aff48985d4ff3fd683a4312bfL335-R435)
- A new callback (`MakeMCPAppModelResultCallback` in `mcp_apps.go`) is added to compact the payload sent to the model for MCP App tools, replacing heavy render payloads with a terminal notice to prevent the model from re-invoking rendering tools unnecessarily. [[1]](diffhunk://#diff-47f0265e8fa992a1daf1e3aa04f2ed446630baac255502ac688dc430f6d19887R1-R72) [[2]](diffhunk://#diff-5e48253d455263dd808978b30a15e6f419f326b1b1561a46c67e51b88e3b071aR119-R124)
- The agent is updated to wire in the MCP App callback only when MCP App tools are present, ensuring efficient handling.

**Testing and validation**
- Comprehensive unit tests are added in `mcp_apps_test.go` to verify correct compaction of responses, error handling, and that only MCP App tools are affected by the new logic.

**Documentation and design**
- A detailed design document (`EP-2046-chat-mcp-ui-widgets.md`) describes the motivation, goals, implementation details, test plan, and open questions for this feature.


go/adk/pkg/agent/mcp_apps.go
```diff
@@ -0,0 +1,72 @@
...
+					continue
+				}
+				part.FunctionResponse.Response = compactMCPAppModelResponse(part.FunctionResponse.Response)
+			}
+		}
+		return nil, nil
+	}
+}
+
+func compactMCPAppModelResponse(response map[string]any) map[string]any {
```
peterj, line 53: are there actual types available we could use here, instead of `map[string]any`?

go/adk/pkg/mcp/registry.go
```diff
@@ -94,9 +94,10 @@ func CreateToolsets(
 	sseTools []adk.SseMcpServerConfig,
 	propagateToken bool,
 	headerProvider DynamicHeaderProvider,
-) []tool.Toolset {
+) ([]tool.Toolset, map[string]bool) {
```
peterj: what does this map store?

go/adk/pkg/mcp/registry.go
```diff
@@ -256,6 +263,100 @@ func createTransport(ctx context.Context, params mcpServerParams) (mcpsdk.Transp
 	return mcpTransport, nil
 }
 
+// isMCPAppOnlyTool reports whether a tool is callable only from the rendered
+// MCP app (hidden from the agent). A tool is app-only when its visibility
+// declares "app" but not "model". Absent visibility defaults to agent-visible.
+func isMCPAppOnlyTool(meta map[string]any) bool {
```
peterj: instead of having boolean, would it make sense to perhaps have 'type' or 'kind'? for example "agent_tool" or "ui_tool" or something like that? this would allow us to extend it later if new 'types' become available or if we want to differentiate between more of them. what do you think?

go/adk/pkg/mcp/registry.go
```diff
@@ -256,6 +263,100 @@ func createTransport(ctx context.Context, params mcpServerParams) (mcpsdk.Transp
...
+		if v == "app" {
+			hasApp = true
+		}
+	}
+	return hasApp
+}
+
+// isMCPAppTool reports whether a tool declares an MCP-UI resourceUri,
+// meaning its result should be rendered as an interactive app in the chat UI.
+func isMCPAppTool(meta map[string]any) bool {
```
peterj: strong type for the map?

go/build.err
```diff
@@ -0,0 +1,38 @@
...
+go: downloading github.com/aws/aws-sdk-go-v2/config v1.32.25
+go: downloading github.com/aws/aws-sdk-go-v2/service/bedrockruntime v1.53.5
+go: downloading google.golang.org/genai v1.60.0
+go: downloading github.com/ollama/ollama v0.30.8
+go: downloading github.com/aws/smithy-go v1.27.1
+go: downloading github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.7.13
+go: downloading github.com/aws/aws-sdk-go-v2/internal/configsources v1.4.29
+go: downloading github.com/aws/aws-sdk-go-v2/credentials v1.19.24
+go: downloading github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.18.29
+go: downloading github.com/aws/aws-sdk-go-v2/service/signin v1.2.0
```
peterj: remove

ui/src/app/api/mcp-apps/[namespace]/[name]/resources/route.ts
```diff
@@ -0,0 +1,16 @@
+import { NextRequest } from "next/server";
+import { mcpAppsBackendPath, optionsResponse, proxyMcpAppsRequest } from "../../../_utils";
```
Comment (peterj) [inline ui/src/app/api/mcp-apps/[namespace]/[name]/resources/route.ts]: you should use the `@` style imports so you can get rid of the `../../../../`

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -204,35 +243,40 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
...
-    e.preventDefault();
-    if (!currentInputMessage.trim() || !selectedAgentName || !selectedNamespace) {
+  const sendChatMessageText = async (
+    userMessageText: string,
+    options: {
+      clearInput?: boolean;
+      restoreInputOnError?: boolean;
+      errorLabel?: string;
+      rethrowOnError?: boolean;
+      fileParts?: FilePart[];
```
peterj: not sure how much work it is, but could we split out the file upload functionality from this PR and have it as a separate one, as I am assuming it's not required for the mcp ui to work

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -204,35 +243,40 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
...
+    } = {},
+  ) => {
+    const fileParts = options.fileParts ?? [];
+    if ((!userMessageText.trim() && fileParts.length === 0) || !selectedAgentName || !selectedNamespace) {
       return;
     }
-
-    // Stop voice recording if active before sending
-    if (isListening) {
```
peterj, line 251: was this voice recording removed?

go/adk/pkg/agent/mcp_apps.go
dimetron: Good call — switched to the typed MCP `CallToolResult` from `modelcontextprotocol/go-sdk`. The callback now decodes the model-facing payload into `mcpsdk.CallToolResult` and rebuilds the success notice with `mcpsdk.TextContent` instead of hand-built `map[string]any` literals (falling back to the original payload if it is not a recognizable MCP result). Done in 4bc1353b.

go/adk/pkg/mcp/registry.go
dimetron: It is the set of agent-visible tool names whose results render as MCP App (UI) widgets — used to attach the model-result compaction callback only to those tools. Gave it a documented named type `MCPAppToolNames` (threaded through the return chain) in 43710996.

go/adk/pkg/mcp/registry.go
dimetron: Agreed. Replaced the `isMCPAppOnlyTool`/`isMCPAppTool` booleans with a single `mcpToolKind` classifier (`agent` / `app` / `app_only`) plus `mcpToolKindOf(meta)`, so call sites switch on the kind and new kinds can be added later without reworking signatures. Behavior unchanged; tests updated. Done in 43710996.

dimetron:
Screenshot 2026-06-19 at 14 44 11

go/adk/pkg/agent/mcp_apps.go
peterj: did you not push these changes? I still see `map[string]any` here.

peterj (changes requested):

go/adk/pkg/mcp/registry.go
```diff
@@ -256,6 +272,146 @@ func createTransport(ctx context.Context, params mcpServerParams) (mcpsdk.Transp
...
+		if len(configuredFilter) > 0 && !configuredFilter[t.Name] {
+			continue
+		}
+		switch mcpToolKindOf(t.Meta) {
+		case mcpToolKindAppOnly:
+			// Hidden from the agent; only the rendered MCP App calls it.
+			continue
+		case mcpToolKindApp:
+			allowedTools = append(allowedTools, t.Name)
+			appToolNames[t.Name] = true
```
peterj: since true/false is the only indicator whether a tools is an UI app tool, could we just rely and use that check later on too? it feels like we're doing this function and then piping the results through initializeToolSet, addToolset and  CreateToolSets only to "use" that in CreateGoogleADKAgentWithSubagentSessionIDs (in agent.go). Couldn't we just filter down to the MCP UI only tools in CreateGoogleADKAgentWithSubagentSessionIDs? that way we get rid of all that piping through multiple functions

ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts
```diff
@@ -0,0 +1,22 @@
+import { NextRequest } from "next/server";
+import { mcpAppsBackendPath, optionsResponse, proxyMcpAppsRequest } from "../../../../../_utils";
```
Comment (peterj) [inline ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts]: still some places with ../.../../../ style imports

dimetron:
Addressed review feedback in `fc81d694` and `271361c0`:

- **File upload split** (`fc81d694`): removed `fileUpload.ts`, `FileAttachment.tsx`, and all chat upload/display wiring from this PR.
- **`go/build.err`**: not present on the branch (removed earlier).
- **`@/` imports**: MCP apps API routes now use `@/app/api/mcp-apps/_utils`.
- **Strong types for tool meta** (`271361c0`): classify tools via `mcpsdk.Meta` + parsed `mcpUIMetadata`; `MakeMCPAppModelResultCallback` takes `mcp.MCPAppToolNames`.
- **Voice input**: still present — `handleSendMessage` stops listening before send (line ~429) and the mic button remains in the composer.

**Re: collecting MCP app tool names during `CreateToolsets`** — we can't defer this to `CreateGoogleADKAgentWithSubagentSessionIDs` without a second `ListTools` round trip per server. The MCP-UI metadata (`ui.resourceUri`, visibility) lives on `mcpsdk.Tool.Meta`, which ADK's `mcptoolset` drops when converting to `tool.Tool`, so classification has to happen at the initial `ListTools` call. `fc81d694` already simplified the plumbing by accumulating into a shared `MCPAppToolNames` map instead of threading per-server return values through `addToolset` → `CreateToolsets`.

go/adk/pkg/mcp/registry.go
dimetron: Good point — the shared accumulator threaded through `CreateToolsets` → `addToolset` → `initializeToolSet` was awkward.

We can't move the *classification* itself into `agent.go` without a second `ListTools` round-trip per server: MCP-UI metadata (`ui.resourceUri`, visibility) lives on `mcpsdk.Tool.Meta`, and ADK `mcptoolset` drops `Meta` when it converts tools to `tool.Tool`. So the kind check still has to happen at `ListTools` time in `agentVisibleToolFilter`.

What we *can* do is stop piping the set through every helper signature. In `1d4c27e6`:

- `CreateToolsets` now returns only `[]tool.Toolset` again
- each MCP toolset is wrapped in `mcpAppToolset` with per-server `appToolNames` recorded during classification
- `agent.go` collects the union in one place: `mcp.MCPAppToolNamesFromToolsets(toolsets)`

That keeps classification where the metadata is available, but moves *collection* to agent wiring time as you suggested.

ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts
```diff
@@ -0,0 +1,22 @@
+import { NextRequest } from "next/server";
+import { mcpAppsBackendPath, optionsResponse, proxyMcpAppsRequest } from "@/app/api/mcp-apps/_utils";
+
+export async function POST(
```
Comment (peterj) [inline ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts]: can we put all these to the mcp-apps.ts actions instead? or do they need to be exposed here as API endpoints? 

(asking because we don't have any API routes so far and this would be the first ones)

ui/src/app/servers/[namespace]/[name]/apps/page.tsx
```diff
@@ -0,0 +1,10 @@
+import { McpAppsInspector } from "@/components/mcp-apps/McpAppsInspector";
+
+export default async function ServerMcpAppsPage({
+  params,
+}: {
+  params: Promise<{ namespace: string; name: string }>;
+}) {
+  const { namespace, name } = await params;
+  return <McpAppsInspector namespace={decodeURIComponent(namespace)} name={decodeURIComponent(name)} />;
```
Comment (peterj) [inline ui/src/app/servers/[namespace]/[name]/apps/page.tsx]: instead of a separate page like that and a context menu on the mcp server item -- can we just list the apps together with tools, just mark them differently or use a different icon.  

Image

go/core/internal/httpserver/handlers/mcpapps.go
```diff
@@ -0,0 +1,289 @@
...
+type mcpAppToolCallRequest struct {
+	Arguments any `json:"arguments,omitempty"`
+}
+
+func NewMCPAppsHandler(base *Base) *MCPAppsHandler {
+	return &MCPAppsHandler{Base: base}
+}
+
+func (h *MCPAppsHandler) HandleListTools(w ErrorResponseWriter, r *http.Request) {
+	namespace, name, ok := h.remoteMCPServerRef(w, r)
```
peterj: are we only supporting RemoteMCPServer CRD types? 

I created a MCPServer CRD that has UI, but the /apps URL fails to render it because it's expecting the RemoteMCPServer

peterj:
another issue I noticed -- I am rendering a UI for weather (using https://servereverything.dev/mcp as the MCP server). The UI renders; but then when I interact wit hthe UI (e.g. click the New York button to show weather), a new request/chat is sent to teh agent and the UI re-renders as a new chat item/response. I was expecting we only render 1 instance of the MCP UI -- if user interacts with the UI we should re-render/update that same instance, not get another UI instance in the chat. 

Screenshot 2026-06-23 at 11 16 30 AM


Also seeing this in the console (not sure if related to the UI re-rendering or not):

```
[AppRenderer] Using provided resource URI: ui://server-everything/weather-dashboard
0s2learxkzu1s.js:78 [AppRenderer] Reading resource HTML from: ui://server-everything/weather-dashboard
0s2learxkzu1s.js:1 Uncaught (in promise) Error: Not connected
    at pN.notification (0s2learxkzu1s.js:1:28732)
    at pN.sendHostContextChange (0s2learxkzu1s.js:78:6241)
    at pN.setHostContext (0s2learxkzu1s.js:78:6178)
    at 0s2learxkzu1s.js:78:16090
    at iv (26bsh0xgt_fiq.js:1:102969)
    at up (26bsh0xgt_fiq.js:1:127008)
    at ud (26bsh0xgt_fiq.js:1:126670)
    at up (26bsh0xgt_fiq.js:1:127053)
    at ud (26bsh0xgt_fiq.js:1:126670)
    at up (26bsh0xgt_fiq.js:1:126988)
notification @ 0s2learxkzu1s.js:1
sendHostContextChange @ 0s2learxkzu1s.js:78
setHostContext @ 0s2learxkzu1s.js:78
(anonymous) @ 0s2learxkzu1s.js:78
```

dimetron:
@peterj - Interesting, I was testing with my own MCP's - let me test this one. Need to make sure we are on the same standard of UI libraries, it might be related to the routing - UI need to go via internal <kagent>/plugins/ or we should allow Origin

peterj:
here's the implementation of the MCP UI I've been using: https://github.com/peterj/servereverything.dev/blob/main/src/widget.html

dimetron:
Screenshot 2026-06-23 at 21 38 54

go/adk/pkg/mcp/registry.go
```diff
@@ -78,9 +79,52 @@ type mcpServerParams struct {
...
-// returning the accumulated toolsets. Errors on individual servers are logged
-// and skipped.
+// MCPAppToolNames is the set of MCP tool names whose results render as
+// interactive MCP App (UI) widgets in the chat (the tool declares a
+// ui.resourceUri and is visible to the agent). It is used as a set, so the bool
+// value is always true and only key presence is meaningful. The agent attaches
+// the model-result compaction callback (see agent.MakeMCPAppModelResultCallback)
+// only to these tools. Collect them from CreateToolsets output via
+// MCPAppToolNamesFromToolsets.
+type MCPAppToolNames map[string]bool
```
EItanya: Why do we need a type alias for this?

go/adk/pkg/mcp/registry.go
```diff
@@ -256,6 +300,159 @@ func createTransport(ctx context.Context, params mcpServerParams) (mcpsdk.Transp
...
+const (
+	// mcpToolKindAgent is a regular tool exposed to the agent/model with no
+	// interactive UI rendering.
+	mcpToolKindAgent mcpToolKind = iota
+	// mcpToolKindApp is an agent-visible tool whose result renders as an
+	// interactive MCP App (UI) widget in the chat (declares a ui.resourceUri).
+	mcpToolKindApp
+	// mcpToolKindAppOnly is hidden from the agent and only callable from within
+	// the rendered MCP App (visibility declares "app" but not "model").
+	mcpToolKindAppOnly
```
EItanya: These names are confusing to me. So to be clear on purpose, `mcpToolKindAppOnly` is a tool which is only meant to be called from within the UI element itself?

go/adk/pkg/mcp/registry.go
```diff
@@ -256,6 +300,159 @@ func createTransport(ctx context.Context, params mcpServerParams) (mcpsdk.Transp
...
+	}
+}
+
+// mcpUIMetadata holds the MCP-UI extension fields read from a tool's _meta.
+type mcpUIMetadata struct {
+	ResourceURI string
+	Visibility  []string
+}
+
+func parseMCPUIMetadata(meta mcpsdk.Meta) mcpUIMetadata {
```
EItanya: Can you link to the part in the MCP spec that defines where these fields are supposed to live.

go/adk/pkg/mcp/registry.go
EItanya, line 1: Can we move these mcp_ui related functions/helpers to a new file, alongside links to the MCP APPs spec so users can read where it is defined, and so we can properly track changes in the future.

go/core/internal/httpserver/handlers/mcpapps.go
EItanya, line 1: Just so I'm clear on this file. The UI element is essentially making calls to MCP resources via raw HTTP rather than actually using the MCP protocol? Why can't it just use the existing MCP server inside of the controller?

EItanya (commented):
This is looking like a great start. I just have some notes and clarifying questions to make sure we get to a clean final state

go/adk/pkg/mcp/registry.go
dimetron: Done — moved all the MCP-UI helpers (`mcpToolKind` + kinds, `mcpToolKindOf`, `parseMCPUIMetadata`/`mcpUIMetadata`, `isAppOnlyVisibility`, `normalizeVisibility`, `agentVisibleToolFilter`, plus `MCPAppToolNames`/`mcpAppToolset`) into a new `go/adk/pkg/mcp/mcp_ui.go`. The file header documents the metadata contract and links the spec:

- Overview: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/extensions/apps/overview.mdx
- Full spec: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

`registry.go` now only keeps transport/toolset wiring. Pushing shortly.

go/adk/pkg/mcp/registry.go
dimetron: These fields are not part of the core MCP spec — they come from the **MCP Apps** extension and live under the tool's `_meta.ui` object (`_meta.ui.resourceUri`, `_meta.ui.visibility`). I've added the links in the new `mcp_ui.go` header:

- Overview (defines `_meta.ui.resourceUri` / preloading / sandbox): https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/extensions/apps/overview.mdx
- Full spec (`2026-01-26`): https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

Per the overview, the tool description carries `_meta.ui.resourceUri` pointing at a `ui://` resource that the host preloads and renders in a sandboxed iframe.

go/adk/pkg/mcp/registry.go
dimetron: Yes, exactly — `mcpToolKindAppOnly` is a tool that is hidden from the model and only invoked from within the rendered MCP App itself (e.g. the widget's own refresh/drill-down buttons). It maps to `_meta.ui.visibility: ["app"]` without `"model"`.

To make the kinds clearer and align with the spec's `visibility` vocabulary (`"model"` / `"app"`), I renamed `mcpToolKindAgent` → `mcpToolKindModel`:

- `mcpToolKindModel` — regular tool visible to the model (the LLM/agent); no UI. (visibility `"model"` or absent)
- `mcpToolKindApp` — model-visible **and** declares `_meta.ui.resourceUri`, so it renders an MCP App widget; the model may call it too.
- `mcpToolKindAppOnly` — visible to the app only, hidden from the model.

Each const now has a doc comment tying it back to the `_meta.ui` fields, in the new `mcp_ui.go`.

go/adk/pkg/mcp/registry.go
dimetron: It's not strictly required — `MCPAppToolNames` is a named `map[string]bool` that travels across a few signatures (`agentVisibleToolFilter` → `mcpAppToolset` → `MCPAppToolNamesFromToolsets` → `agent.MakeMCPAppModelResultCallback`). The alias gives that value one documented meaning ("set of model-visible tool names that render as MCP App widgets; value always true, only key presence matters") instead of a bare `map[string]bool` at each call site, and it's the natural home for that doc comment. No behavior is attached. Happy to inline it back to `map[string]bool` if you'd prefer fewer named types — let me know.

go/core/internal/httpserver/handlers/mcpapps.go
dimetron: Fixed — the handler now resolves both kinds. `resolveRemoteMCPServer` first looks up a `RemoteMCPServer`, and on NotFound falls back to the kmcp `MCPServer` CRD, converting it via `ConvertMCPServerToRemoteMCPServer` so both share one connect path (`mcpapps.go`, `resolveRemoteMCPServer`/`connect`). An `MCPServer` with UI metadata should now render at `/apps`. Could you re-test with your CRD and confirm before we resolve this thread?

ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts
Comment (dimetron) [reply ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts]: These need to be real HTTP route handlers rather than server actions. The rendered MCP App runs in a sandboxed iframe and talks to the host over `fetch` / `postMessage` (per the MCP Apps spec — the app/host channel is JSON-RPC over postMessage, and tool calls are proxied to stable URLs). That requires addressable endpoints (list tools / call tool / read resource) plus `OPTIONS`/CORS, which server actions can't provide — a server action is RPC bound to a server-component render and isn't invokable from the sandboxed app frame. So yes, intentionally the first API routes; they're the host-broker surface for the iframe, not general data fetching. Spec ref: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/extensions/apps/overview.mdx

ui/src/app/servers/[namespace]/[name]/apps/page.tsx
Comment (dimetron) [reply ui/src/app/servers/[namespace]/[name]/apps/page.tsx]: Agreed — will fold MCP App tools into the existing server tools list with a distinct icon/badge instead of a separate `/apps` page + context-menu entry, and drop the dedicated route. One question: do you want the standalone inspector removed entirely, or kept as the target the icon links to? I'll default to removing it and rendering inline unless you'd rather keep an inspector view.

go/core/internal/httpserver/handlers/mcpapps.go
dimetron: Good question — the file is actually doing the opposite of bypassing MCP: it's the controller acting as an **MCP client proxy** so the browser doesn't have to.

Two hops:
- **browser ↔ controller** is plain REST/JSON (`/api/mcp-apps/.../tools`, `/call`, `/resources`). The MCP App runs in a sandboxed iframe (mcp-ui client) and can't open the agent's MCP transport to arbitrary upstreams; it also can't hold the server's auth/headers. So the host exposes a small REST surface, gated by the usual RBAC `Check` on `ToolServer`.
- **controller ↔ MCP server** is the real MCP protocol: `connect()` builds `mcp.NewClient`, `Connect`s over `StreamableClientTransport`/`SSEClientTransport`, then calls `session.ListTools` / `CallTool` / `ReadResource`. Headers/secrets are resolved server-side via `RemoteMCPServer.ResolveHeaders`.

Why not the existing MCP server in the controller: that server exposes the *agent's own* (A2A) tools. The MCP App tools and their `ui://` resources live on the **configured RemoteMCPServer/MCPServer** the agent connects to (often external, e.g. servereverything.dev). The in-controller server doesn't proxy those upstream `ui://` reads or arbitrary upstream tool calls with per-server resolved auth — that's exactly the gap this handler fills (resolve the CR → resolve headers → MCP-client proxy).

If you'd prefer to fold this into the existing MCP server as a dedicated proxy capability instead of a separate handler, I'm open to it — happy to discuss the shape.

ui/src/app/servers/[namespace]/[name]/apps/page.tsx
Comment (dimetron) [reply ui/src/app/servers/[namespace]/[name]/apps/page.tsx]: Follow-up: this is now implemented on the current branch (came in with the latest main merge), so the separate page/menu approach you commented on is gone:

- MCP App tools are listed inline in the server tools grid in ui/src/components/mcp/McpServersView.tsx — each tool picks `const ToolIcon = isApp ? AppWindow : FunctionSquare` (~L347-348) and app tools get an `App` badge. App-ness is resolved lazily on expand via listMcpAppTools.
- The standalone /apps route (ui/src/app/servers/[namespace]/[name]/apps/page.tsx), the McpAppsInspector component, and the "MCP Apps" context-menu entry have all been removed. Only McpAppRenderer.tsx remains (it renders the widget inline in chat).

So no separate page and no extra menu item — apps and tools share one list, differentiated by icon + badge, as you suggested. Could you take a look and resolve if it matches what you had in mind?

ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts
Comment (peterj) [reply ui/src/app/api/mcp-apps/[namespace]/[name]/tools/[toolName]/call/route.ts]: ah, makes sense; it's for the mcp app to call back

dimetron:
### Restored the high-water-mark send guard (commit 9ab57178)

Heads up on the latest push. The merge from `main` pulled in the content-signature send guard from #2034 (`countSendGuardComparableMessages` / `countBackendBackedComparableMessages`), and on this branch it regressed once the MCP UI widgets were in play.

**Symptom:** after a normal chat turn, the next send was blocked with *"New messages loaded — please review before sending"* even though nothing had changed and no other tab was open.

**Root cause:** that guard compares the content *signature* of locally rendered messages against messages re-extracted from the DB. After a turn the UI does not call `reloadSessionFromDB()`, so the live-streamed messages stay in memory. Their `toolResultData.raw_result` serializes slightly differently from the freshly-extracted DB copy — most visibly for MCP App widgets, whose `structuredContent` payloads are large and nested. So `localCount < dbCount` and the guard fired a false positive. This also masked the kanban `show_task_progress` widget: the tool call was blocked before it could ever run.

**Fix:** reverted to the content-independent high-water mark this branch used before the merge. It counts persisted history items across tasks (`task.history.length`) — a value the DB assigns identically for every tab, independent of how each tab rendered them (synthetic tool/artifact/summary cards never land there). We set the mark on load/reload and re-read it after each completed turn (`refreshServerMark`), and block only when the server count exceeds the synced mark, i.e. another tab actually advanced the conversation.

Also removed the now-dead comparable-message helpers from `messageHandlers.ts` and rewrote the send-guard tests for the high-water behavior (same-tab no-block, cross-tab block, post-block proceed, Cmd/Ctrl+Enter). `jest` green (32/32), `eslint` clean.

EItanya:
Posting a quick summary from `codex` that I think is worth discussing. Specifically these are areas where it claims the implementation deviates from the spec. The answer to some of these may be TODO or other, but just wanted to make sure we discussed them.

```
  1. Backend lets app-originated calls invoke tools that may not be app-visible.
     /tmp/kagent-pr-2052/go/core/internal/httpserver/handlers/mcpapps.go:86 accepts any {toolName} and calls it directly at line 120. The UI has a client-side check, but it only
     throws when a registry entry exists and is neither app-only nor agent-visible; if the tool is unknown, it still calls the backend. See /tmp/kagent-pr-2052/ui/src/
     components/mcp-apps/McpAppRenderer.tsx:117. The spec says app tools/call requests must be rejected when the target tool does not include "app" visibility. This needs
     server-side enforcement.

  2. Sandbox/CSP handling is not compliant enough for untrusted UI.
     The spec requires web hosts to use an intermediate sandbox with a different origin and to enforce CSP from resource._meta.ui.csp, with restrictive defaults. This PR
     serves /sandbox_proxy.html from the same origin as the host and loads arbitrary HTML with document.write without applying CSP or permissions from resource metadata. See /
     tmp/kagent-pr-2052/ui/src/components/mcp-apps/McpAppRenderer.tsx:70 and /tmp/kagent-pr-2052/ui/public/sandbox_proxy.html:32. This is the biggest security/spec gap.

  3. MCP Apps capability negotiation is missing.
     The spec defines MCP Apps as an extension negotiated via capabilities.extensions["io.modelcontextprotocol/ui"] with supported MIME types. The Go ADK discovery client
     connects with nil options at /tmp/kagent-pr-2052/go/adk/pkg/mcp/mcp_ui.go:194, and the backend MCP Apps proxy does the same at /tmp/kagent-pr-2052/go/core/internal/
     httpserver/handlers/mcpapps.go:256. A conformant server that only registers UI tools when the client advertises support may not expose them to kagent.

  4. Python runtime path does not appear to hide app-only tools from the model.
     The Go path explicitly filters visibility: ["app"] out of the agent-visible tool list in /tmp/kagent-pr-2052/go/adk/pkg/mcp/mcp_ui.go:215. The Python path records tools
     with mcp_app_resource_uri, but I don’t see visibility parsing/filtering in /tmp/kagent-pr-2052/python/packages/kagent-adk/src/kagent/adk/_mcp_toolset.py:146. Unless
     upstream Google ADK already enforces _meta.ui.visibility, Python agents may expose app-only helper tools to the model.
 ```

File: go/core/internal/httpserver/handlers/mcpapps.go
Comment (EItanya) [reply go/core/internal/httpserver/handlers/mcpapps.go:1]: Sounds good to me, resolving.

File: go/adk/pkg/mcp/registry.go (hunk 14 above)
Comment (EItanya) [reply go/adk/pkg/mcp/registry.go]: I would prefer removing it. The use of a custom type here is a bit arbitrary and the variables will be named anyway.

File: go/adk/pkg/mcp/registry.go (hunk 15 above)
Comment (EItanya) [reply go/adk/pkg/mcp/registry.go]: The names are still confusing to me. How about `mcpToolKindAppInternal`?

**Comment by dimetron** [comment]
## MCP Apps spec-deviation review — addressed in `cf88dc10`

Critical review of the four flagged items, with status, blocking impact, and the fix shipped on this branch. None of the fixes change how widgets render or break the `server-everything` demo.

| # | Claim | Status | Severity | Blocks merge? | Fix possible? | Fix description |
|---|-------|--------|----------|---------------|---------------|-----------------|
| 1 | Backend lets app-originated calls invoke non-app-visible tools | Valid | High (security) | Yes | Yes — small | `HandleCallTool` lists tools on the session it already opens, parses `_meta.ui.visibility`, and rejects only when visibility is present and lacks `"app"` (default / `["app"]` / `["model","app"]` stay allowed). Client check kept as defense-in-depth. |
| 2a | Sandbox applies no CSP (ignores `ui.csp`) | Valid | High (security) | Yes | Yes — small | `@mcp-ui/client` already forwards CSP (`sandbox-resource-ready {html, csp}` + `?csp=`). Host reads `_meta.ui.csp` and passes it via the `sandbox` prop; `sandbox_proxy.html` injects a `<meta http-equiv="Content-Security-Policy">` with the spec restrictive default + `object-src 'none'`. Default keeps `script-src 'self' 'unsafe-inline'`, so widgets still render. |
| 2b | Sandbox is same-origin as host (`allow-same-origin` → no isolation) | Valid | High (security) | Follow-up | Architectural | Serve the proxy from a distinct origin so `allow-same-origin` no longer resolves to the host origin. Deploy/topology change — tracked as a follow-up; boundary stays soft until it lands. |
| 3 | MCP Apps capability negotiation missing | Valid (SHOULD) | Low (interop) | No | Yes — tiny | Both Go MCP clients now advertise `io.modelcontextprotocol/ui` (`mimeTypes: ["text/html;profile=mcp-app"]`) via `ClientCapabilities.Extensions`. Purely additive. |
| 4 | Python runtime doesn't hide app-only tools from the model | Valid | Medium (correctness) | Yes | Yes — small | `KAgentMcpToolset.get_tools` drops tools whose visibility is `["app"]` without `"model"` from the model-visible list (still app-callable), mirroring the Go ADK filter. |

### Implemented (1, 2a, 3, 4)
- **1** — `toolAllowsAppCall` + `visibilityAllowsApp` in `go/core/internal/httpserver/handlers/mcpapps.go`; app calls to non-app-visible tools now return `403` (`404` for unknown). Test: `TestVisibilityAllowsApp`.
- **2a** — `ui/public/sandbox_proxy.html` builds + injects a spec CSP; `McpAppRenderer` captures `_meta.ui.csp` and forwards it via `sandbox`.
- **3** — `mcpUIClientCapabilities` in `go/adk/pkg/mcp/mcp_ui.go` and the `connect` client in `mcpapps.go`.
- **4** — filter in `python/.../_mcp_toolset.py`. Test: `test_get_tools_hides_app_only_tools_from_model`.

### Follow-up (2b)
Cross-origin sandbox isolation is a deployment/topology change, left as a tracked follow-up.

Verified locally: Go (`core` + `adk`) build & tests, Python unit tests (16 passed), UI lint + jest MCP-apps suites.

**Comment by EItanya** [comment]
FYI @dimetron, this is what's failing the snyk job: 
Screenshot 2026-06-30 at 08 55 56

File: go/adk/pkg/mcp/registry.go (hunk 15 above)
Comment (dimetron) [reply go/adk/pkg/mcp/registry.go]: Done in 77501c0 — renamed `mcpToolKindAppOnly` → `mcpToolKindAppInternal` (and its `String()` label to `app_internal`), and updated the classifier/filter usages, the `registry_test.go` cases, and the Python mirror's cross-reference comment. No behavior change.

File: go/core/internal/httpserver/handlers/mcpapps.go [hunk 17]
```diff
@@ -0,0 +1,406 @@
...
+	}
+	return namespace, name, true
+}
+
+// resolveRemoteMCPServer locates the MCP endpoint for the given ref, supporting
+// both RemoteMCPServer (external URL) and the kmcp MCPServer CRD (an in-cluster
+// Deployment+Service). An MCPServer is converted to the same RemoteMCPServer
+// shape the controller uses for tool discovery, so both kinds share one connect
+// path.
+func (h *MCPAppsHandler) resolveRemoteMCPServer(ctx context.Context, namespace, name string) (*v1alpha2.RemoteMCPServer, error) {
```
EItanya: There are a number of issues with this:
1. The name makes it seem like it's only `RemoteMcpServer`
2. The behavior is confusing, if there is a `RemoteMCPServer` with the same `name/ns` as an `MCPServer` than it will be queried, even if the user intended to get the `MCPServer`

go/adk/pkg/mcp/mcp_ui.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// MCPAppToolNames is the set of MCP tool names whose results render as
+// interactive MCP App (UI) widgets in the chat (the tool declares a
+// `_meta.ui.resourceUri` and is visible to the model). It is used as a set, so
+// the bool value is always true and only key presence is meaningful. The agent
+// attaches the model-result compaction callback (see
+// agent.MakeMCPAppModelResultCallback) only to these tools. Collect them from
+// CreateToolsets output via MCPAppToolNamesFromToolsets.
+type MCPAppToolNames map[string]bool
```
EItanya: I thought we discussed getting rid of this type as it is just cosmetic.

EItanya (changes requested):

dimetron:
@EItanya ready for another look when you get a chance. Both threads from your Jul 1 review are addressed in edab7e7f:

- **`resolveRemoteMCPServer`** → renamed to `resolveMCPServerEndpoint` (no longer implies RemoteMCPServer-only) and now takes a `groupKind`. Resolution switches on the kind: `MCPServer` reads only the MCPServer, `RemoteMCPServer` reads only that one — so a RemoteMCPServer and MCPServer sharing a namespace/name no longer collide. The UI threads the selected server's `groupKind` through the list/call/read endpoints; an empty `groupKind` keeps the legacy RemoteMCPServer-then-MCPServer fallback for older callers.
- **`MCPAppToolNames`** type dropped — now plain `map[string]bool` across `mcpAppToolset`, `MCPAppToolNamesFromToolsets`, and `MakeMCPAppModelResultCallback`; the variables carry the meaning.

Branch is up to date with `main` and all CI is green.

EItanya (dismissed):
Approval on the go side :)

peterj (approved):

PR #2056, opened 2026-06-21
feat(auth): add Dex OIDC authenticator for A2A endpoint

QuentinBisson:
## What

Adds `--auth-mode=dex-oidc` to the kagent controller, implementing full Dex JWT validation (signature via JWKS, issuer, audience, expiry) on the A2A endpoint.

## Why

The A2A endpoint previously had no real authentication for human callers. The `unsecure` and `trusted-proxy` modes both accept identity without verifying it. For human-facing A2A sessions driven by a Dex OIDC token (e.g. the `giantswarm-ad` connector on glean), the server must validate the token and reject unauthenticated requests.

## How

- **`DexAuthenticator`** (`go/core/internal/httpserver/auth/dex_authn.go`): fetches the OIDC discovery document at `{issuer}/.well-known/openid-configuration`, registers the JWKS URI in a `lestrrat-go/jwx/v2` auto-refresh cache (15 min default, minimum enforced by httprc), and validates each request's Bearer token against it. User identity defaults to the `email` claim, falling back to `sub`.
- **`UpstreamAuth`**: forwards the original `Authorization: Bearer <dex-token>` and `X-User-Id` headers to agent pods. The Python A2A SDK captures `state['headers']`; kagent's Python executor writes them to ADK session state under `HEADERS_STATE_KEY`, making the token available to `create_header_provider` for MCP calls when `allowed_headers` includes `"authorization"`.
- Two new flags: `--auth-dex-issuer` and `--auth-dex-client-id`. Both are required when `--auth-mode=dex-oidc` is set.

## Operator note

The Dex client ID passed via `--auth-dex-client-id` (e.g. `kagent-a2a`) must be added to muster's `trustedAudiences`. Today only `dex-k8s-authenticator` is listed. On glean the human connector is `giantswarm-ad` and the existing client is `kagent`.

## Tests

15 table-driven tests in `dex_authn_test.go` (valid token with groups, sub fallback, missing header, non-Bearer, expired, wrong issuer, wrong audience, garbage, upstream forwarding, discovery failure cases). Existing `auth_mode_test.go` updated and extended with `dex-oidc` validation cases.


QuentinBisson:
Wrong approach — kagent already runs behind oauth2-proxy with trusted-proxy mode. The A2A endpoint uses the same path as the UI. No code change needed; the only task is registering the Dex client for the A2A audience.

PR #2067, opened 2026-06-22
Add check for schema-agnostic sql

iplay88keys:
## Description

Adds a static check that migration SQL never hard-codes a schema, and flips that gate in the migrations guide from *Target* to enforced. Static-only — runs in the existing `go test` CI, no database required.

## Changes

`go/core/pkg/migrations/cross_track_test.go`:

- **`TestSchemaAgnosticSQL`** — rejects `CREATE`/`DROP SCHEMA`, schema-qualified DDL (`CREATE TABLE foo.bar`), `SET search_path`, and `ALTER ... SET SCHEMA` in every migration file. The schema a migration lands in must come from the connection, not the file.
- **`TestSchemaViolations`** — table-driven unit test with positive and negative cases. Detection is factored into a `schemaViolations` helper; comment stripping via `stripSQLComments`.


go/core/pkg/migrations/cross_track_test.go
```diff
@@ -212,3 +212,124 @@ func TestNoCrossTrackDDL(t *testing.T) {
...
+			got := schemaViolations(tt.sql)
+			if !equalStringSets(got, tt.want) {
+				t.Errorf("schemaViolations() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+// equalStringSets compares two string slices ignoring order and duplicates.
+func equalStringSets(a, b []string) bool {
```
EItanya, line 318: can you use `slices.Equal` for this? Does it matter that duplicates are removed?

go/core/pkg/migrations/cross_track_test.go
```diff
@@ -212,3 +212,124 @@ func TestNoCrossTrackDDL(t *testing.T) {
...
+	for _, track := range tracks {
+		sub, err := fs.Sub(migrations.FS, track)
+		if err != nil {
+			t.Fatalf("fs.Sub(%q): %v", track, err)
+		}
+		err = fs.WalkDir(sub, ".", func(path string, d fs.DirEntry, err error) error {
+			if err != nil || d.IsDir() || !strings.HasSuffix(path, ".sql") {
+				return err
+			}
+			data, err := fs.ReadFile(sub, path)
```
EItanya, line 273: nit: you can use the following to avoid reading the whole thing into memory. Not necessarily important here but keep in mind for future.

```
 f, err := os.Open(path)
  if err != nil {
      return err
  }
  defer f.Close()

  scanner := bufio.NewScanner(f)
  for scanner.Scan() {
      line := scanner.Text()
      // use line
  }
  if err := scanner.Err(); err != nil {
      return err
  }
```

EItanya (approved):

go/core/pkg/migrations/cross_track_test.go
iplay88keys: Ah, yes, of course. Good catch!

PR #2074, opened 2026-06-23
fix: handle missing ate.dev CRDs gracefully in substrate status endpoint

AruneshDwivedi:
When the ate.dev CRDs (WorkerPool, ActorTemplate) are not installed on the cluster, the substrate status endpoint returned HTTP 500 due to an unhandled *meta.NoKindMatchError from the KubeClient.List calls.

Now catches NoKindMatchError for both WorkerPool and ActorTemplate CRDs and returns empty lists instead, so the endpoint returns 200 with empty arrays when the CRDs are absent. Other errors still propagate normally.

Added a test (TestHandleGetSubstrateStatus_MissingCRDs) that uses a fake client returning NoKindMatchError to verify the endpoint returns 200 with empty data.


EItanya:
Thanks so much for the contribution. Now that we've landed on substrate as our sandboxing mechanism I would like to revisit how we handle missing substrate dependencies and unify the approach across the whole app

AruneshDwivedi:
Good catch — changed both CRD-absence log lines from Info to V(1) so they don't spam normal logs when the endpoint is polled frequently. Available at verbose level for debugging.

PR #2076, opened 2026-06-23
fix(controller): reference agent images by tag for private registries

### LINKED ISSUE #2055: [BUG] Pinned Agent images | Ignoring --image-registry / --image-repository in private-registry deployments
### 📋 Prerequisites

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

In >=v0.9.7, the controller pins declarative agent images using digests (registry/repository@sha256:...). 
While the registry and repository are configurable at runtime via _--image-registry_ and _--image-repository,_ the digest itself is injected at controller link time from upstream images.

https://github.com/kagent-dev/kagent/blob/a766d22c96a4d204ed01935f72e7abf903dcf608/scripts/controller-digest-ldflags.sh

There is currently no runtime override for the digest, and _**--image-tag**_ is silently ignored for declarative agents. As a result, in private/mirrored registry setups, the controller generates image references like:

```my-registry.example.com/my-org/kagent/app@sha256:<UPSTREAM_DIGEST>```

If the mirrored image does not retain the exact upstream manifest digest (e.g. due to rebuilds or recompression), all agent pods fail with: `ImagePullBackOff / manifest unknown`

### 🔄 Steps To Reproduce

1. Prepare a private registry (e.g. ECR)
2. Deploy the controller
3. Create a declarative Agent
4. Inspect the generated Deployment. ```kubectl get deploy <agent-name> -o yaml```

### 🤔 Expected Behavior

When _--image-registry / --image-repository_ (Helm: agents.image.*) point to a private registry, the controller should emit an image reference that is resolvable in that registry, e.g.:

- honoring --image-tag, or
- allowing a runtime-provided digest that exists in the private registry

### 📱 Actual Behavior

- Image references are constructed using:
  - runtime registry + repository
  - link-time upstream digest
- --image-tag has no effect for declarative agents

**Result**: invalid image references in private registries

### 💻 Environment

- AWS EKS - eks.29
- Kubernetes - 1.33
- Kagent: 0.7.23
- Model: Bedrock - Anthropic Claude Sonnet 4.5

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

igabi:
## What

Declarative agent pods fail with `ImagePullBackOff` / `manifest unknown` when the controller is configured to use a private registry that does not preserve the upstream image manifest digest.

Since v0.9.7, declarative agent images are pinned by their link-time digest (`registry/repository@sha256:...`), injected via `controller-digest-ldflags.sh`. When `--image-registry` / `--image-repository` point to a private registry, the controller rewrites only the registry and repository segments but keeps the hardcoded **upstream** digest. Many mirrors (e.g. `az acr import` targets, pull-through proxies that re-push) assign a different manifest digest, so the emitted reference does not resolve. `--image-tag` is silently ignored for declarative agents, so there is no way to fall back to a tag.

This is the behaviour reported in #2055.

## Fix

Add a `--pin-runtime-image-digest` flag (env `PIN_RUNTIME_IMAGE_DIGEST`), **defaulting to `true`** so current digest-pinning behaviour is unchanged.

When set to `false`, the Python and Go runtime image resolvers fall back to a tag reference (`registry/repository:tag`) through the existing `ImageConfig.PinnedImage()` helper — restoring the pre-0.9.7 behaviour for operators mirroring images into a registry that does not preserve digests.

Both runtime resolvers now share a small `resolveRuntimeImage` helper, and the "digest not set at link time" error now points operators at the new flag.

The Helm chart exposes this as `controller.pinRuntimeImageDigest` and plumbs it to the controller `ConfigMap` (picked up by the existing env → flag loading).

## Testing

- **Go unit tests** (`go test ./core/internal/controller/translator/agent/...`): added `TestRuntime_PythonRuntime_TagReferenceWhenDigestPinningDisabled` and `TestRuntime_GoRuntime_TagReferenceWhenDigestPinningDisabled`, asserting the rendered deployment references the image by tag (`/app:<tag>`, `/golang-adk:<tag>`) and contains no `@sha256:` when pinning is disabled. Existing digest-pinning tests continue to pass.
- **Helm unit test**: added `tests/controller-configmap_test.yaml` asserting `PIN_RUNTIME_IMAGE_DIGEST` is `"true"` by default and `"false"` when `controller.pinRuntimeImageDigest=false`.
- `go vet` and `gofmt` clean on the changed packages.

## Usage

```yaml
# values.yaml — mirror that does not preserve upstream digests
controller:
  agentImage:
    registry: my.private.registry/mirror
    repository: kagent-dev/kagent/app
  pinRuntimeImageDigest: false   # emit registry/repository:tag instead of @sha256
```

Fixes #2055


EItanya (changes requested):
So I actually think that a different fix is in order here. We switched to digests because substrate does not allow tags, but that surfaced an implementation detail of that system into the broader `kagent`. Rather I think we should switch back to tags for standard deployments, and then allow overriding of digests for substrate. What do you think?

mesutoezdil:
runtime_test.go -- pinruntimeImagedigest=false is tested for the golang-adk path but not for the golang-adk-full path (skills-enabled agents). is this intentional?

Jan-M:
Would be great to get this resolved. We struggle with it as well.

ppeau:
Hi,

We are also blocked by this issue.

This bug potentially affects any organization using private registries with mirroring (ACR import, Harbor, Nexus, Artifactory, etc.). The digest pinning breaks declarative agents as soon as the mirror re-assigns a different digest.

If needed, I'm happy to pick this up and finish it.

It would be great to get this fixed soon.

iplay88keys:
I have a fix for this [here](https://github.com/kagent-dev/kagent/pull/2242).

iplay88keys:
Closed as my [PR](https://github.com/kagent-dev/kagent/pull/2242) has been merged. I will be [backporting](https://github.com/kagent-dev/kagent/pull/2266) it to 0.9.x as well.

PR #2077, opened 2026-06-23
feat(chat): file upload artifacts

dimetron:
## Summary

End-to-end **file upload / download** in kagent chat for the Go ADK runtime, riding the existing A2A message/SSE channel (no new HTTP API, no CRD field). Files travel inline (base64) as A2A `FilePart`s; agent-produced files are surfaced back as downloadable artifacts. See `design/EP-89405-file-upload-artifacts.md`.

- **Go ADK runtime**: wire the in-memory `ArtifactService` into the runner; enable `SaveInputBlobsAsArtifacts`; emit agent-saved artifacts as A2A `FilePart` events driven by the `ArtifactDelta` signal; register `load_artifacts` + a new `save_artifact` tool; add a `fileextract` package (PDF + others) so non-multimodal models still receive document text.
- **Python runtime**: mirror text extraction (`_file_extract.py`) and inject into the OpenAI path.
- **UI**: attach multiple files with type/size validation (images, PDF, text/md, CSV, JSON; 10 MB cap), render image thumbnails + downloadable file chips in both user and agent bubbles.
- **Edge**: add `client_max_body_size` (default `50m`) to the UI nginx config so uploads aren't rejected at the proxy.

## Changes

- `go/adk/pkg/{runner,agent,a2a,tools,fileextract,models}/…` — artifact wiring, save/load tools, extraction, A2A surfacing (+ unit tests)
- `go/core/test/e2e/file_upload_test.go` — e2e upload round trip (a2a-go v2)
- `python/packages/kagent-adk/…` — `_file_extract.py`, `_openai.py` injection (+ tests)
- `ui/src/lib/fileUpload.ts`, `ui/src/components/chat/{ChatInterface,ChatMessage,FileAttachment}.tsx`, `messageHandlers.ts` (+ tests)
- `helm/kagent/{files/nginx.conf,values.yaml,tests/ui-nginx-configmap_test.yaml}` — `clientMaxBodySize`
- `design/EP-89405-file-upload-artifacts.md` — design EP

## Test plan

- [ ] `make -C go test` (adapter, agent, executor, artifacts, save_artifact, fileextract, openai_adk)
- [ ] `make -C go e2e` — `TestE2EFileUploadGoADKAgent` round trip on kind
- [ ] Python: `uv run pytest` (`test_file_extract.py`, `test_openai.py`)
- [ ] UI: unit tests (`fileUpload`, `messageHandlers`, `FileAttachment`, `ChatMessage`)
- [ ] Helm: `ui-nginx-configmap_test.yaml`
- [ ] Manual: attach image/PDF/CSV in chat, confirm thumbnails/chips + agent-produced artifact downloads


ui/src/lib/messageHandlers.ts
```diff
@@ -157,6 +160,28 @@ export function extractMessagesFromTasks(tasks: Task[]): Message[] {
...
+    // Surface agent-produced file artifacts (e.g. save_artifact output). These
+    // live in task.artifacts, not task.history, so the history loop above misses
+    // them; without this they vanish (and become un-downloadable) on reload.
+    for (const artifact of task.artifacts ?? []) {
+      const fileParts = extractFileParts(artifact.parts).filter((fp) => {
+        const key = `${fp.file.name ?? ""}|${isFileWithBytes(fp.file) ? fp.file.bytes.length : (fp.file as { uri?: string }).uri ?? ""}`;
+        if (seenArtifactFiles.has(key)) return false;
+        seenArtifactFiles.add(key);
+        return true;
+      });
```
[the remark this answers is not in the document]
dimetron: Valid — fixed in 64460a0f. The dedupe key now includes the artifact's stable identity (`artifactId`) in addition to file name+size: `${artifact.artifactId}|${name}|${bytesLen|uri}`. Distinct artifacts / new versions (different `artifactId`) are no longer collapsed by a name+size clash, while multiple files within one artifact stay distinct (name component). Added a regression test `keeps distinct artifacts that share name+size`.

ui/src/lib/messageHandlers.ts
```diff
@@ -544,6 +569,22 @@ function isDataPart(part: Part): part is DataPart {
   return part.kind === "data";
 }
 
+/** Type guard for inline-bytes file parts (the only kind kagent transports). */
+export function isFilePart(part: Part): part is FilePart {
+  return part.kind === "file";
+}
```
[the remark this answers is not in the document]
dimetron: Valid (doc clarity) — fixed in 64460a0f. Reworded the doc comment: `isFilePart` is a guard for file parts (`kind === 'file'`); use the separate `isFileWithBytes` guard to assert the inline `bytes` payload.

helm/kagent/values.yaml
```diff
@@ -351,6 +351,20 @@ ui:
...
+  # -- Nginx proxy timeout configuration for the UI sidecar (values are passed
+  # directly to the corresponding nginx directives, e.g. "1800s").
+  nginx:
+    # -- proxy_read_timeout: max time between two successive reads from the upstream.
+    proxyReadTimeout: 1800s
+    # -- proxy_send_timeout: max time between two successive writes to the upstream.
+    proxySendTimeout: 1800s
+    # -- client_max_body_size: max allowed size of the client request body. Increase
+    # to allow large chat messages/attachments (set to 0 to disable the limit).
+    clientMaxBodySize: 50m
```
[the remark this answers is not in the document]
dimetron: Valid — fixed in 64460a0f. `streamTimeoutSeconds`, `nginx.proxyReadTimeout`, and `nginx.proxySendTimeout` were dead config (nginx.conf hardcodes the proxy timeouts; the UI client uses a hardcoded stream timeout). Removed them — and the unrelated oauth2-proxy timeout/cookie-refresh knobs that rode along — keeping only `ui.nginx.clientMaxBodySize`, which is the one value actually templated into nginx.conf for this feature.

design/EP-89405-file-upload-artifacts.md
```diff
@@ -0,0 +1,175 @@
...
+  `runConfig.SaveInputBlobsAsArtifacts = true`; on each event with
+  `Actions.ArtifactDelta`, `Load` each `(name, version)` from the store, set
+  `InlineData.DisplayName`, convert via `ToA2APart`, and emit an A2A
+  artifact-update `FilePart` event (`LastChunk=true`). Load/convert errors are
+  logged and skipped so the turn continues.
+- **`go/adk/pkg/a2a/artifacts.go`** — helpers for building/emitting artifact
+  events from saved ADK parts.
+- **`go/adk/pkg/fileextract/`** (`fileextract.go`, `pdf.go`) — extract text from
+  uploaded documents (PDF and other supported types) so the content is injected
+  for models that can't read the raw file; includes a PDF-extractor timeout.
```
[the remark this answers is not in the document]
dimetron: Valid — fixed in 8ff83805. There is no timeout logic in `go/adk/pkg/fileextract/`, so the "includes a PDF-extractor timeout" claim was removed from the EP.

peterj:
is it expected that I can download the file again after I navigate away from the session? I see the spec says that data is lost on pod restart - not sure if that also means on browser refresh

The download icon is missing: 
Screenshot 2026-06-23 at 2 21 58 PM

go/adk/pkg/tools/save_artifact_tool.go
```diff
@@ -0,0 +1,93 @@
...
+		MIMEType:    mimeType,
+		DisplayName: name,
+	}}
+
+	resp, err := artifacts.Save(ctx, name, part)
+	if err != nil {
+		return nil, fmt.Errorf("failed to save artifact %q: %w", name, err)
+	}
+
+	return map[string]any{
```
peterj, line 86: couldn't we return the `resp` from `artifacts.Save` here?

ui/src/lib/fileUpload.ts
```diff
@@ -0,0 +1,88 @@
...
+  "application/json",
+  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+  "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+  "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+  "application/epub+zip",
+]);
+
+// File extensions allowed as a fallback when the browser reports an empty MIME.
+export const ALLOWED_EXTENSIONS = [
+  ".md",
```
peterj, line 34: yaml? xml?

DEVELOPMENT.md
```diff
@@ -177,3 +177,31 @@ docker run --rm \
   --net=host \
   localhost:5001/kebab:latest
 ```
+
+### File uploads (artifacts)
+
+The chat UI supports uploading files to agents running on the **Go ADK runtime**
+(Python runtime is out of scope). Files travel inline (base64) over the existing
```
peterj: is this still correct that python is out of scope?

mesutoezdil:
artifacts.go:60 -- checkinboundfilesizes fully decodes every base64 payload just to measure its size. a 9.9 mb upload allocates ~10 mb on the heap for a size check that could be avoided with a len(fb.bytes) * 3 / 4 > limit pre-screen. 

and save_artifact_tool.go:60 -- only base64.stdencoding is tried. if the llm produces url-safe (-_) or unpadded base64, the tool call fails with a confusing decode error. should it fall back to base64.urlencoding / base64.rawstdencoding?

go/adk/pkg/tools/save_artifact_tool.go
dimetron: Done in 928105a0 — `Save` now returns `resp`, and the tool result surfaces `resp.Version` in the `version` field, so the caller/LLM gets the persisted artifact version back instead of discarding it.

ui/src/lib/fileUpload.ts
dimetron: Added in 928105a0 — YAML and XML are now in the allowlist (`application/xml`, `text/xml`, `application/yaml`, `application/x-yaml`, `text/yaml`), along with `.xml`/`.yaml`/`.yml` extensions and the `accept` attribute. Both the Go (tabula) and Python (markitdown) extractors already treat them as text, so they round-trip as readable content.

DEVELOPMENT.md
dimetron: Corrected in 928105a0 — Python is not fully out of scope. The doc now states the Go ADK runtime owns the full artifact round trip (persistence, `save_artifact`/`load_artifacts`, surfacing agent-produced files), while the Python runtime mirrors just the text-extraction path so uploaded documents still reach non-multimodal models.

dimetron:
Both addressed in 928105a0:

- `artifacts.go` — `checkInboundFileSizes` now derives the decoded size from the base64 length (`base64DecodedLen`) instead of decoding the whole payload, so a ~10 MB upload no longer allocates ~10 MB on the heap just for the size check. base64 validity is still enforced downstream when the bytes are decoded for persistence.
- `save_artifact_tool.go` — `decodeFlexibleBase64` now falls back across `StdEncoding`/`RawStdEncoding`/`URLEncoding`/`RawURLEncoding`, so url-safe or unpadded base64 produced by the LLM no longer fails with a confusing decode error.

Thanks for the catch — both have regression tests added.

design/EP-89405-file-upload-artifacts.md
```diff
@@ -0,0 +1,174 @@
+# EP-89405: File Upload & Artifact Support in Chat
+
+* Issue: [MSFN-89405] (internal) — feat: file upload feature
```
EItanya: Where did you get `MSFN-89405`?

DEVELOPMENT.md
```diff
@@ -177,3 +177,33 @@ docker run --rm \
   --net=host \
   localhost:5001/kebab:latest
 ```
+
+### File uploads (artifacts)
```
supreme-gg-gg: Can you remove this from development.md? This has been explained in the design doc already.

go/adk/pkg/a2a/artifacts.go
```diff
@@ -0,0 +1,154 @@
...
+		return p
+	case a2atype.FilePart:
+		return &p
+	}
+	return nil
+}
+
+// emitArtifacts loads each artifact named in delta from the artifact service
+// and emits it as an A2A artifact event carrying a FilePart. Load/convert
+// failures are logged and skipped so the turn continues (AC4).
```
supreme-gg-gg: nit: can you remove (AC1-8) from the comments, I find it confusing to run into these abbreviations in the code

go/core/test/e2e/mocks/invoke_file_upload_agent.json
```diff
@@ -0,0 +1,35 @@
+{
+  "openai": [
+    {
+      "name": "file_upload_request",
+      "match": {
+        "match_type": "contains",
+        "message": {
+          "content": "confirm you received the uploaded file",
```
supreme-gg-gg: Shouldn't you match agains the file content `hello from an uploaded file` instead of the user message? Otherwise, this test will always pass whether the file upload pipeline (a2a ingest -> save input as artifact -> `fileextract` -> model adapter) works or not

Also (this is optional but for completeness) having a subtest on the download scenario as well (calling `save_artifact` tool and checking the returned `task.Artifacts`) would cover the entire e2e flow

supreme-gg-gg (commented):
Are you planning to add python support later? I believe it makes most sense to add both together unless there's any blocker on the python path. At least we can make the UI only show the attach button for Go runtime so it doesn't mislead the user into thinking that this is currently supported for all agents.

Also, is there a reason why only OpenAI models have the inline file to text extraction support? This seems like a generic functionality that can be added to other model providers easily, given that part of the motivation behind extracting text is to make it compatible with more models.

dimetron:
> is it expected that I can download the file again after I navigate away from the session? I see the spec says that data is lost on pod restart - not sure if that also means on browser refresh
> 
Current implementation has default in-memory artifact service.

design/EP-89405-file-upload-artifacts.md
dimetron: `MSFN-89405` was an internal tracker id that leaked in when I drafted this — it has no meaning in this repo. Removed from the doc; the EP header now points at the GitHub issue/PR only.

The file name still carries the `89405` digits. Happy to renumber it to a kagent issue number (e.g. `EP-2077-...`) if you'd prefer the EPs keyed consistently to GitHub.

DEVELOPMENT.md
dimetron: Done — the PR no longer touches `DEVELOPMENT.md` at all (`git diff main...HEAD -- DEVELOPMENT.md` is empty). The Go/Python runtime split for the artifact round trip is described in EP-89405 only.

go/adk/pkg/a2a/artifacts.go
dimetron: Done — all `(ACn)` shorthand is gone from code and test comments; each comment now states the behaviour it guarantees in plain language. The `AC1–AC8` labels only survive in the EP doc, where they're actually defined.

go/core/test/e2e/mocks/invoke_file_upload_agent.json
dimetron: Good catch — you were right that matching the typed prompt made the test vacuous. Fixed: the mock now matches `has been saved to the artifacts`.

That string isn't arbitrary. With `SaveInputBlobsAsArtifacts = true`, the ADK runner replaces the inbound inline file part with `"Uploaded file: %s. It has been saved to the artifacts"` before the model is called (`runner/runner.go`). So the phrase reaches the model request *only* if a2a ingest → save-as-artifact actually ran; if the upload were dropped, the request would contain just the typed text and the mock wouldn't match, failing the test.

That substitution is also why I didn't match the raw file content (`hello from an uploaded file`) as suggested — on the Go path the bytes never reach the model request, they're swapped for the placeholder, so that assertion could never pass. The test comment now spells this out.

The download subtest (`save_artifact` → assert `task.Artifacts`) isn't in yet — agreed it would close the loop, tracking it as a follow-up rather than growing this PR further.

dimetron:
### Rebased on `main` — resolves the merge conflict, includes an ADK Go SDK v2 migration

The branch was 131 commits behind and conflicting. Rebased onto `main` (linear history, the `Merge remote-tracking branch 'upstream/main'` commit is gone), so the PR is **mergeable** again. Four commits, all signed off.

The conflict was not just textual — `main` has moved the Go ADK SDK to `google.golang.org/adk/v2`, so this feature's code had to be migrated. Worth a look from a reviewer:

- **ADK v1 → v2 imports** across every file this PR adds or touches (`a2a/artifacts.go`, `tools/save_artifact_tool.go`, `runner/adapter.go`, `agent/agent.go` + tests). `artifact` and `tool/loadartifactstool` both exist in v2 with unchanged shapes.
- **`agent.ToolContext` → `agent.Context`** in `save_artifact_tool.go`: v2's `functiontool.Func` handler takes `agent.Context`, matching the existing `save_memory` / `ask_user` tools.
- **`CreateRunnerConfig`** gained `kagentURL` / `httpClient` params on `main`; the new `adapter_test.go` call site was updated.
- **`ChatInterface.tsx`** — `main` refactored `handleSendMessage` into a parameterised `sendChatMessageText(text, options)` and moved the cross-tab send guard into `@/lib/chatSessionGuard`. File upload now rides along as `options.fileParts` instead of reading `pendingFiles` directly, so the MCP-app send path is unaffected. Dropped this PR's local `RESUBSCRIBE_TASK_STATES` / `ACTIVE_TASK_STATES` in favour of the shared guard module, and kept `main`'s `data-testid` hooks and read-only-share composer.
- **`values.yaml`** — `main` added `ui.streamTimeoutSeconds` and `ui.nginx.proxyReadTimeout` / `proxySendTimeout`. Kept all three and merged this PR's `ui.nginx.clientMaxBodySize` into the same `nginx` block. (Note: these are the knobs I removed earlier in this PR on review feedback; they came back via `main`, where they are wired up.)
- **`executor_test.go`** was an add/add conflict — `main`'s `newAgentMessage` / `newAgentStatusEvent` tests and this PR's upload tests are both kept.
- **`go.mod`** — kept `main`'s newer `testcontainers` v0.43.0 and `golang.org/x/{mod,net}`, added `tsawler/tabula` + `golang.org/x/image`, then re-tidied (no stray `adk` v1 entry).

Verified locally: `go build ./...`, `go vet ./...` and `go test ./adk/...` clean; UI jest suites for chat / file upload / message handlers pass (75 tests); eslint clean on the changed UI files.

Review comments are answered inline above.

go/core/test/e2e/mocks/invoke_file_upload_agent.json
dimetron: Added a focused UI reload regression test in ab0d6510: user-uploaded file parts from task history must survive session reload. The original upload E2E still matches the ADK save-as-artifact placeholder, so it exercises ingest rather than the typed prompt.

dimetron: Added focused UI reload coverage in ab0d6510: user-uploaded file parts from task history must survive session reload. The upload E2E still matches the ADK save-as-artifact placeholder, so it exercises ingest rather than the typed prompt.

dimetron:
shell we rework this on top of new UI @supreme-gg-gg ??

supreme-gg-gg:
> shell we rework this on top of new UI @supreme-gg-gg ??

For sure! Perhaps you can open another PR since this has diverged quite a lot from main now? I'll take at look once you migrated the changes

Charlesthebird:
Adding this to the new UI in https://github.com/kagent-dev/kagent/pull/2936

PR #2084, opened 2026-06-24
Add database upgrade and rollback compatibility tests

### LINKED ISSUE #1637: Add upgrade compatibility tests for database migrations

iplay88keys:
## Description

Closes: #1637

Adds automated coverage that a Helm upgrade from a prior release to the current build is safe: migrations apply cleanly, existing data survives, the resulting schema matches a fresh install, and the migrations reverse cleanly. Also covers the rolling-deploy window where prior-release pods serve against an already-migrated schema, and — at each point in the upgrade/rollback cycle — drives a real agent invocation through the controller so kagent's own query paths are exercised, not just raw SQL.

## Tests

- **Upgrade round-trip** (`TestUpgrade`, gated by `RUN_UPGRADE_TESTS`): installs the prior release, seeds representative data, upgrades to the current build, then asserts the controller rolls out without crashing, migrations reach the target version cleanly, the seeded data survives, the upgraded schema is identical to a clean install, and reversing the migrations restores the prior schema with data intact. It then redeploys the previous release over the migrated schema to confirm the old binary boots and serves rather than crash-looping (ahead-schema tolerance).
- **Rolling compatibility** (`TestRollingUpgradeCompatibility`, gated by `RUN_ROLLING_UPGRADE_TESTS`): runs the upgrade with two controller replicas and verifies a prior-release pod can still read and write while the new image rolls out. It runs regardless of whether migrations changed — the image swap alone can break a rolling deploy — and additionally exercises the old-code/new-schema window when the build adds migrations.
- **Version-matched invocation**: at each state — post-upgrade, old-code-against-the-new-schema, and post-rollback — a representative invoke e2e slice (`TestE2EInvokeInlineAgent`) runs from the tree whose version matches the serving controller (the previous-release slices from a git worktree at that release's tag), so real deploy/invoke/read paths are exercised end to end. Self-skips when the harness (mock-LLM host / previous-release worktree) isn't set up, so a bare `go test` still runs the DB round-trip.

## Targets

Upgrade-from targets are chosen per branch, mirroring how releases ship (a CI matrix; a leg is skipped when it doesn't apply):

- on `main` (building the next minor): the previous minor line's latest patch.
- on a release branch (building a patch): the same line's last patch, and the previous minor's latest.

The resolved upgrade-from version is printed before the build. Override locally with `UPGRADE_FROM_VERSION`.

## Running locally

Requires a kind cluster, then:

```
make create-kind-cluster
make run-upgrade-tests
```

or

```
make create-kind-cluster
make run-rolling-upgrade-tests
```

A run leaves the database migrated, so recreate the cluster between runs.

## Notes

- The previous install pins the bundled Postgres image to whatever the upgrade-from release's own install target shipped, so the baseline matches how that release actually runs rather than a hardcoded guess.
- Query-level backward compatibility (a prior release's queries against the new schema) is covered separately by a static check in #2085.


go/core/test/upgrade/roundtrip_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	db, err := sql.Open("pgx", url)
+	require.NoError(t, err, "open db for %s track", track.name)
+
+	driver, err := migratepgx.WithInstance(db, &migratepgx.Config{MigrationsTable: track.trackingTable})
+	require.NoError(t, err, "build migrate driver for %s track", track.name)
+
+	m, err := migrate.NewWithInstance("iofs", src, "pgx", driver)
+	require.NoError(t, err, "build migrator for %s track", track.name)
+	defer m.Close()
```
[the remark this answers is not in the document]
iplay88keys: Commented by Claude on-behalf-of @iplay88keys

This looks like a false positive — the connection is closed. `migrateTrackTo` has `defer m.Close()`, and in golang-migrate v4.19.1 `migrate.Migrate.Close()` calls the database driver's `Close()`. For the pgx/v5 driver (`database/pgx/v5/pgx.go:211`) that closes both `p.conn` and `p.db`, where `p.db` is exactly the `*sql.DB` passed into `WithInstance(db, …)`. So the handle from `sql.Open` is released via the deferred `m.Close()` — this is the documented `WithInstance` ownership behavior (the migrator takes ownership and closes the db it was given).

The only edge case is if `WithInstance`/`NewWithInstance` themselves returned an error before the `defer` registered, but that path `FailNow`s the test anyway, so there's no connection/FD accumulation across repeated runs.

go/core/test/e2e/upgrade/rolling_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+			if r.err != nil {
+				helmErr, helmOut = r.err, r.out
+				return true
+			}
+			// Helm finished successfully before we caught the window; put the
+			// result back so the final read below still sees it, then keep polling.
+			done <- r
+		default:
+		}
+		state := pgMigrationState(t, env)
```
mesutoezdil: testify runs the condition in a goroutine. pgmigrationstate calls require.nooerror inside it, so a kubectl failure calls t.failnow() there, silently drops the poll, and the test waits 6 minutes with a wrong error message. use assert or return false on error inside condition functions.

go/core/test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,529 @@
...
+		"-n", env.namespace,
+		fmt.Sprintf("--timeout=%s", timeout),
+	)
+}
+
+func waitForPostgresAgentTable(t *testing.T, env upgradeEnv, timeout time.Duration) {
+	t.Helper()
+
+	require.Eventually(t, func() bool {
+		return pgQuery(t, env, "SELECT to_regclass('public.agent') IS NOT NULL") == "t"
```
mesutoezdil: pgquery calls require.noerror inside the condition goroutine, same issue. a kubectl failure silently drops the poll. use assert or return false on error.

go/core/test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,529 @@
...
+		"OPENAI_API_KEY="+env.openAIAPIKey,
+		"KAGENT_DEFAULT_MODEL_PROVIDER=openAI",
+	)
+	return cmd
+}
+
+func loadUpgradeEnv(t *testing.T) upgradeEnv {
+	t.Helper()
+
+	_, filename, _, ok := runtime.Caller(0)
```
mesutoezdil: if this file moves to a different directory depth, filepath.abs gives the wrong path with no error, and make calls in the test fail silently. a repo_root env var would be safer.

Makefile
```diff
@@ -460,6 +460,97 @@ helm-uninstall: ## Uninstall kagent and kagent-crds Helm releases from the kind
...
+# The bundled-Postgres image is selected by the install target's --set flags, not
+# by the chart defaults (the chart ships a non-vector image). So the previous
+# install must use the exact pins the upgrade-from release shipped — otherwise the
+# baseline DB would differ from how that release actually runs, and the upgrade
+# would conflate a DB swap with the migration change under test. Read those flags
+# straight from that release's own helm-install-provider target (via its tagged
+# Makefile) instead of hardcoding values that drift as the bundled image changes.
+# Assumes the flags are literal (no make/env variables); the guard in
+# install-previous-release fails loudly if they can't be read.
+PREV_DB_SET_FLAGS = $(shell git show v$(UPGRADE_FROM_VERSION):Makefile 2>/dev/null | \
```
mesutoezdil: if a future --set value uses a make variable like $(postgres_tag), grep captures the literal string and passes it to helm. the empty-string guard on line 497 does not catch this.

.github/actions/upgrade-test-setup/action.yaml
```diff
@@ -0,0 +1,94 @@
...
+inputs:
+  upgrade-from:
+    description: 'Which release to upgrade from: "adjacent" or "prev-stable".'
+    required: true
+  buildx-builder-name:
+    description: Name for the docker/setup-buildx-action builder.
+    required: true
+  buildx-version:
+    description: Buildx version for docker/setup-buildx-action.
+    required: true
```
EItanya: Why are these required inputs? Can't they be static?

Makefile
```diff
@@ -494,6 +494,100 @@ helm-uninstall: ## Uninstall kagent and kagent-crds Helm releases from the kind
...
+# install) -> reverse schema to target (down files) + data survival.
+# Prerequisite (provided by CI as a separate step; run it locally first): a kind
+# cluster (make create-kind-cluster). The controller tolerates the missing
+# agent-sandbox CRD (the owned-resource watch is skipped), and these tests create
+# no SandboxAgents, so agent-sandbox is not required.
+.PHONY: run-upgrade-tests
+run-upgrade-tests: ## Install the previous release, build current images, and run the upgrade test (migration round-trip)
+	test -n "$(UPGRADE_FROM_VERSION)" || { echo "UPGRADE_FROM_VERSION is empty; set it explicitly or ensure git tags are fetched." >&2; exit 1; }
+	$(MAKE) build
+	$(MAKE) install-previous-release
```
EItanya: Why are these not dependency targets? calling make inside of make seems weird

go/core/test/upgrade/rolling_upgrade_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRollingUpgradeCompatibility(t *testing.T) {
+	if os.Getenv("RUN_ROLLING_UPGRADE_TESTS") != "true" {
+		t.Skip("set RUN_ROLLING_UPGRADE_TESTS=true to run rolling upgrade tests")
+	}
```
EItanya, line 17: This feels weird, can't we have this in a separate package or something?

Makefile
```diff
@@ -494,6 +494,100 @@ helm-uninstall: ## Uninstall kagent and kagent-crds Helm releases from the kind
...
+	VERSION=$(VERSION) \
+	DOCKER_REGISTRY=$(DOCKER_REGISTRY) \
+	KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) \
+	OPENAI_API_KEY="$${OPENAI_API_KEY:-test}" \
+	go test ./core/test/e2e/upgrade -run TestUpgrade -count=1 -timeout=20m -v
+
+.PHONY: run-rolling-upgrade-tests
+run-rolling-upgrade-tests: ## Install the previous release with 2 controller replicas, build the current images, and run the rolling upgrade e2e test
+	$(MAKE) build
+	$(MAKE) install-previous-release UPGRADE_PREV_EXTRA_ARGS="--set controller.replicas=2"
```
EItanya: Same comment about calling `make`

Makefile
```diff
@@ -494,6 +494,100 @@ helm-uninstall: ## Uninstall kagent and kagent-crds Helm releases from the kind
...
+# straight from that release's own helm-install-provider target (via its tagged
+# Makefile) instead of hardcoding values that drift as the bundled image changes.
+# Assumes the flags are literal (no make/env variables); the guard in
+# install-previous-release fails loudly if they can't be read.
+PREV_DB_SET_FLAGS = $(shell git show v$(UPGRADE_FROM_VERSION):Makefile 2>/dev/null | \
+	grep -oE '\-\-set[[:space:]]+database\.postgres\.[^[:space:]\\]+')
+
+.PHONY: install-previous-release
+install-previous-release: ## Install the previous released kagent + kagent-crds charts from the public OCI registry
+	test -n "$(UPGRADE_FROM_VERSION)" || { echo "UPGRADE_FROM_VERSION is empty; set it explicitly or ensure git tags are fetched." >&2; exit 1; }
```
EItanya: What is `test`?

Makefile
```diff
@@ -494,6 +494,100 @@ helm-uninstall: ## Uninstall kagent and kagent-crds Helm releases from the kind
...
+# The bundled-Postgres image is selected by the install target's --set flags, not
+# by the chart defaults (the chart ships a non-vector image). So the previous
+# install must use the exact pins the upgrade-from release shipped — otherwise the
+# baseline DB would differ from how that release actually runs, and the upgrade
+# would conflate a DB swap with the migration change under test. Read those flags
+# straight from that release's own helm-install-provider target (via its tagged
+# Makefile) instead of hardcoding values that drift as the bundled image changes.
+# Assumes the flags are literal (no make/env variables); the guard in
+# install-previous-release fails loudly if they can't be read.
+PREV_DB_SET_FLAGS = $(shell git show v$(UPGRADE_FROM_VERSION):Makefile 2>/dev/null | \
```
EItanya: This means that these will be called every time `make` is used. Should we instead add this specifically to the targets that need it?

EItanya (commented):
My main comment about these tests is that they sidestep the actual `kagent` DB usage. They interact directly with the DB rather than testing how the system behaves. For example they pre-seed the DB with data by directly inserting it rather than performing kagent operations. How much more work would it be to access the data that way?

Makefile
iplay88keys: Great question, I completely agree.

go/core/test/upgrade/rolling_upgrade_test.go
iplay88keys: The concern here is that these upgrade/downgrade Kagent, which breaks the other tests. I mostly didn't want them running on a `go test ./...` or similar call without explicitly invoking them. I am looking into alternative approaches.

Makefile
iplay88keys: `test -n` is the POSIX shell builtin. Switching to `[ -n "$(UPGRADE_FROM_VERSION)" ]` instead to match the rest of the Makefile.

Makefile
iplay88keys: `UPGRADE_FROM_VERSION ?=` is a recursive assignment, so it stays the literal string `$(shell ./scripts/upgrade-from-version.sh)` until something expands it ([docs](https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html)). It doesn't run on every `make`, only when an upgrade target references it. I moved the other one into the recipe.

I kept this one as a variable for a single source of truth and so CI can override it and have it flow through to the targets (including the `install-previous-release` prerequisite).

go/core/test/upgrade/rolling_upgrade_test.go
iplay88keys: I am moving the tests to `go/core/test/upgrade` instead of being within `e2e` so it's a bit more clear that these tests are not run against the same cluster that the `e2e` tests are run against. I think that the env var is the cleanest approach to skipping these specific tests, though I am open to other options.

iplay88keys:
>My main comment about these tests is that they sidestep the actual kagent DB usage. They interact directly with the DB rather than testing how the system behaves. For example they pre-seed the DB with data by directly inserting it rather than performing kagent operations. How much more work would it be to access the data that way?

@EItanya I agree these tests sidestep kagent's real query paths. The scope I'm going for in this PR is migration *correctness*: that the migration files preserve schema and data forward and back. Using Postgres directly lets us seed every table and diff the upgraded schema against a clean install, which there's no API for.

For testing actual system behavior across an upgrade, I've been thinking through our options, but I haven't come up with a great solution. Building out a new test suite which exercises Kagent to test it's behavior and keeping it up-to-date with the old and new versions seems unnecessary when the version-matched e2e suite already is that code, written against each version's API and CRDs. However, running the entire e2e test suite at each state (post-upgrade, old-code-against-the-new-schema, post-rollback) is slow.

The static side of this (that an older release's queries still compile against the new schema) is covered by #2085. So I'd land this PR as the migration-correctness layer and consider doing the version-matched e2e work as a follow-up.

So the question is: do we want to take on the extra cost of of running the entire e2e test suite multiple times for safety right now and optimize it later or start with the subset I have in this PR and expand it over time?

iplay88keys:
I have updated this to run the `TestE2EInvokeInlineAgent` E2E test after each step of the upgrade/rollback cycle.

EItanya (approved):

PR #2085, opened 2026-06-24
Add query-contraction check against the previous release

iplay88keys:
## Description

Adds a static check that a migration on this branch doesn't break the queries a prior release still runs — the backward-compatibility property that makes a one-step rollback (and a rolling deploy) safe.

It compiles a prior release's sqlc queries against the current schema (derived from the migration files) and fails if a migration dropped, renamed, or retyped a column or table an older query still references. Runs in normal CI on every PR — no database, no cluster.

## Targets

Checks two prior versions (deduplicated when they coincide):

- the latest release reachable from HEAD — the adjacent previous release, and
- the previous stable line's latest patch (the `release/vX.Y.x` tip) — the rollback-window floor.

## Running

```
make -C go check-query-contraction
```

Override the targets with `TARGET_VERSIONS` (space-separated, no leading `v`).

## Notes

- Verified it passes today and fails on a simulated column drop.
- Catches column/table/type-shape contractions; it does not catch data rewrites (no DDL to inspect) or semantic changes a query still compiles against.
- Complements the contraction-declaration check in #2067: that one requires destructive DDL to be acknowledged; this one verifies a prior release's queries actually survive the new schema.
- Also documents the existing `sqlc generate` gate (schema derived from migrations) as the forward-direction equivalent for current-release queries.
- This branch carries its own copies of the helper scripts, which are shared with https://github.com/kagent-dev/kagent/pull/2084, so that will require a trivial merge conflict resolution.


go/Makefile
```diff
@@ -33,6 +33,14 @@ generate: controller-gen sqlc-generate ## Generate DeepCopy methods and sqlc que
 sqlc-generate: sqlc ## Generate type-safe Go code from SQL queries.
 	cd core/internal/database && $(SQLC) generate
 
+# Compile the previous release's sqlc queries against the current schema (the
+# migration files) and fail if a migration removed/renamed/retyped a column or
+# table an old query still uses. Static (no database/cluster). Needs git tags
+# fetched. TARGET_VERSIONS (space-separated) overrides the auto-derived release(s).
+.PHONY: check-query-contraction
+check-query-contraction: sqlc ## Verify the previous release's queries still type-check against the current schema
+	SQLC=$(SQLC) ../scripts/check-query-contraction.sh
```
EItanya, line 42: nit: should we just put this script in `go/scripts`?

.github/workflows/ci.yaml
```diff
@@ -169,7 +188,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: "1.26"
+          go-version-file: go/go.mod
```
EItanya, line 191: good idea

scripts/prev-stable-version.sh
EItanya, line 1: Can this be used by both this PR and the upgrade tests?

EItanya (approved):
Just the one nit about script placement since it's only used there

scripts/prev-stable-version.sh
iplay88keys: Yes, it is shared and will be removed from the other PR!

go/Makefile
iplay88keys: That's a good idea, I can move it on the [other PR](https://github.com/kagent-dev/kagent/pull/2084).

iplay88keys: Actually, I'm going to leave it there for now. I had misread which script it was and was thinking it was the shared script between these two PRs. I think it makes sense to leave the `check-query-contraction.sh` script for now and move it if we decide to do a broader move of the other scripts there such as `controller-digest-ldflags.sh`.

PR #2090, opened 2026-06-25
fix(task-store): retry on transient transport errors instead of dropping prompt

### LINKED ISSUE #2086: BYO agent silently drops an incoming prompt when the agent→controller /api/tasks call fails (transient transport error)
## Summary

A BYO agent silently drops an incoming prompt when the agent→controller HTTP call to the task store (`POST/GET /api/tasks`) hits a transient transport error. The user's message is accepted (the UI shows it as sent) but the agent never runs — no LLM/tool/A2A activity, no task or events created in the controller, no reply, and **no error surfaced anywhere the user can see**. The agent only recovers after a pod restart.

## Affected versions / environment

- `kagent-langgraph` **0.9.6** (reproduced here)
- `a2a-sdk` 0.3.23
- Packaged under Solo Enterprise for kagent 0.4.3
- BYO agent (`type: BYO`), `KAGENT_URL` → `kagent-controller`
- Istio **ambient mesh (HBONE)** on the agent→controller hop; EKS

## Symptom

1. Send a prompt to a BYO agent (e.g. from the kagent UI). The UI shows the message as sent.
2. The agent does nothing: no graph execution, no downstream LLM/MCP/A2A calls, no task or events recorded in the controller, no response. The session just looks stuck.
3. The agent container logs show the incoming request being received, but **no subsequent work** — the run is abandoned before the agent graph executes.
4. Restarting the agent pod restores normal behaviour (until the condition recurs).

## Where it occurs

On every incoming prompt, before the agent graph runs, the A2A runtime calls the controller-backed task store (`KAgentTaskStore`): `GET /api/tasks/{id}` followed by `POST /api/tasks`. When that hop raises a transient `httpx.TransportError` — e.g. an idle keep-alive connection reset by the mesh, or the controller pod being rescheduled — the error propagates out of the incoming-request handling and the prompt is lost.

## Scope — not framework-specific

Reproduced with `kagent-langgraph` 0.9.6. The same construct — a controller `httpx.AsyncClient` built in the adapter's `_a2a.py` and wrapped in the shared `KAgentTaskStore`, invoked on every incoming prompt — is also present in the `kagent-adk` adapter (`python/.../kagent/adk/_a2a.py`). So this affects BYO agents using the controller-backed task store **regardless of framework adapter**; the common component is `KAgentTaskStore` / the controller client.

## Impact / severity

Silent loss of a user request on the primary interaction path, with no error surfaced to the user and recovery only via a pod restart. In a service-mesh deployment the triggering condition (idle connection reset / controller reschedule) is routine, so it recurs. In our environment it reproduces **reliably** — the first prompt after the agent→controller connection has been idle is dropped every time.

yashrajshuklaaa:
When the agent  - - > controller HTTP hop raises   **httpx.TransportError**   ( idle keep-alive connection reset by Istio/HBONE mesh , controller pod reschedule , etc ) the error previously propagated uncaught out of   **KAgentTaskStore.get/save**   silently dropping the user prompt with no error surfaced and no recovery short of a pod restart
### Fix : 
introduce   **_request_with_retry( )**    in   **KAgentTaskStore**   that catches **TransportError** calls **aclose( )** to flush the stale connection pool and retries once on a fresh connection. Non-transport HTTP errors (4xx/5xx) are re-raised immediately without retrying. If the transport error persists after all retries it is re-raised so the caller sees a real error rather than a silent drop
fix lives entirely in **kagent-core/_task_store.py** and covers all
four framework adapters (langgraph, adk, openai, crewai) automatically
since they all share **KAgentTaskStore**

Fixes #2086

### DIFF
--- a/python/packages/kagent-core/src/kagent/core/a2a/_task_store.py
+++ b/python/packages/kagent-core/src/kagent/core/a2a/_task_store.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 
 import httpx
 from a2a.server.tasks import TaskStore
@@ -7,6 +8,8 @@ from pydantic import BaseModel
 from typing_extensions import override
 
 from kagent.core.a2a import read_metadata_value
+
+logger = logging.getLogger(__name__)
 
 
 class KAgentTaskResponse(BaseModel):
@@ -25,7 +28,16 @@ class KAgentTaskResponse(BaseModel):
 class KAgentTaskStore(TaskStore):
     """
     A task store that persists A2A tasks to KAgent via REST API.
+
+    Transient transport errors (idle keep-alive connections reset by a service
+    mesh, controller pod restarts, etc.) are handled transparently: each HTTP
+    operation is retried once after closing and re-opening the underlying
+    connection.  Non-transport HTTP errors (4xx/5xx) are surfaced immediately
+    without retrying so real failures are never swallowed.
     """
+
+    # Maximum number of automatic retries for transient transport errors.
+    _MAX_RETRIES = 1
 
     def __init__(self, client: httpx.AsyncClient):
         """Initialize the task store.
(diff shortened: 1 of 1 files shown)


yashrajshuklaaa:
@EItanya PTAL! Merging this will fix the issue where prompts are dropped silently due to transient **httpx.TransportErrors**
I've added **_request_with_retry( )** to clear the connection pool and retry once on fresh connections. All 23 checks are green.

yashrajshuklaaa:
@EItanya @supreme-gg-gg If this looks relevant, pls remove the stale label and review this

Issue #2072, opened 2026-06-22
[FEATURE] Add Azure/Microsoft Foundry model backend support


marosset:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add Azure Foundry model provider/backend support

### ❓ Problem Statement / Motivation

Foundry provides a unified platform for deploying and consuming multiple model families on Azure. kagent users should be able to configure agents against foundry deployments directtly, using Azure-native endpoints and identity.

https://azure.microsoft.com/en-us/products/ai-foundry/models/

Supporting this feature allows users to use their existing Azure AI infrastructure, governance, and identity setup

### 💡 Proposed Solution

Add a dedicated `AzureFoundry` provider with:

- A new provider-specific `AzureFoundryConfig` in `ModelConfig`
- Auth modes for:
  - static API key
  - Azure Workload Identity
  - API key passthrough
- Provider-assisted Azure Workload Identity wiring when kagent owns the generated ServiceAccount:
  - add `azure.workload.identity/use: "true"` pod label
  - add `azure.workload.identity/client-id` ServiceAccount annotation
  - optionally add `azure.workload.identity/tenant-id`
- Standard Kubernetes `ConfigMap`/`Secret` refs for endpoint/client ID/tenant ID values, so ASO (more details below) can export required values without kagent taking a hard dependency on ASO CRDs
- Foundry-specific runtime configuration names rather than reusing Azure OpenAI env vars
- Support for existing `spec.tls` behavior

### 🔄 Alternatives Considered

Reuse or augment the AzureOpenAI provider. Foundry should be a first-class backend with its own config for a better user experience

### 🎯 Affected Service(s)

Multiple services / System-wide

### 📚 Additional Context

Azure Service Operator (ASO) (https://github.com/Azure/azure-service-operator) is a Kubernetes operator that manages Azure resources using Kubernetes custom resources. For this feature, ASO is relevant because users may provision Azure Foundry-related Azure resources from Kubernetes, such as Cognitive Services accounts, Foundry projects, model deployments, user-assigned managed identities, federated identity credentials, and role assignments.

[...]

PR #2091, opened 2026-06-25
Azure foundry backend

marosset:
## Summary

Adds first-class Azure AI Foundry model provider support to kagent for both chat
completions and long-term memory embeddings. The provider is implemented in the Go
ADK and gated to the Go declarative runtime.

Authentication is **implicit** (no auth-type selector, mirroring `GeminiVertexAI`):
if an API key Secret is set it is used, otherwise the Go runtime authenticates with
Azure Workload Identity via `DefaultAzureCredential`, acquiring Cognitive Services
(Foundry) data-plane tokens through the Azure SDK. Workload Identity is wired
through the generic agent deployment fields (pod label + ServiceAccount
annotations) rather than Foundry-specific CRD fields.

## What Changed

- Added `Foundry` as a `v1alpha2` `ModelConfig` provider and ADK runtime model
- Added `FoundryConfig` (`endpoint`, `endpointFrom`, `deployment`, `apiVersion`)
  and the corresponding pod env vars
- Implicit auth: API key (`FOUNDRY_API_KEY`) or Azure Workload Identity via
  `DefaultAzureCredential`, scoped to `https://cognitiveservices.azure.com/.default`
- Provider readiness fails fast when no credential resolves
- Added `endpointFrom` to resolve the endpoint from a `ConfigMap` (Azure Service
  Operator integration)
- Added Foundry memory embedding provider (768-dim), Go runtime only
- Restricted Foundry (chat, summarizer, and memory embedding) to the Go runtime;
  the controller rejects other runtimes
- Added Foundry create/edit support and a curated chat model catalog to the UI
- Added focused unit tests and a mock-based memory-embedding e2e test
- Added user docs (`docs/azure-foundry.md`) and the CRD reference entry

Note: Claude (Anthropic) models on Foundry are not yet supported — planned as a
follow-up.

## Validation

- Verified Foundry Workload Identity end-to-end against a live AKS / Foundry
  deployment (`disableLocalAuth: true`, no keys): agent chat and memory save/load
  both working
- Verified Foundry API-key auth against a Foundry resource
- Verified web UI create/edit behaviors

Fixes #https://github.com/kagent-dev/kagent/issues/2072


go/adk/pkg/models/foundry.go
```diff
@@ -0,0 +1,139 @@
...
+	Endpoint   string
+	Deployment string
+	APIVersion string
+	AuthType   string
+}
+
+const (
+	foundryAuthTypeAPIKey            = "APIKey"
+	foundryAuthTypeWorkloadIdentity  = "WorkloadIdentity"
+	foundryAuthTypeAPIKeyPassthrough = "APIKeyPassthrough"
```
EItanya: Why are these defined here and in `go/api/adk/types.go`?

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -310,6 +312,93 @@ type SAPAICoreConfig struct {
...
 }
 
+// FoundryAuthType identifies the authentication mode for Foundry.
+// +kubebuilder:validation:Enum=APIKey;WorkloadIdentity;APIKeyPassthrough
+type FoundryAuthType string
+
+const (
+	FoundryAuthTypeAPIKey            FoundryAuthType = "APIKey"
+	FoundryAuthTypeWorkloadIdentity  FoundryAuthType = "WorkloadIdentity"
+	FoundryAuthTypeAPIKeyPassthrough FoundryAuthType = "APIKeyPassthrough"
```
EItanya: A third place this is defined...

EItanya:
Hey there, I'm really sorry for not reviewing this earlier. I really thought I had left a review but it's still in pending 🙈, I will review this tomorrow AM. Thanks for the patience!

docs/azure-foundry.md
EItanya, line 1: Can you move this doc under examples? If you want a doc on the website you'll have to add it to https://github.com/kagent-dev/website/

go/adk/pkg/auth/token.go
```diff
@@ -70,10 +70,17 @@ func (s *KAgentTokenService) GetToken() string {
...
+// The kagent token is only a DEFAULT credential for calls to the kagent API. If
+// the caller already set an Authorization header (for example the memory
+// embedding provider attaching a provider-specific API key or an Azure/Foundry
+// Workload Identity bearer token for an external endpoint), it is preserved —
+// otherwise the shared HTTP client would clobber that credential with the
+// kagent token and the external service would reject it.
 func (s *KAgentTokenService) AddHeaders(req *http.Request) {
 	req.Header.Set("X-Agent-Name", s.appName)
-	if token := s.GetToken(); token != "" {
+	if token := s.GetToken(); token != "" && req.Header.Get("Authorization") == "" {
```
EItanya: This essentially just ignores the `KagentTokenService` functionality altogether. I don't like how magical this is, should we just not add this token service when using `Foundry`?

go/adk/pkg/embedding/embedding.go
```diff
@@ -368,3 +375,155 @@ func normalizeL2(vec []float32) []float32 {
...
+const foundryCognitiveServicesScope = "https://cognitiveservices.azure.com/.default"
+
+type foundryTokenCredential interface {
+	GetToken(context.Context, policy.TokenRequestOptions) (azcore.AccessToken, error)
+}
+
+// foundryCredentialFactory constructs the Azure credential for the implicit
+// Workload Identity auth path. It is a package variable so tests can inject a
+// fake credential.
+var foundryCredentialFactory = func() (foundryTokenCredential, error) {
```
EItanya: Can we pass this in rather than having a global `var func`

go/adk/pkg/embedding/embedding.go
```diff
@@ -368,3 +375,155 @@ func normalizeL2(vec []float32) []float32 {
...
+	if deployment == "" {
+		return nil, fmt.Errorf("deployment is required for Foundry embeddings")
+	}
+	apiVersion := p.config.APIVersion
+	if apiVersion == "" {
+		apiVersion = os.Getenv("FOUNDRY_API_VERSION")
+	}
+	if apiVersion == "" {
+		apiVersion = "2024-10-21"
+	}
```
EItanya: The way that these waterfall conditionals work makes sense, but look kinda weird, is there a cleaner way we can do this?

go/adk/pkg/embedding/embedding.go
```diff
@@ -368,3 +375,155 @@ func normalizeL2(vec []float32) []float32 {
...
+
+	requestURL, err := foundryEmbeddingsURL(endpoint, deployment, apiVersion)
+	if err != nil {
+		return nil, err
+	}
+	reqBody := map[string]any{
+		"input":      texts,
+		"dimensions": TargetDimension,
+	}
```
EItanya: Can we add an internal struct for this?

go/adk/pkg/embedding/embedding.go
```diff
@@ -368,3 +375,155 @@ func normalizeL2(vec []float32) []float32 {
...
+// OpenAI-compatible data plane. Authentication is implicit and mirrors the
+// Foundry chat model: FOUNDRY_API_KEY is used when set, otherwise the provider
+// authenticates with DefaultAzureCredential (Azure Workload Identity in-cluster).
+type foundryProvider struct {
+	config     *adk.EmbeddingConfig
+	httpClient *http.Client
+	credential foundryTokenCredential
+}
+
+func (p *foundryProvider) generate(ctx context.Context, texts []string) ([][]float32, error) {
```
EItanya: It feels a bit weird that this function does so much at generate time, rather than construction time. Can we add more of this to a one time op?

go/adk/pkg/embedding/embedding.go
EItanya: In fact you have one defined below

go/adk/pkg/models/foundry.go
```diff
@@ -0,0 +1,152 @@
...
+//   - https://learn.microsoft.com/en-us/azure/ai-foundry/model-inference/concepts/endpoints
+//   - https://learn.microsoft.com/en-us/azure/ai-foundry/foundry-models/concepts/models-sold-directly-by-azure
+//
+// Authentication is implicit: if FOUNDRY_API_KEY is set it is used as the
+// data-plane API key; otherwise the model authenticates with
+// DefaultAzureCredential, which resolves to Azure Workload Identity in-cluster
+// (or the az CLI in local development).
+func NewFoundryModelWithLogger(ctx context.Context, config *FoundryConfig, logger logr.Logger) (*OpenAIModel, error) {
+	endpoint := config.Endpoint
+	if endpoint == "" {
```
EItanya: You are using these env vars a lot, can you define them as consts or getters somewhere for these functions to access? Also, is this the same http client as the embeddings which can have a shared constructor?

go/adk/pkg/models/foundry.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/go-logr/logr"
+	"github.com/openai/openai-go/v3"
+	"github.com/openai/openai-go/v3/option"
+)
+
+// foundryCognitiveServicesScope is the Azure data-plane scope used to request
+// tokens for the Foundry / Azure AI Services OpenAI-compatible API.
+const foundryCognitiveServicesScope = "https://cognitiveservices.azure.com/.default"
```
EItanya: defined twice

go/adk/pkg/models/foundry.go
```diff
@@ -0,0 +1,152 @@
...
+type foundryTokenCredential interface {
+	GetToken(context.Context, policy.TokenRequestOptions) (azcore.AccessToken, error)
+}
+
+// foundryCredentialFactory constructs the Azure credential used for the implicit
+// Workload Identity auth path. It is a package variable so tests can inject a
+// fake credential.
+var foundryCredentialFactory = func() (foundryTokenCredential, error) {
+	return azidentity.NewDefaultAzureCredential(nil)
+}
```
EItanya: This is all the same as embeddings

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -869,6 +894,68 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
...
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: model.Spec.APIKeySecret,
+						},
+						Key: model.Spec.APIKeySecretKey,
+					},
+				},
+			})
+		}
+
+		if endpoint != "" {
```
EItanya: This will always be set when `resolveFoundryEndpoint` returns

go/core/internal/controller/translator/agent/adk_api_translator.go
```diff
@@ -869,6 +894,68 @@ func (a *adkApiTranslator) translateModel(ctx context.Context, namespace, modelC
...
+			modelDeploymentData.EnvVars = append(modelDeploymentData.EnvVars, corev1.EnvVar{
+				Name:  env.FoundryEndpoint.Name(),
+				Value: endpoint,
+			})
+		}
+		if cfg.Deployment != "" {
```
EItanya: This is required by the API

go/core/internal/controller/translator/agent/compiler.go
```diff
@@ -210,12 +210,24 @@ func (a *adkApiTranslator) validateAgent(ctx context.Context, agent v1alpha2.Age
 	return nil
 }
 
+// requireFoundryGoRuntime returns an error if a Foundry model is used with a
+// non-go declarative runtime. Foundry is only supported by the Go ADK runtime.
+func requireFoundryGoRuntime(agent v1alpha2.AgentObject, modelType string) error {
```
EItanya, line 224: Can we make this a `CEL` rule on the API?

marosset:
> Hey there, I'm really sorry for not reviewing this earlier. I really thought I had left a review but it's still in pending 🙈, I will review this tomorrow AM. Thanks for the patience!

Hi @EItanya  - no worries and that for the review. Between your initial review and submitting the comments, I reworked the PR to align how auth works to be closer to how the other backend providers work when APIKey isn't set. This does require that some extra labels / annotations be set for the service account and agent pod. I think these changes should address a lot of the inline comments.

Here is an overview of what I changed
- Simplified the auth model. Auth is now implicit and mirrors `GeminiVertexAI` - if an API Key is set it is used, otherwise the authenticates with `DefaultAzureCredential` (which would use workload identity in-cluster).
- Removed all the azure-specific fields from the CRDs (which should also resolve the copilot comments)
- Removed the foundry-sepcific token bypass. All provider auth now lives in the HTTP clients and there is no special behavior for Foundry. I did make small change so clients could specify its own auth header (which would previously get overwritten) which lets the embedding model http clients reuse it. I'm happy to make further updates here if desired.
- Scoped all these changes under the Go runtime (for now) 


Could you take another look at the changes (and usage in the `azure-foundry.md` doc) with a focus on the overall design? 
If the design looks good I will proceed with the code changes/cleanup and moving the docs.

Thanks in advance!

go/api/v1alpha2/modelconfig_types.go
marosset: these have been removed

go/adk/pkg/models/foundry.go
marosset: these have been removed

docs/azure-foundry.md
marosset: I removed it and will make an update to the website with the same info - thanks!

go/adk/pkg/models/foundry.go
marosset: I consolidated everything and moved the consts, token handleing, and bearer token management to a new azureai package. Both the models and embedding clients now use this.

go/adk/pkg/embedding/embedding.go
marosset: Done, the credential is passed into a newFoundryProvider(cfg, cred) method

go/adk/pkg/embedding/embedding.go
marosset: i cleaned this up and consolidated the logic in a `ResolveFoudry` method in the new/shared azureai pkg

go/adk/pkg/embedding/embedding.go
marosset: this has been removed in the new iteration

go/adk/pkg/embedding/embedding.go
marosset: updated!

go/adk/pkg/models/foundry.go
marosset: Consts are defined in the new azureai pkg and now the model and embeddings share the `azureai.NewOpenAIClient` constructor which builds a openai-go client for both

go/adk/pkg/models/foundry.go
marosset: de-duped

go/adk/pkg/auth/token.go
marosset: These changes are no longer needed. So i'm reverting them

go/core/internal/controller/translator/agent/compiler.go
marosset: I don't think I can make a CEL rule for this (the Agent to get the runtime and the ModelConfig) thoughts?

marosset:
@EItanya Could you take another look when you get a chance? I think the PR is in a much better state now, thanks

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -312,6 +314,48 @@ type SAPAICoreConfig struct {
...
+	// (e.g., https://my-account.cognitiveservices.azure.com/).
+	// Mutually exclusive with EndpointFrom.
+	// +optional
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// EndpointFrom resolves the Foundry endpoint from an external source, such
+	// as a ConfigMap written by Azure Service Operator. Mutually exclusive with
+	// Endpoint.
+	// +optional
+	EndpointFrom *FoundryEndpointSource `json:"endpointFrom,omitempty"`
```
EItanya: Why do we need nested fields here, can't this just be `EndpointFrom *corev1.ConfigMapKeySelector`

go/api/adk/types.go
```diff
@@ -387,20 +418,40 @@ type EmbeddingConfig struct {
...
+	// (account endpoint / deployment name / api-version). They are currently
+	// populated only for the "foundry" provider, whose auth is implicit
+	// (FOUNDRY_API_KEY, else DefaultAzureCredential).
+	//
+	// TODO: these are not Foundry-specific. The "azure_openai" provider uses the
+	// same triplet (v1alpha2.AzureOpenAIConfig azureEndpoint/azureDeployment/
+	// apiVersion) and should be migrated onto these fields. Today
+	// azureOpenAIProvider reads EmbeddingConfig.BaseUrl, which
+	// ModelToEmbeddingConfig never sets for AzureOpenAI, so azure_openai memory
+	// embeddings do not work on the Go runtime.
```
EItanya: How much work would this cleanup be? This does feel like a bit of a hack.

EItanya (changes requested):

go/api/adk/types.go
marosset: I was planning on doing this in a follow up PR but I can make include them in this PR

go/api/v1alpha2/modelconfig_types.go
marosset: That should work. Let me update

marosset: Updated!

go/api/adk/types.go
marosset: I unified auth for foundry and Azure OpenAI. 

I validated that after these changes both Azure OpenAI and Foundry models work for both chat and embedding for both APIkey and implicit (workload identity) auth with in Go and also that Azure OpenAI APIKey auth continues to work in Python.

@EItanya please take another look when you get a chance and thanks!

EItanya (commented):
> First-pass AI review using Codex (GPT-5). These findings should be validated by a maintainer.

I found a few potential issues:

1. **Credential leakage when `OPENAI_API_KEY` is present.** In the API-key branch, `NewOpenAIClient` adds `Api-Key` but does not override the OpenAI SDK's automatic `OPENAI_API_KEY` handling. An agent combining OpenAI with Foundry can therefore send the OpenAI key to the Azure endpoint as `Authorization: Bearer ...` alongside the Foundry `Api-Key`. I reproduced this by running the focused tests with `OPENAI_API_KEY=dummy`; both `TestNewOpenAIClientAPIKey` and `TestFoundryProviderAPIKey` observed the unexpected Authorization header. This shared client also affects Azure OpenAI.

2. **`endpointFrom` changes do not reconcile dependent agents.** The endpoint is read from the ConfigMap and copied into the generated agent config/environment, but the ConfigMap watch only detects references directly present in the Agent spec. It does not follow `Agent -> ModelConfig -> foundry.endpointFrom`. Updating an ASO-managed ConfigMap after the agent becomes ready therefore leaves the running agent on the old endpoint.

3. **The UI cannot safely edit `endpointFrom` configurations.** Edit mode flattens provider values with `String(value)`, turning the selector into `"[object Object]"`. The form then requires an inline `endpoint`, and submission rebuilds `spec.foundry` from the flattened parameters. A valid YAML-created ModelConfig using `endpointFrom` is either blocked from saving or loses the selector when an inline endpoint is entered.

4. **`apiKeyPassthrough` is accepted for Foundry but ignored.** CRD validation permits `apiKeyPassthrough: true`, but the Foundry translator does not copy it into `adk.Foundry.BaseModel`, and the Foundry runtime has no passthrough handling. With no static secret, it silently selects DefaultAzureCredential instead of forwarding the incoming token.

Minor documentation mismatch: the PR description says `docs/azure-foundry.md` was added, but that file is not present in the current diff.

marosset:
> > First-pass AI review using Codex (GPT-5). These findings should be validated by a maintainer.
> 
> I found a few potential issues:
> 
> 1. **Credential leakage when `OPENAI_API_KEY` is present.** In the API-key branch, `NewOpenAIClient` adds `Api-Key` but does not override the OpenAI SDK's automatic `OPENAI_API_KEY` handling. An agent combining OpenAI with Foundry can therefore send the OpenAI key to the Azure endpoint as `Authorization: Bearer ...` alongside the Foundry `Api-Key`. I reproduced this by running the focused tests with `OPENAI_API_KEY=dummy`; both `TestNewOpenAIClientAPIKey` and `TestFoundryProviderAPIKey` observed the unexpected Authorization header. This shared client also affects Azure OpenAI.
> 2. **`endpointFrom` changes do not reconcile dependent agents.** The endpoint is read from the ConfigMap and copied into the generated agent config/environment, but the ConfigMap watch only detects references directly present in the Agent spec. It does not follow `Agent -> ModelConfig -> foundry.endpointFrom`. Updating an ASO-managed ConfigMap after the agent becomes ready therefore leaves the running agent on the old endpoint.
> 3. **The UI cannot safely edit `endpointFrom` configurations.** Edit mode flattens provider values with `String(value)`, turning the selector into `"[object Object]"`. The form then requires an inline `endpoint`, and submission rebuilds `spec.foundry` from the flattened parameters. A valid YAML-created ModelConfig using `endpointFrom` is either blocked from saving or loses the selector when an inline endpoint is entered.
> 4. **`apiKeyPassthrough` is accepted for Foundry but ignored.** CRD validation permits `apiKeyPassthrough: true`, but the Foundry translator does not copy it into `adk.Foundry.BaseModel`, and the Foundry runtime has no passthrough handling. With no static secret, it silently selects DefaultAzureCredential instead of forwarding the incoming token.
> 
> Minor documentation mismatch: the PR description says `docs/azure-foundry.md` was added, but that file is not present in the current diff.

I addressed #1 and #4.
For #3 I had considered that but think that the use endpointFrom use case will be primarily though deployments done outside of the UI (with ASO + yaml based agents) so i think that should be OK as long as it gets documented in the followup PR to Website, what do you think?

marosset:
I also addressed #2, thanks for the reviews so far @EItanya

EItanya (approved):

PR #2106, opened 2026-06-29
feat(sts): send RFC 8707 resource and audience on token exchange

QuentinBisson:
## What

The Go ADK token-propagation plugin (`go/adk/pkg/sts`) always passed `nil`
resource and `nil` audience to the STS token exchange, so an issued token
could not be scoped to a specific backend. This wires two optional inputs
through to the exchange:

- `KAGENT_STS_RESOURCE`: RFC 8707 resource indicator (the target backend).
- `KAGENT_STS_AUDIENCE`: RFC 8693 audience, for STS servers that key on it.

Both are read at agent-runtime startup and passed as `resource`/`audience`
parameters to `NewTokenPropagationPlugin`. Empty values are omitted from the
request, so behaviour is unchanged when neither is set.

## Why

Without a resource/audience, the STS returns a token whose audience is not the
MCP backend, so audience-validating backends reject it. RFC 8707 is the
standard way to scope an exchanged token to its intended resource. The STS
client already serialized `resource`/`audience`; only the call site and the
configuration plumbing were missing.

## Notes

- Backwards compatible: no env set means no `resource`/`audience` sent.
- The env vars are registered in `go/core/pkg/env` for discoverability. Moving
  this into the adk config file (with a first-class `Agent` CRD field) is under
  discussion; see the thread below.

### DIFF
--- a/go/adk/pkg/runner/adapter.go
+++ b/go/adk/pkg/runner/adapter.go
@@ -141,6 +141,24 @@ func buildTokenPropagationPlugin(ctx context.Context, log logr.Logger) (*sts.Tok
 		return nil, fmt.Errorf("failed to initialize STS integration: %w", err)
 	}
 
+	var opts []sts.Option
+	if resource, audience := exchangeTarget(); resource != nil || audience != nil {
+		opts = append(opts, sts.WithExchangeTarget(resource, audience))
+	}
+
 	log.Info("Enabling STS token propagation plugin", "wellKnownURI", stsWellKnownURI)
-	return sts.NewTokenPropagationPlugin(integration, log), nil
+	return sts.NewTokenPropagationPlugin(integration, log, opts...), nil
+}
+
+// exchangeTarget reads the RFC 8707 resource and RFC 8693 audience from the
+// environment. An unset value is returned as nil so it is omitted from the
+// token-exchange request.
+func exchangeTarget() (resource, audience any) {
+	if r := strings.TrimSpace(os.Getenv("KAGENT_TOKEN_RESOURCE")); r != "" {
+		resource = r
+	}
+	if a := strings.TrimSpace(os.Getenv("KAGENT_TOKEN_AUDIENCE")); a != "" {
+		audience = a
+	}
+	return resource, audience
 }
--- a/go/adk/pkg/runner/adapter.go
+++ b/go/adk/pkg/runner/adapter.go
@@ -143,8 +143,8 @@ func buildTokenPropagationPlugin(ctx context.Context, log logr.Logger) (*sts.Tok
 
 	// RFC 8707 resource / RFC 8693 audience scope the exchanged token to a
 	// backend. Empty values are omitted from the exchange request.
-	resource := strings.TrimSpace(os.Getenv("KAGENT_TOKEN_RESOURCE"))
-	audience := strings.TrimSpace(os.Getenv("KAGENT_TOKEN_AUDIENCE"))
+	resource := strings.TrimSpace(os.Getenv("KAGENT_STS_RESOURCE"))
+	audience := strings.TrimSpace(os.Getenv("KAGENT_STS_AUDIENCE"))
 
 	log.Info("Enabling STS token propagation plugin", "wellKnownURI", stsWellKnownURI)
 	return sts.NewTokenPropagationPlugin(integration, log, resource, audience), nil
--- a/go/adk/pkg/runner/adapter.go
+++ b/go/adk/pkg/runner/adapter.go
@@ -124,7 +124,7 @@ func buildTokenPropagationPlugin(ctx context.Context, log logr.Logger) (*sts.Tok
 	// Propagate-only mode: keep parity with Python by enabling plugin without STS exchange.
 	if stsWellKnownURI == "" {
 		log.Info("Enabling token propagation plugin without STS exchange")
-		return sts.NewTokenPropagationPlugin(nil, log, "", ""), nil
+		return sts.NewTokenPropagationPlugin(nil, log, nil, nil), nil
 	}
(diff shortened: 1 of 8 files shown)


QuentinBisson:
Python companion: #2107 (same env vars across both runtimes).

QuentinBisson:
Thanks — addressed in 7dccee6:

- **`any` exchange target (plugin.go):** `WithExchangeTarget` now takes `string` instead of `any`, so a caller can't pass a type the STS client silently drops; empty values are omitted (unset → unscoped, as before). Multi-valued resources are out of scope for this change.
- **Test data race (plugin_test.go):** the handler now sends the captured form over a buffered channel; the test reads it on its own goroutine, so there's no shared-variable race and no `t.Fatalf` from the handler goroutine. Verified with `go test -race`.
- **Hang safety:** the assertion waits on the channel with a 5s timeout instead of reading a variable that may never have been written.

go/adk/pkg/sts/plugin.go
```diff
@@ -38,17 +38,41 @@ type TokenPropagationPlugin struct {
 	mu              sync.RWMutex
 	logger          logr.Logger
 	bufferSeconds   int64
+	resource        any // RFC 8707 resource indicator passed to the STS exchange
+	audience        any // RFC 8693 audience passed to the STS exchange
```
EItanya: Why are these `any`, they can be `string` or `[]string`?

go/adk/pkg/sts/plugin.go
```diff
@@ -38,17 +38,41 @@ type TokenPropagationPlugin struct {
...
+func WithExchangeTarget(resource, audience string) Option {
+	return func(p *TokenPropagationPlugin) {
+		if resource != "" {
+			p.resource = resource
+		}
+		if audience != "" {
+			p.audience = audience
+		}
+	}
 }
```
EItanya: Do we really need to introduce an `options` pattern for 2 strings?

go/adk/pkg/runner/adapter.go
```diff
@@ -123,6 +123,11 @@ func buildTokenPropagationPlugin(ctx context.Context, log logr.Logger) (*sts.Tok
 		return nil, fmt.Errorf("failed to initialize STS integration: %w", err)
 	}
 
+	// RFC 8707 resource / RFC 8693 audience scope the exchanged token to a
+	// backend. Empty values are omitted by WithExchangeTarget.
+	resource := strings.TrimSpace(os.Getenv("KAGENT_TOKEN_RESOURCE"))
+	audience := strings.TrimSpace(os.Getenv("KAGENT_TOKEN_AUDIENCE"))
```
EItanya: Frankly I don't really like adding more env vars if we don't have to, they're hard to track. Can we add a token exchange section to our adk config file instead? This is for the Python impl as well

EItanya (changes requested):

go/adk/pkg/runner/adapter.go
QuentinBisson: Thanks for the review. I'll see what I can do :)

QuentinBisson:
@EItanya thanks for the review. Addressed the two smaller points and pushed:
- Dropped the functional option: `NewTokenPropagationPlugin` now takes `resource, audience string` directly and the fields are typed `[]string`.
- Renamed the env vars to `KAGENT_STS_RESOURCE` / `KAGENT_STS_AUDIENCE` here and in #2107.

On moving these into the adk config file instead of env vars: happy to. Before I wire it, one scoping question so I size the PRs right. My plan is a `tokenExchange: { resource, audience }` section in the agent `config.json` both runtimes already read, populated by a new Agent CRD field the controller renders, with the env vars dropped. Do you want that CRD field plus rendering inside these two PRs, or should these ship env-based as-is and the config section land as a follow-up (Go and Python together)?

go/adk/pkg/sts/plugin.go
QuentinBisson: No this was removed

go/adk/pkg/sts/plugin_test.go
```diff
@@ -139,6 +139,86 @@ func TestBeforeRunCallback_ReusesCachedDynamicActorTokenForExchange(t *testing.T
...
+			}); err != nil {
+				t.Fatalf("BeforeRunCallback() error = %v", err)
+			}
+
+			if gotResource != tt.wantResource {
+				t.Fatalf("resource = %q, want %q", gotResource, tt.wantResource)
+			}
+			if gotAudience != tt.wantAudience {
+				t.Fatalf("audience = %q, want %q", gotAudience, tt.wantAudience)
+			}
```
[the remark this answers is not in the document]
QuentinBisson: resolved

go/adk/pkg/sts/plugin_test.go
```diff
@@ -139,6 +139,86 @@ func TestBeforeRunCallback_ReusesCachedDynamicActorTokenForExchange(t *testing.T
...
+				if err := r.ParseForm(); err != nil {
+					t.Fatalf("ParseForm() error = %v", err)
+				}
+				gotResource = r.FormValue("resource")
+				gotAudience = r.FormValue("audience")
+				_ = json.NewEncoder(w).Encode(map[string]any{
+					"access_token":      "access-token",
+					"issued_token_type": string(TokenTypeJWT),
+				})
+			}))
```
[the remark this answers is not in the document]
QuentinBisson: resolved

go/adk/pkg/sts/plugin.go
QuentinBisson: I updated them to `[]string` to be spec compliant :)

supreme-gg-gg (approved):

PR #2107, opened 2026-06-29
feat(sts): send RFC 8707 resource and audience on token exchange (Python)

QuentinBisson:
## What

The Python ADK token-propagation plugin (`agentsts-adk`) called
`exchange_token` without `resource` or `audience`, so an issued token could
not be scoped to a specific backend. This wires two optional inputs through:

- `KAGENT_STS_RESOURCE`: RFC 8707 resource indicator (the target backend).
- `KAGENT_STS_AUDIENCE`: RFC 8693 audience, for STS servers that key on it.

`ADKTokenPropagationPlugin` now accepts `resource`/`audience`; the CLI reads
them from the environment in `create_sts_integration`. Unset values are
omitted from the request, so behaviour is unchanged when neither is set.

## Why

Without a resource/audience, the STS returns a token whose audience is not the
MCP backend, so audience-validating backends reject it. RFC 8707 is the
standard way to scope an exchanged token to its intended resource. The
`exchange_token` API already accepted `resource`/`audience`; only the plugin
call site and the configuration plumbing were missing.

## Notes

- Backwards compatible: no env set means no `resource`/`audience` sent.
- Mirrors the Go ADK change in #2106; same env var names across runtimes.

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/cli.py
+++ b/python/packages/kagent-adk/src/kagent/adk/cli.py
@@ -26,6 +26,8 @@ app = typer.Typer()
 kagent_url_override = os.getenv("KAGENT_URL")
 sts_well_known_uri = os.getenv("STS_WELL_KNOWN_URI")
 propagate_token = os.getenv("KAGENT_PROPAGATE_TOKEN", "").lower() == "true"
+token_resource = os.getenv("KAGENT_TOKEN_RESOURCE") or None
+token_audience = os.getenv("KAGENT_TOKEN_AUDIENCE") or None
 uvicorn_log_level = os.getenv("UVICORN_LOG_LEVEL", os.getenv("LOG_LEVEL", "info")).lower()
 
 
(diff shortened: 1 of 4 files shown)


python/packages/kagent-adk/src/kagent/adk/cli.py
```diff
@@ -25,6 +25,8 @@
 kagent_url_override = os.getenv("KAGENT_URL")
 sts_well_known_uri = os.getenv("STS_WELL_KNOWN_URI")
 propagate_token = os.getenv("KAGENT_PROPAGATE_TOKEN", "").lower() == "true"
+token_resource = os.getenv("KAGENT_TOKEN_RESOURCE") or None
+token_audience = os.getenv("KAGENT_TOKEN_AUDIENCE") or None
```
EItanya: Can you please rename these `KAGENT_STS_*` so it's more clear what they belong to.

EItanya (changes requested):
See https://github.com/kagent-dev/kagent/pull/2106#discussion_r3492079265

supreme-gg-gg (approved):

PR #2113, opened 2026-06-30
Allow configuration of A2A client timeout

### LINKED ISSUE #2112: [FEATURE] Allow configuration of HTTP client timeout in A2A SDK clients
### 📋 Prerequisites

### 📝 Feature Summary

Allow configuration of HTTP client timeout in A2A SDK clients

The A2A Go SDK (a2aclient package) hardcodes an HTTP client timeout of 3 minutes (defaultRequestTimeout = 3 * time.Minute) when no custom http.Client is provided. This is incompatible with SSE-based streaming agents that legitimately run longer than 3 minutes, causing downstream consumers to receive context deadline exceeded errors even when the agent is still actively streaming responses.

## Current Behaviour

When constructing an A2A client transport without a pre-configured http.Client:

```
// In a2aclient/jsonrpc.go
func NewJSONRPCTransport(url string, client *http.Client) Transport {
    t := &jsonrpcTransport{url: url}
    if client == nil {
        t.httpClient = &http.Client{Timeout: defaultRequestTimeout}  // 3 minutes
    } else {
        t.httpClient = client
    }
    // ...
}
```

The 3-minute limit applies to the entire HTTP request/response cycle, including SSE streaming where events may be spaced minutes apart. This is too short for long-running analysis or computation tasks.

## Impact

Downstream consumers like kagent controller, that rely on the SDK's defaults for streaming A2A calls experience timeout failures:

```
SSE stream error: context deadline exceeded (Client.Timeout or context cancellation while reading body)
```

### 💡 Proposed Solution

Add a timeout configuration option to a2aclient factory functions.

This could be a environment variable that users could pass in the Helm values and the controller starts using to implement the extended timeout limit. 

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

_No response_

Daniel-Vaz:
With this change users can now set `controller.a2aClientTimeout` in their Helm values to override the timeout. Default behavior (empty/0) removes all timeout, preventing the 3-minute SSE cutoff.

Closes #2112

# Problem

The a2a-go SDK applies a hardcoded 3-minute HTTP timeout to all A2A client requests. This causes context deadline exceeded errors when agents take longer than 3 minutes to complete, even during normal SSE streaming operations.

# Proposed Solution

- Add `KAGENT_A2A_CLIENT_TIMEOUT` environment variable (default: 0 = no timeout)
- Create `a2aHTTPClient()` function to build HTTP clients respecting the configured timeout
- Wire the timeout through Helm values (`controller.a2aClientTimeout`) → `ConfigMap` → env var
- Replaces the SDK's 3-minute default with a configurable value suitable for long-running agents

# Changes

- `kagent.go`: Register `KagentA2AClientTimeout` env var
- `a2a_registrar.go`: Implement `a2aHTTPClient()` and use it instead of `debugHTTPClient()`
- `values.yaml`: Add `controller.a2aClientTimeout` Helm value
- `controller-configmap.yaml`: Wire env var to the `ConfigMap`

### DIFF
--- a/go/core/internal/a2a/a2a_registrar.go
+++ b/go/core/internal/a2a/a2a_registrar.go
@@ -216,7 +216,7 @@ func (a *A2ARegistrar) upsertAgentHandler(ctx context.Context, agent v1alpha2.Ag
 
 	provider := resolveProviderName(ctx, a.cache, agent)
 
-	httpClient := debugHTTPClient()
+	httpClient := a2aHTTPClient()
 	if sa, ok := agent.(*v1alpha2.SandboxAgent); ok &&
 		a.substrateSandboxActorBackend != nil {
 		routerURL := a.ateneRouterURL
@@ -274,8 +274,8 @@ func (a *A2ARegistrar) upsertAgentHandler(ctx context.Context, agent v1alpha2.Ag
 	return nil
 }
 
-// debugHTTPClient returns nil in normal operation, letting the a2aclient SDK apply its
-// default 3-minute request timeout. In debug mode it overrides the dial target so all
+// debugHTTPClient returns nil in normal operation, letting the caller apply its
+// configured timeout. In debug mode it overrides the dial target so all
 // A2A traffic is redirected to a fixed address (e.g. a local proxy).
 func debugHTTPClient() *http.Client {
 	debugAddr := env.KagentA2ADebugAddr.Get()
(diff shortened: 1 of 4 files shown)


go/core/internal/a2a/a2a_registrar.go
```diff
@@ -292,6 +292,17 @@ func debugHTTPClient() *http.Client {
 	}
 }
 
+// a2aHTTPClient returns the HTTP client to use for A2A requests to agent pods.
+// It respects KAGENT_A2A_CLIENT_TIMEOUT (default 0 = no timeout), overriding
+// the a2a-go SDK's built-in 3-minute default which is too short for long-running
+// streaming agents.
+func a2aHTTPClient() *http.Client {
```
EItanya: Can we combine this and the existing `debugHTTPClient` functions into a single function so the flow is readable?

Daniel-Vaz: Thanks for the review.
I've merged `debugHTTPClient()` and `a2aHTTPClient()` into a single function.
When possible please do have another look

EItanya (approved):

PR #2133, opened 2026-07-01
fix(security): patch CRITICAL/HIGH CVEs in golang-adk-full and app images

jsonmp-k8:
## Summary

Fixes CRITICAL and HIGH severity vulnerabilities detected by Trivy in the `Scan images` CI workflow.

**golang-adk-full image:**
- `shell-quote` **CVE-2026-9277** (CRITICAL) — Arbitrary code execution; pin to `1.8.4` alongside the existing `lodash-es` override in `go/Dockerfile.full`

**app image:**
- `PyJWT` **CVE-2026-48526** (HIGH) — Authentication bypass via forged tokens; bump constraint from `>=2.12.0` → `>=2.13.0` in `agentsts-core`, `agentsts-adk`, and workspace `constraint-dependencies`; lock resolves to `2.13.0`
- `pyarrow` **CVE-2026-25087** (HIGH) — Denial of Service via Use-After-Free; add workspace constraint `>=23.0.1`; lock resolves to `24.0.0`

**Remaining / blocked:**
- `starlette` CVE-2026-48818 + CVE-2026-54283 (HIGH) — Fix requires starlette `>=1.3.1` (1.x), but `google-adk>=1.28.1,<2` pins `starlette<1.0.0`. A follow-up PR upgrading `agentsts-adk` to `google-adk>=2` is needed to fully close these.

## Changes

| File | Change |
|------|--------|
| `go/Dockerfile.full` | Add `shell-quote@1.8.4` to sandbox-runtime npm overrides |
| `python/packages/agentsts-core/pyproject.toml` | `PyJWT>=2.12.0` → `>=2.13.0` |
| `python/packages/agentsts-adk/pyproject.toml` | `PyJWT>=2.12.0` → `>=2.13.0` |
| `python/pyproject.toml` | Add `pyjwt>=2.13.0` and `pyarrow>=23.0.1` to `constraint-dependencies` |
| `python/uv.lock` | pyjwt `2.12.1`→`2.13.0`, pyarrow `21.0.0`→`24.0.0`, starlette `0.50.0`→`0.52.1` |

## Test plan

- [ ] `Scan images` CI passes for `golang-adk-full` (shell-quote CVE cleared)
- [ ] `Scan images` CI passes for `app` (PyJWT + pyarrow CVEs cleared)
- [ ] All existing unit/integration tests pass


EItanya (approved):
Thanks so much :)

jsonmp-k8:
Thanks for the detailed review — both points are valid. Opened #2138 as the follow-up:

1. **Sample lockfile gap** — you're right that `python/samples/adk/basic/uv.lock` was untouched here and still resolves `starlette 0.47.2`. The image-scan workflow doesn't cover the sample, but the exposure is real. #2138 regenerates that lockfile too (built in an isolated context since the sample's Docker build treats it as standalone).

2. **google-adk major upgrade scope** — you're right that a v1→v2 bump is a large change to fold into a security patch. It isn't scope creep though; the chain is forced:
   - Both starlette CVEs are fixed only in the 1.x line (no 0.x backport)
   - `google-adk<2` hard-pins `starlette<1.0.0`
   - `google-adk 2.2.0` is the earliest 2.x release that supports `starlette>=1.0.1` while staying under kagent-core's existing `opentelemetry-api<1.39.0` pin (2.3.0 would have forced an opentelemetry bump too)
   
   If a VEX / `.trivyignore` route makes more sense — treating the CVEs as non-exploitable in kagent's usage after auditing `StaticFiles` / `application/x-www-form-urlencoded` surface — happy to switch #2138 over. Let me know which direction reviewers prefer.

PR #2134, opened 2026-07-01
fix(helm): keep non-root UID defaults for grafana-mcp and querydoc, document OpenShift override

### LINKED ISSUE #2079: [BUG] Bundled subcharts hardcode securityContext.runAsUser/runAsGroup → not deployable on OpenShift (restricted-v2 dynamic UID range)
## Summary

Several bundled subcharts pin `securityContext.runAsUser` / `runAsGroup` (and one `podSecurityContext`) to fixed UIDs in their **default** values. On OpenShift this makes the chart undeployable out of the box: each namespace is assigned a dynamic UID range and the `restricted-v2` SCC rejects any pod whose `runAsUser` falls outside that range. The affected Deployments stay at `0/1` with a `FailedCreate` event, and every component has to be overridden to an in-range UID by hand.

Relates to #459 ("[FEATURE] OpenShift support", closed as stale).

## Affected defaults (chart `0.9.9`)

| component | file | value |
|-----------|------|-------|
| `grafana-mcp` | `charts/grafana-mcp/values.yaml` (`securityContext`) | `runAsUser: 1000`, `runAsGroup: 999` |
| `querydoc` | `charts/querydoc/values.yaml` (`securityContext`) | `runAsUser: 14000`, `runAsGroup: 14000` |
| `oauth2-proxy` (vendored dependency) | `charts/oauth2-proxy/values.yaml` (`podSecurityContext`) | `runAsUser: 2000`, `runAsGroup: 2000` |
| bundled demo PostgreSQL | `values.yaml` (`...postgresql.podSecurityContext`) | `runAsUser: 999`, `runAsGroup: 999`, `fsGroup: 999` |

## Impact / symptom

On OpenShift (e.g. `restricted-v2` SCC), the API server rejects the pod at admission. Representative event:

```
Error creating: pods "grafana-mcp-..." is forbidden: unable to validate against any
security context constraint: ... runAsUser: Invalid value: 1000: must be in the
ranges: [1001150000, 1001159999]
```

The Deployment then sits at `0/1`. To install the chart we currently have to override each component's `securityContext.runAsUser`/`runAsGroup` to a value inside the namespace's assigned range — which is cluster/namespace-specific and not knowable ahead of time.

## Proposed fix

Don't pin `runAsUser`/`runAsGroup` in the subchart **default** values. Keep the rest of the hardening that is already there (`runAsNonRoot: true`, `capabilities.drop: [ALL]`, `readOnlyRootFilesystem: true`, `seccompProfile`). With no `runAsUser` set:

- **OpenShift** injects a valid in-range UID automatically via the SCC.
- **Vanilla Kubernetes** runs as the image's `USER` and `runAsNonRoot: true` still enforces non-root.

This assumes the `grafana-mcp` / `querydoc` images declare a non-root `USER`. If they don't, the alternative is to make the UID cleanly nullable/overridable (so it can be unset) and document the OpenShift guidance, rather than shipping a hardcoded out-of-range default.

For the vendored `oauth2-proxy` dependency, surfacing/recommending an override (or unsetting it via the umbrella values) would cover it.

I'm happy to send a PR for the `grafana-mcp` and `querydoc` subcharts once you confirm the preferred approach (unset default vs. nullable+documented).

## Environment

- kagent helm chart `0.9.9` (`oci://ghcr.io/kagent-dev/kagent/helm/kagent`)
- OpenShift (ARO), `restricted-v2` SCC with per-namespace dynamic UID ranges

SarthakB11:
## What

#2079 reported that grafana-mcp and querydoc are not deployable on OpenShift out of the box: they pin securityContext.runAsUser/runAsGroup and the restricted-v2 SCC rejects a fixed UID (it assigns a per-namespace range).

Unsetting the defaults is not safe: both charts set podSecurityContext.runAsNonRoot: true, and the kubelet can only confirm non-root from a numeric runAsUser or a numeric image USER. The mcp/grafana image's USER is the name "mcp-grafana" and the doc2vec/mcp (querydoc) image does not run as a numeric non-root user, so with runAsUser unset the pods fail admission with CreateContainerConfigError on vanilla Kubernetes.

This PR keeps the fixed non-root UID/GID defaults (charts install on vanilla Kubernetes) and documents the one-line override for OpenShift, where the SCC then assigns an in-range UID:

    --set tools.grafana-mcp.securityContext.runAsUser=null
    --set tools.grafana-mcp.securityContext.runAsGroup=null

(same for tools.querydoc.securityContext). It also records that the querydoc pin can be dropped once the doc2vec/mcp image ships a numeric non-root USER, opened upstream as kagent-dev/doc2vec#94. This is the fallback described in #2079 for images that do not ship a numeric non-root USER.

## Testing

- helm template: pinned UID by default; field omitted when overridden to null.
- helm unittest: grafana-mcp 10/10, querydoc 26/26.

Fixes #2079

### DIFF
--- a/helm/tools/grafana-mcp/tests/deployment_test.yaml
+++ b/helm/tools/grafana-mcp/tests/deployment_test.yaml
@@ -64,12 +64,10 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].securityContext.readOnlyRootFilesystem
           value: true
-      - equal:
+      - isNull:
           path: spec.template.spec.containers[0].securityContext.runAsUser
-          value: 1000
-      - equal:
+      - isNull:
           path: spec.template.spec.containers[0].securityContext.runAsGroup
-          value: 999
       - isNull:
           path: spec.template.spec.containers[0].securityContext.seccompProfile
 
--- a/helm/tools/grafana-mcp/tests/deployment_test.yaml
+++ b/helm/tools/grafana-mcp/tests/deployment_test.yaml
@@ -64,10 +64,12 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].securityContext.readOnlyRootFilesystem
           value: true
-      - isNull:
+      - equal:
           path: spec.template.spec.containers[0].securityContext.runAsUser
-      - isNull:
+          value: 1000
+      - equal:
           path: spec.template.spec.containers[0].securityContext.runAsGroup
+          value: 999
       - isNull:
           path: spec.template.spec.containers[0].securityContext.seccompProfile
 
--- a/helm/tools/grafana-mcp/values.yaml
+++ b/helm/tools/grafana-mcp/values.yaml
@@ -31,14 +31,15 @@ securityContext:
     drop:
       - ALL
   readOnlyRootFilesystem: true
-  # runAsUser / runAsGroup are left unset so the chart installs on OpenShift,
-  # where the restricted-v2 SCC assigns a per-namespace UID range and rejects a
-  # pinned UID outside it. runAsNonRoot (in podSecurityContext) enforces
-  # non-root: with runAsUser unset the container runs as the image's USER, so
-  # the mcp/grafana image must ship a non-root USER or the pod will fail to
-  # start. Set a fixed UID below on clusters that allow it:
-  # runAsUser: 1000
-  # runAsGroup: 999
+  # A fixed non-root UID/GID is set so the chart installs on vanilla Kubernetes:
+  # runAsNonRoot (in podSecurityContext) needs a numeric non-root user and the
+  # mcp/grafana image does not ship one. On OpenShift, where the restricted-v2
+  # SCC assigns a per-namespace UID range and rejects a pinned UID outside it,
+  # override these to null so the SCC injects an in-range UID:
+  #   --set tools.grafana-mcp.securityContext.runAsUser=null
+  #   --set tools.grafana-mcp.securityContext.runAsGroup=null
+  runAsUser: 1000
+  runAsGroup: 999
 
 tolerations: []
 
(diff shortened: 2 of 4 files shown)


SarthakB11:
Thanks for the review. Force-pushed a fix. Two changes on top of the original commit:

- helm-unittest updates in `helm/tools/querydoc/tests/deployment_test.yaml` and `helm/tools/grafana-mcp/tests/deployment_test.yaml`: the default-container securityContext test now asserts `runAsUser` and `runAsGroup` are unset via `isNull`, matching the new default. The override test in both files still sets `runAsUser` and asserts it renders, so the override path stays covered.
- Values comment reworded: `runAsNonRoot` enforces non-root rather than "keeping" it non-root; with `runAsUser` unset the container runs as the image's `USER`, so `mcp/grafana` and `doc2vec/mcp` must ship a non-root `USER` for the pod to start on non-OpenShift clusters. That constraint is now spelled out.

Also rebased onto main (was BEHIND #2124 / #2125). CI should re-run on the amended commit.

SarthakB11:
Unsetting runAsUser is not safe here: neither image runs as a numeric non-root user, so with runAsNonRoot on, the pods fail admission on vanilla Kubernetes. Kept the pinned defaults and documented the OpenShift null-override. querydoc can be unpinned once its image ships a numeric USER, opened upstream as kagent-dev/doc2vec#94.

SarthakB11:
@EItanya @peterj this got closed by stalebot, but the fix was actually done, kept pinned runAsUser/runAsGroup defaults (unsetting isn't safe since neither image runs as non-root), documented the OpenShift null-override, and opened an upstream issue for querydoc to get a numeric USER so it can be unpinned later. Mergeable, no conflicts. Could one of you reopen it?

PR #2138, opened 2026-07-02
fix(security): patch starlette CVE-2026-48818 and CVE-2026-54283

jsonmp-k8:
## Summary

Follow-up to #2133 (which patched shell-quote, PyJWT, and pyarrow but left starlette on `0.52.1` in main and `0.47.2` in the `samples/adk/basic` image). This PR closes the remaining two HIGH-severity Trivy findings that the `app` image scan is still flagging.

**CVEs closed:**
- `starlette` **CVE-2026-48818** (HIGH) — SSRF and NTLM credential theft via UNC paths in `StaticFiles`. Fix: `starlette>=1.1.0`.
- `starlette` **CVE-2026-54283** (HIGH) — `request.form()` limits silently ignored for `application/x-www-form-urlencoded`, enabling DoS. Fix: `starlette>=1.3.1`.

## Why the google-adk major version bump is unavoidable

The reviewer on #2133 raised a fair concern that upgrading `google-adk` v1 → v2 is a large behavior change for a "security patch". This isn't scope creep — it's the shortest forced chain:

1. Both starlette CVEs are fixed **only** in the starlette 1.x line. There is no 0.x backport.
2. `google-adk<2` hard-pins `starlette<1.0.0`, so it is impossible to install starlette 1.x with google-adk v1.
3. `google-adk 2.2.0` is the earliest 2.x release that supports `starlette>=1.0.1` **and** stays within `kagent-core`'s existing `opentelemetry-api>=1.38.0,<1.39.0` pin (google-adk 2.3.0 would have forced opentelemetry ≥1.39 too). Picking 2.2.0 minimizes the transitive blast radius.

The alternative — filing a VEX / `.trivyignore` if the CVEs are not exploitable in kagent's usage — was considered. It would require an audit of whether kagent-adk / the `app` image exposes `starlette.StaticFiles` or handles `application/x-www-form-urlencoded` form parsing. Happy to switch to that approach instead if reviewers prefer, but the current PR is the "actually fix it" path.

## Changes

| File | Change |
|------|--------|
| `python/packages/kagent-adk/pyproject.toml` | `google-adk>=1.28.1,<2` → `>=2.2.0` |
| `python/packages/agentsts-adk/pyproject.toml` | `google-adk>=1.28.1,<2` → `>=2.2.0` |
| `python/samples/adk/basic/pyproject.toml` | `google-adk>=1.8.0` → `>=2.2.0` |
| `python/pyproject.toml` | Add `starlette>=1.3.1` to workspace `constraint-dependencies` |
| `python/uv.lock` | Regenerated; starlette `0.52.1` → `1.3.1`, google-adk `1.31.1` → `2.2.0`. ~30 unused google-cloud-* packages drop out of the tree. |
| `python/samples/adk/basic/uv.lock` | Regenerated (standalone build context); starlette `0.47.2` → `1.3.1`, google-adk `1.9.0` → `2.3.0` |

## Test plan

- [ ] `Scan images` CI passes on `app` (both starlette CVEs cleared)
- [ ] `python-test (3.10 / 3.11 / 3.12 / 3.13)` all pass on the google-adk v2 line
- [ ] `test-e2e` passes to catch any google-adk 1 → 2 API regressions
- [ ] Manual sanity check that a sample agent (`samples/adk/basic`) still boots


jsonmp-k8:
The `build (golang-adk-full)` failure (and the cascading matrix cancellations) is a **pre-existing infrastructure issue on main**, not introduced by this PR.

Error is identical:
\`\`\`
qemu: uncaught target signal 4 (Illegal instruction) - core dumped
Dockerfile.full:35
ERROR: process ".../npm install ..." did not complete successfully: exit code: 132
\`\`\`

Timeline:
- 14:39 UTC — #2133 (my previous security fix) merged to main → CI **SUCCESS**
- 14:40 UTC — [#2132](https://github.com/kagent-dev/kagent/actions/runs/28598665763) merged to main → **same QEMU crash on same Dockerfile.full:35, same nodejs-26.4.0-r1**
- 15:37 UTC — this PR (#2138) → inherits the broken build

Root cause is apk pulling `nodejs-26.4.0-r1` at build time, whose arm64 binary uses instructions that `docker/setup-qemu-action@v4` (QEMU 10.2.3) cannot emulate. `go/Dockerfile.full` line 30 installs `nodejs` unpinned even though `TOOLS_NODE_VERSION=24` is threaded in as a build arg but never wired into the `apk add` command — so builds silently drifted to nodejs 26 when Wolfi published it.

This PR only touches Python `pyproject.toml` / `uv.lock` files — none of the go Docker builds. The fix belongs in a separate infra PR (either pin `nodejs-24` in the apk add, or pin a QEMU version that supports the newer node arm64 binary).

Happy to open that infra PR too if useful, but flagging this here so the failing checks don't block review of the actual security change.

EItanya:
How is this passing CI, is adk 2.0 backwards compatible?

jsonmp-k8:
-
**What the passing CI actually covers**

- `python-test (3.10 / 3.11 / 3.12 / 3.13)` — all four Python versions install `google-adk 2.2.0` (confirmed in the regenerated `python/uv.lock`; PyPI metadata says `requires-python = ">=3.10"` for 2.x so all four resolve to 2.2.0, not a silent fallback to 1.x).
- `test-e2e` — 23-minute run passes end-to-end.
- `go-unit-tests` — passes.

The kagent surface hitting google-adk that these tests exercise (grep'd from the tree):
`Agent`, `BaseAgent`, `InvocationContext`, `Event`/`EventActions`, `Session`, `McpTool`/`McpToolset`, `SseConnectionParams`/`StreamableHTTPConnectionParams`, `ToolConfirmation`, `ToolContext`, `LlmRequest`/`LlmResponse`, `BaseCodeExecutor`, `CodeExecutionInput`/`CodeExecutionResult`, `a2a.converters.part_converter`, `a2a.executor.a2a_agent_executor`, `agents.remote_a2a_agent`, `flows.llm_flows.functions`, `utils.instructions_utils`. Every one of these still imports and behaves under 2.2.0 in the running tests.

**Known 2.x breaking changes that the tests may not catch**

1. **Session schema**: 2.x-generated sessions are forward-readable by 1.28+ but not backwards-readable by older 1.x. Only relevant if kagent runs mixed-version agents against a shared session store — probably a non-issue but worth calling out.
2. **Event model** and **agent API**: internal shape changes. Our imports still resolve, but if downstream code introspects event dicts by key rather than by helper, that could silently drift.
3. **google-adk 2.2.0 specifically**: `LlmAgent` default model changed from `gemini-2.5-flash` → `gemini-3-flash-preview` (a preview model). Any agent constructed without an explicit `model=` argument will start hitting the preview endpoint. If kagent has agents that omit `model=` on purpose, they'll need to pin.
4. **GenAI SDK v2 rename**: `convert_contents_to_turns` → `convert_contents_to_steps` in `interactions_utils.py`. No kagent code paths use this today (grep confirms), but any user extending on that helper would break.

**Honest bottom line**

Tests exercise the imports kagent actually calls, and they pass. I haven't manually diffed every 2.x method signature against every internal call site — for that level of assurance, the alternative is what I offered earlier: keep google-adk on 1.x and file a VEX / `.trivyignore` for the two starlette CVEs after auditing whether `StaticFiles` and `application/x-www-form-urlencoded` form parsing are actually exposed in kagent's request path. That's a smaller-blast-radius PR if the compatibility risk here feels wrong for a security patch.

Happy to switch to the VEX route — just let me know which shape you'd prefer.

PR #2142, opened 2026-07-03
feat(helm/kagent): allow disabling default modelconfig

anthonyhaussman:
Wrap `modelconfig` and `modelconfig-secret` templates in a conditional check for `.Values.providers`.
This allows users to completely disable the default modelconfig generation by setting providers to null.

Add tests to verify that no documents are rendered when providers is null.

### DIFF
--- a/helm/kagent/templates/modelconfig-secret.yaml
+++ b/helm/kagent/templates/modelconfig-secret.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.providers }}
 {{- $dot := . }}
 {{- $model := index $dot.Values.providers $dot.Values.providers.default  }}
 {{- if and $model.apiKeySecretRef $model.apiKey }}
@@ -12,4 +13,5 @@ metadata:
 type: Opaque
 data:
   {{ $model.apiKeySecretKey | default (printf "%s_API_KEY" $model.provider | upper) }}: {{ $model.apiKey | b64enc }}
+{{- end }}
 {{- end }}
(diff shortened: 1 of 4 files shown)


helm/kagent/templates/modelconfig-secret.yaml
```diff
@@ -1,5 +1,8 @@
+{{- if .Values.providers }}
 {{- $dot := . }}
-{{- $model := index $dot.Values.providers $dot.Values.providers.default  }}
+{{- $defaultProvider := .Values.providers.default | default "openAI" }}
+{{- if hasKey .Values.providers $defaultProvider }}
+{{- $model := index .Values.providers $defaultProvider }}
 {{- if and $model.apiKeySecretRef $model.apiKey }}
```
EItanya, line 6: Won't this line already disable the rendering if these aren't set?

anthonyhaussman: The problem is that if `providers.default` points to a missing key, `index` returns nil.
Setting `providers: {}` value returns when trying to template:
```
Error: template: kagent/templates/modelconfig-secret.yaml:2:14: executing "kagent/templates/modelconfig-secret.yaml" at <index $dot.Values.providers $dot.Values.providers.default>: error calling index: value is nil; should be of type string
```

The check is necessary to prevent Helm from crashing.

anthonyhaussman: Hey @EItanya :wave: 
Just following up on this. Does the explanation above clarify why the `hasKey` check is needed to prevent Helm from crashing?

Let me know if there's anything else you'd like me to change, or if you are waiting on anything else from my side. :pray:

anthonyhaussman: Hi @EItanya,
Just a gentle bump on this thread.
Does my explanation about the `hasKey` check make sense, or would you prefer a different approach to prevent the Helm crash? 

Let me know if you need anything else from my side to resolve this.

PR #2145, opened 2026-07-03
feat: add nodeSelector support to agent helm charts

the-mentor:
## What

Adds an optional `nodeSelector` value to every chart under `helm/agents/`, allowing agent pods to be constrained to nodes with matching labels.

## Why

There was no way to control agent pod scheduling via the Helm charts. This is needed for environments that pin workloads to specific node pools, e.g.:

```yaml
nodeSelector:
  disktype: ssd
```

## How

- Added a `nodeSelector` block (guarded by `{{- with .Values.nodeSelector }}`) to the shared `agent.deploymentSpec` helper in each agent's `_helpers.tpl`. It renders into the Agent CRD's `spec.declarative.deployment.nodeSelector`, which `SharedDeploymentSpec` already supports.
- Added `nodeSelector: {}` (documented) to each agent's `values.yaml`. When unset it is omitted, so there is no change for existing users.

Works standalone and from the main `kagent` chart via subchart value propagation, keyed by dependency name:

```yaml
helm-agent:
  nodeSelector:
    disktype: ssd
k8s-agent:
  nodeSelector:
    disktype: ssd
```

Agents covered: `k8s`, `kgateway`, `istio`, `promql`, `observability`, `argo-rollouts`, `helm`, `cilium-policy`, `cilium-manager`, `cilium-debug`.

## Testing

- Added a `helm-unittest` suite (`helm/kagent/tests/agent-nodeselector_test.yaml`) with 20 tests — for each of the 10 agent charts, one asserting `nodeSelector` is omitted by default and one asserting it renders at `spec.declarative.deployment.nodeSelector` when set. The agent charts depend on parent-chart helpers, so the suite runs through the `kagent` chart (targeting each `charts/<agent>/templates/agent.yaml`) and is picked up by the existing `helm unittest helm/kagent` CI step.
- Full suite passes: `16 test suites / 235 tests`. Confirmed the new tests genuinely guard the change — removing the `nodeSelector` block from the helper makes the override assertions fail.
- `helm template` on a standalone agent chart renders `nodeSelector` under `deployment:` when set, and omits it when unset (default).
- Built the full `helm/kagent` chart via the repo's flow (stamped `Chart.yaml` files + `helm dependency update`) and confirmed `helm template` with `helm-agent.nodeSelector` / `k8s-agent.nodeSelector` overrides propagates the value into the rendered Agent CRDs.


EItanya (approved):

the-mentor:
@EItanya The PR failed on `CI Build / build (golang-adk-full)` this PR only touches the helm chart. 
Should I rebase it on main to resolve this issue ?

EItanya:
> @EItanya The PR failed on `CI Build / build (golang-adk-full)` this PR only touches the helm chart. Should I rebase it on main to resolve this issue ?

No it's a flake, I just updated to kick the jobs

PR #2147, opened 2026-07-03
feat(models): wire maxOutputTokens for native Gemini and Vertex

anthonytwh:
## What

Wire `maxOutputTokens` end to end for the native **Gemini** and **Gemini Vertex AI** model providers.

Previously neither provider could cap a response's output tokens via `ModelConfig`: `GeminiConfig` was an empty struct and `GeminiVertexAIConfig.MaxOutputTokens` was declared but never read by the translator or applied at runtime. Both providers subclass the ADK native Gemini model, which takes generation config from the per-request `LlmRequest.config` rather than the model definition, so nothing plumbed the `ModelConfig`-level setting through to the API.

## Changes

- Add `MaxOutputTokens` to `GeminiConfig` (`v1alpha2`) and regenerate the CRDs (both `go/api/config/crd/bases` and `helm/kagent-crds`).
- Carry it on the `adk.Gemini` / `adk.GeminiVertexAI` serialization structs.
- Copy `Spec.Gemini` / `Spec.GeminiVertexAI.MaxOutputTokens` in the translator.
- Parse it on the Gemini / GeminiVertexAI pydantic models and build the wrappers with it.
- Apply it in a shared `_GeminiGenerationConfigMixin` that injects `max_output_tokens` into `GenerateContentConfig` **without** overriding a value the agent already set on the request.

A per-request value always wins over the model-level default.

## Tests

- Go translator tests: `Test_AdkApiTranslator_GeminiMaxOutputTokens` (native Gemini + Vertex AI) — pass; full translator package green.
- Python unit tests for the mixin: applies the cap when unset, does not override a per-request value, and is a no-op when uncapped — 3 passed.


anthonytwh:
Thanks @copilot — addressed all five comments in ad27cd0b. They were the same point (`maxOutputTokens` accepted 0/negative even though the translator treats `<= 0` as "unset"), so I enforced a positive value at every layer:

- `+kubebuilder:validation:Minimum=1` on `GeminiConfig` and `GeminiVertexAIConfig.MaxOutputTokens`, CRDs regenerated (`minimum: 1` now in both the bases and Helm copies) — Kubernetes rejects invalid values on apply.
- pydantic `ge=1` on the `Gemini` / `GeminiVertexAI` `max_output_tokens` fields so an invalid ModelConfig fails fast at parse time.
- Added unit tests for the parse-time validation.

Go translator tests + Python model tests pass.

anthonytwh:
Thanks @copilot — all five comments make sense and are the same point (reject 0/negative `maxOutputTokens` since the translator treats `<= 0` as "unset"). They're already addressed in commit `75d90bfd` ("validate: reject non-positive Gemini maxOutputTokens"), which landed after the review was posted:

- Go: `+kubebuilder:validation:Minimum=1` on `GeminiConfig` and `GeminiVertexAIConfig.MaxOutputTokens` (`modelconfig_types.go:77,248`)
- CRDs regenerated with `minimum: 1` in both `api/config/crd/bases` and `helm/kagent-crds` (lines 540 & 554)
- Python: `Field(default=None, ge=1)` on `Gemini` and `GeminiVertexAI.max_output_tokens` (`types.py:297,311`) — fails fast at parse time

Added parse-time validation tests as well. Kubernetes now rejects invalid values on apply, and the pydantic models reject them on parse.

supreme-gg-gg (commented):
this lgtm but we will need to add the changes to our go adk agent as well, you can probably do something similar that wraps `*adk.Gemini` and `*adk.GeminiVertexAI` and sets the config

anthonytwh:
Done in 94a662a. Added a `model.LLM` wrapper in the Go ADK (`go/adk/pkg/models/gemini.go`) that mirrors the Python `_GeminiGenerationConfigMixin`:

- `CreateLLM` now wraps both the native Gemini and Gemini Vertex AI models via `WrapGeminiWithGenerationConfig(geminiModel, m.MaxOutputTokens)`.
- The wrapper injects `max_output_tokens` into each request's `GenerateContentConfig` **only when the request hasn't already set it** — a per-request value always wins — and is a no-op when `maxOutputTokens` is unset or non-positive.

Since the ADK Gemini model takes generation config from the per-request `LLMRequest.Config` rather than the model definition, this is the equivalent plumbing to what the Python runtime already does. Added unit tests covering apply-when-unset, don't-override-per-request, nil-config init, and the no-op cases.

go/adk/pkg/agent/agent.go
```diff
@@ -228,10 +228,14 @@ func CreateLLM(ctx context.Context, m adk.Model, log logr.Logger) (adkmodel.LLM,
...
 		}
-		return adkgemini.NewModel(ctx, modelName, &genai.ClientConfig{
+		geminiModel, err := adkgemini.NewModel(ctx, modelName, &genai.ClientConfig{
 			APIKey:     apiKey,
 			HTTPClient: httpClient,
 		})
+		if err != nil {
+			return nil, err
+		}
+		return models.WrapGeminiWithGenerationConfig(geminiModel, m.MaxOutputTokens), nil
```
EItanya: Can we create our own constructor, I don't love the wrapping concept

go/adk/pkg/agent/agent.go
```diff
@@ -246,11 +250,15 @@ func CreateLLM(ctx context.Context, m adk.Model, log logr.Logger) (adkmodel.LLM,
...
-		return adkgemini.NewModel(ctx, modelName, &genai.ClientConfig{
+		geminiModel, err := adkgemini.NewModel(ctx, modelName, &genai.ClientConfig{
 			Backend:  genai.BackendVertexAI,
 			Project:  project,
 			Location: location,
 		})
+		if err != nil {
+			return nil, err
+		}
+		return models.WrapGeminiWithGenerationConfig(geminiModel, m.MaxOutputTokens), nil
```
EItanya: Same concept here

EItanya (commented):
How does ADK recommend doing this, seems a bit weird that you have to wrap the client

anthonytwh:
Good call @EItanya — the wrapper wasn't the idiomatic way. ADK's recommended mechanism is `llmagent.Config.GenerateContentConfig`: the basic request processor clones it into each `LLMRequest.Config` as the seed (`internal/llminternal/basic_processor.go`), which is exactly the "model-level default that later stages can still override" semantics this needed. Same idea as `generate_content_config` on the Python `LlmAgent`.

Reworked in e2559b9:

- Removed the `geminiGenerationConfigModel` wrapper (`go/adk/pkg/models/gemini.go` deleted); `CreateLLM` now returns the native ADK Gemini / Vertex models unwrapped.
- `BuildAgent` sets `GenerateContentConfig` on the `llmagent.Config` via a small `generateContentConfig(adk.Model)` helper that pulls `MaxOutputTokens` from the Gemini / GeminiVertexAI model configs (nil for other providers or when unset).
- Precedence is unchanged: the agent-level value is only the seed for the request config, so anything that mutates the request later (e.g. before-model callbacks) still wins.

Separately, this branch hit an unrelated CI flake: `TestStaticHeaders_OverrideDynamic` in `go/adk/pkg/mcp` failed with `http: CloseIdleConnections called`. Those parallel tests all shared `http.DefaultTransport`, and `httptest.Server.Close` closes the default transport's idle connections, so one test's teardown can kill another test's request. Fixed in 2d48cec by giving each test its own transport.

anthonytwh:
@supreme-gg-gg looking for a final review here

supreme-gg-gg (commented):
lgtm, `GenerateContentConfig` is the right configuration site, this is good to merge after you fix the merge conflict

anthonytwh:
@supreme-gg-gg merge conflicts fixed

supreme-gg-gg (approved):

PR #2149, opened 2026-07-04
feat(adk): expose gen_ai.client.token.usage Prometheus metric

braghettos:
## What

Adds GenAI token-usage instrumentation to the ADK agent, exposing the OTel-semconv metric **`gen_ai.client.token.usage`** via the **native Prometheus client library** (`github.com/prometheus/client_golang`) — matching kagent's existing Go metrics approach on the controller side. First of the three observability gaps in #2148.

> **Updated per review** ([#2149 review](https://github.com/kagent-dev/kagent/pull/2149#pullrequestreview-4637109986)): switched from the initial OTLP `MeterProvider` to the Prometheus client library, keeping OTel-semconv naming.

## How

- **`pkg/telemetry/metrics.go`** — a `gen_ai_client_token_usage` histogram (Prometheus name for the semconv metric `gen_ai.client.token.usage`) with labels, following GenAI semantic conventions (semconv 1.40.0), mapped to Prometheus naming (dots → underscores):
  - `gen_ai_token_type` — `input` / `output`
  - `gen_ai_request_model` — e.g. `gpt-4o`
  - `gen_ai_provider_name` — e.g. `openai`, `anthropic`
  - `gen_ai_operation_name` — `chat` (the chat-completion operation that produces these tokens)
- **`pkg/a2a/server/server.go`** — serves the metrics at **`/metrics`** on the agent's A2A mux (excluded from request tracing).
- **`pkg/a2a/executor.go`** — records token counts from the A2A event loop where `UsageMetadata` is surfaced: `input = PromptTokenCount`, `output = CandidatesTokenCount + ThoughtsTokenCount`.
- **`cmd/main.go` / `pkg/config`** — resolve the request model + provider labels from the agent config.

## Semconv alignment

Addresses the naming points raised on #1412: uses `gen_ai.provider.name` (not `gen_ai.system`), includes `gen_ai.request.model`, and sets `gen_ai.operation.name` (`chat`) rather than omitting it. Naming targets semconv 1.40.0.

## Testing

- `pkg/telemetry/metrics_test.go` — asserts the histogram records one series per token type, skips zero counts, and that the `/metrics` handler serves the series with the expected labels (via `prometheus/client_golang/testutil` + `httptest`).
- `go build ./adk/...`, `go test -race -skip 'TestE2E.*' ./adk/...`, and `golangci-lint run` all pass.

## Notes

- `prometheus/client_golang` is already a dependency; no new direct requirements (only a transitive `godebug` from the test util).
[...]

Issue #2148, opened 2026-07-04
[FEATURE] Close three observability gaps: GenAI token metrics, controller reconcile Events, and a controller log→OTLP bridge (all gated default-OFF)


braghettos:
**Affected Service(s):** Multiple services / System-wide (App Service + Controller Service)

## 📋 Prerequisites
- [x] I have searched the existing issues to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow the Code of Conduct

## 📝 Feature Summary

Add three small, independent, **default-OFF** observability improvements that build directly on the OTel pipeline landed in #1433 / #1292:

1. **GenAI token-usage metrics** in the ADK (`go/adk`) — the first `MeterProvider` in kagent, emitting the OpenTelemetry GenAI-semconv instrument `gen_ai.client.token.usage`.
2. **Kubernetes Events on reconcile** for the `Agent` / `ModelConfig` / `MCPServerTool` / `RemoteMCPServer` controllers (`go/core`), mirroring the pattern the harness/substrate controller already uses.
3. **Controller log → OTLP bridge** (`go/core`) — an `otelzap` tee on the controller zap core so logs can be exported over OTLP additively, without changing stdout.

Each is additive and byte-identical to current behavior when its gate is unset. Happy to split this into **three separate PRs** (they're fully independent) — filing one issue first to align on approach per CONTRIBUTING.

## ❓ Problem Statement / Motivation

kagent already has solid **tracing** (controller distributed tracing + A2A propagation, #1433) and standardized OTel env-var naming (#1292), but three observability primitives are still missing on `main`:

- **No metrics pipeline at all.** There is no `MeterProvider` anywhere in the repo (`go/adk/pkg/telemetry/` contains only `attributes.go` + `tracing.go`). Every model adapter already computes token usage — `go/adk/pkg/models/{openai,anthropic,bedrock,ollama,sapaicore}_adk.go` all populate `genai.GenerateContentResponseUsageMetadata` (`PromptTokenCount` / `CandidatesTokenCount` / `ThoughtsTokenCount`) — but that data is never recorded to a metric, so operators can't graph or alert on token spend / cost per model or provider without scraping traces. This is the still-unchecked "Metrics: LLM calls/token-count" item from the (now closed) observability EPIC #130, and the abandoned Prometheus attempt in #1412 (closed unmerged) shows the demand is real; this proposes a standards-aligned OTLP/semconv take instead of a scraped `/metrics` endpoint.
[...]

PR #2150, opened 2026-07-04
feat(core): emit Kubernetes Events on Agent/ModelConfig/MCPServerTool/RemoteMCPServer reconciles

braghettos:
## What

Adds Kubernetes Events on reconcile to the **Agent**, **ModelConfig**, **MCPServerTool**, and **RemoteMCPServer** controllers, so users can see reconcile outcomes with `kubectl describe` / `kubectl get events`. This is the second of the three observability gaps proposed in #2148.

Today these four controllers emit no Events (unlike `agentharness-substrate-controller`, which already does), so a bad Agent/ModelConfig/MCPServer gives no signal in `kubectl describe`.

## How

Each controller gains an optional `Client` + `Recorder events.EventRecorder` and a small nil-guarded `recordEvent` helper that re-fetches the reconciled object and emits an Event against it, mirroring the existing substrate-controller pattern.

Emitted events:

| Controller | Success | Failure |
|---|---|---|
| Agent | `Normal Accepted` | `Warning ReconcileFailed` |
| ModelConfig | `Normal Accepted` | `Warning ReconcileFailed` |
| RemoteMCPServer | `Normal Accepted` | `Warning ReconcileFailed` |
| MCPServerTool | `Normal ToolsDiscovered` | `Warning ValidationFailed` (spec validation) / `Warning ReconcileFailed` (transient) |

Recorders are wired from `pkg/app/app.go` via `mgr.GetEventRecorder(...)`.

## Additive / no behaviour change

- `recordEvent` is a **no-op** when either `Recorder` or `Client` is nil, so existing tests and any caller that doesn't wire them are unaffected.
- No new gate/env var — Events are a cheap, additive status surface.
- RBAC is already covered by the leader-election role (`events: create,patch`), so no RBAC changes are needed.
- No new dependencies (`go.mod` unchanged); `k8s.io/client-go/tools/events` and `apierrors` are already used in-tree.

## Testing

- New `mcp_server_tool_controller_events_test.go` — asserts the correct event type/reason is emitted for each of the three MCPServerTool outcomes (ToolsDiscovered / ValidationFailed / ReconcileFailed) using a fake client + `events.NewFakeRecorder`, plus a no-Recorder no-panic case.
- `go build ./core/...`, `go test -race -skip 'TestE2E.*' ./core/internal/controller/...`, and `golangci-lint run` all pass.

## Notes

Opening as **draft / work-in-progress** per CONTRIBUTING; happy to adjust event reasons/actions. The remaining gap from #2148 (controller log→OTLP bridge) will follow as a separate PR. Companion PR for gap 1 (token metrics): #2149.

Refs: #2148

### DIFF
--- a/go/core/internal/controller/agent_controller.go
+++ b/go/core/internal/controller/agent_controller.go
@@ -19,9 +19,11 @@ package controller
 import (
 	"context"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -44,6 +46,12 @@ type AgentController struct {
 	Scheme        *runtime.Scheme
 	Reconciler    reconciler.KagentReconciler
 	AdkTranslator agent_translator.AdkApiTranslator
+	// Client is used to fetch the reconciled object so Events can be emitted
+	// against it. Optional; event emission is skipped when nil.
+	Client client.Client
+	// Recorder emits Kubernetes Events on reconcile transitions. Optional;
+	// event emission is skipped when nil, keeping behaviour unchanged.
+	Recorder events.EventRecorder
 }
 
 // +kubebuilder:rbac:groups=kagent.dev,resources=agents,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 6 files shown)


braghettos:
Addressed the Copilot feedback about Event flooding: the MCPServerTool and RemoteMCPServer controllers requeue every 60s, so they now emit the **Normal success Event only on a meaningful transition** — a pending spec change (`generation != observedGeneration`, captured before reconcile) — instead of on every periodic refresh. Warning Events (ReconcileFailed / ValidationFailed) still emit each time. Added a test asserting no success Event on a periodic no-op refresh. The Agent/ModelConfig controllers are unaffected since they don't requeue periodically.

EItanya (commented):
Overall the functionality makes sense. However I'm wondering if we should add this directly to the `reconciler` since that code already does the `Get` required to ensure that the object exists.

braghettos:
Thanks @EItanya — good point. The reconciler does already `Get` the object and, in the status-update path, computes a `conditionChanged` flag and syncs `observedGeneration`, so it has a cleaner "meaningful transition" signal than the extra `Get` I added in the controller (`generation != observedGeneration`). Emitting there would drop the duplicate `Get` and let the Event fire exactly on a condition transition.

One wrinkle to flag before moving it: the **Event reason** is partly decided in the controller today. For example `MCPServerToolController` classifies the reconciler's error (`errors.As(err, *ValidationError)`) to choose `ValidationFailed` vs `ReconcileFailed`. To move Events fully into the reconciler, that classification (and the per-object reasons: `Accepted` / `ToolsDiscovered` / `ValidationFailed` / `ReconcileFailed`) would move there too, and the reconciler would need an `EventRecorder` injected (it's a single shared service across the four controllers).

I'm happy to do that — probably emitting from the reconciler's status-finalizing step where `conditionChanged` is already known, keyed off the condition reason. Want me to go ahead with that shape, or would you prefer to keep the recorder at the controller level (the more conventional controller-runtime placement) and just avoid the double `Get` some other way? Either works for me.

braghettos:
@EItanya rebased onto latest `main` (0 behind, CI re-running). Wanted to make the case for keeping Event emission at the controller layer, but I'm happy to move it if you'd still prefer otherwise.

**Why controller-level fits here:**
1. **There's already precedent** — `agentharness_substrate_controller.go` emits its Events directly from the controller via `r.Recorder.Eventf(...)`. Keeping these four consistent with that avoids two different conventions in the same package.
2. **The `KagentReconciler` interface stays signal-agnostic.** Its methods return only `error`; it has no notion of Event *reasons* or Normal-vs-Warning outcomes. Moving emission in would push presentation concerns (reason strings, the `ToolsDiscovered`/`ValidationFailed`/`ReconcileFailed` taxonomy) down into shared reconcile logic that several controllers share.
3. **Outcome classification already lives in the controller.** MCPServerTool distinguishes a *validation* failure from a *reconcile* failure at the controller boundary — that's where the Event reason is decided. The reconciler's `Get` ensures existence, but it doesn't classify the failure mode the Event needs.
4. The duplicate `Get` I added is only to read `generation`/`observedGeneration` for the "meaningful transition" gate — cheap (cached client) and scoped to the two periodically-requeuing controllers.

That said — if you'd rather centralize it in the reconciler (e.g. pass the recorder in and surface a typed outcome), I'll refactor it that way; just confirm the direction and I'll push. Otherwise this is ready for another look. Thanks!

PR #2151, opened 2026-07-04
feat(core): add controller log→OTLP bridge (gated OTEL_LOGGING_ENABLED)

braghettos:
## What

Adds an optional **controller log → OTLP bridge**: the controller's own zap logs can be exported over OTLP alongside the traces already emitted by `InitTracerProvider`, enabling log/trace correlation in the controller. This is the third and final observability gap proposed in #2148.

Today `go/core/internal/telemetry/` only sets up traces; the controller's zap logs are stdout-only, so they can't be shipped to an OTLP backend or correlated with spans.

## How

- **New `internal/telemetry/logging.go`:**
  - `InitLoggerProvider` builds an OTLP `LoggerProvider` via `autoexport.NewLogExporter` (same env-driven exporter/endpoint resolution as `InitTracerProvider`) and registers it as the global OTel logger provider.
  - `ControllerZapOpts` returns a `RawZapOpts(zap.WrapCore(zapcore.NewTee(core, bridgeCore)))` option that **additively tees** the stdout zap core with an `otelzap` bridge core bound to that provider — stdout logging is preserved, OTLP is added.
- **`tracing.go`:** extracted the shared `newTelemetryResource` helper so traces and logs carry identical resource attributes (no behaviour change to `InitTracerProvider`).
- **`pkg/app/app.go`:** initialize the logger provider *before* building the controller logger (so the bridge binds to the configured global provider), then add the tee via `ControllerZapOpts`.

## Gating (default-OFF, additive)

Enabled only when `OTEL_LOGGING_ENABLED=true` (the flag already registered in `pkg/env`). When unset, both `InitLoggerProvider` and `ControllerZapOpts` are no-ops and the controller logger is byte-identical to today.

## Testing

- New `logging_test.go` — asserts `ControllerZapOpts()` returns no options when disabled, returns a working teed logger when enabled (no panic with the default no-op global provider), and that `InitLoggerProvider`'s shutdown is a safe no-op when disabled.
- `go build ./core/...`, `go test -race ./core/internal/telemetry/...`, and `golangci-lint run` all pass.

## Notes

- Adds `go.opentelemetry.io/contrib/bridges/otelzap v0.19.0` (matches the existing `otel/log v0.20.0`) and promotes `otel/log` from indirect to direct; no other dependency or version changes.
- Opening as **draft / work-in-progress** per CONTRIBUTING. Companion PRs from #2148: token metrics #2149, reconcile Events #2150.

Refs: #2148

### DIFF
--- a/docs/controller-otlp-logs.md
+++ b/docs/controller-otlp-logs.md
@@ -56,10 +56,20 @@ service:
 export OTEL_LOGGING_ENABLED=true
 export OTEL_EXPORTER_OTLP_ENDPOINT=http://<collector>:4318
 export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
-# controller logs now appear in the collector's debug output as LogRecords,
-# with the same resource attributes (service.name=kagent-controller, service.version, ...)
-# as the controller's traces.
+# controller logs now appear in the collector's debug output as LogRecords.
 ```
 
-<!-- TODO: add a screenshot of the collector debug exporter showing controller LogRecords once a
-     cluster/collector is available to capture one. -->
+The controller's log records show up in the `debug` exporter with the shared resource attributes
+(`service.name=kagent-controller`, …) and the bridge scope:
+
+```text
+Resource attributes:
+     -> k8s.namespace.name: Str(kagent)
+     -> k8s.pod.name: Str(kagent-controller-...)
+     -> service.name: Str(kagent-controller)
+     -> service.version: Str(v0.0.0-demo)
+InstrumentationScope github.com/kagent-dev/kagent/go/core
+LogRecord  SeverityText: info   Body: Str(reconciling Agent)             { controller=agent, agent=k8s-agent, namespace=kagent }
+LogRecord  SeverityText: info   Body: Str(Agent reconciled successfully) { controller=agent, agent=k8s-agent }
+LogRecord  SeverityText: error  Body: Str(example error log)             { reason=demo }
+```
--- a/docs/controller-otlp-logs.md
+++ b/docs/controller-otlp-logs.md
@@ -1,0 +1,65 @@
+# Exporting controller logs over OTLP
+
+The kagent controller can export its own logs to an OpenTelemetry (OTLP) backend, in addition to
+stdout. This is useful for shipping controller logs to the same collector/backend as its traces.
+It is **off by default**.
+
+## Enabling
+
+Set the standard OTel environment variables on the controller:
+
+| Variable | Purpose |
+| --- | --- |
+| `OTEL_LOGGING_ENABLED=true` | turn on the OTLP log pipeline |
+| `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` | collector endpoint |
+| `OTEL_EXPORTER_OTLP_PROTOCOL` | `grpc` (default) or `http/protobuf` |
+
+Logs still go to stdout unchanged — the OTLP export is additive (a tee on the controller's zap core
+via the [otelzap bridge](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelzap)).
+
+### Severity
+
+The OTLP pipeline ships the **same levels as stdout**: a min-severity processor
+([`minsev`](https://pkg.go.dev/go.opentelemetry.io/contrib/processors/minsev)) is set to the
+controller's configured `--zap-log-level`, so enabling export at `error` level does not start
+shipping `info`/`debug` records.
+
+### Trace correlation (note)
+
+Records carry `trace_id`/`span_id` only when the log call passes the request `context.Context` as a
+field. controller-runtime's `logr` → `zapr` path does not thread the reconcile context into fields,
+so logs emitted via `log.FromContext(ctx)` are **not** automatically span-linked. Full automatic
+correlation would require a separate mechanism and is out of scope for this feature.
+
+## Testing it with an OTel Collector
+
+A collector with the `debug` exporter is the quickest way to confirm logs arrive:
+
+```yaml
+# otel-collector-config.yaml
+receivers:
+  otlp:
+    protocols:
+      grpc: { endpoint: 0.0.0.0:4317 }
+      http: { endpoint: 0.0.0.0:4318 }
+exporters:
+  debug: { verbosity: detailed }
+service:
+  pipelines:
+    logs:
+      receivers: [otlp]
+      exporters: [debug]
+```
+
+```bash
+# run the collector, then point the controller at it and enable logging
+export OTEL_LOGGING_ENABLED=true
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://<collector>:4318
+export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+# controller logs now appear in the collector's debug output as LogRecords,
+# with the same resource attributes (service.name=kagent-controller, service.version, ...)
+# as the controller's traces.
+```
+
+<!-- TODO: add a screenshot of the collector debug exporter showing controller LogRecords once a
+     cluster/collector is available to capture one. -->
--- a/docs/controller-otlp-logs.md
+++ b/docs/controller-otlp-logs.md
@@ -4,15 +4,27 @@ The kagent controller can export its own logs to an OpenTelemetry (OTLP) backend
 stdout. This is useful for shipping controller logs to the same collector/backend as its traces.
 It is **off by default**.
 
+> **Note — avoid double ingestion.** The OTLP export is *additive*: logs still go to stdout
+> unchanged. If you already collect the controller's stdout (e.g. a Fluent Bit / Vector / Loki
+> agent scraping pod logs), enabling this ships the same records a second time over OTLP and your
+> backend will ingest them twice. Enable it only if OTLP is your primary log path, or drop the
+> controller's stdout logs at your node agent.
+
 ## Enabling
 
-Set the standard OTel environment variables on the controller:
+Set the following environment variables on the **controller**:
 
 | Variable | Purpose |
 | --- | --- |
-| `OTEL_LOGGING_ENABLED=true` | turn on the OTLP log pipeline |
+| `KAGENT_CONTROLLER_OTLP_LOGS_ENABLED=true` | turn on the controller's OTLP log pipeline |
 | `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` | collector endpoint |
 | `OTEL_EXPORTER_OTLP_PROTOCOL` | `grpc` (default) or `http/protobuf` |
+
+> This switch is **decoupled from agent logging**. `OTEL_LOGGING_ENABLED` enables gen_ai
+> input/output logging on *agents* and is forwarded to agent pods; it does **not** control the
+> controller's own log export. The controller uses its own non-`OTEL_`-prefixed flag
+> (`KAGENT_CONTROLLER_OTLP_LOGS_ENABLED`) precisely so the two are configured independently. Via
+> Helm these are `otel.logging.enabled` (agents) and `otel.logging.controller.enabled` (controller).
 
 Logs still go to stdout unchanged — the OTLP export is additive (a tee on the controller's zap core
 via the [otelzap bridge](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelzap)).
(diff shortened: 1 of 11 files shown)


go/core/internal/telemetry/logging.go
```diff
@@ -0,0 +1,71 @@
...
+// the global OTLP LoggerProvider while preserving stdout logging. When disabled
+// it returns no options, leaving the logger byte-identical to upstream.
+//
+// InitLoggerProvider must be called first so the bridge core binds to the
+// configured global LoggerProvider.
+func ControllerZapOpts() []crzap.Opts {
+	if !env.OtelLoggingEnabled.Get() {
+		return nil
+	}
+	bridgeCore := otelzap.NewCore(loggerBridgeName,
```
krisztianfekete, line 80: I don't think this actually gets us log-trace correlation as-is, which is what you're trying to deliver?

`otelzap.Core` only adds trace_id/span_id to a record when a context.Context is passed as a zap field. controller-runtime logs through logr -> zapr, and `log.FromContext(ctx)` only threads values in via `WithValues`, so I am not sure these records contain spans linked to them.

go/core/internal/telemetry/logging.go
```diff
@@ -0,0 +1,71 @@
...
+		otelzap.WithLoggerProvider(logglobal.GetLoggerProvider()),
+	)
+	return []crzap.Opts{
+		crzap.RawZapOpts(zap.WrapCore(func(core zapcore.Core) zapcore.Core {
+			return zapcore.NewTee(core, bridgeCore)
```
krisztianfekete, line 85: The OTel log SDK puts severity filtering in a processor, not in the exporter or the bridge, so someone running at error level only expects error+ logs, but when they enable this flag it will ship all their info/debug logs to the backend. 

We can fix it via https://pkg.go.dev/go.opentelemetry.io/contrib/processors/minsev

go/core/internal/telemetry/tracing.go
```diff
@@ -38,6 +38,29 @@ func InitTracerProvider(ctx context.Context, serviceVersion string) (func(contex
...
+	))
+
+	return tp.Shutdown, nil
+}
+
+// newTelemetryResource builds the OTel resource shared by the controller's
+// signal pipelines (traces and logs), keeping their resource attributes
+// identical.
+func newTelemetryResource(ctx context.Context, serviceVersion string) (*resource.Resource, error) {
 	instanceID, err := os.Hostname()
```
krisztianfekete, line 63: Since `newTelemetryResource` is now called once by `InitTracerProvider` and once by `InitLoggerProvider`, and `os.Hostname()` falls back to a fresh `uuid.New()` on failure, it's possible that processes can end up with two different `service.instance.ids` across the two ttelemetry signals. 

Let's build the resource once in `app.go` and pass the same `*resource.Resource` into both initializers instead.

go/core/internal/telemetry/logging_test.go
```diff
@@ -0,0 +1,51 @@
...
+	t.Setenv("OTEL_LOGGING_ENABLED", "false")
+	if opts := ControllerZapOpts(); opts != nil {
+		t.Fatalf("expected nil opts when logging disabled, got %d", len(opts))
+	}
+}
+
+// TestControllerZapOpts_EnabledTeesLogger verifies that when OTEL_LOGGING_ENABLED
+// is set, an otelzap bridge option is returned and the resulting logger is usable
+// (the bridge tees onto the stdout core without panicking).
+func TestControllerZapOpts_EnabledTeesLogger(t *testing.T) {
```
krisztianfekete, line 31: Let's also test 

1. a record actually having trace_id when there's an active span, and
2. severity gating vs the stdout level

krisztianfekete (changes requested):
Thanks for the PR!
Added some comments, and just like in https://github.com/kagent-dev/kagent/pull/2149, we'd like to see docs/examples to set it up, and some screenshots showing that the changes have been tested in a real setup. An OTel collector + debugexporter is perfect for this purpose if you don't want to ingest the logs into a proper logging backend.

braghettos:
Thanks @krisztianfekete — great catches. Pushed changes:

- **Log-trace correlation**: you're right. `otelzap` only attaches `trace_id`/`span_id` when a log call passes a `context.Context` as a field, and controller-runtime's `logr`→`zapr` path doesn't thread the reconcile context into fields — so `log.FromContext(ctx)` records aren't automatically span-linked. Rather than claim something that doesn't hold, I've **reframed the feature as OTLP controller-log export** (ship controller logs to an OTLP backend alongside traces, same resource) and documented the correlation caveat in the code, tests, and docs. Automatic correlation for all controller logs would need a separate mechanism — happy to explore it as a follow-up if you think it's worth it. Added a test showing the span *is* attached when the ctx is passed as a field (and absent otherwise).
- **Severity**: moved filtering into a `minsev` processor tied to the controller's `--zap-log-level`, so enabling export at `error` level no longer ships `info`/`debug`. Test asserts sub-threshold records are dropped.
- **Duplicate `service.instance.id`**: the resource is now built once (`NewTelemetryResource`) in `app.go` and passed into both `InitTracerProvider` and `InitLoggerProvider`.
- **Docs**: added `docs/controller-otlp-logs.md` — enabling, severity behaviour, the correlation caveat, and an OTel collector + `debug` exporter setup for testing. (Screenshot placeholder left in the doc — I don't currently have a cluster/collector wired to capture one; happy to add.)
- **Copilot (os.Exit vs deferred shutdown)**: the telemetry shutdowns run on graceful termination (manager stop); the fatal startup `os.Exit` paths crash-fast without flushing (minimal buffered telemetry, error already on stderr). Noted this in a comment. Let me know if you'd prefer a `run() error` refactor so those paths flush too.

braghettos:
### Tested with an OTel collector + debug exporter ✅

Drove the log bridge against a local OTel Collector (`otlp` receiver → `debug` exporter, per your suggestion) with `OTEL_LOGGING_ENABLED=true`. The controller's log records arrive over OTLP with the shared resource (`service.name=kagent-controller`, same as its traces) and the bridge scope:

```text
Resource attributes:
     -> k8s.namespace.name: Str(kagent)
     -> k8s.pod.name: Str(kagent-controller-...)
     -> service.name: Str(kagent-controller)
     -> service.version: Str(v0.0.0-demo)
InstrumentationScope github.com/kagent-dev/kagent/go/core
LogRecord  SeverityText: info   Body: Str(reconciling Agent)             { controller=agent, agent=k8s-agent, namespace=kagent }
LogRecord  SeverityText: info   Body: Str(Agent reconciled successfully) { controller=agent, agent=k8s-agent }
LogRecord  SeverityText: error  Body: Str(example error log)             { reason=demo }
```

Stdout logging is unchanged (additive tee). Added this to `docs/controller-otlp-logs.md`.

krisztianfekete:
@braghettos I asked for a screenshot to ensure that the changes have been tested manually. The one in the docs seems to be a synthetic one, which is ~fine-ish (would be better to have a proper one there too), but please provide one on the PR as well. Thank you!

braghettos:
@krisztianfekete added a screenshot from a real manual run (captured from an actual terminal, not a render) — the controller log bridge exporting `LogRecord`s over OTLP to an OTel collector's `debug` exporter, with the shared resource (`service.name=kagent-controller`, real `service.instance.id`, SDK version) and the `github.com/kagent-dev/kagent/go/core` scope, real timestamps and attributes. Let me know if you'd like anything else on this one!

image (15)

go/core/internal/telemetry/logging_test.go
krisztianfekete: @braghettos, can you please also add an e2e test?

krisztianfekete (commented):
Thanks, LGTM, added one more comment about an e2e test.

braghettos:
@krisztianfekete added an end-to-end test in `4e71a00` — `TestInitLoggerProvider_E2E_ExportsControllerLogsOverOTLP`.

It stands up a real in-process **OTLP/gRPC `LogsService` collector**, points the actual `autoexport` OTLP exporter at it via the standard `OTEL_*` env vars, builds the controller logger through `ControllerZapOpts`, emits a log line, and asserts it round-trips over the wire as a `LogRecord` carrying the shared resource (`service.name=kagent-controller`) and the bridge scope (`github.com/kagent-dev/kagent/go/core`).

Unlike the existing unit tests (which use an in-memory `logtest` processor), this exercises the genuine pipeline end to end:
`zap core → otelzap bridge → LoggerProvider → minsev processor → batch processor → OTLP/gRPC exporter → collector`.

It's hermetic (random loopback port, no external infra) so it runs in normal CI. If you'd instead prefer a full kind-cluster e2e in `test/e2e` (deploy the controller with `OTEL_LOGGING_ENABLED` + a collector pod, trigger a reconcile, assert the logs land), happy to add that too — just let me know which scope you want.

krisztianfekete (commented):
Thanks, can you please rebase the PR?

braghettos:
@krisztianfekete rebased onto latest `main` (now 0 commits behind). Only conflict was in `go/go.mod` — took main's `grpc` bump and re-ran `go mod tidy`. Telemetry tests still green with `-race`. Ready when you are 🙏

docs/controller-otlp-logs.md
```diff
@@ -0,0 +1,75 @@
+# Exporting controller logs over OTLP
+
+The kagent controller can export its own logs to an OpenTelemetry (OTLP) backend, in addition to
```
krisztianfekete, line 3: > in addition to stdout

This can mean double ingestion if users already have a stdout-based agent, so we should call this out.

docs/controller-otlp-logs.md
```diff
@@ -0,0 +1,75 @@
...
+via the [otelzap bridge](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelzap)).
+
+### Severity
+
+The OTLP pipeline ships the **same levels as stdout**: a min-severity processor
+([`minsev`](https://pkg.go.dev/go.opentelemetry.io/contrib/processors/minsev)) is set to the
+controller's configured `--zap-log-level`, so enabling export at `error` level does not start
+shipping `info`/`debug` records.
+
+### Trace correlation (note)
```
krisztianfekete, line 39: I guess we could emit structured JSON to stdout with trace_id/span_id as fields so at least stdout-based agents/collectors can have this. Do you mind opening an issue for this?

docs/controller-otlp-logs.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Enabling
+
+Set the standard OTel environment variables on the controller:
+
+| Variable | Purpose |
+| --- | --- |
+| `OTEL_LOGGING_ENABLED=true` | turn on the OTLP log pipeline |
+| `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` | collector endpoint |
+| `OTEL_EXPORTER_OTLP_PROTOCOL` | `grpc` (default) or `http/protobuf` |
```
krisztianfekete: These are the same env vars that one can enable to switch to logging gen_ai input/outputs instead of spans in actual agents. The two should be decoupled and configurable individually. We should also expose these via Helm values.

krisztianfekete (changes requested):

braghettos:
Thanks @krisztianfekete — addressed all three in `f39b9ef` (rebased onto latest `main`):

**1. Decoupled controller log export from agent gen_ai logging** — this was a real coupling, not just naming: `collectOtelEnvFromProcess` forwards *every* `OTEL_*` var from the controller into agent pods, so `OTEL_LOGGING_ENABLED` did double duty (controller export **and** agent gen_ai I/O logging). The controller's export now uses its own **`KAGENT_CONTROLLER_OTLP_LOGS_ENABLED`** flag — intentionally not `OTEL_`-prefixed so it is *not* propagated to agents. `OTEL_LOGGING_ENABLED` keeps its agent-only meaning.

**Helm** — both are now individual values:
- `otel.logging.enabled` → `OTEL_LOGGING_ENABLED` (agents, gen_ai logging)
- `otel.logging.controller.enabled` → `KAGENT_CONTROLLER_OTLP_LOGS_ENABLED` (controller export)

**2. Double-ingestion caveat** — added a callout in the docs: since export is additive over stdout, anyone already scraping the controller's stdout (Fluent Bit/Vector/Loki) would ingest twice; enable only if OTLP is the primary path, or drop stdout at the node agent.

**3. Structured JSON stdout w/ `trace_id`/`span_id`** — opened **#2349** as the follow-up and linked it from the docs. Good idea — it gives stdout-based setups correlation without the OTLP pipeline.

CI re-running. Let me know if the Helm value shape works for you or you'd prefer a different layout.

PR #2153, opened 2026-07-05
fix(go-adk): add per-call session isolation for Agent tools

### LINKED ISSUE #2137: [BUG] Subagent session isolation
### 📋 Prerequisites

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

# EP: Per-call session isolation for Agent tools (`isolateSessions`)

## Goal

Let a declarative agent fan out **parallel, isolated** sub-agent calls. Today every
call the parent makes to a given sub-agent (Agent tool) reuses **one** A2A
`context_id`, so all calls land in a **single** sub-agent session and are processed
as one conversation instead of independent sessions.

Add an opt-in `isolateSessions` flag on Agent tools. When enabled, each call to that
sub-agent uses a **fresh `context_id`**, so every invocation runs in its own isolated
sub-agent session — the Go equivalent of ADK's `AgentTool` "fresh session/context per
invocation" (the pattern the Python coordinator/dispatcher uses for per-repository
context isolation).

## Problem (root cause)

`go/adk/pkg/tools/remote_a2a_tool.go` generates `lastContextID` **once**, at tool
construction (pod startup), and stamps it on **every** outbound A2A message:

```go
state := &remoteA2AState{ /* ... */ lastContextID: a2atype.NewContextID() }
// ...
message.ContextID = s.lastContextID   // same id for every call, forever
```

The worker derives its session id straight from that context id
(`go/adk/pkg/a2a/executor.go`: `sessionID := reqCtx.ContextID`). So N parallel calls
→ 1 shared worker session, serialized/interleaved rather than N parallel sessions.

Observed: a coordinator emitted 11 sub-agent calls in one model turn; all 11 collapsed
into a single worker session (11 A2A tasks in one context) instead of 11 isolated
sessions.

The Python low-level tool (`_remote_a2a_tool.py`) has the identical shared-context
behavior; the Python side only gets isolation because a custom dispatcher wraps the
worker in ADK `AgentTool`, which mints a fresh session per call. This EP brings the
same capability to the Go declarative runtime without a custom agent.

## Design

`isolateSessions` is a per-Agent-tool boolean.

- **`false` (default):** unchanged. One stable `context_id` per tool for its lifetime
  → session continuity for stateful sub-agents.
- **`true`:** `context_id` is minted per call via `a2atype.NewContextID()` → each call
  is an isolated sub-agent session; parallel fan-out no longer shares state/history.

Cross-conversation/turn continuity that stateful workers need does **not** depend on
the message `context_id`: it rides the `x-kagent-root-context-id` header
(`lineageHeadersInterceptor`), which stays stable per conversation regardless of this
flag. So isolation only changes which sub-agent *session* a turn is recorded in.

### UI linking

The executor pre-stamps outgoing `function_call` parts with a pre-known
`subagent_session_id` from a startup map (`subagentSessionIDs`, keyed by tool name).
With per-call ids there is no single pre-known id, so an isolated tool contributes
**no** map entry: the UI links each sub-agent card via the per-call
`subagent_session_id` returned in the tool's `function_response` (already consumed by
`AgentCallDisplay`).

## Changes

| File | Change |
|------|--------|
| `go/api/v1alpha2/agent_types.go` | Add `IsolateSessions *bool` to `Tool` (+ CEL validation: only for `type=Agent`) |
| `go/api/adk/types.go` | Add `IsolateSessions bool` (`isolate_sessions`) to `RemoteAgentConfig` |
| `go/core/internal/controller/translator/agent/compiler.go` | Copy `tool.IsolateSessions` → `RemoteAgentConfig.IsolateSessions` |
| `go/adk/pkg/agent/agent.go` | Pass flag to `NewKAgentRemoteA2ATool`; skip stamp-map entry when isolated |
| `go/adk/pkg/tools/remote_a2a_tool.go` | `isolateSessions` field; per-call `context_id`; thread it into `processResult` |
| `python/packages/kagent-adk/src/kagent/adk/types.py` | Add `isolate_sessions` field for config-schema parity (Go runtime honors it) |
| generated | CRD YAML + DeepCopy via `make controller-manifests`; golden files via `UPDATE_GOLDEN=true make -C go test` |

## Deployment / configuration

```yaml
kind: Agent
metadata:
  name: coordinator
spec:
  declarative:
    runtime: go
    tools:
      - type: Agent
        agent:
          name: worker
        isolateSessions: true   # each worker call = its own isolated session
```

- Unset / `false` → current behavior (one shared sub-agent session).
- `true` → fresh sub-agent session per call; enables parallel isolated fan-out.

## Scope / non-goals

- **Runtime:** honored by the **Go** declarative runtime. The field is accepted by the
  Python config model (schema parity) but the Python low-level tool is unchanged.
- **Not** a concurrency limiter — see `max-concurrency.md` for capping parallel
  invocations per pod. Isolation and concurrency limiting are complementary.
- No change to HITL resume: resume still targets the original sub-agent session via
  the `context_id` stored in the confirmation payload.

## Behavior notes & caveats

- Stateful sub-agents that rely on **reused** session history across calls must keep
  `isolateSessions: false` (or key state on `x-kagent-root-context-id`).
- With isolation on, a sub-agent's per-call sessions have `source='agent'` and are
  hidden from the worker's sidebar by design; they render inline in the parent chat.
- Parallelism is bounded by how many tool calls the parent LLM emits in a turn and by
  any `KAGENT_MAX_CONCURRENCY` on the worker pod.

### 🔄 Steps To Reproduce

1. Create WorkerAgent  
2. Create Coordinator Agent with WorkerAgent as Tool
3. Ask coordinator agent to call Worker with 3 tasks in parallel 

### 🤔 Expected Behavior

_No response_

### 📱 Actual Behavior

_No response_

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

yashrajshuklaaa:
## Summary

Fixes #2137

The bug : every call a parent agent makes to a sub-agent (Agent tool) reuses the same A2A **context_id** since it's minted once when the tool is built. worker just uses that **context_id** as its session id directly ( **executor.go: sessionID := reqCtx.ContextID** ) so if a coordinator fires off N parallel calls to the same sub-agent in one turn , they all pile into a single shared session instead of getting their own.

Fix is an opt-in **IsolateSessions** flag on Agent-type tools :

```yaml
spec:
  declarative:
    tools:
      - type: Agent
        agent:
          name: worker
        isolateSessions: true   # each call to worker gets its own session
```

Default behavior (flag unset/false) doesn't change  
one **context_id** for the tool's whole lifetime so stateful sub-agents keep their session continuity. with the flag on we mint a fresh **context_id** on every call so each invocation is isolated.

Doesn't touch the **x-kagent-root-context-id** header stuff. that's still what carries cross-turn continuity and it stays stable either way.

## What changed

- `agent_types.go`: added `Tool.IsolateSessions *bool` plus a CEL rule so it can only be set when `type: Agent`
- `adk/types.go`: `RemoteAgentConfig.IsolateSessions bool`
- `compiler.go`: passes the flag through into `RemoteAgentConfig`
- `agent.go`: forwards the flag to `NewKAgentRemoteA2ATool`. also skips adding an entry to `subagentSessionIDs` when isolated since there's no single session id to pre-stamp function_call parts with anymore
- `remote_a2a_tool.go`: added `isolateSessions` plus a small `nextContextID()` helper that either returns the stable id or mints a new one. This gets reported back as `subagent_session_id`
- Python `types.py`: added `isolate_sessions` for schema parity only, it doesn't do anything on that side yet (matches what the issue scoped out)
- Regenerated CRDs/deepcopy with `make controller-manifests`
- Added a test for the new context-id logic and a golden fixture (`agent_with_isolated_session_tool`) to check it flows through the whole translation pipeline

## About the UI

Isolated tools don't have one fixed session id, so the executor can't pre-populate the stamp map for them at startup. Instead the UI grabs the session id per call from `subagent_session_id` in the function_response. `AgentCallDisplay` already reads that field, so nothing new needed there.

## Not doing in this PR

- Not a concurrency limiter, that's separate (`max-concurrency.md`)
- HITL resume is unaffected, it still uses the context_id from the confirmation payload
- Only the Go runtime respects this flag right now. Python accepts it for config parity but the low-level tool doesn't use it yet

## Tests

Ran `go test ./adk/pkg/tools/... ./adk/pkg/agent/... ./api/...` and the golden translator tests. Everything passes. 
Confirmed `isolate_sessions: true` shows up correctly in the generated `config.json` for the new fixture.

## One small thing I noticed

In `handleResume`, `processResult` now sets `subagent_session_id` from the contextID I pass in, but there's older code right after it that sets the same key again with a fallback to `lastContextID`. Not wrong, just a bit redundant now since it writes the same value twice in the normal case. Left it as is since it's harmless but flagging it in case someone wants it cleaned up.


yashrajshuklaaa:
PTAL @EItanya

go/adk/pkg/tools/remote_a2a_tool.go
```diff
@@ -358,7 +392,7 @@ func (s *remoteA2AState) processResult(ctx adkagent.ToolContext, result a2atype.
```
mesutoezdil, line 357: message responses from handleFirstCall never include subagent_session_id because processResult does not set it here, unlike the task case.

go/adk/pkg/tools/remote_a2a_tool.go
```diff
@@ -322,7 +351,7 @@ func (s *remoteA2AState) handleResume(ctx adkagent.ToolContext) (map[string]any,
 		return map[string]any{"error": fmt.Sprintf("Remote agent '%s' resume failed: %v", subagentName, err)}, nil
 	}
 
-	ret, retErr := s.processResult(ctx, result)
+	ret, retErr := s.processResult(ctx, contextID, result)
 	// Prefer the context_id from the confirmation payload (the original subagent
```
mesutoezdil: when isolateSessions is true and hitlPayload.ContextID is empty, the fallback s.lastContextID is a construction-time id that was never used in any actual a2a call.

EItanya (commented):
Thanks so much for the contribution, however I did find a bug we need to fix and clean up

I think the current approach fixes the session-collision issue, but it also highlights an existing design smell: we have two different mechanisms for linking an AgentCall card to subagent activity.

Today we pre-generate a subagent `context_id` when constructing the remote tool, return that from `NewKAgentRemoteA2ATool`, store it in `subagentSessionIDs`, and stamp it onto the `function_call` metadata so the UI can show the nested Activity panel. With `isolateSessions`, that model no longer fits because the actual subagent session is per invocation, not per tool instance. Returning `""` from the constructor for isolated tools is a sign that the constructor-time session ID is the wrong abstraction.

I’d suggest simplifying this so both shared and isolated modes use the same UI path:

1. Stop treating the constructor-time context ID as UI metadata.
2. Don’t return a subagent session ID from `NewKAgentRemoteA2ATool`.
3. Don’t special-case isolated tools in `CreateGoogleADKAgentWithSubagentSessionIDs`.
4. Always return the actual invocation’s subagent session ID from the tool result:

```go
return remoteA2AResponse{
    Result:            text,
    SubagentSessionID: contextID,
}
```

5. Also include it for `input_required` / HITL:

```go
return remoteA2AResponse{
    Status:            "pending",
    WaitingFor:        "subagent_approval",
    Subagent:          s.name,
    SubagentSessionID: task.ContextID,
}
```

Then the UI can use one source of truth for both modes: `function_response.response.subagent_session_id`. That ID is the actual remote A2A `context_id` used by the subagent session.

This also lets the Go code be simpler:

```go
type remoteA2AState struct {
    // ...
    sharedContextID string
    isolateSessions bool
}

func (s *remoteA2AState) contextIDForCall() string {
    if s.isolateSessions {
        return a2atype.NewContextID()
    }
    return s.sharedContextID
}
```

`lastContextID` should probably be renamed to `sharedContextID`, since it is not really “last”; in shared mode it is a stable session ID, and in isolated mode it should not be used.

While touching this path, I’d also prefer replacing the ad-hoc `map[string]any` tool responses with a typed response struct. Right now fields like `result`, `error`, `status`, `waiting_for`, `subagent`, `subagent_session_id`, and `kagent_usage_metadata` are implicit string keys spread across several branches. That makes it easy to forget `subagent_session_id` in one path, which is exactly the kind of issue this PR risks for `input_required`.

Something like:

```go
type remoteA2AResponse struct {
    Result              string         `json:"result,omitempty"`
    Error               string         `json:"error,omitempty"`
    Status              string         `json:"status,omitempty"`
    WaitingFor          string         `json:"waiting_for,omitempty"`
    Subagent            string         `json:"subagent,omitempty"`
    SubagentSessionID   string         `json:"subagent_session_id,omitempty"`
    KAgentUsageMetadata map[string]any `json:"kagent_usage_metadata,omitempty"`
}
```

Each path can return the same typed shape, and `subagent_session_id` becomes an explicit field rather than a magic map key. That would make this change safer and easier to review.

The one behavior change is that the nested Activity panel would appear once the tool response/pending response exists, rather than being pre-linked from the initial `function_call` event. I think that is the cleaner and more honest model. If we need live subagent activity before a response exists, that should be modeled explicitly with a per-invocation “tool started”/running event that carries the generated context ID. Constructor-time pre-stamping only works for shared sessions and is exactly what breaks down here.

yashrajshuklaaa:
Thanks for the thorough review, makes sense. Pushed a follow up commit for this:

- NewKAgentRemoteA2ATool doesn't return a session id anymore. Dropped the constructor time context_id as UI metadata entirely, per your suggestion.
- Every response path now sets SubagentSessionID consistently (Message, completed Task, input_required, failed Task, and the unrecognised result fallback), using a typed remoteA2AResponse struct instead of ad hoc map[string]any literals. functiontool.New is generic so it infers the output schema straight from the struct, no manual marshaling needed.
- Renamed lastContextID to sharedContextID and nextContextID() to contextIDForCall(), matching what you proposed.
- agent.go no longer pulls a session id from the constructor. Left the subagentSessionIDs/stampSubagentSessionID plumbing in the executor as is (just always gets an empty map now) with a comment explaining why, instead of ripping it out. Felt like a bigger cleanup outside this issue's scope tbh. Happy to do that in a follow up PR if you'd rather see it gone now.
- Checked the UI side before touching this, ToolCallDisplay.tsx already falls back to reading subagent_session_id from the function_response when it's missing from the function_call's DataPart metadata (~line 294). So no UI changes needed. Only behavior change is the Activity panel links once a response/pending state exists instead of pre-linking from the initial function_call, which matches what you described as the cleaner model.

Also added TestProcessResult_SetsSubagentSessionIDOnEveryBranch to cover the branches that were missing the field before.

go/adk/pkg/tools/remote_a2a_tool.go
yashrajshuklaaa: Fixed, processResult's Message branch sets SubagentSessionID now too, same as every other branch

go/adk/pkg/tools/remote_a2a_tool.go
yashrajshuklaaa: Good catch, removed that fallback. handleResume just delegates to processResult with the payload's contextID now, so no more case where an isolated tool falls back to a construction time id that was never actually used in any real A2A call.

dimetron:
Thank you @yashrajshuklaaa  for taking care of https://github.com/kagent-dev/kagent/issues/2137 some additional changes I've discovered during my testing

## Additional Required changes

### 1. [required] Response struct is missing failure-linkage and structured fields

The PR struct carries only `result/error/status/waiting_for/subagent/
subagent_session_id/kagent_usage_metadata`. For failed/canceled/rejected
children the parent also needs:

- `subagent_task_id`, `subagent_task_state` — link to the exact failed task.
- `error_type`, `error_step`, `error_message`, `remediation_hint` — child-provided
  structured failure passed through instead of buried in text, so the coordinator
  can act without re-parsing.

Without these, a failed sub-agent is not debuggable from the parent session
(the original motivation for the fan-out investigation). Add them to the struct
and populate on the terminal-failure branch.

### 2. [required] `processResult` must switch explicitly on terminal state

The PR keeps the old shape where any `*Task` state other than `input_required`
and `failed` falls through to "completed". A `working`/`submitted` task after the
call ends is **not** success. Switch explicitly:
`completed` → result; `failed`/`canceled`/`rejected` → typed error;
`input_required` → HITL; anything else → `subagent_non_terminal` error.

dimetron:
@yashrajshuklaaa just for the reference https://github.com/kagent-dev/kagent/pull/2202

yashrajshuklaaa:
These are good additions, thanks for testing it out @dimetron. I think it's a separate concern from what #2137 asked for (session isolation vs structured failure reporting) so my plan is to pick these up in a focused follow up PR once this one lands and link back to this thread when I open it.
@EItanya curious what you think too , happy to fold it into this PR instead if you'd rather see it land together

yashrajshuklaaa:
Also keeps git log/blame cleaner, a commit called "add per-call session isolation" quietly also adding a failure reporting schema would be confusing to trace back later.

yashrajshuklaaa:
@EItanya PTAL

supreme-gg-gg (commented):
@yashrajshuklaaa the changes look good overall, pls resolve the merge conflict and I'll take another look / approve. A few follow ups:

1. I agree we need to address @dimetron's comments regarding terminal state handling and failure fields
2. I think it's a fair tradeoff to drop constructor-time pre-stamping of session ID for a single response-based UI path, with that, we should get rid of the dead plumbings like `stampSubagentSessionID`, `subsagentSessionIDs` threading in `agent.go, executor.go, converter.go`

There's a regression regarding how HITL activity is supposed to work that is quite important to address. Without pre-stamped IDs, subagent activity panels will not open until a `function_response` supplies the session ID. HITL events surfaced from subagents returns this `function_response` containing `status: "pending" + subagent_session_id`, but it is currently skipped in the UI since it wasn't needed before due to pre-stamping IDs. One of the key motivations of having subagent activity viewing is for the user to know what the subagent is doing before approving its tool calls, which no longer works.

https://github.com/kagent-dev/kagent/blob/eed7287b3e8280b70239fc6c988ced48d97641ea/ui/src/lib/messageHandlers.ts#L940-L950

A simple fix that I came up with is to check if it comes from a pending subagent and do not skip it: 

```diff
             } else if (partType === "function_response") {
               const responseData = (data as { response?: Record<string, unknown> })?.response;
               const responseStatus = responseData?.status as string | undefined;
-              if (responseStatus === "confirmation_requested" || responseStatus === "pending") {
+              const toolData = data as unknown as ToolResponseData;
+              const isPendingAgentSession =
+                responseStatus === "pending" &&
+                isAgentToolName(toolData.name) &&
+                typeof responseData?.subagent_session_id === "string";
+              if (
+                (responseStatus === "confirmation_requested" || responseStatus === "pending") &&
+                !isPendingAgentSession
+              ) {
                 continue;
               }
-              const toolData = data as unknown as ToolResponseData;
               const source = getSourceFromMetadata(adkMetadata, defaultAgentSource);
               processFunctionResponsePart(toolData, statusUpdate.contextId, statusUpdate.taskId, source);
             }
```

I've tried this out and it preserves the old behaviour where when a subagent requires an approval, the user can view its activity.

EItanya:
Hey there, unit tests are failing

yashrajshuklaaa:
@supreme-gg-gg @EItanya thanks for the review!

yashrajshuklaaa:
@supreme-gg-gg @EItanya removed the dead plumbing
**stampSubagentSessionID** / s**tampSubagentSessionIDOnDataPart** gone from converter.go, SubagentSessionIDs dropped from executor.go, **CreateGoogleADKAgentWithSubagentSessionID**s now just returns (agent.Agent, error). build/vet/test all passing. PTAL

go/adk/pkg/agent/agent.go
```diff
@@ -43,11 +43,11 @@ func CreateGoogleADKAgent(ctx context.Context, agentConfig *adk.AgentConfig, age
 // outbound A2A events). Callers that only need the agent can use
 // CreateGoogleADKAgent.
 // Optional stsPlugin can be provided for token propagation to MCP tools.
-func CreateGoogleADKAgentWithSubagentSessionIDs(ctx context.Context, agentConfig *adk.AgentConfig, agentName string, stsPlugin *sts.TokenPropagationPlugin, extraTools ...tool.Tool) (agent.Agent, map[string]string, error) {
+func CreateGoogleADKAgentWithSubagentSessionIDs(ctx context.Context, agentConfig *adk.AgentConfig, agentName string, stsPlugin *sts.TokenPropagationPlugin, extraTools ...tool.Tool) (agent.Agent, error) {
```
supreme-gg-gg: Can you rename this to `CreateGoogleADKAgent` since subagent session IDs are removed?

supreme-gg-gg (changes requested):
This looks great! One nit and this is good to merge

go/adk/pkg/agent/agent.go
yashrajshuklaaa: Done : )

supreme-gg-gg (approved):
lgtm, thanks for the PR!

EItanya:
This will need to be backported to make it into 0.10

PR #2165, opened 2026-07-06
feat(helm): add UI HTTPRoute

younsl:
## Summary

Adds an optional Gateway API `HTTPRoute` for the UI. It is off by default so existing installs are unaffected.

- `ui.httpRoute`: lets users front the UI with a Gateway API implementation such as kgateway or Istio or Envoy Gateway instead of the OpenShift `Route` or the bundled oauth2-proxy.

### HTTPRoute (`ui.httpRoute`)

- New template `templates/ui-httproute.yaml` gated behind `ui.httpRoute.enabled` (default `false`).
- Supports `parentRefs` and `hostnames` and per-rule `matches` / `filters` / `timeouts`.
- When a rule omits `backendRefs` it defaults to the UI `Service` on `ui.service.ports.port`.

```gotemplate
{{- if .Values.ui.httpRoute.enabled }}
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
...
{{- end }}
```

## Testing

Added helm-unittest suite `tests/ui-httproute_test.yaml`.

```
helm unittest helm/kagent
```

## References

- Kubernetes Gateway API concept: https://kubernetes.io/docs/concepts/services-networking/gateway/
- Gateway API project docs: https://gateway-api.sigs.k8s.io/
- HTTPRoute resource: https://gateway-api.sigs.k8s.io/api-types/httproute/

## Notes

- No change to existing rendered output when `ui.httpRoute.enabled` stays at its default.


EItanya (changes requested):
Hey there, I think there are a mixture of concerns here that I'd like to tease apart. 
1. The HTTPRoute seems generally useful and I'd like to discuss the pros/cons
2. The arbitrary extra objects is unrelated and should be split up into a separate PR to debate on its own merit

younsl:
Thanks for the review. Agreed on splitting these up. This PR now only contains the HTTPRoute change. I dropped the extraObjects commit and will open a separate PR for it so it can be debated on its own merit. Happy to discuss the pros and cons of the HTTPRoute here.

EItanya (approved):
I got signoff from gw-api expert @keithmattix

PR #2168, opened 2026-07-06
Add kagent db migrate CLI for out-of-band migrations

iplay88keys:
## Description

Adds a `kagent db migrate` command group for running and inspecting database migrations out-of-band from server startup.

```
kagent db migrate up                    # apply all pending migrations, every registered source
kagent db migrate status [--output json] # aggregate + per-source applied/pending breakdown
kagent db migrate version               # per-source current versions (--source filters)
kagent db migrate down N  --source core # roll back the N most-recent migrations on one source
kagent db migrate goto V  --source core # move one source to version V (0 = empty schema)
kagent db migrate force V --source core # assert a version without running SQL (recovery)
```

- New exported packages `go/core/pkg/cli/db` and `go/core/pkg/cli/db/migrate`; `NewCommand(sources ...migrations.Source)` takes the source list explicitly so downstream consumers can wire their own sources.
- The connection comes from `--db-url`, falling back to the controller's `POSTGRES_DATABASE_URL` env var. The vector track is gated on the controller's `DATABASE_VECTOR_ENABLED` env var (default enabled).
- New `migrations.WithMigrator` export: the CLI opens migrators through the orchestrator's own schema handling, tracking-table config, and advisory-lock identity, so a CLI invocation racing a booting server serializes.
- `up`, `down`, and `goto` refuse a dirty source and report the `force` invocation that clears it; the CLI never auto-recovers (startup remains the automatic tier). `status`/`version` report dirty state instead of refusing.
- `force 0` clears the version record entirely (recovery when a track's first migration fails dirty); any other value must be a shipped migration version.
- `down` on an empty track reports "no migrations to roll back" instead of golang-migrate's raw "file does not exist".
- `status --output json` has a frozen field shape locked by `TestStatusJSONShape`.
- Help output for the `db` subtree hides the root's server-oriented persistent flags (`--kagent-url`, `-n`, `-o`, `--timeout`, etc.); db commands talk to Postgres directly.

## Testing

Version:
```
❯ ./go/core/bin/kagent-local db migrate version
core: 6
vector: 3
```

Fresh install:
```
❯ ./go/core/bin/kagent-local db migrate status
9 migration(s) applied, 0 pending
  core: 6 applied (at v6), 0 pending
  vector: 3 applied (at v3), 0 pending

kagent=# \dt
                   List of tables
 Schema |           Name           | Type  | Owner
--------+--------------------------+-------+--------
⋮
(17 rows)
```

Down (Rollback):
```
kagent=# \dt
                   List of tables
 Schema |           Name           | Type  | Owner
--------+--------------------------+-------+--------
⋮
(17 rows)

❯ ./go/core/bin/kagent-local db migrate down 1 --source core 
rolled back 1 migration(s)

❯ ./go/core/bin/kagent-local db migrate status              
8 migration(s) applied, 1 pending
  core: 5 applied (at v5), 1 pending
  vector: 3 applied (at v3), 0 pending

kagent=# \dt
                   List of tables
 Schema |           Name           | Type  | Owner
--------+--------------------------+-------+--------
⋮
(15 rows)
```

Goto:
```
❯ ./go/core/bin/kagent-local db migrate goto 1 --source core
schema is at version 1

kagent=# \dt
                   List of tables
 Schema |           Name           | Type  | Owner
--------+--------------------------+-------+--------
 public | memory                   | table | kagent
 public | schema_migrations        | table | kagent
 public | vector_schema_migrations | table | kagent
(3 rows)
```

Force:
```
kagent=# select * from schema_migrations;
 version | dirty
---------+-------
       6 | f
(1 row)

❯ ./go/core/bin/kagent-local db migrate force 1 --source core 
version 1 marked as applied

kagent=# select * from schema_migrations;
 version | dirty
---------+-------
       1 | f
(1 row)
```

Goto + Up:
```
❯ ./go/core/bin/kagent-local db migrate goto 1 --source core
schema is at version 1

❯ ./go/core/bin/kagent-local db migrate goto 1 --source vector
schema is at version 1

❯ ./go/core/bin/kagent-local db migrate status                
2 migration(s) applied, 7 pending
  core: 1 applied (at v1), 5 pending
  vector: 1 applied (at v1), 2 pending

❯ ./go/core/bin/kagent-local db migrate up                    
applied 7 migration(s); schema is up to date
```


go/core/cli/cmd/kagent/main.go
```diff
@@ -449,11 +452,37 @@ Examples:
...
+// newDBCommand builds the `kagent db` command over the built-in migration
+// tracks. The vector track is gated on the same DATABASE_VECTOR_ENABLED env
+// var the controller reads for --database-vector-enabled, with the same
+// default (enabled), so the CLI operates on the tracks the server migrates.
+// An invalid value is reported only when a db subcommand actually runs, so
+// unrelated commands aren't polluted by warnings during construction.
+func newDBCommand() *cobra.Command {
+	vectorEnabled := true
+	var envWarning string
+	if v := os.Getenv("DATABASE_VECTOR_ENABLED"); v != "" {
```
EItanya: Can we somehow read this from the live cluster rather than requiring information from the user. We can use this as a fallback of course

iplay88keys: Good question, I will look into that.

iplay88keys: I have a way to look up the kagent's config. The only issue is that if the kubeconfig is pointing to a Kagent which is attached to a different postgres than you are connecting to via `--db-url`, the looked-up value of `VECTOR_ENABLED` could silently differ from what's in postgres. I can't detect a mismatch, but I can print it in the cli output as stderr:
```
❯ go run ./core/cli/cmd/kagent db migrate status   
resolved vector track from cluster context "kind-cluster": configmap kagent/kagent-controller has DATABASE_VECTOR_ENABLED=false (set DATABASE_VECTOR_ENABLED to override)

9 migration(s) applied, 0 pending
  core: 6 applied (at v6), 0 pending
  vector: 3 applied (at v3), 0 pending
```

go/core/cli/cmd/kagent/main.go
```diff
@@ -452,35 +456,73 @@ Examples:
...
+// subcommand runs (never during command construction, so unrelated commands
+// do no work and print no warnings). The vector track is gated, in order of
+// precedence, on: the DATABASE_VECTOR_ENABLED env var (explicit operator
+// intent, works without a cluster), the controller's configmap on the live
+// cluster (the same value the server reads), and finally the controller's
+// default (enabled).
+func migrationSources(cfg *config.Config) dbmigrate.SourcesFunc {
+	return func(ctx context.Context) ([]migrations.Source, error) {
+		vectorEnabled := true
+		if v := os.Getenv("DATABASE_VECTOR_ENABLED"); v != "" {
```
EItanya: Why do you have this and the check inside of `clusterVectorEnabled`

iplay88keys: They are technically different sources (env var override for the cli vs the deployed controller's configmap key). The env var wins when both are set. I've made that more clear in the code.

EItanya (approved):

PR #2171, opened 2026-07-07
feat(sandbox-agents): Store session state in `durableDir` volume for declarative sandbox agents

jmhbh:
**Overview**
- Sandbox Agents using the `python` and `go` declarative runtimes are now configured by default to use a local store for session data. By default, the sqlite db now points to the Sandbox Agent's `durableDir` volume. 
- Session metadata is mirrored to the postgres database to support APIs that fetch session information ie: `GET /sessions/{id}`. 
- A new `KAGENT_SESSION_DB_URL` was added to control  the `go` and `python` declarative runtimes behavior to choose which session service type. 
- local session store for BYO agents are explicitly not supported since it requires the BYO agents user to implement support for local session store. If BYO agents implement local session store and would like to leverage this feature using the same aforementioned env var, the `kagent.dev/local-session-storage` annotation has been added to override existing behavior.

**Testing**
- Tested both python and go runtimes using the sandbox agent CR across multiple turns with config rollout (the change was to the prompt to tell the agent to respond in french) in between turns to ensure data history remains intact across config rollout.

Python runtime sqlite db history:
[python-runtime-sql.md](https://github.com/user-attachments/files/29812074/python-runtime-sql.md)
Screenshot 2026-07-08 at 1 53 08 PM

Go runtime sqlite db history
[go-runtime-sql.md](https://github.com/user-attachments/files/29812091/go-runtime-sql.md)
Screenshot 2026-07-08 at 1 52 57 PM


jmhbh:
WIP, still need to add support for the go runtime.

go/adk/cmd/main.go
```diff
@@ -150,8 +150,21 @@ func main() {
...
 	// The executor needs a session service for its BeforeExecute callback
 	// (session creation/lookup). This must be created before the executor.
-	var sessionService *session.KAgentSessionService
-	if kagentURL != "" {
+	// KAGENT_SESSION_DB_URL selects the durable-dir local store (substrate sandbox agents:
+	// session state lives in a sqlite DB inside the actor's durableDir volume); otherwise
+	// session state is stored in the postgres database.
+	var sessionService session.Service
+	var localSessions *session.LocalSessionService
+	if sessionDBURL := os.Getenv("KAGENT_SESSION_DB_URL"); sessionDBURL != "" {
```
EItanya: Can this be in the config struct? That way it could also work for BYO

go/core/internal/a2a/substrate_sandbox_transport.go
```diff
@@ -94,6 +107,31 @@ func (t *substrateSandboxSessionRoundTripper) RoundTrip(req *http.Request) (*htt
 	return resp, nil
 }
 
+// ensureSessionRow materializes the controller-side session row for sessions that arrive via
+// raw A2A. Ensures that headless sessions have a row.
```
EItanya: What's a headless session?

go/core/internal/a2a/substrate_sandbox_transport.go
```diff
@@ -94,6 +107,31 @@ func (t *substrateSandboxSessionRoundTripper) RoundTrip(req *http.Request) (*htt
...
+func (t *substrateSandboxSessionRoundTripper) ensureSessionRow(ctx context.Context, sessionID, userID string) error {
+	if t.db == nil || !substrate.SandboxAgentUsesDurableDirSessions(t.sandboxAgent) {
+		return nil
+	}
+	agentID := utils.ConvertToPythonIdentifier(t.sandboxAgent.Namespace + "/" + t.sandboxAgent.Name)
+	sessions, err := t.db.ListSessionsForAgentAllUsers(ctx, agentID)
```
EItanya: Why aren't we calling for the user session?

EItanya (commented):
A few questions, but looks good overall

go/core/internal/a2a/substrate_sandbox_transport.go
jmhbh: when an a2a requests is made without going through the UI

go/core/pkg/sandboxbackend/substrate/agent_lifecycle_test.go
```diff
@@ -262,6 +262,88 @@ func TestBuildSandboxAgentActorTemplate(t *testing.T) {
...
+			require.Equal(t, "/health", c.Readyz.HTTPGet.Path)
+			require.Equal(t, substrateKagentListenPort, c.Readyz.HTTPGet.Port)
+			var dbURL string
+			for _, e := range c.Env {
+				if e.Name == sessionDBURLEnv {
+					require.NotNil(t, e.Value)
+					dbURL = *e.Value
+				}
+			}
+			require.Equal(t, tc.wantDBURL, dbURL)
```
mesutoezdil: thse tests fail as-is: testbuildsandboxagentactortemplatedurabledirsessions and the KAGENT_SESSION_DB_URL check in TestBuildSandboxAgentActorTemplate all get epty string for dbURL insted of the expected URL. 

probabyl bcs of the local env slice (which appends KAGENT_SESSION_DB_URL) is never passed to the Env: field in the ActorTemplateSpec -- append(containerEnv, kagentContainer.Env...) is used there instead?

go/core/internal/a2a/substrate_sandbox_transport.go
jmhbh: there's no reason not to get session by user id here. Fixed

go/core/pkg/sandboxbackend/substrate/agent_lifecycle_test.go
jmhbh: yup, thanks I fixed this

mesutoezdil: > yup, thanks I fixed this

thx for writing your own sentence :) one misses human interaction.
every pr in this repo has turned into a total llm dumping ground today :) crazy

mesutoezdil: dco is red

jmhbh: thanks, I'll get that passing as well just need to circle back with Eitan on the config comment above. Not sure if he is referring to the kagent config secret that we mount to declarative agents or something else

go/adk/pkg/session/local.go
```diff
@@ -0,0 +1,106 @@
...
+	"gorm.io/gorm"
+)
+
+// Service is the session surface kagent wires through the runtime:
+// Implemented by KAgentSessionService (HTTP → postgre database)
+// and LocalSessionService (sqlite in the actor's durableDir volume).
+type Service interface {
+	adksession.Service
+	GetSession(ctx context.Context, appName, userID, sessionID string) (adksession.Session, error)
+	CreateSession(ctx context.Context, appName, userID string, state map[string]any, sessionID string) error
```
EItanya: I looked around through the code and I don't understand why we have these. They're only used by one caller, I don't think we should keep them on the interface.

go/core/internal/a2a/substrate_sandbox_transport_test.go
```diff
@@ -57,6 +63,107 @@ func (t *trackingReadCloser) Close() error {
 	return nil
 }
 
+// fakeSessionDB stubs the two database.Client methods ensureSessionRow uses; the embedded nil
+// interface panics on anything else.
+type fakeSessionDB struct {
+	database.Client
```
EItanya: Why does this need a db client?

EItanya: Can we use a fake postgres with a testcontainer?

go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -87,7 +87,9 @@ func (a *adkApiTranslator) BuildManifest(
 	if err != nil {
 		return nil, err
 	}
-	outputs.Manifest = append(outputs.Manifest, configSecret.secret)
+	if !manifestCtx.runInSandbox() {
+		outputs.Manifest = append(outputs.Manifest, configSecret.secret)
+	}
```
EItanya: I thought we were always adding the config secret? Or we're doing it via env var for now?

go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
```diff
@@ -63,71 +85,131 @@ func (p *Lifecycle) buildSandboxAgentActorTemplate(
...
+	sum := sha256.Sum256(raw)
+	return fmt.Sprintf("%x", sum[:8]), nil
+}
+
+// SandboxAgentUsesDurableDirSessions reports whether the agent's ADK session state lives in a
+// sqlite DB in a durableDir volume inside the session actor instead of the controller database.
+// This is how ALL declarative substrate agents (python and go runtimes) store session state.
+// Since storing session state is a runtime implementation, it has to be explicitly supported by
+// BYO agents by setting the kagent.dev/local-session-storage annotation to true.
+func SandboxAgentUsesDurableDirSessions(sa *v1alpha2.SandboxAgent) bool {
```
EItanya: Why go through this song and dance and not just do it automatically? The user has to create a new `SandboxAgent` anyway and we can explain what they need to do. I don't want to create a new API unless we absolutely haave to

jmhbh: originally, this was so we would only include the db session config for declarative agents but given that it needs to be respected by the BYO implementation and we are also including the config for BYO now, I've removed this.

go/core/internal/controller/translator/agent/manifest_builder.go
jmhbh: yeah we are, this guard was an artifact of the previous implementation and is now removed. Previously the substrate backend cloned the secret rendered by `buildConfigSecret` and patched session_db_url into the copy, so both same-named secrets landed in the manifest and this guard suppressed one of them. The url is now set in `CompileAgent` before the secret is rendered so the secret is only built once and this guard is now removed.

go/adk/pkg/session/local.go
jmhbh: removed.

go/core/internal/a2a/substrate_sandbox_transport_test.go
jmhbh: we use the client because the a2a call creates a row in the postgres db for session metadata. I also updated the test to run against the existing dbtest testcontainer helper

EItanya (approved):

PR #2173, opened 2026-07-07
fix(adk): annotate serialize_metadata_value as str | dict

anxkhn:
### What

`serialize_metadata_value` in `python/packages/kagent-adk/src/kagent/adk/converters/event_converter.py` accepted arbitrary metadata values, but its annotation and docstring said it always returned `str`. Pydantic values were already returned as `dict` objects through `model_dump(...)`, and that structured metadata is intentional because consumers such as the UI read fields like token usage directly.

### Fix

Update the helper contract to return `str | dict[str, Any]` and document the structured metadata behavior. For Pydantic values, keep returning a dict, but serialize with `model_dump(mode="json", exclude_none=True, by_alias=True)` so the metadata is JSON-compatible, omits `None` fields, and preserves aliases. Non-Pydantic values still fall back to `str(value)`.

### Tests

Expanded `TestSerializeMetadataValue` to cover:

- structured Pydantic metadata staying as a dict,
- aliases via `by_alias=True`,
- omitted `None` fields via `exclude_none=True`,
- enum values serialized in JSON mode, and
- plain values still serializing to strings.

Verification:

- `cd python && uv run ruff format packages/kagent-adk/src/kagent/adk/converters/event_converter.py packages/kagent-adk/tests/unittests/converters/test_event_converter.py`
- `cd python && uv run ruff check packages/kagent-adk/src/kagent/adk/converters/event_converter.py packages/kagent-adk/tests/unittests/converters/test_event_converter.py`
- `cd python && uv run pytest packages/kagent-adk/tests/unittests/` -> 341 passed, 1 skipped
- `cd python && uv run pytest packages/kagent-adk/tests/unittests/converters/test_event_converter.py` -> 4 passed


supreme-gg-gg (changes requested):
There's nothing wrong with returning dict here. A2A metadata allows them and they are needed in the frontend to properly show things like token usage: 

https://github.com/kagent-dev/kagent/blob/3eff08bd3c06ebe14a1ac25da6e5de95e32a9446/ui/src/lib/messageHandlers.ts#L390-L398

A better fix would be to update the type annotation to `str | dict[str, Any]` and the docstring.

If you trace the langgraph example you cited, you will find that even though `serialize_metadata_value` does return string it is in fact never used in the converter with any metadata, so it's effectively dead code.

python/packages/kagent-adk/src/kagent/adk/converters/event_converter.py
```diff
@@ -43,7 +43,7 @@ def serialize_metadata_value(value: Any) -> str:
     """
     if hasattr(value, "model_dump"):
         try:
-            return value.model_dump(exclude_none=True, by_alias=True)
+            return str(value.model_dump(exclude_none=True, by_alias=True))
```
mesutoezdil: str(model_dump(...)) produces pthon dict repr with single-quoted strings, not valid json. 
any caller doing json.parse() on this metadata value will get a syntaxerror. 
if the goal is a json string, model_dump_json(exclude_none=true, by_alias=true) is the right call.

supreme-gg-gg:
@anxkhn pls update PR title and description

python/packages/kagent-adk/src/kagent/adk/converters/event_converter.py
```diff
@@ -32,14 +32,20 @@
...
 
     Args:
       value: The value to serialize.
 
     Returns:
-      String representation of the value.
+      The value's ``model_dump`` dict for Pydantic values, otherwise its string
+      representation.
     """
     if hasattr(value, "model_dump"):
```
mesutoezdil, line 51: test line 50

mesutoezdil, line 51: model_dump() returns python-mode values that are not json-serializable.. 
if grounding_metadata or similar has enum fields, a2a serialization throws typeerror. 
use model_dump(mode='json', ...) instead

python/packages/kagent-adk/tests/unittests/converters/test_event_converter.py
```diff
@@ -135,3 +136,25 @@ def test_message_carries_task_and_context_ids(self):
...
+
+        class _Model(BaseModel):
+            a: int
+            b: str
+
+        value = _Model(a=1, b="x")
+
+        result = serialize_metadata_value(value)
+
+        assert result == value.model_dump(exclude_none=True, by_alias=True)
```
mesutoezdil: the model has no optional fields and no aliases, so exclude_none=True and by_alias=True have no observable effect. you could remove those kwargs from the implementation and this test would still pass

python/packages/kagent-adk/tests/unittests/converters/test_event_converter.py
```diff
@@ -4,8 +4,9 @@
 from a2a.types import TaskState, TaskStatusUpdateEvent
 from google.genai import types as genai_types
 from kagent.core.a2a import get_kagent_metadata_key
+from pydantic import BaseModel
```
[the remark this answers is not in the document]
anxkhn: Ruff keeps this import order, so I left the file formatter-compliant.

python/packages/kagent-adk/tests/unittests/converters/test_event_converter.py
```diff
@@ -135,3 +136,26 @@ def test_message_carries_task_and_context_ids(self):
...
         assert message.context_id == "ctx-xyz"
+
+
+class TestSerializeMetadataValue:
+    """Test cases for serialize_metadata_value."""
+
+    def test_pydantic_value_serializes_to_str(self):
+        """A pydantic value must serialize to a str, matching the -> str
+        annotation and docstring, so metadata values stay strings like their
+        siblings (e.g. error_code, which is stringified explicitly)."""
```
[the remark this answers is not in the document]
anxkhn: Updated the docstring to capitalize Pydantic.

python/packages/kagent-adk/src/kagent/adk/converters/event_converter.py
anxkhn: Changed direction to keep structured dict metadata instead of stringifying it.

python/packages/kagent-adk/src/kagent/adk/converters/event_converter.py
anxkhn: Superseded by the updated implementation on the latest commit.

anxkhn: Updated to use `model_dump(mode="json", exclude_none=True, by_alias=True)`.

python/packages/kagent-adk/tests/unittests/converters/test_event_converter.py
anxkhn: Expanded the test model to cover aliases, omitted `None`, and enum JSON serialization.

anxkhn:
Updated the title and description to reflect the structured metadata fix.

supreme-gg-gg (approved):

EItanya (approved):

Issue #2028, opened 2026-06-16
Secure the kagent UI


davidkarlsen:
The kagent UI/dashboard is exposed without authentication. We should have a story for securing access to it.

Discussion: https://discord.com/channels/1346225185166065826/1346225185841221644/1516428842057601165

Two things that came up:
- A `HTTPRoute` + `NetworkPolicy` to gate/restrict access.
- The OpenShift Route ([openshift-route.yaml](https://github.com/kagent-dev/kagent/blob/main/helm/kagent/templates/openshift-route.yaml)) renders whenever the `route.openshift.io/v1` API is present — there's no value to toggle it off and no auth in front of it, so on OpenShift it's effectively always exposed.

davidkarlsen:
Linking the two halves of this so they read as one story.

This issue is about *exposure* — keeping the UI/Route from being reachable with nothing in front of it (HTTPRoute + NetworkPolicy, and the always-on OpenShift Route). That's the edge: authenticate, and don't let traffic bypass the proxy.

The complementary half is *authorization* — once a user is in, who's allowed to see and invoke which agents. The controller currently ships a no-op authorizer (everyone can do everything), so the edge work above gets you authentication but not access control.

I've posted a concrete design proposal for that half on #1270 — a CEL-based authorizer behind the existing `Authorizer` interface, with per-agent policy on the CR: https://github.com/kagent-dev/kagent/issues/1270#issuecomment-4778333681

So the full picture: gate + authenticate at the edge (this issue) → authorize per-resource inside the controller (#1270).

PR #2174, opened 2026-07-07
feat(helm): add extraObjects for arbitrary manifests

younsl:
## Summary

Adds an optional top-level `extraObjects` value that renders arbitrary Kubernetes manifests within the kagent chart. It is empty by default so existing installs stay unaffected. Each entry is rendered through `tpl` so values can reference the release context such as `{{ include "kagent.fullname" . }}` or `{{ .Release.Namespace }}`. Both map and multi-line string entries are supported.

This lets users manage companion resources like `ExternalSecret` or `HTTPRoute` in the same chart lifecycle. It follows a common pattern used by community charts such as:

- grafana
- kube-prometheus-stack
- external-secrets

## Changes

- New template `templates/extra-objects.yaml` ranging over `.Values.extraObjects` (default `[]`).
- Documented `extraObjects` in `values.yaml` with an `ExternalSecret` example.
- Added helm-unittest suite `tests/extra-objects_test.yaml`.

## Testing

`helm unittest helm/kagent` passes. The new suite covers the empty default map entries multiple documents `tpl` evaluation against the release context and multi-line string entries.

## Notes

This change was originally intended for #2165 but was split into its own PR at a maintainer request so each concern can be reviewed on its own.

### DIFF
--- a/helm/kagent/values.yaml
+++ b/helm/kagent/values.yaml
@@ -865,21 +865,24 @@ otel:
 # Both map and multi-line string entries are supported. Use this to manage
 # resources such as ExternalSecret, HTTPRoute, or NetworkPolicy within the same
 # chart lifecycle without maintaining a separate chart.
+#
+# To use, replace the empty list below with your manifests, e.g.:
+#   extraObjects:
+#     - apiVersion: external-secrets.io/v1
+#       kind: ExternalSecret
+#       metadata:
+#         name: '{{ include "kagent.fullname" . }}-openai'
+#         namespace: '{{ .Release.Namespace }}'
+#       spec:
+#         secretStoreRef:
+#           name: aws-secretsmanager
+#           kind: ClusterSecretStore
+#         target:
+#           name: kagent-openai
+#         data:
+#           - secretKey: OPENAI_API_KEY
+#             remoteRef:
+#               key: prod/kagent/openai
+#               property: api_key
 # @default -- []
 extraObjects: []
-  # - apiVersion: external-secrets.io/v1
-  #   kind: ExternalSecret
-  #   metadata:
-  #     name: '{{ include "kagent.fullname" . }}-openai'
-  #     namespace: '{{ .Release.Namespace }}'
-  #   spec:
-  #     secretStoreRef:
-  #       name: aws-secretsmanager
-  #       kind: ClusterSecretStore
-  #     target:
-  #       name: kagent-openai
-  #     data:
-  #       - secretKey: OPENAI_API_KEY
-  #         remoteRef:
-  #           key: prod/kagent/openai
-  #           property: api_key
--- a/helm/kagent/values.yaml
+++ b/helm/kagent/values.yaml
@@ -854,3 +854,32 @@ otel:
         endpoint: ""
         timeout: 15000 # milliseconds
         insecure: true
+
+# ==============================================================================
+# EXTRA OBJECTS
+# ==============================================================================
+
+# -- Additional arbitrary Kubernetes manifests to deploy alongside the chart.
+# Each list entry is rendered through `tpl`, so values may reference the release
+# context (e.g. `{{ include "kagent.fullname" . }}`, `{{ .Release.Namespace }}`).
+# Both map and multi-line string entries are supported. Use this to manage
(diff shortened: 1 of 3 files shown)


younsl:
@EItanya PTAL

helm/kagent/values.yaml
```diff
@@ -854,3 +854,32 @@ otel:
...
+# ==============================================================================
+
+# -- Additional arbitrary Kubernetes manifests to deploy alongside the chart.
+# Each list entry is rendered through `tpl`, so values may reference the release
+# context (e.g. `{{ include "kagent.fullname" . }}`, `{{ .Release.Namespace }}`).
+# Both map and multi-line string entries are supported. Use this to manage
+# resources such as ExternalSecret, HTTPRoute, or NetworkPolicy within the same
+# chart lifecycle without maintaining a separate chart.
+# @default -- []
+extraObjects: []
```
mesutoezdil, line 888: uncommenting the example lines without also changing `extraObjects: []` to `extraObjects:` produces a yaml parse error. two edits are required, but only one is obvious.

EItanya (commented):
Personally this feels like a weird pattern to me. I'd be interested to hear how other projects and other systems are dealing with this problem today.

helm/kagent/values.yaml
younsl: Good catch. Fixed in 22e0968a. The commented example was indented under `extraObjects: []`, so uncommenting it alone left the empty list in place and produced a YAML parse error. I moved the example above the key as a self-contained reference block (a commented `extraObjects:` with its own indented children), so now there is only one obvious action: replace the `[]` value directly. No hidden second edit.

younsl:
Fair question. This mirrors an established Helm pattern: a top level list rendered through tpl, used for companion resources like ExternalSecret, HTTPRoute, Istio DestinationRule, or NetworkPolicy that need to share the release lifecycle.

Many popular community Helm charts follow this pattern through values named like extraObjects or extraManifests. A few examples with the same shape:
- [Argo CD extraObjects](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml)
- [kube-prometheus-stack extraManifests](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml)
- [Bitnami extraDeploy](https://github.com/bitnami/charts/blob/main/bitnami/redis/values.yaml)

The alternative is a second chart, which duplicates values, needs its own helm upgrade, and can drift from the app version. extraObjects keeps them atomic.

Happy to rename to extraDeploy or extraManifests for familiarity if you prefer. Let me know.

younsl:
For context on my own use case: I needed to manage a separate HTTPRoute resource for the kagent UI as part of the same kagent chart release, rather than maintaining it in a detached manifest or a second chart. extraObjects lets that HTTPRoute live and upgrade alongside the app it fronts.

EItanya:
Ok that all makes sense, I’ll do a little more research of my own. Since we merged that other HTTPRoute helm change do you still need this PR?

younsl:
Thanks for looking into this. I think it is still worth keeping. The HTTPRoute change covers UI routing, but extraObjects is the general way to add other related resources that may not each need their own value.

A couple of cases I have in mind:
- A PrometheusRule handled by the Prometheus Operator, so kagent's alerts upgrade together with the app instead of living in a separate file.
- The NetworkPolicy from #2028 for locking down the UI, which users could add right away without a dedicated value.

Adding a new value for every resource type does not scale well, while extraObjects stays generic. That said, I am happy to follow your call if you prefer a different direction.

elihai-risk:
+1 — this would cover three real use cases we just brought as separate PRs (#2238 UI Ingress, #2240 Istio Sidecar/ServiceEntries/JWT, #2241 Vault Secrets Operator secrets), all now closed in favor of this. We run kagent on EKS behind Istio + Gatekeeper with Vault-sourced secrets, and `extraObjects` with `tpl` rendering is exactly the escape hatch that keeps such org-specific resources in the same chart lifecycle without bloating the core chart. Happy to test this branch against our production values if it helps it land.

EItanya (approved):

younsl:
Thanks for the detailed review. Your feedback made the change cleaner, and it's great that it also resolved the related issues. Looking forward to contributing more.

PR #2175, opened 2026-07-07
feat(models): add maxCompletionTokens for OpenAI

anthonytwh:
## What

Add a `maxCompletionTokens` field to the OpenAI `ModelConfig` that maps to OpenAI's **`max_completion_tokens`** request parameter.

## Why

OpenAI's `max_tokens` parameter is deprecated for the Chat Completions API and is **rejected outright by reasoning models** (GPT-5 / o-series):

```
400 Bad Request: Unsupported parameter: 'max_tokens' is not supported with this model.
Use 'max_completion_tokens' instead.
```

Today `OpenAIConfig` only exposes `maxTokens` (→ `max_tokens`), so there is **no way to cap output tokens for GPT-5 / o-series models** — setting `maxTokens` on such a model makes every request fail. (Per OpenAI's SDK docstrings, `max_tokens` "is now deprecated in favor of `max_completion_tokens`, and is not compatible with o-series models"; `max_completion_tokens` is "an upper bound … including visible output tokens and reasoning tokens".)

## Approach — additive, non-breaking

`maxTokens` is left intact and unchanged:
- **Backward compatible** — existing ModelConfigs behave identically.
- **OpenAI-compatible endpoints** — `BaseOpenAI` also serves `baseUrl`-overridden OpenAI-compatible servers, some of which still expect `max_tokens`. Keeping both lets operators choose the right one per model.

The two fields are independent; set `maxCompletionTokens` for reasoning models, `maxTokens` for legacy/compatible endpoints.

## Changes

- `OpenAIConfig.MaxCompletionTokens` (v1alpha2), `+kubebuilder:validation:Minimum=1`; CRDs regenerated (bases + Helm).
- `adk.OpenAI` serialization struct carries `max_completion_tokens`.
- Controller translator copies `Spec.OpenAI.MaxCompletionTokens`.
- Python ADK OpenAI client emits `max_completion_tokens`; `OpenAI` config model gains the field (`ge=1`).
- Go ADK OpenAI model (`applyOpenAIConfig` + `CreateLLM`) emits `max_completion_tokens` via the openai-go SDK.

## Tests

- Go translator test: `Test_AdkApiTranslator_OpenAIMaxCompletionTokens` — verifies the field flows through and `max_tokens` stays unset.
- Go ADK apply test: `TestApplyOpenAIConfig/config_with_max_completion_tokens`.
- Python client test: `test_generate_content_async_with_max_completion_tokens` — asserts `max_completion_tokens` is forwarded and `max_tokens` is absent.

All pass; `go build ./api/... ./adk/... ./core/internal/controller/...` is green and CRD generation is clean.

## Note

Scope is the native OpenAI provider (the one that hits `api.openai.com`). Azure OpenAI reasoning deployments have the same underlying need; happy to extend `AzureOpenAIConfig` in a follow-up if maintainers prefer.


anthonytwh:
Good catch @copilot — fixed in 0b813ae5. `maxTokens` and `maxCompletionTokens` are now mutually exclusive at request-build time in both the Python (`_openai.py`) and Go (`openai_adk.go`) clients: exactly one parameter is sent, with `max_completion_tokens` taking precedence when both are configured.

I went with modern-parameter precedence rather than the `base_url`-based selection, because `base_url` being overridden doesn't reliably indicate a non-OpenAI backend — it's commonly a proxy/gateway that fronts the real OpenAI API (which still needs `max_completion_tokens` for reasoning models). Precedence is deterministic and never sends both. Single-field configs — including compat endpoints that only set `maxTokens` — are unchanged. Added precedence tests for both runtimes.

If you'd prefer an even stronger guard, I can also add a CRD-level CEL validation rejecting configs that set both fields — happy to include that if maintainers want it.

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -151,10 +151,20 @@ type OpenAIConfig struct {
...
 	// +optional
 	MaxTokens int `json:"maxTokens,omitempty"`
 
+	// Maximum completion tokens to generate. Sent as the OpenAI
+	// `max_completion_tokens` request parameter (an upper bound on visible
+	// output plus reasoning tokens). This is the parameter reasoning models
+	// (GPT-5 / o-series) require in place of the deprecated maxTokens.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	MaxCompletionTokens int `json:"maxCompletionTokens,omitempty"`
```
supreme-gg-gg, line 171: We should add a validation rule that only one of `MaxCompletionTokens` or `MaxTokens` is provided (if I understood properly from the description they should be mutually exclusive)?

supreme-gg-gg (changes requested):
Makes sense overall, one comment before we can merge this

mesutoezdil:
> Good catch @copilot — fixed in [0b813ae](https://github.com/kagent-dev/kagent/commit/0b813ae5c94aa217b6a9575421963753a3d0da3d). `maxTokens` and `maxCompletionTokens` are now mutually exclusive at request-build time in both the Python (`_openai.py`) and Go (`openai_adk.go`) clients: exactly one parameter is sent, with `max_completion_tokens` taking precedence when both are configured.
> 
> I went with modern-parameter precedence rather than the `base_url`-based selection, because `base_url` being overridden doesn't reliably indicate a non-OpenAI backend — it's commonly a proxy/gateway that fronts the real OpenAI API (which still needs `max_completion_tokens` for reasoning models). Precedence is deterministic and never sends both. Single-field configs — including compat endpoints that only set `maxTokens` — are unchanged. Added precedence tests for both runtimes.
> 
> If you'd prefer an even stronger guard, I can also add a CRD-level CEL validation rejecting configs that set both fields — happy to include that if maintainers want it.

seeing one llm's output pitted against another's looks quite funny from the outside :)
everything has become tragicomic in recent years.

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -151,10 +151,20 @@ type OpenAIConfig struct {
 	// +optional
 	Temperature string `json:"temperature,omitempty"`
 
-	// Maximum tokens to generate
+	// Maximum tokens to generate. Sent as the OpenAI `max_tokens` request
+	// parameter, which is deprecated and rejected by reasoning models
+	// (GPT-5 / o-series). For those models set maxCompletionTokens instead.
 	// +optional
 	MaxTokens int `json:"maxTokens,omitempty"`
```
mesutoezdil, line 162: maxTokens has no +kubebuilder:validation:Minimum=1 marker but maxCompletionTokens below does.. 
value like maxTokens: 0 or maxTokens: -1 passes crd validation silently.. and is then ignored by the translator
adding the same minimum marker here would make the two fields consistent in my oprinio

go/api/v1alpha2/modelconfig_types.go
anthonytwh: Done in 9283c15. Added a type-level CEL rule on `OpenAIConfig`:

```
+kubebuilder:validation:XValidation:message="maxTokens and maxCompletionTokens are mutually exclusive",rule="!(has(self.maxTokens) && has(self.maxCompletionTokens))"
```

So a ModelConfig that sets both is now rejected at admission (matching the existing XValidation pattern used for the TLSConfig rules in this file). The runtime-level precedence guard in the Python/Go clients stays as defense-in-depth for OpenAI-compatible paths that don't go through CRD admission. Added an envtest-based CEL test (`modelconfig_cel_test.go`) covering both-set rejection plus the single-field and neither-field cases.

go/api/v1alpha2/modelconfig_types.go
anthonytwh: Good catch — fixed in 9283c15. `maxTokens` now carries `+kubebuilder:validation:Minimum=1` too, so `maxTokens: 0`/`-1` is rejected at admission instead of silently ignored by the translator. The regenerated CRD bases/helm templates reflect `minimum: 1` on both fields, and the new CEL test asserts a negative value on either field is rejected.

mesutoezdil: > Good catch — fixed in [9283c15](https://github.com/kagent-dev/kagent/commit/9283c159a9eda3511b607ee62960b23adebe8c11). `maxTokens` now carries `+kubebuilder:validation:Minimum=1` too, so `maxTokens: 0`/`-1` is rejected at admission instead of silently ignored by the translator. The regenerated CRD bases/helm templates reflect `minimum: 1` on both fields, and the new CEL test asserts a negative value on either field is rejected.

why do you post llm outputs here instead of writing your own sentences? i think everyone should maintain healthy communication as a sign of respect for the person they are interacting with..

anthonytwh: It is automatic, no disrespect here.. Thanks for your comment on the PR.

EItanya (commented):
LGTM, I'll leave to @supreme-gg-gg for final approval

anthonytwh:
@supreme-gg-gg looking for a final review here

supreme-gg-gg (commented):
@anthonytwh can you fix merge conflict and I'll approve, looks good

supreme-gg-gg (approved):

PR #2176, opened 2026-07-07
Force-flush trace spans before the A2A response completes for substrate

iplay88keys:
# Description

Agent Substrate checkpoints an actor as soon as its A2A response completes. Spans still buffered by OpenTelemetry's `BatchSpanProcessor` at that point remain frozen until the actor resumes, or can be lost when the snapshot does not preserve process state. This makes the session's latest trace delayed or missing.

The flush must happen after HTTP instrumentation ends the inbound server span, but before the terminal response message triggers checkpointing. Flushing from the agent executor is too early because the server span is still open.

### Go ADK

- Add `telemetry.ForceFlush(ctx)` to flush providers that support it.
- Wrap the instrumented A2A handler so flushing happens after the `otelhttp` server span ends and before `ServeHTTP` returns.
- Use a detached timeout because the request context may already be canceled.
- Skip health and agent-card requests.

### Python ADK

- Add `kagent.core.tracing.force_flush()`.
- Wrap the FastAPI ASGI stack and hold only the terminal response body or trailer message.
- Let the instrumented application return and end its server span, flush in a worker thread, then send the held terminal message.
- Skip health, thread-dump, and agent-card requests.

Both implementations default to a 3-second flush timeout, configurable with `KAGENT_TRACE_FLUSH_TIMEOUT_MS`. Unsupported tracer providers are a no-op, and exporter failures do not fail the A2A response. Streaming response chunks are unaffected; only response completion waits for the bounded flush.

### Helm image helpers

- Include the optional `image.name` segment in the `grafana-mcp` and `querydoc` image helpers.
- Preserve existing two-segment image references when `image.name` is unset.


python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -198,6 +199,24 @@ async def execute(
...
+            try:
+                await asyncio.to_thread(force_flush_tracing)
+            except asyncio.CancelledError:
+                # A cancellation landing during this await (same cross-task
+                # cancel-scope corruption handled above) must not escape
+                # execute() — that would 500 an otherwise-complete request.
+                # The flush is best-effort; clear the cancellation and move on.
+                current_task = asyncio.current_task()
+                if current_task is not None:
+                    while current_task.uncancel() > 0:
```
mesutoezdil: Task.uncancel() was added in python 3.11 but pyproject.toml declares requires-python = '>=3.10'

go/adk/pkg/a2a/executor.go
```diff
@@ -165,7 +165,13 @@ func (e *KAgentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.RequestCont
 	}
 	ctx = telemetry.SetKAgentSpanAttributes(ctx, spanAttributes)
 	ctx, invocationSpan := telemetry.StartInvocationSpan(ctx)
-	defer invocationSpan.End()
+	defer func() {
+		invocationSpan.End()
```
krisztianfekete: Hm, I don't think invocation is the root span. Not sure if we can flush after the middleware span though in case of substrate. Can you please try it?

go/adk/pkg/telemetry/tracing.go
```diff
@@ -38,6 +38,26 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
...
+// until the session's next resume (or forever, for a session's last message).
+// Uses its own detached timeout because the request context is typically
+// already canceled by the time deferred cleanup runs.
+func ForceFlush(ctx context.Context) {
+	type flusher interface{ ForceFlush(context.Context) error }
+	fp, ok := otel.GetTracerProvider().(flusher)
+	if !ok {
+		return
+	}
+	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 3*time.Second)
```
krisztianfekete: Maybe we should make this configurable, 3s seems a lot in some cases, especially if users expect "user-visible chat latency".

iplay88keys: Good thinking, I will add that. I didn't think it should block the response due to it being in a defer/finally, but I might be wrong about that in the unhappy path case.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
iplay88keys: Good catch. I think we can bump to python 3.11 in a follow-up PR. There's already a current reference to `uncancel()` [here](https://github.com/kagent-dev/kagent/blob/b04769e8bc0fa52c8d4145229218c6995dea2f9d/python/packages/kagent-adk/src/kagent/adk/_agent_executor.py#L189), so this isn't new.

go/adk/pkg/a2a/executor.go
iplay88keys: Good catch, I have implemented a fix for this.

Before (note that the `POST` would come through on the next invocation, but that was not ideal):
```
generate_content ...
invoke_agent ...
invocation ...
```

After:
```
generate_content ...
invoke_agent ...
invocation ...
POST / ...
```

go/adk/pkg/a2a/server/server.go
```diff
@@ -57,6 +59,14 @@ func NewA2AServer(agentCard a2atype.AgentCard, executor a2asrv.AgentExecutor, lo
...
 		}),
 	)
+	// Flush buffered spans after the otelhttp server span ends (when the inner
+	// handler returns) but before net/http closes the response body: Agent
+	// Substrate checkpoints the actor as soon as the body closes, and a flush
+	// issued inside the executor can never include the still-open server span.
+	flushingHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		instrumentedHandler.ServeHTTP(w, r)
+		telemetry.ForceFlush(r.Context())
+	})
```
krisztianfekete: We probably want something like this cover only the A2A path:
https://github.com/kagent-dev/kagent/blob/003663c4f50d88aa3882a13425f0a9a8c3d9bdb4/go/adk/pkg/a2a/server/server.go#L51-L58

go/adk/pkg/a2a/executor.go
krisztianfekete: Now that we have the server-side changes, we can probably revert this to avoid the duplicated work and clean up the stale commit?

go/adk/pkg/a2a/server/server.go
krisztianfekete, line 1: We don't have a similar mechanism in place for Python. Is that supported via substrate?

go/adk/pkg/a2a/executor.go
iplay88keys: Oh, yes, of course. Will do!

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
iplay88keys: Python bumped to 3.11 min [here](https://github.com/kagent-dev/kagent/pull/2242) with a backport to the 0.9.x line [here](https://github.com/kagent-dev/kagent/pull/2266).

go/adk/pkg/a2a/server/server.go
iplay88keys: Looking into this.

go/adk/pkg/a2a/server/server.go
iplay88keys: Agreed, I'll add that.

go/adk/pkg/a2a/server/server.go
iplay88keys: I have added similar logic to the python side and see the following spans: `POST /`, `openai.chat`, `generate_content`, `call_llm`, `invoke_agent`, `invocation`, and A2A queue spans.

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
```diff
@@ -106,6 +107,104 @@ def _instrument_google_generativeai():
...
+                    message_type == "http.response.body"
+                    and not expecting_trailers
+                    and not message.get("more_body", False)
+                ) or (message_type == "http.response.trailers" and not message.get("more_trailers", False))
+                if is_terminal:
+                    terminal_message = message
+                else:
+                    await send(message)
+
+            await inner(scope, receive, hold_terminal_message)
```
krisztianfekete: Can we wrap this into try/finally, and add a test?

krisztianfekete (commented):
Can we gate this to substrate only?

krisztianfekete (approved):
Thanks, LGTM!

krisztianfekete:
I've been looking into this (and now https://github.com/agent-substrate/substrate/issues/503, cc. @EItanya) and I think long term something like this makes sense:

- merge this as-is for now as it's substrate-focused and better than before
- we should recommend node-local collectors with `file_storage` configured as sending queues
- once we have the lifecycles implemented, we can use those intead of what we do in this PR
- (we can also decrease the export interval)

EItanya (approved):

PR #2180, opened 2026-07-08
feat(ui-test): Playwright ui tests foundation stage 0

cjlg-soloio:
### Description

Adds the **Stage 0 foundation** for Playwright page-level E2E testing of the kagent UI, running against a mocked backend with no cluster required.

The suite targets the one gap nothing else covers — **multi-step user flows across components**:

| Tool | Covers |
|------|--------|
| Storybook / Chromatic | Component render states |
| Jest | Logic |
| Cypress | Page-load smoke |
| **Playwright (this PR)** | **Cross-component user flows** |

**The core design challenge:** the UI fetches data server-side via Next.js server actions, so browser-level `page.route("**/api/**")` intercepts nothing. Instead, this PR boots a dependency-free stub backend and redirects Next's server-side fetch to it via `BACKEND_INTERNAL_URL` — which `getBackendUrl()` in `ui/src/lib/utils.ts` already honors. Playwright's `webServer` block starts both the stub (`:8899`) and `next dev` (`:8001`) automatically.

This is **foundation-only**: config, stub backend, shared fixture, CI workflow, and a single smoke test. The helper/driver library and per-test scenarios land in Stage 1; feature flows in Stage 3 (see `ui/playwright/README.md`).

### Files

**New**
| File | Purpose |
|------|---------|
| `ui/playwright.config.ts` | Boots stub + dev server; chromium project; CI-aware retries/workers; trace & screenshot on failure |
| `ui/playwright/mocks/server.mjs` | Dependency-free stub backend (Node `http` only); serves happy-path payloads mirroring `src/types/index.ts`; exposes `/__mock/health` & `/__mock/reset` |
| `ui/playwright/mocks/data.ts` | Typed spec-side data builders mirroring the stub's shapes |
| `ui/playwright/fixtures/test.ts` | Shared `test`/`expect` fixture; resets stub per test and bypasses first-run onboarding |
| `ui/playwright/tests/smoke.spec.ts` | Stage 0 smoke test — asserts the Agents list renders against the mock |
| `ui/playwright/tsconfig.json` | Extends UI tsconfig with node + Playwright types |
| `ui/playwright/README.md` | Test strategy, mocking rationale, conventions, roadmap |
| `.github/workflows/ui-playwright.yaml` | Standalone CI workflow (paths-filtered to `ui/**`); installs chromium, runs `test:pw`, uploads report |

**Modified**
- `ui/package.json` — `@playwright/test` dev dep + `test:pw` / `test:pw:ui` / `test:pw:debug` scripts
- `ui/package-lock.json` — lockfile updates
- `ui/.gitignore` — ignores `playwright-report`, `playwright/test-results`, `playwright/.cache`

### Testing

```
cd ui
npm install
npx playwright install chromium   # first time only
npm run test:pw
```

Playwright starts the stub backend and dev server automatically. The smoke test passes when the home page renders the Agents, no "Error Encountered" state appears, and no uncaught page errors occur. CI runs the same on any `ui/**` change.

### Change type

test / ci — new test infrastructure and CI workflow. No product or runtime code changed.

### Changelog

`Stage 1 of 3 playwright tests for ui`


ui/playwright/mocks/server.mjs
```diff
@@ -0,0 +1,120 @@
...
+    "Content-Type": "application/json",
+    "Content-Length": Buffer.byteLength(payload),
+  });
+  res.end(payload);
+};
+
+const server = createServer((req, res) => {
+  const { method, url } = req;
+  const pathname = url.split("?")[0];
+
```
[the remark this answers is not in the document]
cjlg-soloio: Defaulted req.method and req.url before parsing so a missing url can't throw on split.
---
*🤖 written by Claude*

ui/playwright/fixtures/test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { test as base, expect } from "@playwright/test";
+
+const STUB_URL = "http://127.0.0.1:8899";
+
+export const test = base.extend({
+  page: async ({ page, request }, use) => {
+    await request.post(`${STUB_URL}/__mock/reset`).catch(() => {
+      // Stub not reachable (e.g. reused external server) — non-fatal for Stage 0.
+    });
```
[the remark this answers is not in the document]
cjlg-soloio: Replaced the swallowed catch with a try/catch that rethrows when process.env.CI is set, so a failed reset fails fast in CI while staying non-fatal locally.
---
*🤖 written by Claude*

.github/workflows/ui-playwright.yaml
```diff
@@ -0,0 +1,57 @@
...
+    paths:
+      - "ui/**"
+  workflow_dispatch:
+
+jobs:
+  playwright:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
```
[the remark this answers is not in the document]
cjlg-soloio: Dropped pull-requests: write since no step writes to the PR; the job keeps only contents: read.
---
*🤖 written by Claude*

ui/playwright.config.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const CI = !!process.env.CI;
+
+const STUB_PORT = 8899;
+const STUB_URL = `http://127.0.0.1:${STUB_PORT}`;
+const APP_URL = "http://localhost:8001";
+
+export default defineConfig({
+  testDir: "./playwright/tests",
+  outputDir: "./playwright/test-results",
```
mesutoezdil, line 23: With workers: 1 this is a no-op, but raising workers before Stage 1 isolation lands will cause tests to race against the shared stub. Consider setting this to false until then.

mesutoezdil:
dco is red

ui/playwright.config.ts
cjlg-soloio: Good call — Fixed in 08b3cfcf — set fullyParallel: false to match workers: 1, so both stay off until Stage 1 isolation lands. Thanks @mesutoezdil !

mesutoezdil: > Good call — Fixed in [08b3cfc](https://github.com/kagent-dev/kagent/commit/08b3cfcf36fe2bf971fa24ab4c30a5c8d51bf952) — set fullyParallel: false to match workers: 1, so both stay off until Stage 1 isolation lands. Thanks @mesutoezdil !

pls write your own sentences when interacting with me, not llm outputs. thx!

cjlg-soloio: my bad! I will keep it human moving forward 🤝

ui/playwright/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "types": ["node", "@playwright/test"]
+  },
+  "include": ["**/*.ts", "**/*.mjs"]
```
mesutoezdil: mjs without allowJs: true is a no-op, typescript won't check server.mjs

ui/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+  use: {
+    baseURL: APP_URL,
+    screenshot: "only-on-failure",
+    trace: "retain-on-failure",
+    video: "off",
+  },
+  projects: [{ name: "chromium", use: { ...devices["Desktop Chrome"] } }],
+  webServer: [
+    {
+      command: "node playwright/mocks/server.mjs",
```
mesutoezdil, line 44: server.mjs reads STUB_PORT from env but this block doesn't pin it, so a shell-set STUB_PORT makes the stub bind to the wrong port and the health check times out

ui/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+      url: `${STUB_URL}/__mock/health`,
+      reuseExistingServer: !CI,
+      timeout: 30_000,
+      stdout: "pipe",
+      stderr: "pipe",
+    },
+    {
+      command: "npm run dev",
+      url: APP_URL,
+      reuseExistingServer: !CI,
```
mesutoezdil: if npm run dev is already running without BACKEND_INTERNAL_URL, reuseExistingServer reuses it and the stub is bypassed silently

ui/playwright/mocks/data.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import type {
+  AgentResponse,
+  ModelConfig,
+  ProviderModelsResponse,
+  ToolsResponse,
+} from "@/types";
+
+export type Namespace = { name: string; status: string };
+export type ToolServerListEntry = { ref: string; groupKind: string; discoveredTools: unknown[] };
```
mesutoezdil: discoveredTools is typed as unknown[] but the real ToolServerResponse from @/types uses DiscoveredTool[]. the file comment promises compile-time drift detection, but this local type bypasses it

ui/playwright/tsconfig.json
cjlg-soloio: removed it

ui/playwright.config.ts
cjlg-soloio: I pinned STUB_PORT in the stub's webServer env so it always goes to the health endpoint or whatever BACKEND_INTERNAL_URL is set to

ui/playwright.config.ts
cjlg-soloio: changed it to boot always on BACKEND_INTERNAL_URL and made it fail loudly if 8001 is already taken

ui/playwright/mocks/data.ts
cjlg-soloio: good catch! using RemoteMCPServerResponse now

cjlg-soloio: I wrote that sentence! 😅

.github/workflows/ui-playwright.yaml
```diff
@@ -0,0 +1,56 @@
...
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: "24"
```
Charlesthebird: This can use the `.nvmrc` file:

```suggestion
          node-version-file: "ui/.nvmrc"
```

.github/workflows/ui-playwright.yaml
```diff
@@ -0,0 +1,56 @@
...
+          cache: "npm"
+          cache-dependency-path: ui/package-lock.json
+
+      - name: Install dependencies
+        working-directory: ./ui
+        env:
+          CYPRESS_INSTALL_BINARY: 0
```
Charlesthebird: Might be able to be removed with the removal of Cypress.

```suggestion
```

ui/playwright/mocks/data.ts
Charlesthebird, line 1: If this isn't used it should be removed (it's okay to leave in if it's used or being removed in subsequent steps).

ui/playwright/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Layer | Tool | Not Playwright's job |
+|---|---|---|
+| Atoms (`src/components/ui/*`) | shadcn primitives | skip |
+| Visual / render states | Storybook + Vitest-browser + Chromatic | skip |
+| Unit / logic | Jest (`*.test.ts(x)`) | skip |
+| Page-load smoke (`h1` renders, page reachable, onboarding steps 1–2, model-edit page opens) | Cypress (`cypress/e2e/smoke.cy.ts`) | skip |
+| **Multi-step flows: form submission, payload correctness, streaming, wizard completion, error/edge states** | **Playwright (this suite)** | — |
+
+Rule: if Cypress / Chromatic / Jest already assert it, Playwright does not.
```
Charlesthebird: Cypress can be removed in this PR (the smoke test, npm package, and any other references).

Jest cannot - there are a lot of Jest unit tests across the repo.

There are also Storybook tests that check that each component can be rendered and passes a11y checks. These tests internally use Vitest and Playwright because that's what Storybook does. These shouldn't be removed. It might make sense to migrate Jest to Vitest and then just use one vitest config for all of it, but that's outside the scope of this PR. Also, these are just dev dependencies, so using 2 different tools (vitest and jest) doesn't cause much of an issue aside from 2 different configs, which are relatively low maintenance.

Charlesthebird (changes requested):
Looks good overall, just a few requests before merging:

- [x] The package-lock diff due to node version not matching .nvmrc should be resolved.
- [x] Cypress can be removed here.
- [x] There should be a playwright video output.

ui/playwright/mocks/data.ts
cjlg-soloio: keeping for the upcoming PR

ui/package.json
```diff
@@ -10,8 +10,9 @@
     "test": "jest",
     "test:vitest": "vitest run",
     "test:watch": "jest --watch",
-    "test:e2e:cypress": "cypress run --e2e",
-    "test:e2e": "start-server-and-test dev http://localhost:8001 test:e2e:cypress",
+    "test:pw": "playwright test",
```
Charlesthebird, line 13: We might want to keep in the test:e2e script here but point it to the playwright one.

```suggestion
    "test:pw": "playwright test",
    "test:e2e": "npm run test:pw",
```

ui/package-lock.json
Charlesthebird, line 1: I still see some diff of some extra metadata because my npm version is:
```
❯ npm --version
11.6.2
```

But this isn't a blocker - there's no functional difference and when `npm i` is run again using npm v10 it would revert back. We should probably have `packageManager` set in the `package.json` to avoid that (maybe could add this in to follow-up).

Charlesthebird (commented):
Thanks for the updates!

Charlesthebird (approved):

Charlesthebird (approved):

Charlesthebird (approved):

peterj (dismissed):

peterj:
@cjlg-soloio there are some CI failures: https://github.com/kagent-dev/kagent/actions/runs/29391678324/job/87276477369?pr=2180

cjlg-soloio:
Superseded by #2262. This PR's branch name contained parentheses (`feat(ui)--...`) which broke the Docker buildx cache-scope shell interpolation in CI (`/bin/sh: Syntax error: "(" unexpected`), failing all build/e2e/upgrade jobs. Reopened from a paren-free branch (`cjlg-soloio/playwrightTestsStage1`) in #2262 with the same changes plus the ui-tests lint fix.

PR #2191, opened 2026-07-09
feat: allow read-only shares to read A2A tasks

QuentinBisson:
## What

Read-only share tokens can now read tasks over A2A (`ListTasks`, `GetTask`). They were blocked from every A2A method because `shareTokenMiddleware` rejects all non-GET requests on the A2A path, and A2A is JSON-RPC over POST. Read-only is the default share mode (`session_shares.go` defaults `read_only` to `true`), so the share-scoped task read path added in #2187 was unreachable for the common case.

## How

Read-only enforcement for A2A moves from the transport verb to the request handler:

- Reads pass through: `ListTasks`, `GetTask`, push-config get/list, `SubscribeToTask`.
- Mutating methods are rejected under a read-only share via `requireWritableShare`: `SendMessage`, `SendStreamingMessage`, `CancelTask`, `CreateTaskPushConfig`, `DeleteTaskPushConfig`.
- The session REST path keeps its existing verb-based read-only block.

The middleware no longer blanket-blocks non-GET A2A requests for read-only shares, so the per-method guards are what stop a read-only share from mutating a session. Guarding at the handler covers both wires (v0 and v1 route the same methods to it).

## Behavior change

A read-only share attempting a mutating A2A method now returns a JSON-RPC error result instead of an HTTP 403.

Follows #2187 (store-backed `ListTasks`); branched from `main`, no code dependency on it.


EItanya:
@onematchfox would love your thoughts as well

onematchfox:
> @onematchfox would love your thoughts as well

I'm on PTO at the moment so haven't looked at the code/implementation itself but the general ask/fix makes sense to me. My original implementation focussed on the controller API endpoints used to read state from the Kagent DB rather than the semantics of similar functionality that could be achieved directly via A2A methods.

QuentinBisson:
Out of draft and ready for review

go/core/internal/a2a/passthrough_handler.go
```diff
@@ -89,6 +105,12 @@ func (h *PassthroughRequestHandler) SubscribeToTask(ctx context.Context, req *a2
 }
 
 func (h *PassthroughRequestHandler) SendStreamingMessage(ctx context.Context, req *a2atype.SendMessageRequest) iter.Seq2[a2atype.Event, error] {
+	if err := requireWritableShare(ctx); err != nil {
+		return func(yield func(a2atype.Event, error) bool) {
+			var zero a2atype.Event
+			yield(zero, err)
+		}
+	}
```
EItanya, line 113: Why not return an error here like the other functions?

QuentinBisson: `SendStreamingMessage` returns `iter.Seq2[a2atype.Event, error]`, not `(result, error)`, so there's no error return to hand back here. 

With this interface the only way to surface the error to the caller is to yield it from the sequence, which is what the closure does.

EItanya (approved):

PR #2194, opened 2026-07-09
fix: parse migration filenames the same way golang-migrate does

mesutoezdil:
sourceFileVersions and maxEmbeddedVersion both used custom parsing that accepted filenames golang-migrate would skip, like one with no underscore after the number. Both now use source.DefaultParse so they can't drift from what golang-migrate actually runs.

### DIFF
--- a/go/core/pkg/cli/db/migrate/migrate.go
+++ b/go/core/pkg/cli/db/migrate/migrate.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/source"
 	"github.com/spf13/cobra"
 
 	"github.com/kagent-dev/kagent/go/core/pkg/migrations"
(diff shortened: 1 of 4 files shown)


EItanya:
@iplay88keys

iplay88keys:
It would be good to align the approaches, but this is the wrong direction. If we go with this approach, something like  `000007test.up.sql` would be counted, but `golang-migrate` would silently skip that. The `golang-migrate` regex [requires underscores](https://github.com/golang-migrate/migrate/blob/89e308c340a2c069c3260f268307d38a7db88227/source/parse.go#L22), but [skips](https://github.com/golang-migrate/migrate/blob/89e308c340a2c069c3260f268307d38a7db88227/source/iofs/iofs.go#L60-L63) any files that don't match that.

So I think the right approach would be to call `golang-migrate's` `source.DefaultParse(name)` directly to make sure that it can't drift.

iplay88keys:
Just wanted to follow up on this as I saw you had pushed a change:

1. `maxEmbeddedVersion` in `runner.go` still uses `fmt.Sscanf` and should also be switched to use `source.DefaultParse`.
2. The PR title/body are stale at this point.

Thanks!

mesutoezdil:
> Just wanted to follow up on this as I saw you had pushed a change:
> 
> 1. `maxEmbeddedVersion` in `runner.go` still uses `fmt.Sscanf` and should also be switched to use `source.DefaultParse`.
> 2. The PR title/body are stale at this point.
> 
> Thanks!

thx, it is done now

iplay88keys (approved):

iplay88keys:
Sorry, I meant to merge this in yesterday. If you wouldn't mind doing `merge` commits instead of rebasing + force pushing, that would make it a bit clearer what (if anything has changed).

mesutoezdil:
> Sorry, I meant to merge this in yesterday. If you wouldn't mind doing `merge` commits instead of rebasing + force pushing, that would make it a bit clearer what (if anything has changed).

Sure, I rebase to keep up with main since the repo moves fast, but I’ll switch to merge commits for that from now on 👍🏽

PR #2201, opened 2026-07-10
feat(helm): allow custom annotations on the default ModelConfig

### LINKED ISSUE #2200: [FEATURE] Allow custom annotations on the Helm-generated default ModelConfig
### 📋 Prerequisites

### 📝 Feature Summary

Add a values knob to set arbitrary metadata.annotations on the ModelConfig rendered by helm/kagent/templates/modelconfig.yaml.

The Helm chart renders a default ModelConfig (name from `kagent.defaultModelConfigName`) from `providers[.Values.providers.default]`. Its metadata currently emits only `name`, `namespace`, and `labels` -- there is no way to attach annotations to it via values.

Downstream distributions and platform integrations often need to attach their own annotations to this object. In our case we run a UI in front of kagent that reads a display-name annotation off the ModelConfig to render a friendly model label; without an annotations knob there is no supported way to set it on the chart-generated default ModelConfig.

Today the only workarounds are to define a *separate* ModelConfig outside the chart (leaving the chart's default ModelConfig unused but still created — it has no disable toggle), or to post-patch a chart-owned object across release boundaries, which fights the Helm/Flux owner on every reconcile.

Anyone packaging kagent under their own platform, or integrating with tooling that keys off resource annotations, would benefit.

### 💡 Proposed Solution

Add an optional `annotations` map on the provider entry and render it onto the ModelConfig metadata. Per-provider fits the current template, since the default ModelConfig is derived from the single `$model := index .Values.providers .Values.providers.default` object it already reads.

Template change in helm/kagent/templates/modelconfig.yaml:

```nohighlight
    metadata:
      name: {{ include "kagent.defaultModelConfigName" $dot | quote }}
      namespace: {{ include "kagent.namespace" . }}
      labels:
        {{- include "kagent.labels" $dot | nindent 4 }}
      {{- with $model.annotations }}
      annotations:
        {{- toYaml . | nindent 4 }}
      {{- end }}
```

Usage:

```yaml
    providers:
      anthropic:
        provider: Anthropic
        model: claude-sonnet-4-6
        annotations:
          example.com/display-name: "Claude Sonnet 4.6"
```

### 🔄 Alternatives Considered

- Define a separate, self-managed ModelConfig instead of using the chart's default. Works, but the chart still creates its (now unused) default ModelConfig — it is rendered unconditionally with no disable toggle -- so operators end up with a stray, un-annotated object.
- Post-render / Flux post-build patch to inject the annotation into the chart-owned object. Crosses release/owner boundaries and conflicts with Helm's field manager on each reconcile.
- A dedicated top-level `defaultModelConfig.annotations` block instead of per-provider. Also fine; per-provider was chosen as the smaller change that travels with the existing `$model` object.
- Custom metadata (labels and annotations) for all generated resources.

### 🎯 Affected Service(s)

Not Sure

### 📚 Additional Context

Current template (`helm/kagent/templates/modelconfig.yaml` on main) emits only name/namespace/labels in metadata; spec reads provider, model, apiKeySecretRef, apiKeySecretKey, defaultHeaders, tls, config.

marians:
## What

Adds an optional per-provider `annotations` map that is rendered onto the metadata of the Helm-generated default ModelConfig (the one derived from `providers.default`).

## Why

Today `helm/kagent/templates/modelconfig.yaml` emits only `name`/`namespace`/`labels` — there is no supported way to attach annotations to the default ModelConfig via values. Downstream distributions and tooling that key off resource annotations (e.g. a UI reading a display-name annotation) currently have to either post-render the chart-owned object or define a separate, self-managed ModelConfig (leaving the chart's unconditional default ModelConfig created but unused). See #2200.

## How

```yaml
providers:
  anthropic:
    provider: Anthropic
    model: claude-haiku-4-5
    annotations:
      example.com/display-name: "Claude Haiku 4.5"
```

The map is generic (no vendor-specific keys) and omitted entirely when unset.

## Testing

`helm template` verified in two cases: with `annotations` set the block renders under `metadata.annotations`; without it the output is byte-identical to before (backward-compatible).

Closes #2200


helm/kagent/values.yaml
```diff
@@ -486,6 +486,12 @@ providers:
     #   caCertSecretRef: ""
     #   caCertSecretKey: ""
     #   disableSystemCAs: false
+    # Optional annotations added to the generated default ModelConfig's metadata.
```
EItanya: Can we add this key under `providers` so that we can have a default `annotations: {}` key to show the API?

EItanya (changes requested):

helm/kagent/values.yaml
marians: Sure. Done in https://github.com/kagent-dev/kagent/pull/2201/commits/3479892cb4e99bc95df274b19d7497babdacf63d. Thanks for the review!

EItanya: You did not do what i asked, i asked to move the key up a level

marians: A misunderstanding, sorry! Please check again.

EItanya:
Last thing, please add a helm unit test

EItanya (approved):

marians:
Thanks for merging!

PR #2202, opened 2026-07-10
feat(go-adk): add per-call session isolation for Agent tools

dimetron:
Add an isolateSessions flag to Agent tools so a declarative coordinator can fan out parallel, isolated sub-agent calls: when set, a fresh A2A context_id is minted per call instead of reusing one stable id for the tool's lifetime.

Also replace the per-branch map[string]any tool return with a single typed remoteA2AResponse so subagent_session_id is reported on every outcome (success, input_required, failure, non-terminal, no-result). processResult now switches explicitly on the terminal task state and passes through child-provided structured failure fields with sub-agent task linkage.

- v1alpha2 Tool.IsolateSessions (*bool, Agent-only via CEL) -> adk RemoteAgentConfig.IsolateSessions via the translator; Python config parity.
- NewKAgentRemoteA2ATool takes a RemoteA2AToolConfig struct; isolated tools return an empty stamp id and add no subagentSessionIDs entry (UI links via the per-call id in the function_response instead).
- Unit tests for contextIDForCall, every processResult branch, structured failure passthrough, and the translator wiring.

Design: specs/a2a_sessions/EP_2153.md

### DIFF
--- a/design/EP-2153-a2a-session-isolation.md
+++ b/design/EP-2153-a2a-session-isolation.md
@@ -1,0 +1,230 @@
+# EP-2153: Per-call session isolation + robust sub-agent session reporting
+
+* Issue: [#2153](https://github.com/kagent-dev/kagent/issues/2153)
+* Upstream PR: https://github.com/kagent-dev/kagent/pull/2153
+
+Status: implemented
+
+## Goal
+
+Let a declarative agent fan out **parallel, isolated** sub-agent calls, and make
+each of those calls **observable** in the UI by always reporting the sub-agent
+session it ran in.
+
+Two capabilities, delivered together:
+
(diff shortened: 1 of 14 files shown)


dimetron:
### Rebased on `main` — merge conflict resolved

This was conflicting in `go/adk/pkg/tools/remote_a2a_tool.go`. Rebased onto `main` (single commit, signed off), so the PR is **mergeable** again. The design doc moved from `specs/a2a_sessions/EP_2153.md` to `design/EP-2153-a2a-session-isolation.md` to match the repo layout.

`go build ./...`, `go vet ./...` and `go test ./adk/... ./api/...` are clean against current `main`.

Not stale — please take another look when you have a moment.

PR #2204, opened 2026-07-10
fix(mcp): include SandboxAgents in MCP agent listing

### LINKED ISSUE #2123: [BUG] MCP list_agents / invoke_agent exclude all SandboxAgents (only lists Agent CRs, hardcodes reason=="DeploymentReady")
### 📋 Prerequisites

### 🎯 Affected Service(s)

### 🚦 Impact/Severity

Blocker

### Summary

Substrate-backed `SandboxAgent`s that are fully `Ready` (golden actor built, `Accepted=True`, reconciled into the kagent DB) are **never returned by the MCP `list_agents` tool and can't be reached via `invoke_agent`** — `invoke_agent` returns `agent <ns>/<name> not found or not ready`. The kagent REST API (`/agents`) and UI list them correctly; only the MCP path drops them.

Root cause is in `MCPHandler.listReadyAgents` at `go/core/internal/mcp/mcp_handler.go` (current `main`):

```go
// listReadyAgents returns agents that are accepted and deployment-ready.
func (h *MCPHandler) listReadyAgents(ctx context.Context) ([]AgentSummary, error) {
	agentList := &v1alpha2.AgentList{}                       // (a) Agent CRs only
	if err := h.kubeClient.List(ctx, agentList); err != nil {
		return nil, err
	}
	agents := make([]AgentSummary, 0, len(agentList.Items))
	for _, agent := range agentList.Items {
		deploymentReady := false
		accepted := false
		for _, condition := range agent.Status.Conditions {
			if condition.Type == "Ready" && condition.Reason == "DeploymentReady" && condition.Status == "True" { // (b)
				deploymentReady = true
			}
			if condition.Type == "Accepted" && condition.Status == "True" {
				accepted = true
			}
		}
		if !accepted || !deploymentReady {
			continue
		}
		...
	}
}
```

Two compounding problems:

- **(a) Wrong CR type — `SandboxAgent`s are never listed.** It lists `v1alpha2.AgentList` only; `SandboxAgentList` is never enumerated. So every substrate SandboxAgent is invisible to `listReadyAgents` regardless of status.
- **(b) Hardcoded `condition.Reason == "DeploymentReady"`.** Declarative `Agent`s report `Ready` with reason `DeploymentReady`; substrate `SandboxAgent`s report `Ready` with reason `WorkloadReady` (set by `reconcileSandboxAgentStatus` → `AgentReadyReasonWorkloadReady` when `sandboxBackend.ComputeReady` returns True). So even if `SandboxAgent` were listed, this reason check would still drop it.

`listReadyAgents` feeds both `handleListAgents` (the `list_agents` tool) and `readAgentsResource` (the `kagent://agents` resource), so both the tool and the resource are affected.

### 🔄 Steps To Reproduce

1. Install kagent 0.9.10 with substrate enabled and a `SandboxAgent` whose golden actor builds successfully (e.g. the `k8s-agent` from `examples/substrate-openclaw`, runtime `go`).
2. Wait until `kubectl get sandboxagent <name>` shows `READY=True` (reason `WorkloadReady`) and `ACCEPTED=True` (reason `Reconciled`). The agent is in the kagent Postgres `agents` table and appears in the REST `/agents` API + UI.
3. Call the MCP `list_agents` tool → the SandboxAgent is absent. Call `invoke_agent` with `<ns>/<name>` → `Failed to send A2A message: agent <ns>/<name> not found or not ready`.

```
# declarative Agent
$ kubectl get agent <decl-agent> -o jsonpath='{.status.conditions[?(@.type=="Ready")].reason}'
DeploymentReady
# substrate SandboxAgent
$ kubectl get sandboxagent <sub-agent> -o jsonpath='{.status.conditions[?(@.type=="Ready")].reason}'
WorkloadReady
```

### 🤔 Expected Behavior

MCP `list_agents` / `invoke_agent` treat a `SandboxAgent` that is `Accepted=True` and `Ready=True` as invocable, the same as a declarative `Agent`.

### 📱 Actual Behavior

_No response_

### 💻 Environment

- kagent: v0.9.10 (also reproduced on `main` — `listReadyAgents` is unchanged there)
- Substrate: v0.0.7
- Run on Kubernetes 1.36.2 (homelab), substrate installed into a non-`ate-system` namespace

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

This was the final blocker after fixing a stack of `ate-system`-hardcoding bugs (conn-spec, jwks-url, atelet-discovery namespace, env-source RBAC namespace, `SandboxConfig`/`runsc`, and a Cilium egress policy) that had kept the SandboxAgents at `Ready=False`. Once those were resolved the agents reached `Ready=True` but remained un-invocable via MCP, which traced to this `listReadyAgents` limitation.

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

lubingtan:
## Summary
- include Ready/Accepted SandboxAgent resources in MCP `list_agents` and `kagent://agents`
- accept both `DeploymentReady` and `WorkloadReady` Ready reasons when filtering MCP-visible agents
- let MCP `invoke_agent` resolve SandboxAgent clients registered under the sandbox route key

Fixes #2123.

## Reproduction
Added a focused MCP unit test that fails on main: a regular Agent is listed, but an Accepted/Ready SandboxAgent with `WorkloadReady` is omitted.

## Tests
- `GOCACHE=/tmp/go-build GOFLAGS=-buildvcs=false go test ./internal/mcp -run TestListReadyAgentsIncludesSandboxAgents -count=1`
- `GOCACHE=/tmp/go-build GOFLAGS=-buildvcs=false go test ./internal/a2a -run TestAgentClientRegistrySendMessageFallsBackToSandboxRoute -count=1`
- `GOCACHE=/tmp/go-build GOFLAGS=-buildvcs=false go test ./internal/a2a -count=1`
- `GOCACHE=/tmp/go-build GOFLAGS=-buildvcs=false go test ./internal/mcp -count=1`


go/core/internal/a2a/agent_client_registry_test.go
```diff
@@ -0,0 +1,59 @@
...
+	a2aclient "github.com/a2aproject/a2a-go/v2/a2aclient"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAgentClientRegistrySendMessageFallsBackToSandboxRoute(t *testing.T) {
+	called := false
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		called = true
+		var rpcReq map[string]any
+		require.NoError(t, json.NewDecoder(r.Body).Decode(&rpcReq))
```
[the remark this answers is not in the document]
lubingtan: Updated in 0b81aba1: `called` is now an `atomic.Bool`, with the handler using `called.Store(true)` so the test no longer has an unsynchronized cross-goroutine write/read. Verified with `go test -race ./internal/a2a -run TestAgentClientRegistrySendMessageFallsBackToSandboxRoute -count=1` and `go test -race ./internal/a2a -count=1`.

go/core/internal/a2a/agent_client_registry_test.go
```diff
@@ -0,0 +1,59 @@
...
+	registry.set(routeKey(true, "default", "sandbox-agent"), client)
+
+	_, err = registry.SendMessage(
+		context.Background(),
+		"default",
+		"sandbox-agent",
+		&a2atype.SendMessageRequest{Message: a2atype.NewMessage(a2atype.MessageRoleUser, a2atype.NewTextPart("hello"))},
+	)
+	require.NoError(t, err)
+	require.True(t, called)
```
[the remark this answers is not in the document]
lubingtan: Updated in 0b81aba1: the assertion now reads the atomic value via `called.Load()`. I also ran the focused test and the full `internal/a2a` package with `-race`.

EItanya (changes requested):
Unfortunately this only masks the problem that exists, and we really need to figure out how to really fix this. If an `Agent` exists with that `Name`, `Namespace` it will return that even if you intended to get the `SandboxAgent`. I think we need to start including the `GroupKind` in order to properly key the entries

lubingtan:
@EItanya thanks, that makes sense. I agree the current fallback from `namespace/name` to the sandbox route can select the wrong target when an `Agent` and `SandboxAgent` share the same namespace/name.

Before pushing another revision, I want to confirm the shape I am thinking about:

1. `list_agents` / `kagent://agents` should return each entry with both `ref` and `groupKind`, for example:
   - `ref: "default/my-agent", groupKind: "Agent.kagent.dev"`
   - `ref: "default/my-agent", groupKind: "SandboxAgent.kagent.dev"`
2. `invoke_agent` should accept an optional `groupKind` argument. If omitted, it defaults to `Agent.kagent.dev` for backward compatibility with existing clients.
3. `AgentClientRegistry` should route by `(groupKind, namespace, name)` instead of falling back from the regular Agent key to the SandboxAgent key.
4. Tests should cover the same namespace/name collision case and assert that `groupKind: SandboxAgent.kagent.dev` invokes the SandboxAgent backend, not the regular Agent backend.

Does that match what you had in mind, or would you prefer a different MCP input shape (for example encoding the kind directly into the `agent` ref instead of a separate `groupKind` field)?

lubingtan:
@EItanya implemented the GroupKind-based routing in [`2d3b17b4`](https://github.com/LuBingtan/kagent/commit/2d3b17b4066b5689814b066d93c393cef142c08a).

- `list_agents` and `kagent://agents` now return `groupKind`.
- `invoke_agent` accepts `groupKind`, defaulting to `Agent.kagent.dev` for backward compatibility.
- The registry now selects the exact `(groupKind, namespace, name)` route and no longer falls back to a SandboxAgent route.
- The regression test covers an `Agent` and `SandboxAgent` with the same namespace/name and verifies that the SandboxAgent backend is selected.

I also checked that `routeKeyForGroupKind` delegates to the existing `routeKey`, preserving the current route keys: `namespace/name` for `Agent.kagent.dev` and `sandboxes/namespace/name` for `SandboxAgent.kagent.dev`. This matches the A2A registrar, which derives the same keys from workload mode.

One remaining extension point: if a new `AgentObject` kind is added, or an existing kind can map to a different workload mode, its GroupKind-to-route mapping must be added explicitly. Unsupported kinds return an error instead of silently falling back to a potentially wrong target.

Verified with `go test ./internal/a2a -count=1` and `go test ./internal/mcp -count=1`.

go/core/internal/a2a/agent_client_registry.go
```diff
@@ -38,14 +39,43 @@ func (r *AgentClientRegistry) Register(namespace, name string, c *a2aclient.Clie
...
+// RegisterForGroupKind adds or replaces the A2A client for the given agent GroupKind.
+func (r *AgentClientRegistry) RegisterForGroupKind(groupKind, namespace, name string, c *a2aclient.Client) error {
+	key, err := routeKeyForGroupKind(groupKind, namespace, name)
+	if err != nil {
+		return err
+	}
+	r.set(key, c)
+	return nil
+}
+
```
EItanya: I don't like adding functionality like this just for tests, and it seems to be only called by the tests

go/core/internal/a2a/agent_client_registry.go
```diff
@@ -38,14 +39,43 @@ func (r *AgentClientRegistry) Register(namespace, name string, c *a2aclient.Clie
...
 // SendMessage invokes an agent directly via its cached A2A client.
 func (r *AgentClientRegistry) SendMessage(ctx context.Context, namespace, name string, req *a2atype.SendMessageRequest) (a2atype.SendMessageResult, error) {
-	key := namespace + "/" + name
+	return r.SendMessageForGroupKind(ctx, schema.GroupKind{Group: "kagent.dev", Kind: "Agent"}.String(), namespace, name, req)
+}
+
+// SendMessageForGroupKind invokes an agent directly via its cached A2A client.
+func (r *AgentClientRegistry) SendMessageForGroupKind(ctx context.Context, groupKind, namespace, name string, req *a2atype.SendMessageRequest) (a2atype.SendMessageResult, error) {
```
EItanya: This is a classic `claudism`. Let's just break the `SendMessage` API and add the new param

EItanya (changes requested):

go/core/internal/a2a/agent_client_registry.go
lubingtan: Updated in a6e6be8d: removed RegisterForGroupKind. The production registrar continues to register its workload-derived route key via set, and the A2A test uses that internal route setup directly, so no test-only registration API remains.

go/core/internal/a2a/agent_client_registry.go
lubingtan: Updated in a6e6be8d: removed SendMessageForGroupKind and changed SendMessage to require groupKind directly. The MCP handler now calls the revised API. Verified with go test ./internal/a2a ./internal/mcp -count=1 and the focused A2A route test under -race.

lubingtan:
Hi @EItanya would you pls take a look again? thanks!

go/core/internal/a2a/agent_client_registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+			ProtocolBinding: a2atype.TransportProtocolJSONRPC,
+			ProtocolVersion: a2atype.Version,
+		}},
+		a2aclient.WithJSONRPCTransport(&http.Client{}),
+	)
+	require.NoError(t, err)
+
+	registry := NewAgentClientRegistry()
+	sandboxGroupKind := schema.GroupKind{Group: "kagent.dev", Kind: "SandboxAgent"}.String()
+	registry.set(routeKey(true, "default", "sandbox-agent"), client)
```
iplay88keys: Could we make this the collision case by registering both a regular Agent client and a SandboxAgent client with the same namespace/name, then asserting each `GroupKind` reaches the correct backend? With only the sandbox key present, this test doesn’t protect against accidentally preferring the regular Agent when both exist which is the case that prompted GroupKind routing.

go/core/internal/mcp/mcp_handler_test.go
```diff
@@ -82,6 +85,54 @@ func TestListAgentsInputSchemaHasProperties(t *testing.T) {
...
+		},
+		Status: v1alpha2.AgentStatus{
+			Conditions: []metav1.Condition{
+				{Type: v1alpha2.AgentConditionTypeAccepted, Status: metav1.ConditionTrue},
+				{Type: v1alpha2.AgentConditionTypeReady, Status: metav1.ConditionTrue, Reason: reconciler.AgentReadyReasonDeploymentReady},
+			},
+		},
+	}
+	sandboxAgent := &v1alpha2.SandboxAgent{
+		ObjectMeta: metav1.ObjectMeta{Name: "sandbox-agent", Namespace: "default"},
```
iplay88keys: Could the Agent and SandboxAgent use the same name here? That would verify `list_agents` preserves both colliding refs and distinguishes them by `groupKind`.

shrinedogg:
@LuBingtan do you need any assistance with this PR? I'm sure there's a larger impact than just me for this bug, so I'm happy to help, if I can!

go/core/internal/a2a/agent_client_registry_test.go
lubingtan: Updated in 65624710: the test now registers regular Agent and SandboxAgent clients under the same namespace/name with separate backends. It invokes each GroupKind and asserts that only the matching backend receives the request.

go/core/internal/mcp/mcp_handler_test.go
lubingtan: Updated in 65624710: both resources now use default/shared-agent. The expected list contains two entries with the same ref and distinct GroupKind values.

lubingtan:
@shrinedogg thanks, I really appreciate the offer. I missed the most recent review comments, so I did not follow up with the updates as promptly as I should have.

I have now rebased the PR and addressed the feedback, including collision coverage for an `Agent` and `SandboxAgent` with the same namespace/name. If you have a setup that reproduces the issue, testing this branch and sharing feedback on the `groupKind` MCP interface would be very helpful.

lubingtan:
@EItanya  @iplay88keys Please help to take a look again when you have a chance, thanks!

iplay88keys:
@LuBingtan, since #2422 removed the deployment-backed Agent API on main, MCP now operates only on `v1alpha3` `SandboxAgents` and this fix is no longer applicable there. The original bug still exists on r`elease/v0.10.x`. Could we close this PR and open a focused backport against `release/v0.10.x`? Retargeting the current branch directly would include substantial unrelated main-branch changes.

lubingtan:
Closing this PR because the current main branch no longer has the deployment-backed Agent API.

PR #2205, opened 2026-07-10
Gate SandboxAgent actor readiness on the app serving traffic

iplay88keys:
# Description
Gates SandboxAgent actor readiness on the agent actually serving HTTP traffic, instead of only on a golden snapshot existing.

`buildSandboxAgentActorTemplate` now sets the substrate `Readyz` probe on the kagent container:

- Path: `/.well-known/agent-card.json` — the same path the Kubernetes Deployment readiness probe already uses for declarative and BYO agents, and the standard A2A agent-card location. The path is set explicitly because substrate's `Readyz` default (`/readyz`) is not served by either ADK runtime or by typical BYO A2A servers.
- Port: `80` (`substrateKagentListenPort`), the port substrate routes actor traffic to.

With `Readyz` set, substrate does not report the actor ready — and `Run`/`Restore` RPCs do not return success — until the container answers 200. A SandboxAgent whose app crashes on startup or listens on the wrong port now stays `Ready=False` with the ActorTemplate never reaching `PhaseReady`, instead of reporting `Ready=True` with a golden snapshot that captured only the pause container and a first chat that fails or hangs.

# Behavior changes (not breaking)

- **One-time golden rebuild on upgrade.** The ActorTemplate name/config-hash is unchanged, so existing SandboxAgents roll through the spec-drift recreate path once: the template is deleted and recreated, and a fresh golden snapshot is built on the next reconcile. This is a single rebuild per existing agent, not a hot loop.
- **Previously-"Ready" broken agents become NotReady.** Any agent that reported `Ready=True` but never actually served traffic on port 80 will now (correctly) report `Ready=False` with an `ActorTemplateNotReady` reason.
- Known interaction: on current substrate, a golden build that fails (because the app never becomes ready) can hold a worker slot until capacity is reclaimed. That is a pre-existing substrate behavior for failed golden builds; this change makes such failures visible at build time instead of at first chat.


EItanya (dismissed):
This brings up another issue that we need e2e tests for this functionality

EItanya (approved):

PR #2207, opened 2026-07-10
fix(db): cascade soft-delete tasks and events on session delete

mesutoezdil:
deleting a session left related task, event, and session_share rows untouched. this wraps DeleteSession in a transaction that soft-deletes tasks and events and hard-deletes session shares before marking the session itself as deleted.

also adds a migration making session id unique among live sessions. ids were only unique per user, so two users could hold the same session id and deleting one would have cascaded into the other user's tasks and events.


go/core/internal/database/client_postgres.go
```diff
@@ -146,7 +152,24 @@ func (c *postgresClient) ListSessionsForAgentAllUsers(ctx context.Context, agent
 }
 
 func (c *postgresClient) DeleteSession(ctx context.Context, sessionID, userID string) error {
-	return c.q.SoftDeleteSession(ctx, dbgen.SoftDeleteSessionParams{ID: sessionID, UserID: userID})
+	return c.withTx(ctx, func(q *dbgen.Queries) error {
+		if _, err := q.GetSession(ctx, dbgen.GetSessionParams{ID: sessionID, UserID: userID}); err != nil {
```
EItanya, line 170: Any reason we're not using CASCADE deletes via SQL for this?

mesutoezdil: > Any reason we're not using CASCADE deletes via SQL for this?

no fk exists between session and task/event/session_share right now
even if it did, tasks and events r soft deleted, and ON DELETE CASCADE only fires on a real DELETE and not an UPDATE.       
and session itself is soft deleted too, so cascade would never trigger
that's why this is explicit app code instead.

mesutoezdil:
closing. go/core/internal/service/session is gone on main, the legacy session/task/event service was removed as part of the v2 rewrite (conversations api, #2598). the fix this covers no longer applies to code that exists.

PR #2209, opened 2026-07-11
fix(helm): make the cluster DNS domain configurable

### LINKED ISSUE #1767: [BUG] Cluster domain assumed to be cluster.local in helmcharts etc
### 📋 Prerequisites

### 🎯 Affected Service(s)

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

I'm configuring my k3s setup with `mycluster.k8s.somedomain.net` as my `cluster-domain`.

This means `servicename.somenamespace.svc.cluster.local` does not resolve, but `servicename.somenamespace.svc.mycluster.k8s.somedomain.net` and `servicename.somenamespace.svc` do.

In a minimal deployment from Helm chart, I manually edited the resulting `Deployment`s and removed `cluster.local` in `NEXT_PUBLIC_BACKEND_URL` and `POSTGRES_DATABASE_URL`. This complicates pushes, and would require hacks upon an automated push such as via ArgoCD.

Ideally either the helmcharts would expose the clusterdomain in values yaml, or they'd omit it and depend on search-domains being set up correctly in CoreDNS (as I did above).

I did not yet get to a functioning system, and I'll postpone it for later, so I don't know if this assumption about `cluster-domain` propagated anywhere else, but a quick search revealed at least one yaml file in `tools` dir to mention `cluster.local`.

### 🔄 Steps To Reproduce

```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  namespace: kagent
  name: kagent-crds
spec:
  targetNamespace: kagent
  createNamespace: true
  version: "0.9.0"  # https://github.com/kagent-dev/kagent/pkgs/container/kagent%2Fhelm%2Fkagent-crds
  repo: ghcr.io/kagent-dev
  chart: kagent/helm/kagent-crds
  valuesContent: |
    # n/a

---

apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  namespace: kagent
  name: kagent
spec:
  targetNamespace: kagent
  createNamespace: true
  version: "0.9.0"  # https://github.com/kagent-dev/kagent/pkgs/container/kagent%2Fhelm%2Fkagent
  repo: ghcr.io/kagent-dev
  chart: kagent/helm/kagent
  valuesContent: |
    providers:
      default: ollama
      # default: openAI
      openAI:
        apiKey: 12345
      # gemini, anthropic, azureOpenAI

    controller:
      env: {}
      # envFrom: [ { secretRef: { name: controller-secrets } } ]

    ipv6:
      enabled: true
```

Note, this is actually processed by ArgoCD appset, but applying on k3s with `kubectl apply -f` should have a comparable result. As noted, k3s has been initialized with a custom `cluster-domain`.

### 🤔 Expected Behavior

_No response_

### 📱 Actual Behavior

_No response_

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

shekhargit1912:
The chart hardcoded `cluster.local` when building in-cluster service URLs, so installs on clusters configured with a different DNS domain rendered unreachable addresses that had to be patched by hand after every upgrade.

Add a `clusterDomain` value (default `cluster.local`) and use it to build the A2A base URL, the nginx controller upstream, and the bundled PostgreSQL connection string. Existing installs are unaffected: when the value is unset the templates render exactly as before.

Fixes #1767


EItanya (commented):
Rather than adding this, should we just remove the DNS domain completely. Usually just `name.namespace` will do the trick

shekhargit1912:
> Rather than adding this, should we just remove the DNS domain completely. Usually just `name.namespace` will do the trick
@EItanya
Good point - short names resolve in-cluster via the resolv.conf search domains, so dropping the suffix would fix #1767 without any new config. Before I switch the PR over: are any of these three URLs (the A2A base URL, the nginx upstream authority, and the Postgres connection string) ever consumed outside the pod's DNS context, or do they all resolve strictly in-cluster? If they're all in-cluster, I'm happy to drop clusterDomain entirely and just use name.namespace.svc. Want me to rework it that way?

EItanya:
> > Rather than adding this, should we just remove the DNS domain completely. Usually just `name.namespace` will do the trick
> > @EItanya
> > Good point - short names resolve in-cluster via the resolv.conf search domains, so dropping the suffix would fix #1767 without any new config. Before I switch the PR over: are any of these three URLs (the A2A base URL, the nginx upstream authority, and the Postgres connection string) ever consumed outside the pod's DNS context, or do they all resolve strictly in-cluster? If they're all in-cluster, I'm happy to drop clusterDomain entirely and just use name.namespace.svc. Want me to rework it that way?

AFAIK they're all in cluster, `cluster.local` marks it that way.

shekhargit1912:
> > > Rather than adding this, should we just remove the DNS domain completely. Usually just `name.namespace` will do the trick
> > > @EItanya
> > > Good point - short names resolve in-cluster via the resolv.conf search domains, so dropping the suffix would fix #1767 without any new config. Before I switch the PR over: are any of these three URLs (the A2A base URL, the nginx upstream authority, and the Postgres connection string) ever consumed outside the pod's DNS context, or do they all resolve strictly in-cluster? If they're all in-cluster, I'm happy to drop clusterDomain entirely and just use name.namespace.svc. Want me to rework it that way?
> 
> AFAIK they're all in cluster, `cluster.local` marks it that way.
@EItanya 
Thanks for confirming! I'll rework the PR to drop clusterDomain entirely and switch the A2A base URL, nginx upstream, and Postgres host to the name.namespace.svc short form. That fixes the original issue on clusters with custom DNS domains without adding a new value. Will push the update shortl

shekhargit1912:
> AFAIK they're all in cluster, `cluster.local` marks it that way.
-
@EItanya 
Thanks for confirming! Reworked as suggested - dropped clusterDomain entirely and switched the A2A base URL, nginx upstream, and bundled Postgres URL to the name.namespace.svc short form, so they resolve on any cluster regardless of its DNS domain. Tests updated accordingly

EItanya (dismissed):

shekhargit1912:
@EItanya thanks for merging main in! The helm-unit-tests failure was the ui-nginx-configmap test still expecting the cluster.local suffix in the nginx upstream - fixed in 1f63fe5 to match the new name.namespace.svc short form. All helm tests pass locally now

shekhargit1912:
@EItanya sorry - my test fix in 1f63fe5 landed right after your approval and dismissed it as stale. The only change since your review is that one-line nginx test assertion, and all 31 checks are green now. When you get a chance, could you re-approve? Thanks!

EItanya (approved):

PR #2224, opened 2026-07-13
feat (UI): improve UX by grouping tool calls in the chat

peterj:
Group the tool calls in the chat for better readability of the chat. The expanded view is the same as what we currently have, the new part is the grouping of the tool calls.

image


dimetron:
@peterj - What about subagent Tool Call and MCP App calls? Should we separated those ??

peterj:
i’ll check the sub agents - the mcp app calls are rendered outside of the
group

EItanya (approved):

PR #2229, opened 2026-07-13
fix: render solid rule borders in disabled-SSL warning banner

anxkhn:
## What

When SSL verification is disabled, `create_ssl_context` (in
`python/packages/kagent-adk/src/kagent/adk/models/_ssl.py`) logs a warning banner
that is meant to be framed by three 60-character rules of `=`. Because of an
operator-precedence issue those rules render as a jagged column of single `=`
signs instead of solid lines. This makes a security-relevant warning look
malformed in the logs. This PR fixes the borders and adds a regression test.

## Why it happens

Each rule is written as a string literal immediately followed by `"=" * 60`, for
example:

```python
logger.warning(
    "\n"
    "=" * 60 + "\n"
    ...
)
```

Python applies implicit string-literal concatenation *before* the `*` repetition,
so `"\n" "=" * 60` parses as `("\n=") * 60`, i.e. sixty `\n=` pairs, rather than a
newline followed by a 60-character rule. All three rules in the banner (top,
separator, and bottom) are affected, so the logged warning looks like this:

```
=
=
=
... (60 rows)
⚠️  SSL VERIFICATION DISABLED ⚠️
=
=
... (60 rows)
```

## The fix

Add an explicit `+` before each `"=" * 60` so the repetition applies to `"="`
alone. All three borders now render as solid 60-character rules:

```
============================================================
⚠️  SSL VERIFICATION DISABLED ⚠️
============================================================
SSL certificate verification is disabled.
This should ONLY be used in development/testing.
Production deployments MUST use proper certificates.
============================================================
```

The change is limited to the string framing; no logging behavior, return value,
or TLS logic changes.

## Testing

Added `test_ssl_context_disabled_warning_banner_borders`, which captures the
warning via `caplog` and asserts the banner contains exactly three solid
60-character `=` rules and no stray single-`=` rows. It fails on the current code
and passes with the fix.

```
cd python
uv run pytest packages/kagent-adk/tests/unittests/models/test_ssl.py -q
# 7 passed
```

`ruff check` and `ruff format --diff` are clean on both changed files.


peterj:
makes sense, but we probably don't really need a unit test for this

peterj (approved):

anxkhn:
Hey @peterj can you have a look at it again?

PR #2232, opened 2026-07-13
fix(crewai): align memory storage save with LongTermMemory contract

anxkhn:
### What

`KagentMemoryStorage.save` names its third parameter `timestamp`, but CrewAI's
`LongTermMemory` invokes its storage backend with that value passed as the
keyword argument `datetime`:

```python
# crewai/memory/long_term/long_term_memory.py (LongTermMemory.save)
self.storage.save(
    task_description=item.task,
    score=metadata["quality"],
    metadata=metadata,
    datetime=item.datetime,
)
```

The reference backend `LTMSQLiteStorage.save(self, task_description, metadata,
datetime, score)` names that parameter `datetime`, so any storage that plugs
into `LongTermMemory` has to use the same keyword. Because `KagentMemoryStorage`
uses `timestamp` instead, every long-term-memory write raises:

```
TypeError: KagentMemoryStorage.save() got an unexpected keyword argument 'datetime'
```

`_executor.py` wires `KagentMemoryStorage` as the crew's `LongTermMemory`
backend when memory is enabled, so this fires on any memory-enabled CrewAI crew
(`crew.memory=True`) the first time it persists a long-term-memory item. The
exception propagates out of `LongTermMemory.save` after a `MemorySaveFailedEvent`
is emitted.

### Fix

Rename the parameter from `timestamp` to `datetime` so the signature matches the
storage contract exactly. This is a two-line change and also aligns `save` with
the sibling `load`, which already reads and writes the `datetime` key.

### Tests

Added `packages/kagent-crewai/tests/test_memory.py`:

- `test_save_signature_matches_crewai_storage_contract` asserts
  `KagentMemoryStorage.save`'s parameters match the reference
  `LTMSQLiteStorage.save`.
- `test_long_term_memory_save_posts_datetime` drives a real
  `LongTermMemory(KagentMemoryStorage(...)).save(LongTermMemoryItem(...))` with
  the HTTP client mocked and asserts the item's datetime is forwarded to the
  backend in `memory_data["datetime"]`.

A new `tests/conftest.py` sets `KAGENT_URL` / `KAGENT_NAME` / `KAGENT_NAMESPACE`
defaults so the package imports without a running backend. Both tests fail on
`main` (the second with the exact upstream `TypeError`) and pass with the fix.

Run:

```
cd python
uv run pytest packages/kagent-crewai/tests/
```

### DIFF
--- a/python/packages/kagent-crewai/src/kagent/crewai/_memory.py
+++ b/python/packages/kagent-crewai/src/kagent/crewai/_memory.py
@@ -26,7 +26,7 @@ class KagentMemoryStorage:
         self.user_id = user_id
         self.base_url = base_url
 
-    def save(self, task_description: str, metadata: dict, timestamp: str, score: float) -> None:
+    def save(self, task_description: str, metadata: dict, datetime: str, score: float) -> None:
         """
         Saves a memory item to the Kagent backend.
         The agent_id is expected to be in the metadata.
@@ -39,7 +39,7 @@ class KagentMemoryStorage:
                 "task_description": task_description,
                 "score": score,
                 "metadata": metadata,
-                "datetime": timestamp,
+                "datetime": datetime,
             },
         )
 
--- a/python/packages/kagent-crewai/tests/test_memory.py
+++ b/python/packages/kagent-crewai/tests/test_memory.py
@@ -1,0 +1,54 @@
+import inspect
+from unittest.mock import MagicMock, patch
+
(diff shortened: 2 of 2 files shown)


peterj (dismissed):

EItanya (dismissed):

anxkhn:

anxkhn:
@EItanya @peterj rebased onto current main and reran the crewai memory tests. please re-review when convenient.

PR #2234, opened 2026-07-14
fix(go-adk): truncate A2A session name on rune boundaries

anxkhn:
### What

`extractSessionName` in `go/adk/pkg/a2a/executor.go` derives a session name from the
first text part of an inbound A2A message and caps it at `sessionNameMaxLength` (20).
The cap used a byte-wise slice:

```go
if len(tp.Text) > sessionNameMaxLength {
    return tp.Text[:sessionNameMaxLength] + "..."
}
```

Both `len()` and the `[:20]` slice operate on bytes, not runes. A first message longer
than 20 bytes that begins with multi-byte UTF-8 (CJK, emoji, accented Latin) gets cut
in the middle of a rune, producing an invalid-UTF-8 string.

This is only ~6-7 non-ASCII characters, so it is easy to hit for non-English users.
For example `こんにちは世界` (7 CJK runes = 21 bytes) truncates to
`こんにちは世\xe7\x95...`, which is not valid UTF-8.

### Why it matters

The derived name flows into `state[StateKeySessionName]` and is persisted via
`CreateSession`:

- On the **Postgres** session store, a `text` column rejects invalid UTF-8, so the
  session-create call fails and the whole turn is aborted with an error, a valid
  non-English first message can break the conversation.
- On **SQLite**, the invalid bytes are stored as mojibake and shown in the UI.

### The fix

Count the cap in runes and slice on a rune boundary, so the result is always valid
UTF-8:

```go
if utf8.RuneCountInString(tp.Text) > sessionNameMaxLength {
    return string([]rune(tp.Text)[:sessionNameMaxLength]) + "..."
}
```

Pure-ASCII behavior is unchanged (for ASCII, byte count == rune count), and the `"..."`
suffix is kept. The only additions are the `unicode/utf8` import and a short doc comment
explaining why truncation is rune-aware.

### Tests

Adds a table-driven `TestExtractSessionName` colocated in
`go/adk/pkg/a2a/executor_test.go` covering:

- nil message / no parts / short ASCII (unchanged behavior)
- ASCII exactly at the 20-rune limit (not truncated)
- ASCII over the limit (truncated with `...`)
- multi-byte over the 20-**byte** limit but under the rune limit (must stay valid UTF-8)
- long multi-byte truncated on a rune boundary
- skipping an empty text part and using the first non-empty one

Each case asserts both the exact result and `utf8.ValidString(got)`. The multi-byte
cases fail against the old byte-slice code (they return invalid UTF-8) and pass with the
fix; the ASCII cases pass either way.

Verified locally:

```
cd go && go test -race -run TestExtractSessionName -v ./adk/pkg/a2a/   # PASS (8/8)
cd go && go test -race -skip 'TestE2E.*' ./adk/...                     # ok, no regressions
cd go && go build ./adk/... && go vet ./adk/pkg/a2a/                   # clean
```

This is a small (<100 LoC) bug fix with unit coverage and no CRD/API change, so no E2E
change is needed.

### DIFF
--- a/go/adk/pkg/a2a/executor.go
+++ b/go/adk/pkg/a2a/executor.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"os"
 	"strings"
+	"unicode/utf8"
 
 	a2atype "github.com/a2aproject/a2a-go/a2a"
 	"github.com/a2aproject/a2a-go/a2asrv"
@@ -424,14 +425,17 @@ func (e *KAgentExecutor) Cancel(ctx context.Context, reqCtx *a2asrv.RequestConte
 }
 
 // extractSessionName extracts session name from the first text part of a message.
+// Truncation is rune-aware: cutting on a byte boundary can split a multi-byte
+// UTF-8 rune and yield an invalid-UTF-8 name, which the Postgres session store
+// rejects on session create. sessionNameMaxLength is therefore counted in runes.
 func extractSessionName(message *a2atype.Message) string {
 	if message == nil {
 		return ""
 	}
 	for _, part := range message.Parts {
 		if tp, ok := part.(a2atype.TextPart); ok && tp.Text != "" {
-			if len(tp.Text) > sessionNameMaxLength {
-				return tp.Text[:sessionNameMaxLength] + "..."
+			if utf8.RuneCountInString(tp.Text) > sessionNameMaxLength {
+				return string([]rune(tp.Text)[:sessionNameMaxLength]) + "..."
 			}
 			return tp.Text
 		}
--- a/go/adk/pkg/a2a/executor.go
+++ b/go/adk/pkg/a2a/executor.go
@@ -6,7 +6,6 @@ import (
 	"maps"
 	"os"
 	"strings"
-	"unicode/utf8"
 
 	a2atype "github.com/a2aproject/a2a-go/a2a"
 	"github.com/a2aproject/a2a-go/a2asrv"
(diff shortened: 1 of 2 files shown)


go/adk/pkg/a2a/executor.go
```diff
@@ -424,14 +425,17 @@ func (e *KAgentExecutor) Cancel(ctx context.Context, reqCtx *a2asrv.RequestConte
...
 func extractSessionName(message *a2atype.Message) string {
 	if message == nil {
 		return ""
 	}
 	for _, part := range message.Parts {
 		if tp, ok := part.(a2atype.TextPart); ok && tp.Text != "" {
-			if len(tp.Text) > sessionNameMaxLength {
-				return tp.Text[:sessionNameMaxLength] + "..."
+			if utf8.RuneCountInString(tp.Text) > sessionNameMaxLength {
+				return string([]rune(tp.Text)[:sessionNameMaxLength]) + "..."
```
mesutoezdil: the text is not trimmed before truncating while the python runtime and the ui both strip whitespace first, so the same message can produce a different session name here, add strings.trimspace before counting runes.

anxkhn:
trimmed whitespace before truncation and switched to bounded rune scanning without allocating the full rune slice.

anxkhn:
@mesutoezdil please re-review when convenient.

anxkhn:

anxkhn:
@mesutoezdil rebased onto current main and reran the focused session-name tests. please re-review when convenient.

PR #2246, opened 2026-07-14
Resolve agent kind name collisions with kind-qualified DB ids and per-kind API routes

iplay88keys:
# Description

An `Agent`, `SandboxAgent`, and `AgentHarness` sharing a `namespace/name` currently overwrite each other's row in the shared `agent` table (last reconcile wins, no error) and resolve interchangeably on every surface that addresses agents by `namespace/name`. This PR gives each kind a distinct identity in the database and resolves kind from per-kind API routes.

## Database identity

- Agent rows keep their existing id (`ns__NS__name`). `SandboxAgent` and `AgentHarness` rows are kind-qualified: `sandboxagents__NS__ns__NS__name`, `agentharnesses__NS__ns__NS__name`. The prefixes mirror the API route resources.
- Migration `000007` re-keys existing sandbox/harness rows and their sessions. The down migration deletes the kind-qualified rows and their sessions/events outright (experimental data carries no rollback guarantee); this keeps the up migration re-runnable after a 0.9.x rollback window recreates bare rows. Normal-agent rows are untouched, so a 0.10 → 0.9.x rollback leaves normal agents fully functional; kind-qualified rows are inert under 0.9.x until the down migration is run (agent lists are built from Kubernetes, not the DB).
- `utils.AgentDBID` / `ParseAgentDBID` / `QualifiedAgentRef` are the only places that know the prefix format.

## API routes

- Legacy routes are frozen: `GET/PUT/DELETE /api/agents...` and `GET /api/sessions/agent/{ns}/{name}` resolve as kind `Agent` only, exactly as before this PR. No `groupKind` query parameter or body field anywhere.
- Per-kind item routes (`/api/sandboxagents/{ns}/{name}`, `/api/agentharnesses/{ns}/{name}`) already existed; this PR adds the missing list routes (`GET /api/sandboxagents`, `GET /api/agentharnesses`) and per-kind session lists (`GET /api/{sandboxagents,agentharnesses}/{ns}/{name}/sessions`).
- Session payload schema is unchanged. `agent_ref` accepts kind-qualified values (`sandboxagents/ns/name`, `agentharnesses/ns/name`); a bare `ns/name` still means Agent, so every existing payload keeps its meaning.

## UI

- Sandbox chats live under `/sandbox-agents/[ns]/[name]/chat`; kind rides in the path and survives navigation.
- All data fetching uses the per-kind endpoints; session creation sends kind-qualified `agent_ref` values via `sessionAgentRefFor`.
- Substrate icon added to the agent cards page: (shown when hovered)
image

## Agent-as-tool names

- The generated `RemoteAgentConfig.Name` is now kind-qualified for sandbox tool agents, so a parent referencing both an `Agent` and a `SandboxAgent` with the same name gets two distinct tools. Agents keep the historical bare name.

## Behavior changes visible to existing clients

- `GET /api/agents` (merged list) returns kind-qualified `id` values for `SandboxAgent`/`AgentHarness` entries. `Agent` entries are identical to before.
- An existing `SandboxAgent` used as an agent tool gets a new tool name after upgrade; system prompts addressing the old bare name need updating.

## Rollback and backwards compatibility

Normal `Agent` resources have full rollback support; `SandboxAgent`/`AgentHarness` are experimental and exempt. Concretely:

- **Helm rollback to 0.9.x (no migration action needed).** The DB stays at schema version 7; the 0.9.x migration runner detects it is behind the schema and starts in compatibility mode without touching it. Normal agents are fully functional — their rows, sessions, and routes are unchanged by this PR. Kind-qualified sandbox/harness rows and their sessions are inert (0.9.x computes bare ids and builds agent lists from Kubernetes, so they neither resolve nor render). The 0.9.x reconciler recreates bare rows for live sandbox/harness resources, so those resources work again with 0.9.x semantics and fresh (empty) session history.
- **Re-upgrade after a rollback window.** The DB is already at version 7, so no migration re-runs. Qualified rows resume serving and pre-rollback sandbox/harness history returns. Bare rows created during the window remain as inert cruft (nothing resolves bare ids for those kinds anymore) — harmless, not rendered.
- **Explicit down migration (operator-run only).** Deletes kind-qualified sandbox/harness rows and their sessions/events outright; Agent rows are untouched. Chat history for the experimental kinds does not survive an explicit downgrade — this is the documented exemption, and it is what keeps the down migration collision-free (it cannot abort) and the up migration re-runnable afterward.
- **API compatibility for existing clients.** Legacy routes are byte-identical for kind `Agent`, including error bodies (the delete 404 body deliberately matches the previous release). The only value-level changes old clients can observe are listed under "Behavior changes" above, and both are confined to the experimental kinds.

## Out of scope

- MCP `list_agents`/`invoke_agent` and the A2A client registry remain kind-blind; #2204 covers both.
- The Go API client and `kagent` CLI (`agent get`, `invoke`) address kind `Agent` only, matching the frozen legacy routes.
- Harness chats have no kind-scoped UI prefix yet; an `AgentHarness` sharing a name with an `Agent` is not reachable in chat (tracked as a follow-up).


iplay88keys:
Closing in favor of a broader fix to these issues.

PR #2249, opened 2026-07-15
fix(mcp): include SandboxAgents in list_agents and invoke_agent

anxkhn:
## What

`list_agents`, the `kagent://agents` MCP resource, and `invoke_agent` now return
`SandboxAgent`s (substrate-backed agents) in addition to declarative `Agent`s. A
fully ready `SandboxAgent` was previously invisible to all three, even though it
shows up in the REST `/agents` API and the UI.

## Why

`MCPHandler.listReadyAgents` (`go/core/internal/mcp/mcp_handler.go`) had two
compounding faults:

1. It listed only `v1alpha2.AgentList`, so `SandboxAgent` CRs were never
   enumerated at all.
2. It gated readiness on the string literal `condition.Reason == "DeploymentReady"`.

A `SandboxAgent` reports its `Ready` condition with reason `WorkloadReady`, while
a declarative `Agent` uses `DeploymentReady`:

- `reconcileSandboxAgentStatus` sets `Reason = AgentReadyReasonWorkloadReady`
  when the workload is ready (`reconciler.go`).
- `reconcileAgentStatus` sets `Reason = AgentReadyReasonDeploymentReady`
  (`reconciler.go`).

So a `SandboxAgent` with `Accepted=True` and `Ready=True` (reason `WorkloadReady`)
was dropped by both faults. `listReadyAgents` feeds `handleListAgents` (the
`list_agents` tool) and `readAgentsResource` (the `kagent://agents` resource), and
`invoke_agent` relies on the same readiness notion, so such an agent was reported
as "not found or not ready" over MCP while being fully usable via the REST API.

The correct, two-reason, interface-based readiness check already exists elsewhere
in the codebase and is what this change reuses:

- `isAgentReady` in the A2A registrar (`a2a/a2a_registrar.go`) accepts both
  `AgentReadyReasonDeploymentReady` and `AgentReadyReasonWorkloadReady` over the
  `AgentObject` interface.
- The REST agents handler (`httpserver/handlers/agents.go`) does the same.

## Fix

- List both `v1alpha2.AgentList` and `v1alpha2.SandboxAgentList`, and evaluate
  readiness through a small `readyAgentSummary(v1alpha2.AgentObject)` helper that
  operates on the shared `AgentObject` interface.
- The helper accepts both `reconciler.AgentReadyReasonDeploymentReady` and
  `reconciler.AgentReadyReasonWorkloadReady`, and uses the typed condition-type
  constants (`v1alpha2.AgentConditionTypeReady` / `...Accepted`) with
  `metav1.ConditionTrue`, matching the A2A registrar and the REST handler instead
  of re-implementing the check with hardcoded strings.
- Updated the `list_agents` tool and `kagent://agents` resource descriptions from
  "accepted + deploymentReady" to "accepted + ready" now that both workload types
  are covered.

No CRD/API types change, so no code generation is required.

## Tests

Added `TestListReadyAgents_IncludesSandboxAgents` in
`go/core/internal/mcp/mcp_handler_test.go` (table of fake objects via the
controller-runtime fake client already used in this file). It asserts that:

- a ready declarative `Agent` (`DeploymentReady`) is listed,
- a ready `SandboxAgent` (`WorkloadReady`) is listed,
- a not-ready `SandboxAgent` is excluded, and
- the `SandboxAgent`'s description is surfaced.

The test fails against the current code (the `SandboxAgent` is absent and its
description is empty) and passes with the fix. Revert the `mcp_handler.go` hunks
to reproduce.

## How to verify

```bash
cd go
go build ./core/...
go test -race ./core/internal/mcp/...   # incl. TestListReadyAgents_IncludesSandboxAgents
make -C go lint                         # from repo root; 0 issues
```

## Scope

Limited to the MCP agent-listing path. It reuses the existing readiness semantics
(the same two `Ready` reasons the REST API and A2A registrar already honor) rather
than introducing new behavior, so declarative `Agent` results are unchanged and
only genuinely ready `SandboxAgent`s are added.

## Checklist

- [x] Commit is DCO signed-off (`git commit -s`).
- [x] `go build ./core/...`, `go test -race ./core/internal/mcp/...`, and
      `make -C go lint` all pass.
- [x] Unit test added; fails without the fix, passes with it.
- [x] Diff limited to `go/core/internal/mcp/mcp_handler.go` and its test.

### DIFF
--- a/go/core/internal/mcp/mcp_handler.go
+++ b/go/core/internal/mcp/mcp_handler.go
@@ -11,10 +11,12 @@ import (
 	"github.com/google/jsonschema-go/jsonschema"
 	"github.com/kagent-dev/kagent/go/api/v1alpha2"
 	"github.com/kagent-dev/kagent/go/core/internal/a2a"
+	"github.com/kagent-dev/kagent/go/core/internal/controller/reconciler"
 	"github.com/kagent-dev/kagent/go/core/internal/version"
 	"github.com/kagent-dev/kagent/go/core/pkg/auth"
 	"github.com/kagent-dev/kagent/go/core/pkg/env"
 	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -83,7 +85,7 @@ func NewMCPHandler(kubeClient client.Client, agentClients *a2a.AgentClientRegist
 		server,
 		&mcpsdk.Tool{
 			Name:        "list_agents",
-			Description: "List invokable kagent agents (accepted + deploymentReady)",
+			Description: "List invokable kagent agents (accepted + ready)",
 			InputSchema: &jsonschema.Schema{
 				Type:                 "object",
 				Properties:           map[string]*jsonschema.Schema{},
@@ -108,7 +110,7 @@ func NewMCPHandler(kubeClient client.Client, agentClients *a2a.AgentClientRegist
 		&mcpsdk.Resource{
 			URI:         "kagent://agents",
 			Name:        "agents",
-			Description: "List of invokable kagent agents (accepted + deploymentReady)",
+			Description: "List of invokable kagent agents (accepted + ready)",
 			MIMEType:    "application/json",
 		},
 		handler.readAgentsResource,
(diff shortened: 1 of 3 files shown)


the-mentor:
@anxkhn maybe this PR https://github.com/kagent-dev/kagent/pull/2204 is the same thing?

go/core/internal/mcp/mcp_handler.go
```diff
@@ -129,35 +131,66 @@ func NewMCPHandler(kubeClient client.Client, agentClients *a2a.AgentClientRegist
...
+		}
+		if condition.Type == v1alpha2.AgentConditionTypeAccepted && condition.Status == metav1.ConditionTrue {
+			accepted = true
+		}
+	}
+	if !accepted || !ready {
+		return AgentSummary{}, false
+	}
+	summary := AgentSummary{
+		Ref: agent.GetNamespace() + "/" + agent.GetName(),
```
mesutoezdil: sandbox agents are registered in the a2a client registry under sandboxes/ns/name, so the ns/name ref listed here still fails in invoke_agent with not found or not ready.

anxkhn:
addressed the registry key mismatch and typed condition feedback; sandbox refs now use sandboxes/namespace/name and invoke through the same key.

anxkhn:
@mesutoezdil please re-review when convenient.

anxkhn:

mesutoezdil:
> @mesutoezdil please re-review when convenient.

thx, i would, but #2204 which is already assigned to the reporter of #2123 and has active review from EItanya might be worth consolidating into that one instead. FIFO..

anxkhn:
@mesutoezdil rebased onto current main and revalidated the sandbox agent listing and invocation paths. please re-review when convenient.

anxkhn:
closing in favor of #2204, which predates this pr, is owned by the issue assignee, and covers the same sandbox agent listing and invocation fix.

PR #2251, opened 2026-07-15
fix(controller): default agent deployments to restricted Pod Security

### LINKED ISSUE #2244: Default regular Agent Deployments to satisfy restricted Pod Security
## Problem

Regular `Agent` resources (`kagent.dev/v1alpha2`) created through the kagent UI/CLI can render Deployments whose pod templates omit security context defaults required by Kubernetes Pod Security Standards `restricted` admission.

On a cluster enforcing `restricted:latest`, the generated agent pods are rejected before startup. Example admission failure from a `Declarative` Python agent:

```text
violates PodSecurity "restricted:latest":
allowPrivilegeEscalation != false
unrestricted capabilities (container "kagent" must set securityContext.capabilities.drop=["ALL"])
runAsNonRoot != true
seccompProfile must be RuntimeDefault or Localhost
```

The installed kagent control-plane workloads are already compliant, for example the controller pod has:

```yaml
securityContext:
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
containers:
- name: controller
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
```

But controller-created `Agent` Deployments can render like this:

```yaml
spec:
  template:
    spec:
      securityContext: {}
      containers:
      - name: kagent
        # no container securityContext
```

As a result, newly created agents from the GUI/CLI are accepted at the kagent CRD level but never become ready on restricted clusters.

## Reproducer

Create a regular declarative agent in a namespace enforcing restricted Pod Security:

```bash
kubectl label ns kagent pod-security.kubernetes.io/enforce=restricted --overwrite
```

Then create a basic `Agent` through the UI/CLI, or apply a minimal CR:

```yaml
kind: Agent
metadata:
  name: pss-restricted-test
  namespace: kagent
spec:
  type: Declarative
  description: PSS restricted smoke test
  declarative:
    runtime: python
    systemMessage: "smoke test"
```

The controller creates a Deployment, but the ReplicaSet cannot create pods because the generated pod template does not meet PSS restricted requirements.

## Workaround

Adding the deployment security settings to the `Agent` CR works:

```yaml
spec:
  declarative:
    deployment:
      podSecurityContext:
        runAsNonRoot: true
        seccompProfile:
          type: RuntimeDefault
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
          - ALL
        runAsNonRoot: true
```

For BYO agents, the same shape is available under:

```yaml
spec:
  byo:
    deployment:
      podSecurityContext: ...
      securityContext: ...
```

This is accepted by the installed `v1alpha2` CRD and is enough for the generated Deployment to satisfy restricted admission.

## Requested feature

Please make regular UI/CLI-created `Agent` resources PSS-restricted-compatible by default.

Potential approaches:

1. Add default security context fields to Agent creation payloads in the UI and CLI.
2. Extend `controller.agentDeployment` Helm defaults to include `podSecurityContext` and `securityContext`, and have the controller apply those defaults when an Agent omits them.
3. Apply built-in controller defaults matching the already-compliant kagent control-plane chart defaults.

Suggested default:

```yaml
podSecurityContext:
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
  runAsNonRoot: true
```

`readOnlyRootFilesystem: true` would be nice as a harder default only if the agent runtime image supports it without requiring writable root paths. The minimal request here is to pass Kubernetes PSS `restricted` admission.

## Related issue

Related but not identical to #1900, which covers `SandboxAgent` runtime identity. This request is for regular `Agent` resources rendered into Deployments by the kagent controller.

## Environment

- kagent chart/app: 0.9.11
- CRD: `kagent.dev/v1alpha2`
- Kubernetes: cluster enforcing Pod Security Standards `restricted:latest`

reyshazni:
## Problem

Agent Deployments created through the kagent UI or CLI produce pod templates with empty `securityContext` at both pod and container level. On clusters that enforce the Kubernetes `restricted` Pod Security Standard (the recommended default since Kubernetes 1.25), admission rejects these pods before they start. Agents never become ready.

The kagent control-plane pods are already compliant, which makes this inconsistency confusing for users.

Fixes #2244

## Root Cause

`buildContainerSecurityContext` returns `nil` when the Agent CR has no `SecurityContext` and code-exec isolation is not needed. `buildPodTemplate` passes `dep.PodSecurityContext` directly, which is also `nil` by default. The result is a pod spec with no security context at either level.

## Fix

When the user does not supply security context fields, the controller now injects restricted-PSS-compliant defaults:

**Container level:**
- `runAsNonRoot: true`
- `allowPrivilegeEscalation: false`
- `capabilities.drop: [ALL]`
- `seccompProfile: RuntimeDefault`

**Pod level:**
- `runAsNonRoot: true`
- `seccompProfile: RuntimeDefault`

User-supplied values always take precedence. The existing `base != nil` path in `buildContainerSecurityContext` and the nil check in `defaultPodSecurityContext` preserve any explicit user configuration unchanged.

The code-exec isolation path (skills with sandbox) is also unchanged: when `needCodeExecIsolation` is true, the controller still sets `Privileged: true` as before.

## Testing

- Updated `TestSecurityContext_OnlyPodSecurityContext` to verify that the container now receives restricted defaults instead of nil
- Updated `TestSecurityContext_OnlyContainerSecurityContext` to verify that the pod now receives restricted defaults instead of nil
- Added `TestSecurityContext_NilSecurityContext_RestrictedDefaults` covering the common case where no SecurityContext is set at all
- All 6 security context tests pass
- Existing tests for skills/sandbox (`SkillsDefaultPrivilegedSandbox`, `SkillsPSSRestricted`) continue to pass without modification

### DIFF
--- a/go/core/internal/controller/translator/agent/manifest_builder.go
+++ b/go/core/internal/controller/translator/agent/manifest_builder.go
@@ -488,10 +488,37 @@ func buildContainerSecurityContext(
 	}
 
 	if !needCodeExecIsolation {
-		return nil
+		return defaultRestrictedContainerSecurityContext()
 	}
 
 	return &corev1.SecurityContext{Privileged: new(true)}
+}
+
+// defaultRestrictedContainerSecurityContext returns a restricted-PSS-compliant container SecurityContext.
+func defaultRestrictedContainerSecurityContext() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		RunAsNonRoot:             new(true),
+		AllowPrivilegeEscalation: new(false),
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		},
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		},
+	}
+}
+
+// defaultPodSecurityContext returns the provided PodSecurityContext or a restricted-PSS-compliant default if nil.
+func defaultPodSecurityContext(provided *corev1.PodSecurityContext) *corev1.PodSecurityContext {
+	if provided != nil {
+		return provided
+	}
+	return &corev1.PodSecurityContext{
+		RunAsNonRoot: new(true),
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		},
+	}
 }
 
 func buildPodTemplate(
@@ -526,7 +553,7 @@ func buildPodTemplate(
 		Spec: corev1.PodSpec{
 			ServiceAccountName: *dep.ServiceAccountName,
 			ImagePullSecrets:   dep.ImagePullSecrets,
-			SecurityContext:    dep.PodSecurityContext,
+			SecurityContext:    defaultPodSecurityContext(dep.PodSecurityContext),
 			InitContainers:     runtimeInputs.initContainers,
 			Containers: append([]corev1.Container{{
 				Name:            "kagent",
--- a/go/core/internal/controller/translator/agent/security_context_test.go
+++ b/go/core/internal/controller/translator/agent/security_context_test.go
@@ -195,9 +195,15 @@ func TestSecurityContext_OnlyPodSecurityContext(t *testing.T) {
 	assert.Equal(t, int64(2000), *podSecurityContext.RunAsUser)
 	assert.Equal(t, int64(2000), *podSecurityContext.RunAsGroup)
 
-	// Container security context should be nil if not specified
+	// Container security context should get restricted-PSS defaults when not specified
 	containerSecurityContext := podTemplate.Spec.Containers[0].SecurityContext
-	assert.Nil(t, containerSecurityContext, "Container securityContext should be nil when not specified")
+	require.NotNil(t, containerSecurityContext, "Container securityContext should get restricted defaults")
+	assert.True(t, *containerSecurityContext.RunAsNonRoot, "Default runAsNonRoot should be true")
+	assert.False(t, *containerSecurityContext.AllowPrivilegeEscalation, "Default allowPrivilegeEscalation should be false")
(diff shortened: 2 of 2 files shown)


go/core/internal/controller/translator/agent/manifest_builder.go
```diff
@@ -488,12 +488,39 @@ func buildContainerSecurityContext(
 	}
 
 	if !needCodeExecIsolation {
-		return nil
+		return defaultRestrictedContainerSecurityContext()
```
mesutoezdil, line 491: forcing runasnonroot true by default breaks byo agents whose images run as root on any cluster, so consider applying these defaults only to kagent owned images or making them opt in.

Tazmainiandevil:
Reproduced the underlying issue independently (kind cluster, `restricted` PSA namespace, BYO agent) — pod rejected with `allowPrivilegeEscalation`/`capabilities`/`runAsNonRoot`/`seccompProfile` all unset, matches the issue description exactly.

On @mesutoezdil's opt-in suggestion — agree that's the right call, and I think it holds up cleanly once you separate two cases.

In a `restricted`-PSA namespace, hardening isn't really optional — the apiserver enforces it at admission regardless of what kagent defaults to. A root-requiring BYO image can't run there today, and no default/override choice changes that either way.

The risk case is non-restricted namespaces, where kagent currently sets nothing and root-requiring images work fine as-is. Defaulting to hardened everywhere (this PR's current approach, override-to-loosen) hardens namespaces that never asked for it and can surprise existing root-requiring BYO deployments there for no gain — those namespaces were never going to reject the pod anyway.

So: opt-in, not enforced. Render the hardened securityContext block only when the user (or a namespace/PSA-aware toggle) actually asks for it. Zero behavior change for existing deployments, and restricted-namespace users get compliant pods by opting in.

PR #2260, opened 2026-07-15
fix: set --allowed-hosts for bundled grafana-mcp deployment

### LINKED ISSUE #2166: [BUG] RemoteMCPServer initialization fails with Forbidden when using bundled kagent-grafana-mcp defaults
### 📋 Prerequisites

- [ ] I am using the latest version of the software

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

No impact (Default)

### 🐛 Bug Description

The bundled `kagent-grafana-mcp` deployment fails to initialize as a `RemoteMCPServer` unless `--allowed-hosts` is explicitly configured for the `mcp-grafana` server.

In the default Helm chart configuration, `kagent-grafana-mcp` is exposed as a Kubernetes Service and the `RemoteMCPServer` points to `http://kagent-grafana-mcp.kagent:8000/mcp`.

However, mcp-grafana validates the HTTP Host header for streamable-http transports as part of DNS rebinding protection.
Without `--allowed-hosts`, requests using the Kubernetes Service DNS name are rejected with `Forbidden`.
As a result, the default kagent-grafana-mcp deployment provided by kagent does not work out of the box.

This appears to be related to [grafana/mcp-grafana#957](https://github.com/grafana/mcp-grafana/pull/957), which added Host header validation for streamable-http transports as DNS rebinding protection.

As of this report, the change is not part of any tagged release, but the `docker.io/mcp/grafana:latest` image is built from `main` and therefore already includes it (image digest `sha256:fef799be2e8b59b9551382038d0efcafa0849aa1942e46bbb2538dd12d7c8a8f`, built from commit `583994f`). 

### 🔄 Steps To Reproduce

The issue is reproducible immediately after installing kagent with the bundled Grafana MCP integration and creating the bundled RemoteMCPServer resource, without any custom configuration.

The kagent-grafana-mcp deployment starts mcp-grafana with:

```yaml
args:
  - --transport
  - streamable-http
```

Additional arguments can be passed through `.Values.args`, but the default `values.yaml` does not document or set `--allowed-hosts`.

The default RemoteMCPServer URL is:

```text
http://kagent-grafana-mcp.kagent:8000/mcp
```

### 🤔 Expected Behavior

The bundled kagent-grafana-mcp deployment should work with the bundled RemoteMCPServer configuration without requiring additional Helm arguments.

Possible approaches:

1. Automatically configure `--allowed-hosts` for the Kubernetes Service DNS names used by the bundled RemoteMCPServer, or
2. Expose a dedicated `allowedHosts` value and document that it is required for streamable-http deployments, or
3. Document the required `.Values.args` configuration in the Grafana MCP setup instructions.

### 📱 Actual Behavior

The RemoteMCPServer fails to reconcile:

```yaml
status:
  conditions:
  - type: Accepted
    status: "False"
    reason: ReconcileFailed
    message: 'failed to fetch tools for toolServer kagent/kagent-grafana-mcp:
      failed to connect client for toolServer kagent/kagent-grafana-mcp:
      calling "initialize": sending "initialize": Forbidden'
```

The kagent controller logs show:

```text
"failed to upsert tool server for remote mcp server","remoteMCPServer":"kagent/kagent-grafana-mcp","duration":0.032398094,"error":"failed to fetch tools for toolServer kagent/kagent-grafana-mcp: failed to connect client for toolServer kagent/kagent-grafana-mcp: calling \"initialize\": sending \"initialize\": Forbidden"
```

This suggests that the failure occurs while kagent-controller initializes the MCP connection to kagent-grafana-mcp. 
No request reaches Grafana at this stage, so Grafana connectivity or authentication is not involved.

### 💻 Environment

- OS version: AmazonLinux 2023
- Kubernetes: 1.35.6
- Kubernetes Provider: AWS (EKS)
- kagent version: v0.9.9
- Namespace: kagent
- MCP server: kagent-grafana-mcp
- MCP server image: docker.io/mcp/grafana:latest
  - Digest: sha256:fef799be2e8b59b9551382038d0efcafa0849aa1942e46bbb2538dd12d7c8a8f
  - `mcp-grafana -version` output: `(devel)`
- RemoteMCPServer: kagent/kagent-grafana-mcp
- MCP transport: streamable-http

### 🔧 CLI Bug Report

N/A — this is a server-side bug, not a CLI issue, so kagent bug-report does not apply.

### Workaround

Adding `--allowed-hosts` to the kagent-grafana-mcp deployment resolves the Forbidden error.

```yaml
args:
  - --transport
  - streamable-http
  - --allowed-hosts=kagent-grafana-mcp.kagent:8000
```

After applying this, the RemoteMCPServer is reconciled successfully and the Grafana MCP tools become visible to agents.

### Related issue

This is also related to #2040, but not fully solved by image pinning alone. Pinning `mcp-grafana` would reduce the risk of silent behavior changes from `latest`, but once a version containing grafana/mcp-grafana#957 is used, the chart still needs to configure or document `--allowed-hosts` for the bundled Kubernetes Service hostname.

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

muskiteer:
## Description

**What does this PR do?**
This PR fixes an out-of-the-box initialization failure for the bundled `grafana-mcp` deployment. 

**Why is it needed?**
Currently, when `kagent-grafana-mcp` is enabled in the default Helm chart, it fails to initialize as a `RemoteMCPServer` because of a `Forbidden` error. This happens because recent versions of `mcp-grafana` validate the HTTP Host header for `streamable-http` transports as DNS rebinding protection (introduced in `grafana/mcp-grafana#957`). Without passing `--allowed-hosts` matching the internal Kubernetes service DNS name, requests from the Kagent controller are rejected. 

**How does this fix it?**
This automatically resolves the Kubernetes Service DNS name natively generated by the chart (`<fullname>.<namespace>:<port>`) and appends it to the container's `--allowed-hosts` arguments in the deployment spec. This ensures the bundled integration works immediately without manual configuration.

## Testing

**How was this tested?**
1. Deployed a local Kind cluster (`make create-kind-cluster`).
2. Installed kagent with the default chart values (`make helm-install`), which enabled `grafana-mcp`.
3. Verified that the `kagent-grafana-mcp` RemoteMCPServer successfully transitions to `Accepted: True` instead of `ReconcileFailed: Forbidden`.

Fixes #2166


helm/tools/grafana-mcp/templates/deployment.yaml
```diff
@@ -40,6 +40,7 @@ spec:
           args:
             - --transport
             - streamable-http
+            - --allowed-hosts={{ include "grafana-mcp.fullname" . }}.{{ .Release.Namespace }}:{{ .Values.service.port }}
```
onematchfox: Whilst this is a sensible default, please can we ensure this is configurable. I may want to add additional hosts here (e.g. if the MCP is exposed via something like agentgateway) or even to just set it to `*`. This is currently possioble using something like the following but will break if this change is merged as is:

``` yaml
args:
  - --allowed-hosts=*
```

mesutoezdil: the allowlist only matches the exact name.namespace:port host form, so clients using the svc or svc.cluster.local dns variants or localhost via port-forward now get 403, a comma separated list covering those would be safer.

muskiteer:
I've made allowedHosts configurable and added sensible defaults for the Kubernetes Service DNS variants while still allowing users to override them (including `["*"]`). I've also added tests covering the default and custom configurations.

Happy to receive any further feedback @mesutoezdil @onematchfox

helm/tools/grafana-mcp/templates/_helpers.tpl
```diff
@@ -75,4 +75,18 @@ Join registry/repository for grafana-mcp image, skipping empty segments, then ap
...
+{{- end -}}
+
+{{- define "grafana-mcp.allowedHosts" -}}
+{{- if .Values.allowedHosts -}}
+{{- join "," .Values.allowedHosts -}}
+{{- else -}}
+{{- $fullname := include "grafana-mcp.fullname" . -}}
+{{- $ns := .Release.Namespace -}}
+{{- $port := .Values.service.port | int -}}
+{{- join "," (list (printf "%s.%s:%d" $fullname $ns $port)
```
mesutoezdil: nit: maybe add the plain svc name too (sth like `<name>:8000` w/o the ns), clients in the same ns usually hit it that way and the host header wont match this list

mesutoezdil:
and branch is out-of-date

PR #2265, opened 2026-07-15
fix(security): scope task get, create, and delete to the owner

mesutoezdil:
Task get, create, and delete had no owner check at all. Any user could read another user's task by id, delete it, or overwrite its data by posting a task with the same id.

Task now has a user_id column, backfilled from its session history (including deleted sessions, so a deleted owner can't shift a task to whoever reuses that session id later). All three endpoints check it, plus ListTasksForSession and the session activity touch, same pattern already used for session and event. A NULL owner is only treated as readable and claimable when the session id unambiguously belongs to the caller, which also covers tasks written by old pods mid rolling upgrade.

Also fixed a related bug: upserting a task whose id belonged to a soft-deleted row hit the ON CONFLICT guard, matched zero rows, and still reported success. The upsert now returns the written id and treats no rows as an explicit conflict instead of a silent no-op.


go/core/pkg/migrations/core/000007_task_owner.down.sql
iplay88keys, line 1: These up/down files should be `go/core/pkg/migrations/core/000007_task_owner.{up,down}.sql`, not `8`.

go/core/pkg/migrations/core/000007_task_owner.up.sql
```diff
@@ -0,0 +1,20 @@
...
+    SELECT id, MIN(user_id) AS user_id
+    FROM session
+    WHERE deleted_at IS NULL
+    GROUP BY id
+    HAVING COUNT(*) = 1
+)
+UPDATE task
+SET user_id = unique_session.user_id
+FROM unique_session
+WHERE unique_session.id = task.session_id AND task.user_id IS NULL;
```
iplay88keys, line 21: We should add some tests covering this logic. Something like `go/core/pkg/migrations/migration_000007_test.go` [here](https://github.com/kagent-dev/kagent/pull/2246/changes#diff-d327905775c8fba51ceead5c09999fc66d31fb30135c869ca7bee4c20e18aff5).

go/core/internal/database/queries/tasks.sql
```diff
@@ -15,13 +18,14 @@ ORDER BY created_at ASC;
...
-INSERT INTO task (id, data, session_id, protocol_version, created_at, updated_at)
-VALUES ($1, $2, $3, $4, NOW(), NOW())
+INSERT INTO task (id, data, session_id, protocol_version, user_id, created_at, updated_at)
+VALUES ($1, $2, $3, $4, $5, NOW(), NOW())
 ON CONFLICT (id) DO UPDATE SET
     data             = EXCLUDED.data,
     session_id       = EXCLUDED.session_id,
     protocol_version = EXCLUDED.protocol_version,
     updated_at       = NOW()
+WHERE task.user_id = EXCLUDED.user_id OR task.user_id IS NULL
```
iplay88keys: We should probably avoid treating a `NULL` owner as writable by any caller. This preserves the same ownership ambiguity that this change is trying to resolve. I wonder if we should reject those updates or only claim the task after unambiguously proving the session belongs to the caller. Though that might still have security implications still 🤔

go/core/pkg/migrations/core/000008_task_owner.up.sql
```diff
@@ -0,0 +1,20 @@
...
+ALTER TABLE task ADD COLUMN IF NOT EXISTS user_id TEXT;
+
+-- session.id is not globally unique (session's key is (id, user_id)), so only
+-- backfill from a session id that maps to exactly one live user. Anything
+-- ambiguous or gone is left NULL, which the new owner checks then hide from
+-- everyone rather than guessing wrong.
+WITH unique_session AS (
+    SELECT id, MIN(user_id) AS user_id
+    FROM session
+    WHERE deleted_at IS NULL
```
iplay88keys: Filtering out deleted sessions before testing the uniqueness can misalign ownership:
```
session "s1", Alice, deleted
session "s1", Bob, active

task "t1", session_id = "s1"
```

The filter removes Alice, making it appear that Bob is the only owner.  Every task referencing "s1" would be assigned to Bob in that case, even if "t1" was owned by Alice.

We should determine uniqueness across all historical owners and have no pre-filter, leaving the task as `NULL` when there are multiple users who have used that session ID.

iplay88keys (commented):
I agree this is important to fix. Just a few gaps that I see in addition to the comments I've added:
- Scope `ListTasksForSession` and the `session` `updated_at` update by both `session_id` and `user_id`.
- During a rolling upgrade, old pods can create new `NULL`-owned tasks after the backfill, which new pods cannot read.

go/core/pkg/migrations/core/000007_task_owner.down.sql
mesutoezdil: thx, done

go/core/pkg/migrations/core/000007_task_owner.up.sql
mesutoezdil: thx, done

go/core/internal/database/queries/tasks.sql
mesutoezdil: thx, done

go/core/pkg/migrations/core/000008_task_owner.up.sql
mesutoezdil: thx, done

go/core/internal/database/queries/tasks.sql
```diff
@@ -1,6 +1,17 @@
...
+-- included) and that user is the caller. Anything ambiguous stays hidden
+-- rather than guessed. This mirrors the backfill rule in migration
+-- 000007_task_owner.
+
 -- name: GetTask :one
 SELECT * FROM task
-WHERE id = $1 AND user_id = $2 AND deleted_at IS NULL
+WHERE task.id = $1 AND task.deleted_at IS NULL
+  AND (task.user_id = $2 OR (task.user_id IS NULL AND $2 = (
+      SELECT MIN(s.user_id) FROM session s WHERE s.id = task.session_id
```
iplay88keys: There's another potential issue here (as well as for list, upsert, and delete). Consider this situation:
```
task "t1", session_id = "s1", user_id = NULL
no session "s1" exists

Bob creates session "s1"
```

Bob is now the owner and can list, read, and claim the pre-existing task. I was able to reproduce this locally by creating an orphaned task, creating a session as another user and reading that task as that user.

Maybe we should require evidence that a session existed when the task was created by comparing the `created_at`, for example? Or make it so that unresolved NULL tasks are inaccessible?

I'm not sure how likely this is to occur, but something to consider.

go/core/internal/database/queries/tasks.sql
iplay88keys, line 25: Soft-deleted tasks still have the `task.id` primary key. This can result in the following:
1. Alice deletes task `t1`
2. Bob tries to create task `t1`
3. The create is blocked by `ON CONFLICT (id) DO UPDATE SET`, but this is not a sql error as it just affects zero rows
4. `GetTaskOwner` filters out the deleted row and returns no rows as a result
5. `CheckTaskOwner` interprets "no rows" as success
6. Bob gets a `201 created`, but no task was stored.

In a similar case, if Alice re-used the deleted id, the update would be allowed (by the `WHERE task.user_id = EXCLUDED.user_id`), but `deleted_at` is never cleared, so Alice would receive `201 created` and the task would remain deleted.

I think we would need to return an explicit conflict in these cases as I'm not sure that resurrecting the rows if the owner matches would be the right approach here either.

mesutoezdil: i went a bit different way here, hopefully more robust. fixed in 036477db by making the upsert itself report the outcome instead of adding a separate check. UpsertTask now returns the upserted id and the ON CONFLICT guard also excludes soft-deleted rows, so reusing a deleted id affects zero rows and client maps the empty result to 409. deleted ids stay burned for everyone, also for the original owner, resurrecting felt too surprising to me. detection happens in same statement as the write, so there is no window between check and upsert. also added TestDeletedTaskIdCannotBeReused covering both scenarios.

iplay88keys: That makes sense to me. Thanks!

go/core/internal/database/queries/tasks.sql
mesutoezdil: looks better now?

iplay88keys (approved):
I think this is good overall. Although it is unlikely, I think we should probably note somewhere the risk of losing access to a task so that risk is not completely hidden. This work will go into the 0.10.0 release, so at least it's more likely to be seen by users upgrading (more visible than in a patch release).

PR #2267, opened 2026-07-15
feat: add list_files and grep_file skills tools

brandonkeung:
## Summary
- Adds native, in-process `list_files` and `grep_file` tools to both the Go (`go/adk/pkg/skills`, `go/adk/pkg/tools`) and Python (`kagent-skills`, `kagent-adk`) agent runtimes, alongside the existing `read_file`/`write_file`/`edit_file`/`bash` skills tools.
- Gives agents safe, non-privileged file visibility without depending on `bash`, which some deployments disable for privilege/security reasons.
- Hardened across several review passes; fixes along the way:
  - Go: a symlink-resolution inconsistency in `resolveReadPath`/`resolveWritePath`/`resolveEditPath` could reject valid paths under a symlinked session root.
  - Go: `NewSkillsTools` failed entirely (dropping all tools) when the bash command executor couldn't be constructed, instead of just omitting bash.
  - Go: `filepath.WalkDir` root resolution fixed so an unresolved symlink root is actually recursed into; an err-shadowing bug that silently truncated results on a `WalkDir` failure fixed.
  - Both: `grep_file` no longer hangs indefinitely on a FIFO (or other non-regular file) with no writer connected, in both the recursive walk and single-target paths.
  - Both: matched lines are capped to 2000 chars (matching `read_file`'s existing convention) — previously unbounded in Python, and could fail the whole search past 1MB in Go.
  - Python: `GrepFileTool` now uses a dedicated thread pool instead of the shared default pool, since a hung regex match can't be forcibly killed and would otherwise starve unrelated work.
  - Both: a single unreadable file or subdirectory is now a skip rather than an abort of the whole search, so one bad entry doesn't discard matches already found elsewhere in the tree — annotated as `"no matches found (N entries could not be read)"` so a systemic failure isn't indistinguishable from a genuinely empty search.
  - Both: a fully unreadable search root now surfaces a real error instead of a misleadingly confident empty result.
  - Per maintainer feedback, `list_files`/`grep_file` are now **disabled by default** and gated behind `KAGENT_ENABLE_FILE_SEARCH_TOOLS` (set on the Agent's `env` to opt in) — mirroring how `bash` is already effectively opt-in. `read_file`/`write_file`/`edit_file`/`bash`/`skills` are unaffected.

## Test plan
- [x] Go: `TestGrepContent` (`go/adk/pkg/skills/shell_test.go`) — matches, recursion, symlink escape/resolution, FIFO hang safety, unreadable file/subdirectory/root handling, output truncation — and `TestListFilesAndGrepFileTools_RunThroughADK`, `TestNewSkillsTools_OmitsBashWithoutSRTSettings` (`go/adk/pkg/tools/skills_test.go`)
- [x] Python: unit tests in `kagent-skills`/`kagent-adk` covering the same scenarios (path traversal, recursive/ignore-case, symlink escape, FIFO hang safety, unreadable file/subdirectory/root handling, truncation, `GrepFileTool` timeout behavior)
- [x] All new regression tests verified to fail against the pre-fix code, confirming they exercise the actual bug
- [x] Verified end-to-end in a live kind cluster via the browser UI for both runtimes (dedicated Go- and Python-runtime test agents), including the FIFO-hang scenario and the unreadable-subdirectory/root scenarios, each invoking `list_files`/`grep_file` by name and returning correct output
- [x] Go/Python unit tests for the `KAGENT_ENABLE_FILE_SEARCH_TOOLS` flag's default-off/opt-in behavior (`skills_test.go`, `test_skill_execution.py`, new `test_skills_plugin.py`)
- [x] Verified end-to-end on a live kind cluster via A2A and the browser UI: dedicated Go- and Python-runtime test agents, with and without `KAGENT_ENABLE_FILE_SEARCH_TOOLS` set, confirming correct tool registration and that `list_files`/`grep_file` execute correctly when enabled

### Known gaps

- **No shared Go/Python conformance fixture.** 6 of the 10 review findings were
  runtime drift between the two implementations. That parity is currently held by
  prose comments naming the other runtime's identifiers, which is weaker than it
  should be for behavior that must match. Tracked in #<issue>.
- **`EditFileContent` ambiguity threshold diverges** (pre-existing, untouched by
  this PR): Go errors only when the trimmed `old_string` is under 5 characters;
  Python always errors on multiple matches without `replace_all`.


go/adk/pkg/tools/shell.go
```diff
@@ -108,6 +110,201 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+	if err != nil {
+		return "", err
+	}
+
+	if len(entries) == 0 {
+		return "Directory is empty.", nil
+	}
+
+	var result strings.Builder
+	for _, entry := range entries {
```
mesutoezdil, line 221: fixed now.

EItanya (commented):
Just out of curiosity, why do we need this? Can't it just use the shell tools? This is already in a sandbox so why would shell be disabled?

brandonkeung:
> Just out of curiosity, why do we need this? Can't it just use the shell tools? This is already in a sandbox so why would shell be disabled?

Some deployments turn bash off anyway for extra tightness (fewer arbitrary-command-execution surfaces even inside a sandbox), and list_files/grep_file were originally always-on regardless of that choice, which undercut it. Pushed a change gating both behind KAGENT_ENABLE_FILE_SEARCH_TOOLS (default off) so they follow the same opt-in posture as bash instead of being an end-run around it.

go/adk/pkg/skills/shell.go
```diff
@@ -108,6 +110,201 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+				return nil
+			}
+			switch classifyWalkEntry(root, p, d) {
+			case walkEntryUnreadable:
+				skipped++
+			case walkEntryGrep:
+				// A read error on one file (permission denied, a line
+				// exceeding the scan buffer, etc.) shouldn't abort matches
+				// already found elsewhere in the tree.
+				if grepErr := grepFile(p); grepErr != nil {
```
mesutoezdil: fixed now.

go/adk/pkg/skills/shell.go
```diff
@@ -108,6 +110,201 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+					skipped++
+				}
+			}
+			return nil
+		})
+		if err == nil && skipped > 0 && result.Len() == 0 {
```
mesutoezdil: fixed now.

python/packages/kagent-skills/src/kagent/skills/shell.py
```diff
@@ -126,6 +126,140 @@ def edit_file_content(
...
+                # identical to a genuinely empty search -- see the note below.
+                skipped += 1
+                continue
+    else:
+        if not file_or_dir_path.is_file():
+            raise OSError(f"{file_or_dir_path} is not a regular file")
+        results.extend(grep_file(file_or_dir_path))
+
+    if not results:
+        if skipped:
```
mesutoezdil, line 321: fixed now.

mesutoezdil:
2 more things i saw while reading

- shell.go ReadFileContent (unrelated to this pr) never bumps the scanner buffer like GrepContent does right below it, so read_file just errors out completely on a file w one line over 64kb instead of truncating it

- shell.go EditFileContent (also pre existing) only errors on ambiguous multi match when trimmed old_string is under 5 chars, but the py twin errors on any count>1 no matter the length, so edit_file acts diff depending on which runtime the agent is on

python/packages/kagent-skills/src/kagent/skills/shell.py
brandonkeung: addressed in [e3f9c76](https://github.com/kagent-dev/kagent/pull/2267/commits/e3f9c76d181ce4571a01d16d07ca693a869529e0)

go/adk/pkg/skills/shell.go
brandonkeung: addressed in [e3f9c76](https://github.com/kagent-dev/kagent/pull/2267/commits/e3f9c76d181ce4571a01d16d07ca693a869529e0)

go/adk/pkg/skills/shell.go
brandonkeung: addressed in [e3f9c76](https://github.com/kagent-dev/kagent/pull/2267/commits/e3f9c76d181ce4571a01d16d07ca693a869529e0)

go/adk/pkg/tools/shell.go
brandonkeung: addressed in [e3f9c76](https://github.com/kagent-dev/kagent/pull/2267/commits/e3f9c76d181ce4571a01d16d07ca693a869529e0)

brandonkeung:
bump (prevent auto close PR)

go/adk/pkg/tools/shell.go
```diff
@@ -108,6 +110,230 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+		// follows symlinks) to classify those correctly, matching Python's
+		// pathlib Path.is_dir(), which follows symlinks by default.
+		if entry.Type()&fs.ModeSymlink != 0 {
+			if target, statErr := os.Stat(filepath.Join(path, entry.Name())); statErr == nil && target.IsDir() {
+				fmt.Fprintf(&result, "%s/\n", entry.Name())
+				continue
+			}
+		}
+
+		info, err := entry.Info()
```
mesutoezdil, line 248: wrong size for a symlink. shows symlink size, not file size.

go/adk/pkg/tools/shell.go
```diff
@@ -108,6 +110,230 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+	if statErr != nil {
+		return walkEntryUnreadable, ""
+	}
+	if fi.IsDir() {
+		// p is a symlink to a directory: WalkDir doesn't recurse into
+		// symlinked directories, and grepFile would fail trying to read
+		// one as a file, so skip it rather than treating it as an error.
+		return walkEntrySkip, ""
+	}
+	if !fi.Mode().IsRegular() {
```
mesutoezdil: read_file can hang on a fifo. no fix like grep_file's.

python/packages/kagent-skills/src/kagent/skills/shell.py
```diff
@@ -126,6 +126,143 @@ def edit_file_content(
...
+                raise OSError(f"{file_or_dir_path} could not be read: {walk_err}") from walk_err
+        # A nested subdirectory that couldn't be read shouldn't abort
+        # matches already found in sibling directories -- just count it.
+        skipped += len(walk_errors)
+
+        for entry in sorted(entries):
+            # is_file() follows symlinks and checks S_ISREG, so this also
+            # excludes FIFOs/sockets/devices -- opening one for reading can
+            # block indefinitely (e.g. a FIFO with no writer connected).
+            if not entry.is_file():
```
mesutoezdil: broken symlink skipped silently. no skip count here.

python/packages/kagent-skills/src/kagent/skills/shell.py
```diff
@@ -126,6 +126,143 @@ def edit_file_content(
...
+                continue
+            try:
+                # Skip entries whose symlink-resolved target escapes the
+                # root being searched, so a symlink can't be used to read
+                # files outside the requested directory.
+                safe_entry = _validate_path(entry, allowed_root)
```
mesutoezdil: allowed_root too wide. symlink can escape this folder.

go/adk/pkg/tools/shell.go
```diff
@@ -108,6 +110,230 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+		}
+		defer file.Close()
+
+		scanner := bufio.NewScanner(file)
+		scanner.Buffer(make([]byte, 64*1024), 1024*1024)
+		lineNum := 1
+		for scanner.Scan() {
+			if line := scanner.Text(); re.MatchString(line) {
+				if len(line) > 2000 {
+					line = line[:2000] + "..."
```
mesutoezdil: cuts by byte. can break utf-8.

go/adk/pkg/tools/shell.go
```diff
@@ -108,6 +110,230 @@ func EditFileContent(path string, oldString, newString string, replaceAll bool)
...
+		expr = "(?i)" + expr
+	}
+	re, err := regexp.Compile(expr)
+	if err != nil {
+		return "", fmt.Errorf("invalid pattern: %w", err)
+	}
+
+	info, err := os.Stat(path)
+	if err != nil {
+		return "", err
```
mesutoezdil: error not wrapped. line above wraps it.

go/adk/pkg/tools/skills.go
```diff
@@ -242,7 +375,7 @@ func resolveReadPath(sessionID, skillsDirectory, requestedPath string) (string,
 		return "", err
 	}
 
-	sessionRoot, err := filepath.Abs(sessionPath)
+	sessionRoot, err := filepath.EvalSymlinks(sessionPath)
```
mesutoezdil, line 385: three functions repeat same steps. share one function?

python/packages/kagent-adk/src/kagent/adk/tools/skills_toolset.py
```diff
@@ -50,19 +53,30 @@ def __init__(self, skills_directory: str | Path):
         self.read_file_tool = ReadFileTool(skills_directory)
         self.write_file_tool = WriteFileTool()
         self.edit_file_tool = EditFileTool()
+        # list_files/grep_file are opt-in: they give an agent broad
+        # filesystem visibility, so some deployments want them off unless
+        # explicitly enabled, same as bash. A single list (rather than two
+        # separately-nullable attributes) keeps "both present or both
+        # absent" a structural guarantee instead of a convention the two
+        # attributes have to be kept in sync by hand.
+        self._file_search_tools: list[BaseTool] = (
```
mesutoezdil, line 62: bash always added. go skips it. match go?

go/adk/pkg/tools/skills.go
```diff
@@ -199,31 +258,105 @@ func NewSkillsTools(skillsDirectory string) ([]tool.Tool, error) {
...
+			if strings.TrimSpace(in.Path) == "" {
+				return "Error: No file path provided", nil
+			}
+
+			path, err := resolveReadPath(ctx.SessionID(), absSkillsDir, in.Path)
+			if err != nil {
+				return fmt.Sprintf("Error searching %s: %v", strings.TrimSpace(in.Path), err), nil
+			}
+
+			content, err := skillruntime.GrepContent(path, in.Pattern, in.Recursive, in.IgnoreCase)
```
mesutoezdil: no timeout here. python has 30s.

go/adk/pkg/tools/skills.go
```diff
@@ -12,6 +13,31 @@ import (
...
 
+// enableFileSearchToolsEnv gates the list_files and grep_file tools, which
+// are opt-in (disabled by default): they let an agent walk the filesystem
+// under its session/skills roots without a shell, which some deployments
+// want to keep off by default alongside bash rather than enable implicitly.
+//
+// Also registered (separately, for `kagent env` CLI discoverability only,
+// not read here) as KagentEnableFileSearchTools in go/core/pkg/env/kagent.go
+// -- keep both string literals in sync if this name ever changes.
+const enableFileSearchToolsEnv = "KAGENT_ENABLE_FILE_SEARCH_TOOLS"
```
mesutoezdil, line 25: same string in two files. share one constant?

mesutoezdil (commented):
some questions and small findings from a close read.

mesutoezdil:
resolve conflicts pls

go/adk/pkg/tools/shell.go
brandonkeung: Correct. `entry.Info()` is Lstat-based, so for a symlink it returned the length of the stored target path rather than the file's size — and `entry.IsDir()` reported the link rather than its target, so a symlinked directory also lost its trailing `/`.

Fixed in 07e4d627. `ListDirContent` now branches on `entry.Type()&fs.ModeSymlink` and uses `os.Stat`, which follows the link — matching Python's `Path.is_dir()`/`Path.stat()`, which follow by default. A broken link prints the bare name, as Python does when its `stat()` raises there.

go/adk/pkg/tools/shell.go:233-246. Covered by `TestListDirContent`.

go/adk/pkg/tools/shell.go
brandonkeung: Correct — and rather than copy grep's check across, both tools now share one reader, `scanFileLines`, which stats before opening and rejects anything non-regular. Putting it there means no future caller can omit it.

That also closed a second divergence in the same seam that wasn't visible from this line: `read_file` was on bufio's default 64KB token limit while grep set a larger buffer, so `read_file` failed outright on any file with a long line (minified JSON, for instance). Both now share `maxLineBytes`.

d5fb0090, go/adk/pkg/tools/shell.go:38-71. Tests: `TestReadFileContent_FIFOReturnsErrorInsteadOfHanging`, `TestReadFileContent_LongLineIsTruncatedNotFatal`.

python/packages/kagent-skills/src/kagent/skills/shell.py
brandonkeung: Correct. Fixed in 07e4d627, with the classification made explicit in 0e303ee3.

A walk entry now resolves to one of three outcomes — GREP / SKIP / UNREADABLE — via `_classify_walk_entry`. A broken symlink is UNREADABLE, so it increments `skipped` and appears in the trailer. A FIFO stays a silent SKIP: it isn't a read failure, and not blocking on it is the point.

shell.py:43-90 and shell.py:302-321. Tests: `test_grep_content_counts_broken_symlink_as_unreadable`, `test_classify_walk_entry_covers_each_outcome`, `test_grep_content_does_not_count_fifo_as_unreadable`.

python/packages/kagent-skills/src/kagent/skills/shell.py
brandonkeung: Correct, and stricter than the nearby Copilot finding: validating `allowed_root` once let a symlinked entry resolve outside the directory actually being searched. The bound should be the searched root, not the allowed root.

Each entry is now resolved and checked with `is_relative_to(root)` before it is opened — 07e4d627, shell.py:86-90. Go's equivalent is `classifyWalkEntry` in go/adk/pkg/tools/grep.go.

Tests: `test_grep_content_recursive_skips_symlinks_that_escape_root` and `..._escape_the_searched_directory`.

go/adk/pkg/tools/shell.go
brandonkeung: Correct. Go string slicing is byte-indexed, so `s[:n]` could split a multi-byte sequence and leave a U+FFFD. It also silently made the limit mean *bytes* — so the same line truncated at a different point than in the Python runtime (whose `str` slicing is per-code-point) and than the tool description promises, which says "characters".

Replaced with `truncateRunes`, which ranges the string (one iteration per rune, with `i` the byte index where that rune starts) and cuts on a rune boundary. 07e4d627, shell.go:82-94. `TestTruncateRunes`.

go/adk/pkg/tools/shell.go
brandonkeung: Correct — `ListDirContent` returned the bare `os.ReadDir` error while everything around it wrapped, so the caller lost the path. Now `fmt.Errorf("failed to read directory %q: %w", path, err)`. 07e4d627, shell.go:211-214.

go/adk/pkg/tools/skills.go
brandonkeung: Agreed, and the dedup mattered less than what it exposed: the three copies had already drifted from each other.

They now share `resolveSandboxedPath`, parameterized by a `pathPolicy` value, so every difference is data instead of a divergent copy:

    readPolicy  = pathPolicy{filepath.EvalSymlinks, true, "the allowed roots"}
    editPolicy  = pathPolicy{filepath.EvalSymlinks, false, "the writable session directory"}
    writePolicy = pathPolicy{resolvePathWithExistingParents, false, "the writable session directory"}

Read is the only one permitted outside the session directory; write is the only one that resolves through not-yet-existing parents. d5fb0090, skills.go:338-420.

python/packages/kagent-adk/src/kagent/adk/tools/skills_toolset.py
brandonkeung: I checked the Go path and the two already match — Go registers bash unconditionally too. The `if fileSearchEnabled` next to it only appends a hint to bash's *description*; the tool itself is constructed outside that branch. That was true at e3f9c76 as well, so this is a clarification rather than a change.

The gate is deliberately narrow: list_files/grep_file grant filesystem visibility, whereas bash already implies it — gating bash on this flag would change unrelated existing behavior.

Since the code read ambiguously enough to prompt the question, both runtimes now say so at the gate — "Note this gate is theirs alone -- bash below is always registered." — at skills.go:234-236 and skills_toolset.py:56-58.

go/adk/pkg/tools/skills.go
brandonkeung: Half fixed, half a deliberate difference — worth separating the two.

Fixed: `GrepContent` now takes a `context.Context` and checks it once per walk entry (not per line — a single file's scan is bounded by its size, the entry count isn't), so a recursive search is abortable by whatever deadline the caller already set. d5fb0090, grep.go:95 and :140.

Deliberate: no fixed 30s cap on top of that. Python's cap exists for a hazard Go doesn't have — Python's `re` backtracks, so an adversarial pattern can hang on a single line. Go uses RE2: linear-time, no backtracking. That leaves only the walk unbounded, and its cost scales with the tree the deployment itself provisioned, so a fixed cap would break large legitimate searches without closing a distinct risk.

The reasoning is recorded in the comment at skills.go:283-296 so it doesn't get rediscovered later as an oversight.

go/adk/pkg/tools/skills.go
brandonkeung: They shouldn't share one, because they aren't quite the same thing — but you're right that an unenforced duplicate literal is a trap, so it's now pinned by a test rather than by trust.

The `go/core/pkg/env` entry exists only for `kagent env` CLI discoverability; it is never read to decide the gate. The ADK reads the var directly, and deliberately parses it differently: `BoolVar.Get()` uses `strconv.ParseBool` (var.go:186), which rejects e.g. "tRue", while `fileSearchToolsEnabled` strips-and-lowercases into ("1","t","true") to exactly match Python's `file_search_tools_enabled()`. Routing the gate through the registry would silently change which literal values enable the tools, and split Go from Python.

So instead: `TestEnableFileSearchToolsEnvMatchesRegistry` asserts the two literals are equal, and each site carries a comment naming the other. skills.go:15-25, core/pkg/env/kagent.go:83-87, skills_test.go:25.

EItanya (commented):
🤖 AI-generated review.

I found three reproducible issues at commit `b4f76af0`:

1. **Python’s regex timeout cannot prevent the runtime from hanging.** [file_tools.py:262](https://github.com/kagent-dev/kagent/blob/b4f76af0f3af1b17715260f4cd2db3195dbaeeb6/python/packages/kagent-adk/src/kagent/adk/tools/file_tools.py#L262)  
   Python’s regex engine holds the interpreter lock during catastrophic backtracking, blocking the event loop too. Searching `(a+)+$` against 32 `a`s followed by `!` blocked both a 50 ms timeout and an independent heartbeat until my watchdog killed it after 10 seconds. The existing test uses `sleep`, which releases that lock. Use a terminable subprocess or an engine with enforceable matching limits.

2. **An entry’s permission error discards successful matches.** [shell.py:302](https://github.com/kagent-dev/kagent/blob/b4f76af0f3af1b17715260f4cd2db3195dbaeeb6/python/packages/kagent-skills/src/kagent/skills/shell.py#L302)  
   A directory with mode `0444` allows listing filenames but prevents inspecting them. `entry.is_file()` raises `PermissionError` outside the per-entry handler, aborting the search and losing earlier sibling matches. Classify stat/resolve failures as unreadable entries and continue.

3. **Python silently changes regex patterns.** [file_tools.py:248](https://github.com/kagent-dev/kagent/blob/b4f76af0f3af1b17715260f4cd2db3195dbaeeb6/python/packages/kagent-adk/src/kagent/adk/tools/file_tools.py#L248)  
   `.strip()` removes meaningful whitespace: `"foo "` incorrectly matches `"foobar"`. Preserve the original pattern when validating and executing it.

Go tools tests passed. All 54 selected Python tests passed after correcting a local interpreter-path issue. The three additional reproduction tests fail as described. No live-cluster validation was run.

brandonkeung:
> 🤖 AI-generated review.
> 
> I found three reproducible issues at commit `b4f76af0`:
> 
> 1. **Python’s regex timeout cannot prevent the runtime from hanging.** [file_tools.py:262](https://github.com/kagent-dev/kagent/blob/b4f76af0f3af1b17715260f4cd2db3195dbaeeb6/python/packages/kagent-adk/src/kagent/adk/tools/file_tools.py#L262)
>    Python’s regex engine holds the interpreter lock during catastrophic backtracking, blocking the event loop too. Searching `(a+)+$` against 32 `a`s followed by `!` blocked both a 50 ms timeout and an independent heartbeat until my watchdog killed it after 10 seconds. The existing test uses `sleep`, which releases that lock. Use a terminable subprocess or an engine with enforceable matching limits.
> 2. **An entry’s permission error discards successful matches.** [shell.py:302](https://github.com/kagent-dev/kagent/blob/b4f76af0f3af1b17715260f4cd2db3195dbaeeb6/python/packages/kagent-skills/src/kagent/skills/shell.py#L302)
>    A directory with mode `0444` allows listing filenames but prevents inspecting them. `entry.is_file()` raises `PermissionError` outside the per-entry handler, aborting the search and losing earlier sibling matches. Classify stat/resolve failures as unreadable entries and continue.
> 3. **Python silently changes regex patterns.** [file_tools.py:248](https://github.com/kagent-dev/kagent/blob/b4f76af0f3af1b17715260f4cd2db3195dbaeeb6/python/packages/kagent-adk/src/kagent/adk/tools/file_tools.py#L248)
>    `.strip()` removes meaningful whitespace: `"foo "` incorrectly matches `"foobar"`. Preserve the original pattern when validating and executing it.
> 
> Go tools tests passed. All 54 selected Python tests passed after correcting a local interpreter-path issue. The three additional reproduction tests fail as described. No live-cluster validation was run.

Thanks — all three reproduced, and all three are fixed.

**1. The regex timeout couldn't fire.** Confirmed exactly as described. `re` holds the
GIL while matching, so the match froze every thread including the event loop, and
`asyncio.wait_for` never got scheduled to fire. My repro: a 50ms heartbeat ticked twice,
then went silent the instant `(a+)+$` dispatched; the 0.5s timeout never fired and the
process needed a hard kill at 8s. You were right about the test too — it faked slowness
with `time.sleep`, which *releases* the GIL, so it only ever proved that `wait_for` works
when the worker cooperates.

Fixed in 028a5fa by moving the match to a worker process (`_run_with_deadline`,
file_tools.py:67). A process has its own GIL, so the loop stays responsive and the
deadline is honored — and unlike a thread the runaway can be killed, which it is:
`shutdown()` won't stop a worker mid-match, so `_kill_workers` (file_tools.py:37)
terminates it. Verified in-cluster: the same pattern now returns the proper error at 33s
(30s + overhead), /proc shows no leaked worker, and the next grep on that agent succeeds.

Two consequences I'd rather state than leave in the diff:

- **`grep_file` costs ~100ms more per call** (measured 113/100/100ms vs ~0ms warm). A
  shared pool avoids that, but then a timeout must recycle shared state — killing workers
  belonging to unrelated in-flight greps and surfacing `BrokenProcessPool` to callers that
  did nothing wrong. At ~100ms inside a multi-second model turn I didn't think that was
  worth buying. Happy to revisit if you disagree.
- **Mock-based tests for `grep_file` no longer work.** A `MagicMock` can't be pickled to a
  worker, and the child imports the real module regardless. The two tests that patched
  `grep_content` are now real-file tests — stronger, but worth knowing before adding more.

Tests: `test_grep_file_tool_times_out_on_catastrophic_backtracking` runs a genuinely
pathological match and asserts the event loop kept ticking.
`test_grep_work_runs_in_a_separate_process` is a fast tripwire, because the behavioral
test *hangs* rather than fails if the offload regresses to a thread, and there's no
pytest-timeout plugin here to cut it short.

**2. A permission error discarded accumulated matches.** Confirmed. `Path.is_file()`
swallows ENOENT/ENOTDIR/EBADF/ELOOP but *raises* on EACCES, and the classifier call sat
outside the `try` guarding the per-file read — so one unstattable entry aborted the search
and threw away matches already found in sibling directories.

Worth noting why the existing permission tests missed it: they use mode 0o000, where
`os.walk` itself fails and the error is already handled by the walk_errors path. Your
0o444 is the case that slips through — readable, so `os.walk` lists the children, but not
searchable, so stat on each child is denied.

Fixed in 0b0bfe8 (shell.py:71-85): stat failures are classified UNREADABLE, so the caller
counts them and keeps walking — matching Go's `classifyWalkEntry`, which already maps every
EvalSymlinks/Stat failure to `walkEntryUnreadable`. Tests:
`test_classify_walk_entry_counts_a_permission_error_as_unreadable` and
`test_grep_content_keeps_sibling_matches_when_an_entry_cannot_be_statted`. I also corrected
the docstring, which claimed this divergence was "not reachable through any case we could
construct" — `chmod 0444` reaches it trivially.

**3. The pattern was silently rewritten.** Confirmed: `"foo "` became `"foo"`, which matches
`foobar` — text the pattern excludes. Fixed in 028a5fa (file_tools.py:317,322): trim only to
decide emptiness, search with what was asked for. That's the split skills.go already makes,
so the runtimes no longer disagree on identical input. Test:
`test_grep_file_tool_passes_the_pattern_through_verbatim` uses a file containing both
`foobar` and `foo bar` and asserts only line 2 matches.

**Cross-runtime check.** Since two of these were Go/Python divergences, I ran the same two
scenarios against the Go runtime through the full stack (AgentInstance → Substrate actor →
Go ADK) and got byte-identical output:

    fixA/aaa_ok/match.txt:1:foo readable

    (1 entries could not be read)

    fixC/data.txt:2:foo bar

@EItanya

EItanya (approved):

PR #2269, opened 2026-07-16
feat: make agent Deployment rollout strategy configurable

### LINKED ISSUE #2253: [FEATURE] Make the agent Deployment rollout strategy configurable in `SharedDeploymentSpec`
## Problem

The agent Deployment's update strategy is hardcoded and cannot be overridden from the Agent CR. In [`go/core/internal/controller/translator/agent/manifest_builder.go:599-604`](https://github.com/kagent-dev/kagent/blob/main/go/core/internal/controller/translator/agent/manifest_builder.go#L599-L604) every agent Deployment is rendered with:

```go
Strategy: appsv1.DeploymentStrategy{
    Type: appsv1.RollingUpdateDeploymentStrategyType,
    RollingUpdate: &appsv1.RollingUpdateDeployment{
        MaxUnavailable: &intstr.IntOrString{Type: intstr.Int, IntVal: 0},
        MaxSurge:       &intstr.IntOrString{Type: intstr.Int, IntVal: 1},
    },
},
```

`SharedDeploymentSpec` ([`go/api/v1alpha2/agent_types.go:433`](https://github.com/kagent-dev/kagent/blob/main/go/api/v1alpha2/agent_types.go#L433)) already exposes many pod/deployment knobs — `Replicas`, `Volumes`, `VolumeMounts`, `Affinity`, `Tolerations`, `NodeSelector`, `SecurityContext`, `PodSecurityContext`, `ExtraContainers` — but there is no field for the rollout strategy, so operators cannot change it.

## Why this matters

`RollingUpdate{maxUnavailable: 0, maxSurge: 1}` is a reasonable default for a stateless agent, but it wedges any agent that mounts a **`ReadWriteOnce` volume**. On upgrade the surge pod is created *before* the old pod terminates; if it schedules to a different node than the one holding the RWO volume, the new pod stays `Pending` on `Multi-Attach`/`FailedAttachVolume` and the rollout stalls indefinitely.

Any operator attaching persistent, node-bound storage to an agent (via the existing `volumes`/`volumeMounts` fields) hits this. The generic k8s answer — `strategy.type: Recreate`, which terminates the old pod before starting the new one so the volume detaches cleanly — is exactly what cannot be expressed today.

RWO storage is the common default on many clusters (RWX requires EFS/NFS/CephFS or similar), so "just use RWX" is not always available, and `maxUnavailable: 0` is precisely the setting that makes RWO unusable across a rollout.

## Proposal

Add an optional rollout-strategy field to `SharedDeploymentSpec` and use it in `manifest_builder.go` instead of the hardcoded literal, falling back to the current `RollingUpdate{0,1}` when unset (fully backward compatible). For example:

```go
// Strategy overrides the agent Deployment update strategy.
// Defaults to RollingUpdate{maxUnavailable: 0, maxSurge: 1} when unset.
// +optional
DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
```

An operator with an RWO-backed agent would then set:

```yaml
spec:
  declarative:
    deployment:
      deploymentStrategy:
        type: Recreate
```

Exposing the whole `DeploymentStrategy` keeps parity with core k8s and covers the `Recreate` case as well as tuned `RollingUpdate` values; a narrower `strategyType` enum would also solve the RWO case if a smaller surface is preferred.

## Prior art

This is the same "extend `SharedDeploymentSpec` with one first-class, backward-compatible field" shape that was already accepted in [#1646 (Add `ExtraContainers` to `SharedDeploymentSpec`)](https://github.com/kagent-dev/kagent/issues/1646).

Happy to open the PR for this if the direction and the field shape (full `DeploymentStrategy` vs. a `strategyType` enum) look good.

paurosello:
Fixes #2253

## Problem

The agent Deployment's update strategy is hardcoded to `RollingUpdate{maxUnavailable: 0, maxSurge: 1}` in the manifest builder. That default wedges any agent mounting a `ReadWriteOnce` volume: on upgrade the surge pod is created before the old pod terminates, and if it lands on a different node it stays `Pending` on `Multi-Attach`/`FailedAttachVolume` errors indefinitely. The generic Kubernetes answer — `strategy.type: Recreate` — cannot be expressed today. See #2253 for the full motivation.

## Changes

- Add an optional `deploymentStrategy` field (`*appsv1.DeploymentStrategy`) to `SharedDeploymentSpec` (v1alpha2), used by the manifest builder instead of the hardcoded literal.
- **Backward compatible defaulting**: when unset, the strategy is exactly the previous `RollingUpdate{maxUnavailable: 0, maxSurge: 1}`. When set with type `RollingUpdate` (or type omitted), any unset `rollingUpdate` fields are merged with those defaults. This also means the emitted Deployment always matches what the API server stores — emitting `nil` fields verbatim would let the server default them to `25%`, and the reconciler's `DeepEqual` comparison would then fire an `Update` on every reconcile, looping forever. `Recreate` is passed through as-is.
- Defaulting lives in `getDeploymentStrategyOrDefault` next to the other `getDefault*` helpers in the resolve functions, and deep-copies the input so the built Deployment never aliases the Agent object.
- **CEL validation**: `type` is restricted to `RollingUpdate`/`Recreate` (omitted means RollingUpdate), and `type: Recreate` combined with a `rollingUpdate` block is rejected at Agent admission (the apps/v1 API server would otherwise reject the Deployment later, surfacing only as a reconcile failure). No SandboxAgent-specific rule: the deployment block is ignored for `SandboxAgent` today and a larger API change is planned post 0.10.x.

## Example

```yaml
spec:
  declarative:
    deployment:
      deploymentStrategy:
        type: Recreate
```

## Testing

- Table-driven unit test for `getDeploymentStrategyOrDefault` (nil fallback, Recreate passthrough, empty type, partial `rollingUpdate` merge, aliasing regression check).
- Translator golden test `agent_with_deployment_strategy` (Recreate); existing goldens unchanged, confirming byte-identical default output.
- Envtest CEL test (`TestDeploymentStrategyCELValidation`, 5 cases) exercising the validation rules against a real kube-apiserver.
- CRDs regenerated via `make controller-manifests` (including the `helm/kagent-crds` copies).


go/api/v1alpha2/agent_types.go
```diff
@@ -484,6 +485,16 @@ type SharedDeploymentSpec struct {
...
+	// DeploymentStrategy overrides the agent Deployment update strategy.
+	// Useful for agents mounting ReadWriteOnce volumes, which require the
+	// Recreate strategy to avoid multi-attach errors during rollouts.
+	// When unset, defaults to RollingUpdate with maxUnavailable 0 and maxSurge 1.
+	// When set with type RollingUpdate (or with type omitted), any unset
+	// rollingUpdate fields default the same way: maxUnavailable to 0 and
+	// maxSurge to 1. Recreate is passed through as-is.
+	// +kubebuilder:validation:XValidation:rule="!(has(self.type) && self.type == 'Recreate' && has(self.rollingUpdate))",message="rollingUpdate may not be specified when strategy type is Recreate"
+	// +optional
+	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
```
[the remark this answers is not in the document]
paurosello: Good catch — addressed in b32f6843. Added a field-level CEL rule restricting `type` to `RollingUpdate`/`Recreate` (omitted still means RollingUpdate), regenerated the CRDs, and added an envtest case rejecting an invalid type.

go/api/v1alpha2/sandboxagent_types.go
```diff
@@ -38,6 +38,7 @@ type SandboxAgent struct {
 }
 
 // +kubebuilder:validation:XValidation:rule="!has(self.skills)",message="spec.skills is not supported for sandbox agents"
+// +kubebuilder:validation:XValidation:rule="!(has(self.declarative) && has(self.declarative.deployment) && has(self.declarative.deployment.deploymentStrategy)) && !(has(self.byo) && has(self.byo.deployment) && has(self.byo.deployment.deploymentStrategy))",message="deploymentStrategy is not supported for sandbox agents"
```
peterj: since sandbox agents don't support deployment strategy, it might make sense to move the deploymentStrategy out of the shared spec and just have it in the declarative spec. what do you think?

go/core/internal/controller/translator/agent/deployments.go
```diff
@@ -67,6 +70,40 @@ func getDefaultResources(spec *corev1.ResourceRequirements) corev1.ResourceRequi
 	return *spec
 }
 
+// getDefaultDeploymentStrategy returns the Deployment update strategy to use,
+// falling back to RollingUpdate{maxUnavailable: 0, maxSurge: 1} when unset.
+// A partial RollingUpdate strategy is merged with those defaults so the emitted
+// Deployment matches what the API server stores, avoiding reconcile hotloops.
+// The input is deep-copied so the result never aliases the Agent CR object.
+func getDefaultDeploymentStrategy(strategy *appsv1.DeploymentStrategy) appsv1.DeploymentStrategy {
```
peterj: nit: perhaps `getDeploymentStrategyOrDefault`

go/core/internal/controller/translator/agent/deployments.go
```diff
@@ -67,6 +70,40 @@ func getDefaultResources(spec *corev1.ResourceRequirements) corev1.ResourceRequi
...
+	if strategy == nil {
+		return appsv1.DeploymentStrategy{
```
peterj: you probably don't need this check for nil strategy here. you could create the `out` if strategy is not nil and then set the values like you're already setting them (if none of them are set, you're return the default values anyway). also, you can early return if the type is recreate.

peterj (changes requested):
added a couple of smaller comments

go/api/v1alpha2/sandboxagent_types.go
paurosello: makes sense

go/api/v1alpha2/agent_types.go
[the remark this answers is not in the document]
paurosello: should be fixed now

go/api/v1alpha2/sandboxagent_types.go
paurosello: I just checked and this would then affect BYO agent and moving it to declarative doesn't remove it from the sandbox schema as SandboxAgentSpec inlines AgentSpec

go/core/internal/controller/translator/agent/deployments.go
paurosello: changed

go/api/v1alpha2/sandboxagent_types.go
EItanya: Unfortunately this is more nuanced, we really need to remove all deployment related config from the SandboxAgent, but that’s a much larger change

go/core/internal/controller/translator/agent/deployments.go
paurosello: Good call, simplified

go/api/v1alpha2/sandboxagent_types.go
EItanya: Can you just remove this block, the whole deployment block is ignored for SandboxAgent today. We are currently planning a larger API change post `0.10.x`

paurosello: Removed in 7276a9d9 — dropped the CEL rule and its envtest cases, CRDs regenerated.

go/core/internal/controller/translator/agent/deployments.go
```diff
@@ -67,6 +70,35 @@ func getDefaultResources(spec *corev1.ResourceRequirements) corev1.ResourceRequi
...
+	}
+	// Recreate carries no rollingUpdate block; pass it through as-is.
+	if out.Type != appsv1.RollingUpdateDeploymentStrategyType {
+		return out
+	}
+	if out.RollingUpdate == nil {
+		out.RollingUpdate = &appsv1.RollingUpdateDeployment{}
+	}
+	if out.RollingUpdate.MaxUnavailable == nil {
+		out.RollingUpdate.MaxUnavailable = &intstr.IntOrString{Type: intstr.Int, IntVal: 0}
```
mesutoezdil, line 93: when the user sets only maxsurge to 0 this defaults maxunavailable to 0 too and the apps api rejects a 0/0 rolling update, so the agent passes admission but the deployment fails on every reconcile, consider skipping this default when maxsurge is 0.

paurosello: maxUnavailable now defaults to 1 when maxSurge is 0, and an explicit 0/0 rolling update is rejected at admission via CEL. Added unit + envtest coverage.

paurosello:
can we reopen this one?

PR #2274, opened 2026-07-16
refactor: hide pgx errors behind database.ErrNotFound

QuentinBisson:
## Why

Follow-up to the review discussion on #2187: the `ErrNotFound` sentinel was dropped there so the pgx-error cleanup could land on its own, covering all the places pgx leaks through the `database.Client` interface rather than just one.

Today, callers outside the database package match driver errors directly (`errors.Is(err, pgx.ErrNoRows)` in the share-token middleware, error-response logging, and sandbox session materialization). That couples HTTP and A2A code to the storage driver. It also means REST handlers cannot tell a missing row from a backend failure, so several endpoints answer 404 during a database outage.

## What changed

**Sentinel and boundary (no behavior change):**

- `go/api/database/errors.go`: exported `ErrNotFound` sentinel. Implementations wrap it with call-site context; callers match with `errors.Is`.
- `client_postgres.go`: every single-row read (`GetAgent`, `GetSession`, `GetSessionShareByToken`, `GetTask`, `GetPushNotification`, `GetTool`, `GetToolServer`) maps `pgx.ErrNoRows` to the sentinel via a small `notFoundOr` helper. Other errors pass through unchanged. `GetCrewAIFlowState` keeps its existing nil-on-missing contract.
- Consumers migrated off pgx: `httpserver/middleware.go` (share token 403 vs 500), `httpserver/middleware_error.go` (log-level decision), `a2a/substrate_sandbox_transport.go` (ensureSessionRow). No non-test file outside the database package imports pgx for error matching anymore; `pkg/migrations` keeps its own driver-level handling, which is not client-API leakage.

**REST not-found gating (behavior fix):**

- Session, session-share, and task handlers mapped every database read error to 404, so a backend outage read as "session/task not found" to API clients that treat 404 as a definitive signal. `RespondNotFoundOrError` in `handlers/helpers.go` writes 404 only for `ErrNotFound` and 500 for anything else; all ten blanket sites in `sessions.go`, `session_shares.go`, and `tasks.go` now go through it.

**5xx bodies no longer echo the underlying error (behavior fix):**

- `RespondWithError` in `httpserver/middleware_error.go` previously appended the underlying error text to *every* response body. It now does so only for sub-500 responses; 5xx bodies carry the generic message and the underlying error is logged server-side instead. This is a deliberate information-disclosure fix, and it applies to the whole API surface that funnels through this writer, not just the not-found sites touched here. Clients or tests that parsed detail out of 5xx bodies will see the generic message; 4xx bodies are unchanged.

## Tests

- `TestSingleRowReadsMapMissingToErrNotFound` covers all seven getters against a live postgres.
- `TestRespondNotFoundOrError` covers the 404/500 split.
- `TestRespondWithErrorBody` pins the new boundary: 4xx keeps the underlying detail, 5xx returns only the generic message.
- Existing middleware, share, session, and transport tests updated to match on the sentinel.

## What this unblocks

#2187 can map `errors.Is(err, database.ErrNotFound)` on the `contextId` path of `ListTasks` to `a2a.ErrTaskNotFound`, giving A2A clients a typed not-found (`-32001`) instead of a generic internal error. External clients probing a session over `tasks/list` (or `GET /api/sessions/{id}`) can then tell "gone" from "backend failure".

### DIFF
--- a/go/core/internal/httpserver/middleware_error.go
+++ b/go/core/internal/httpserver/middleware_error.go
@@ -8,7 +8,7 @@ import (
 	"net"
 	"net/http"
 
-	"github.com/jackc/pgx/v5"
+	"github.com/kagent-dev/kagent/go/api/database"
 	apierrors "github.com/kagent-dev/kagent/go/core/internal/httpserver/errors"
 	"github.com/kagent-dev/kagent/go/core/internal/httpserver/handlers"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
@@ -67,7 +67,7 @@ func (w *errorResponseWriter) RespondWithError(err error) {
 		underlying = err
 	}
 
-	if underlying != nil && !errors.Is(underlying, pgx.ErrNoRows) {
+	if underlying != nil && !errors.Is(underlying, database.ErrNotFound) {
 		log.Error(underlying, message)
 	} else {
 		log.Info(message)
--- a/go/core/internal/httpserver/middleware_error.go
+++ b/go/core/internal/httpserver/middleware_error.go
@@ -76,7 +76,9 @@ func (w *errorResponseWriter) RespondWithError(err error) {
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(statusCode)
 	errMsg := message
-	if underlying != nil {
+	// 5xx bodies carry only the generic message: the underlying error is
+	// backend-internal detail (already logged above) and must not reach clients.
+	if underlying != nil && statusCode < http.StatusInternalServerError {
 		errMsg = message + ": " + underlying.Error()
 	}
 	json.NewEncoder(w).Encode(map[string]string{"error": errMsg}) //nolint:errcheck
--- a/go/api/database/errors.go
+++ b/go/api/database/errors.go
@@ -1,0 +1,8 @@
+package database
+
+import "errors"
+
+// ErrNotFound reports that the requested record does not exist (or is not
+// visible to the given user). Match with errors.Is; implementations wrap it
+// with call-site context.
+var ErrNotFound = errors.New("record not found")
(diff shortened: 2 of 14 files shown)


go/core/internal/httpserver/middleware_error.go
```diff
@@ -67,7 +67,7 @@ func (w *errorResponseWriter) RespondWithError(err error) {
 		underlying = err
 	}
 
-	if underlying != nil && !errors.Is(underlying, pgx.ErrNoRows) {
+	if underlying != nil && !errors.Is(underlying, database.ErrNotFound) {
 		log.Error(underlying, message)
 	} else {
```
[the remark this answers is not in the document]
QuentinBisson: Fixed in 8dab542: 5xx bodies now carry only the generic message; the underlying error is logged server-side (that path already ran through log.Error above). Non-5xx responses keep the detail, which callers rely on for 400/403/404 context.

go/core/internal/httpserver/middleware_error.go
```diff
@@ -76,7 +76,9 @@ func (w *errorResponseWriter) RespondWithError(err error) {
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(statusCode)
 	errMsg := message
-	if underlying != nil {
+	// 5xx bodies carry only the generic message: the underlying error is
+	// backend-internal detail (already logged above) and must not reach clients.
+	if underlying != nil && statusCode < http.StatusInternalServerError {
```
mesutoezdil, line 81: this strips the underlying detail from every existing 500 response across the api, not just the new sites, so it deserves a mention in the pr description since clients or tests may rely on that text.

QuentinBisson: Good catch, that's intentional but I'd under-documented it. I updated the PR description and added a test

EItanya (approved):

PR #2275, opened 2026-07-16
Bump minimum python version to 3.11

iplay88keys:
# Description

Raises the minimum supported Python version from 3.10 to 3.11 across the Python workspace. `kagent-adk` already calls `asyncio.Task.uncancel()` (added in Python 3.11) in its A2A agent executor (#2176), so 3.10 was broken in practice; this makes the declared floor match. 3.10 also reaches [end of life](https://devguide.python.org/versions/) in October 2026.

- `requires-python`: `>=3.10` → `>=3.11` in all 15 workspace pyprojects (root, packages, samples). The existing `<3.14` caps on the crewai packages are unchanged.
- ruff/black `target-version`: `py310` → `py311` where set.
- CI: `python-test` matrix drops 3.10 (now 3.11/3.12/3.13); `python-lint` installs 3.11.
- `uv.lock` regenerated: removes `async-timeout`, `exceptiongroup`, and `backports-asyncio-runner` (3.10-only stdlib backports) and collapses the numpy 2.2.6/2.3.2 split resolution to 2.3.2 only.
- Removes the now-dead `from datetime import UTC` try/except 3.10 fallbacks in `kagent-langgraph` and `kagent-openai` (ruff UP017 flags them at the py311 target; its autofix mangles the fallback branch, so they are deleted instead). The `typing.override` fallbacks stay — those need 3.12.
- CLI Python scaffolding template (`go/core/cli/internal/mcp/frameworks/python/templates/pyproject.toml.tmpl`): ruff/black `target-version` and mypy `python_version` bumped `py310`/`3.10` → `py312`/`3.12` to match the template's own `requires-python = ">=3.12"`.

## Test plan

- CI `python-test` matrix (3.11–3.13) runs the full test suite.
- `uv lock` resolves cleanly (272 packages).


iplay88keys:
@EItanya, `python-test (3.10)` is marked as required and will need to be removed from the required status checks in the repo's ruleset for `main`.

peterj:
removed 3.10 from the status checks

peterj (approved):

EItanya (approved):

PR #2280, opened 2026-07-17
emit OTel spans for memory operations (Go + Python ADK, controller DB)

### LINKED ISSUE #1909: [FEATURE] emit OTel spans for memory operations (save / load / prefetch / extractor)
### 📋 Prerequisites

### 📝 Feature Summary

Add dedicated OTel spans for the memory subsystem (memory.write / memory.read / memory.embed / memory.consolidate / memory.evict), alongside the existing gen_ai.* spans on invoke_agent. Purely additive instrumentation — no behavior change, no new runtime dependencies.

kagent's platform-level OTel pipeline is already excellent: A2A metadata propagates as span attributes (v0.9.3), the controller's `invoke_agent` span carries `gen_ai.agent.*` + `gen_ai.provider.name` per the OTel GenAI semconv (verified live against a v0.9.4 deployment, Go OTel SDK 1.43.0), and `helm/kagent/values.yaml` exposes a clean `otel.tracing` block.

What's missing is **dedicated spans for the memory subsystem**: `save_memory`, `load_memory`, `prefetch_memory`, and the auto-extractor that fires every 5th user message.

Without dedicated spans, memory operations are visible only as opaque HTTP POSTs against `/api/sessions/{ctx-id}/events` — which is fine for HTTP-level latency but makes it impossible to compare kagent against other agent-memory backends on retrieval latency, embedding cost, or write amplification.

**Verified baseline (Dynatrace, 2026-05-20, kagent v0.9.4)** — 48 h DQL scan in a live tenant:

| Surface                     | Emitted today |
| --------------------------- | --- |
| Agent invocation            | `invoke_agent` span with `gen_ai.agent.id`, `gen_ai.agent.name`, `gen_ai.operation.name=invoke_agent`, `gen_ai.provider.name=ollama`. Service: `kagent-controller` v0.9.4, OTel Go SDK 1.43.0. |
| HTTP session/task surface   | `POST /api/sessions/{ctx-id}/events`, `POST /api/tasks`, `GET /api/agents` via `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` v0.68.0. HTTP-standard attrs only. |
| A2A worker (Python runtime) | `a2a.server.events.in_memory_queue_manager.*` via Python A2A SDK auto-instrumentation (in-process queue, not user-facing memory). |

**Confirmed gap:** zero spans named `memory.*` across the window; the session-events HTTP span carries no `memory.operation` / `memory.store.kind`; no `memory_*` metric names.

**Who benefits:** kagent operators (retrieval-latency + embedding-cost visibility), the agent-memory benchmarking workstream (apples-to-apples comparison across OSS backends), and the OTel GenAI semconv WG (a reference implementation to point at when `memory.*` is proposed upstream).

### Spans (new)

| Span name            | Kind     | Required attributes                                       | Optional |
| -------------------- | -------- | --------------------------------------------------------- | --- |
| `memory.write`       | INTERNAL | `memory.operation`, `memory.store.kind`                   | `memory.tenant`, `memory.input.size_bytes`, `memory.extracted.facts_count` |
| `memory.read`        | INTERNAL | `memory.operation`, `memory.store.kind`, `memory.query.k` | `memory.tenant`, `memory.results.count`, `memory.top_similarity` |
| `memory.embed`       | INTERNAL | `memory.embedder.model`                                   | `memory.embed.token_count` |
| `memory.consolidate` | INTERNAL | `memory.consolidate.kind`                                 | `memory.consolidate.input_items` |
| `memory.evict`       | INTERNAL | `memory.evict.reason`                                     | `memory.evict.count` |

`memory.read.kind=prefetch` distinguishes recall-before-LLM-dispatch reads from explicit `load_memory` tool calls.

### Resource attributes (set once per process)

- `memory.sut.name=kagent`
- `memory.sut.architecture=vector`
- `memory.sut.store_backend=pgvector`
- `memory.sut.version` (git SHA or release tag)

### Reuse of existing `gen_ai.*`

For embedding calls inside `memory.embed` and any LLM dispatch inside the auto-extractor we reuse the GenAI semconv kagent already emits on `invoke_agent`:

- `gen_ai.system`, `gen_ai.request.model`
- `gen_ai.operation.name` (extended with `memory.write.extract`, `memory.read.rerank`)
- `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`

No new GenAI conventions — we slot in alongside what's already there.

### Parent-span hygiene

Memory-read spans emitted during a request must be **children of the existing `invoke_agent` span** when the recall happens before LLM dispatch. Keeps the trace tree connected with what users already see in Dynatrace / Honeycomb / Tempo and avoids orphan trees.

### Files 

- `go/internal/memory/store.go` — wrap `save_memory`, `load_memory`, `prefetch_memory`.
- `go/internal/memory/extractor.go` — `memory.consolidate` on the auto-extractor (fires every 5th message).
- **Session-events handler** behind `POST /api/sessions/{ctx-id}/events` — wrap with a child `memory.write` span when the payload is a memory-bearing event. **Highest-leverage single change** — the HTTP span already exists; we just attach business-level semantics.
- `go/pkg/telemetry/spans.go` — add `MemoryOperationName` constants reusing the existing tracer (same provider that wires `invoke_agent` in v0.9.4).
- `docs/observability/memory.md` (NEW).
- `helm/kagent/values.yaml` — document new memory span names under the existing `otel.tracing` block.

Targets the declarative runtime. The BYO ADK runtime documents Memory API as unsupported and is out of scope.

### Question:

1. **Naming sanity check.** Are `memory.read/write/embed/consolidate/evict` reasonable next to the existing `gen_ai.*` envelope on `invoke_agent`?
2. **PR shape.** Single PR with all six files, or phased (handler-only first, then extractor + tools)?
3. **Runtime targeting.** Confirm declarative runtime is the right target (BYO ADK out of scope per docs).
4. **Helm doc placement.** OK to extend the existing `otel.tracing` block in `values.yaml`, or do you want a new `otel.memory` sub-block?

I'll wait for a signal here before creating any PR.

### 🔄 Alternatives Considered

1. **Wait for upstream OTel GenAI semconv to land memory-* natively.** Viable but slow — the GenAI WG cadence has been ~quarterly. We'd rather ship a reference implementation kagent operators can use today and migrate when upstream solidifies. We're tracking `memory-semconv v0.1.0` as the interim contract.
2. **HTTP-attribute overload on `POST /api/sessions/{ctx-id}/events`.** Add `memory.operation` / `memory.store.kind` as attributes on the existing HTTP span instead of creating new spans. Rejected: doesn't model `prefetch_memory` (no HTTP boundary) or the auto-extractor (background tick); also conflates HTTP timing with retrieval timing in dashboards.
3. **Metrics-first (counters + histograms) instead of spans.** Useful but insufficient — metrics can't show parent→child causality (which `invoke_agent` triggered which `memory.read` with which `memory.query.k`). Spans first; metrics derivable from span attributes later via OTel Collector connectors.
4. **Custom kagent-specific attribute namespace (`kagent.memory.*`).** Rejected: locks operators into kagent-specific dashboard queries instead of OTel-portable ones. Following the `gen_ai.*` precedent kagent already adopts.

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

- **Why now / context.** These conventions are being drafted as `memory-semconv v0.1.0` across six OSS agent-memory projects (MemPalace, kagent, sympozium, Graphiti, Mem0, Letta) so a benchmark harness can compare retrieval latency, write amplification, and bi-temporal invalidation churn apples-to-apples. The plan is to ship the convention in **two** implementations first, then propose it upstream to the **OpenTelemetry Semantic Conventions GenAI WG** — same pattern `gen_ai.*` followed.
- **Why kagent is the strongest CNCF-context candidate to land it first:**
  1. The OTel scaffolding is already empirically verified in production.
  2. The `gen_ai.*` envelope already in place proves the maintainers accept OTel semconv as the canonical naming source.
  3. The contribution is purely additive — no behavior change, no new runtime dependencies, OTel Go SDK 1.43.0 already in `go.mod`.
- **Talk credibility note (informational).** This contribution is referenced in an upcoming KubeCon + OSS Summit talk benchmarking OSS agent-memory solutions on Kubernetes (Cognee / MemOS / Honcho / kagent / sympozium / MemPalace). Whatever maintainers decide here is what gets cited — acceptance is not a precondition for the talk.
- **Suggested labels:** `area/observability`, `kind/proposal`. (Add `good first issue` if maintainers feel that fits — totally optional.)

henrikrexed:
Closes #1909

## Why

kagent's platform-level OpenTelemetry is already strong: the controller's `invoke_agent` span carries the `gen_ai.*` GenAI semconv attributes, and HTTP auto-instrumentation covers the session/task endpoints. But the **memory subsystem is a blind spot** — long-term memory (pgvector-backed) reads, writes, and consolidation emit **no dedicated spans**, so operators can't see when an agent recalled or stored a memory, how many items came back, or whether a recall was filtered out by the similarity gate. Memory is increasingly where agent behaviour is decided, so this is exactly the layer teams need to observe and, eventually, govern.

This PR closes that gap for **both** agent runtimes (Go ADK and Python ADK), plus the controller's data-access layer, and does it purely additively — no behaviour change, no new runtime dependencies beyond `opentelemetry-api`, and everything exports only when `OTEL_TRACING_ENABLED=true`.

## What changed

**Memory spans** (children of the existing `invoke_agent` trace, so the tree stays connected with the `gen_ai.*` spans you already see):

| Span | Emitted by | `memory.operation` |
|------|-----------|--------------------|
| `memory.read` | recall / `search_memory` | `prefetch` (child of `invoke_agent`) |
| `memory.write` | the `save_memory` tool's live write path + session ingestion | `save` |
| `memory.consolidate` | LLM fact extraction on the summarizing path | `extract` |

- **Go ADK** (`go/adk/pkg/memory`, `go/adk/pkg/telemetry/memory.go`) and **Python ADK** (`_memory_service.py`, `_memory_telemetry.py`) emit the same convention.
- **Controller pgvector DB layer** (`go/core/internal/database`) emits `db.memory.*` child spans, cardinality- and PII-guarded (never puts raw embedding vectors or query/content text on attributes).

**Attribute convention** — each span carries `memory.operation`, `memory.scope`, `memory.source`, `memory.index_ref`, `memory.injection_result` (`injected`/`filtered` from the min-score gate), `memory.item.count`, and a `memory.sut.{name,architecture,store_backend}` descriptor (`kagent`/`vector`/`pgvector`). Only attributes kagent can populate truthfully today are emitted; a governance vocabulary (`memory.status`, `memory.authority`, `promote`/`revoke`) is intentionally reserved, not stubbed, until the subsystem models memory lifecycle and injection authority.

**Trace hygiene / configurable auto-instrumentation** — the a2a-SDK and httpx-client auto-instrumentations become **opt-out** via helm (`otel.tracing.a2aSdkInstrumentation` / `httpxClientInstrumentation`, both default `true` for parity), forwarded to agent pods as `OTEL_INSTRUMENTATION_*` env. The delegation `execute_tool` span is enriched with `a2a.*` attributes (subagent, context lineage, task id/state). Trace continuity is preserved even with the instrumentations disabled: request event-hooks re-inject W3C trace context (correlation headers only, no extra spans), so agent→controller memory calls and agent→agent hops stay in one trace.

Docs for the new span names live under `helm/kagent/values.yaml` (`otel.tracing`).

## Tests

- **Go** — `go/adk/pkg/memory` (read span injected/filtered + attributes; write span on the tool path) and `go/core/internal/database` (asserts the `db.memory.search` span fires and never leaks embedding vectors / raw text). `go build` + `go vet` clean.
- **Python** — 51 unit tests pass (`ruff` clean): memory read/write spans + attributes, a2a delegation attributes, remote-a2a trace continuation, and outbound trace-context propagation, using in-memory span exporters.
- **Live-verified** end-to-end in a Dynatrace tenant on a real cluster: `memory.read`/`memory.write` land from both the Go and Python runtimes with the expected attributes and correct parent-span nesting (`memory.read` under `invoke_agent`; `memory.write` under the `save_memory` `execute_tool` span).

## What this unblocks

- **Memory observability as a first-class signal** — operators can see recall/store rate, hit-vs-filtered ratios, and per-operation latency across both runtimes, in whatever backend they already use (Dynatrace / Honeycomb / Tempo / Jaeger).
- **A path to memory *governance*** — the reserved `memory.*` vocabulary gives a forward-compatible contract; when kagent models memory lifecycle/authority, those attributes drop onto the same spans with no span-name churn.
- **Trace-cost control** — teams can dial trace verbosity (drop a2a/httpx plumbing spans) without losing cross-agent/-service correlation.
- **A concrete reference implementation** for the memory-operation span convention discussed in #1909.

image


python/packages/kagent-adk/src/kagent/adk/_memory_service.py
```diff
@@ -337,60 +390,71 @@ async def _summarize_session_content_async(
...
+                                "LLM returned valid JSON but not a list of strings. Falling back to full text."
+                            )
+                    except json.JSONDecodeError:
+                        logger.warning(
+                            "Failed to parse LLM output as JSON. Falling back to full text. Output: %s", summary_text
+                        )
+                        pass
+
+                logger.warning("Empty summary or invalid format returned, using original content")
+                span.set_attribute(mtel.ATTR_MEMORY_ITEM_COUNT, 0)
```
mesutoezdil: go's summarizecontent stamps item.count 1 on this same fallback because the original content is still stored as one item, so the two runtimes disagree here and it would be good to align them.

mesutoezdil:
Could you have a look at the comments of bots above?
and DCO is still red
and + branch is not up-to-date

henrikrexed:
I have updated the branch (now feba0d2a on the updated branch)
The Python fallback now stamps memory.item.count=1 (the original content is stored as one item), matching Go's summarizeContent and the Python success path (number of items returned). 
Also fixed the exception fallback and updated the docstring so both runtimes agree on every memory.consolidate path.

henrikrexed:
Thanks @Copilot — all 8 comments addressed in `6f6e6e36`:

- **`_memory_telemetry.py`** — `record_span_error` now imports `Status`/`StatusCode` explicitly from `opentelemetry.trace` and passes a `Status` instance to `set_status()`.
- **`_memory_service.py`** (4 early-return paths, lines 111/117/185/191) — each now sets `memory.item.count=0` so the attribute is always present.
- **`_propagation.py`** — docstring corrected: httpx client auto-instrumentation is **enabled by default (opt-out)**, matching `_utils.py`.
- **`client_postgres.go`** — `db.*` attributes now use semconv v1.39.0 constants (`DBSystemNamePostgreSQL`, `DBOperationName`, `DBCollectionName`); keys/values byte-identical.
- **`kagent_service.go`** — `memory.consolidate` sets `memory.item.count` before every return and records errors via `recordSpanError`.

python/packages/kagent-adk/src/kagent/adk/_memory_service.py
henrikrexed: Good catch , aligned in `036721f7` (now `feba0d2a` on the updated branch). The Python fallback now stamps `memory.item.count=1` (the original content is stored as one item), matching Go's `summarizeContent` and the Python success path (number of items returned). Also fixed the exception fallback and updated the docstring so both runtimes agree on every `memory.consolidate` path.

PR #2284, opened 2026-07-18
feat(ui): shell-style message history recall in the chat composer

Feelings0220:
ArrowUp on the composer's first line steps back through this session's previously sent messages (consecutive duplicates collapsed); ArrowDown steps forward and finally restores the unsent draft. Guards keep normal editing intact: recall only fires with a collapsed cursor on the first/last line, never during IME composition, and any manual edit or send exits history-browsing mode. The cursor lands at the end of each recalled message.

### DIFF
--- a/ui/src/components/chat/ChatInterface.tsx
+++ b/ui/src/components/chat/ChatInterface.tsx
@@ -171,6 +171,27 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
   // Shared call_id -> is_error lookup so each group summary is O(group size).
   const toolResultsByCallId = useMemo(() => buildToolCallResultsIndex(allMessages), [allMessages]);
 
+  // Shell-style composer history: this session's user message texts, oldest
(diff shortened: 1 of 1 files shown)


ui/src/components/chat/ChatInterface.tsx
```diff
@@ -925,6 +947,50 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
...
+
+        if (e.key === "ArrowUp" && collapsed && onFirstLine) {
+          if (historyIndexRef.current === null) {
+            historyDraftRef.current = value;
+            historyIndexRef.current = userMessageHistory.length;
+          }
+          if (historyIndexRef.current > 0) {
+            e.preventDefault();
+            historyIndexRef.current -= 1;
+            const recalled = userMessageHistory[historyIndexRef.current];
```
mesutoezdil, line 1038: historyIndexRef survives changes to userMessageHistory, so after a session reload shrinks the list this can read past the end and put undefined into the textarea, clamping the index to the current length would be safer.

PR #2285, opened 2026-07-20
reduce redundant fetching and optimize rendering (ui)

peterj:
- reduce redundant api requests 
- moving inital agent, model, mcp and prompt library loading to server-rendered routes
- automatically refresh agent status when agent is created


ui/src/components/sidebars/GroupedChats.tsx
```diff
@@ -38,61 +35,24 @@ export default function GroupedChats({
...
-    const onSuspended = () => void refresh();
-    window.addEventListener("harness-session-suspended", onSuspended);
-
-    return () => {
-      cancelled = true;
-      clearInterval(interval);
-      window.removeEventListener("harness-session-suspended", onSuspended);
-    };
-  }, [isHarness, sessions, agentNamespace, agentName]);
+  const sessionStatuses = useMemo<Record<string, SessionActorState>>(() => {
```
mesutoezdil, line 48: The status API is per session, so this stamps the single polled session's state onto every chat in the sidebar and other chats can show the wrong indicator.

peterj: harness chats are multiplexed as ACP sessions inside a single harness actor, so every session ID resolves to the same actor ID (even if there are multiple sessions/chats).

EItanya (approved):

Issue #2197, opened 2026-07-09
A2A ListTasks: push all-sessions listing down to SQL


QuentinBisson:
Follow-up to #2187 (Copilot review finding).

## Problem

The store-backed `ListTasks` handles an all-sessions query (empty `contextId`) by loading every session for the user, then every task per session, then filtering, sorting, and paginating in memory. Latency and memory scale with the user's total task count even for a small page, and it issues N+1 queries (one `ListSessions` plus one `ListTasksForSession` per session).

## Proposed fix

Add a single sqlc query that does the work server-side:

```sql
-- name: ListTasksForUser :many
SELECT task.*, COUNT(*) OVER() AS total
FROM task
JOIN session ON session.id = task.session_id
WHERE session.user_id = $1
  AND task.deleted_at IS NULL
  AND session.deleted_at IS NULL
ORDER BY task.id
LIMIT $2 OFFSET $3;
```

and a `ListTasksForUser` method on the database client, then have `collectUserTasks` use it for the all-sessions path. The single-session path can reuse the same shape with a `session_id` predicate.

## Open question: status filters

`status` and `statusTimestampAfter` live inside `task.data`, which is a `TEXT` column, so filtering them in SQL needs either:

1. `(data::jsonb -> 'status' ->> 'state')` casts per row (no index, and the cast errors on any malformed row), or
2. a migration adding generated columns (e.g. `status_state`, `status_timestamp`) or converting `data` to `JSONB`, with indexes.

Option 2 is the durable one but needs a new numbered migration (migration files are immutable in CI). A middle ground that already removes the N+1 and unbounded scan: push only the join + LIMIT/OFFSET down and keep the two JSON filters in memory on the fetched page window when no filter is set; when a filter is set, fall back to the current in-memory path until the schema supports it.

EItanya:
This makes sense to me. Can we start with option 1 for status for now?

QuentinBisson:
Resolved by #2598: the cross-session ListTasks path was removed and the new `a2agateway` listing filters, paginates and counts in SQL. Closing, #2302 is closed for the same reason.

PR #2302, opened 2026-07-20
perf(a2a): push ListTasks all-sessions query down to SQL

QuentinBisson:
Follow-up to #2187, addressing #2197 (Copilot review finding).

## What

`collectUserTasks` served an all-sessions ListTasks query by loading every session for the user, then every task per session, then filtering, sorting, and paginating in memory. An N+1 query whose result set and app-side memory scaled with the user's total task count even for a small page.

This replaces that path with a single `ListTasksForUser` sqlc query that joins `session` to `task`, filters, orders by `task.id`, and paginates server-side, returning `COUNT(*) OVER()` as the total. The app now materializes one page instead of the user's entire task set.

This is a memory/transfer win, not (yet) a DB-latency win. As option 1 (no schema change) there is still no index on `session.user_id`, so the join drives off a sequential scan of the whole `session` table, and the status filter casts `task.data::jsonb` per row. See "Remaining latency" below.

## Ownership

The query applies both gates the Go loop applied: the session belongs to the caller, and the task itself does under the `task.user_id` rule from #2265 (including the NULL-owner fallback for rows written before the owner column existed). `session.id` is not globally unique (the session key is `(id, user_id)`), so the session join alone would have handed a foreign task parked in the caller's session to the caller.

## Status filtering (issue option 1)

`status` and `statusTimestampAfter` are filtered in SQL via a `jsonb` cast on `task.data`, with no schema migration (option 1 from the issue, per the request to "start with option 1 for status for now").

Only the v1 vocabulary is matched (`TASK_STATE_WORKING`, etc.). Since the v1.0 cutover in #2354, `parseVersionedTask` rejects any row whose `protocol_version` is not `1.0`, so a legacy-shaped row cannot be returned by any read path regardless of what the filter matches. Carrying a second legacy spelling in the query would not make such a row reachable.

`data` is always a JSON object (`json.Marshal` output), so `::jsonb` never errors. The timestamp cast is guarded by a `CASE` (ordered evaluation) matching a full RFC3339 timestamp, so an absent, null, empty, wrong-typed, or non-timestamp-shaped value is filtered out rather than erroring the query. A syntactically valid but out-of-range value (`2026-02-30T00:00:00Z`) would still error the cast; every writer of `task.data` formats from a `time.Time`, so that shape is not produced in practice, and option 2's generated column would reject it at write time instead.

`ListTasksForUser` and `CountTasksForUser` carry byte-identical WHERE clauses.

## Behavior preserved

The single-session path keeps its fail-closed `GetSession` ownership check (a missing or other-user context is an error, not an empty page). Ordering by `task.id` matches the prior in-memory sort, so page tokens stay stable.

`COUNT(*) OVER()` rides on the returned rows, so a page requested past the end of the set (e.g. rows deleted between paginated calls) comes back empty and would carry no total. `CountTasksForUser` recovers the true filtered count in that case, matching the old in-memory behavior.

## Remaining latency

Measured on a seeded database (2000 users, 50k sessions, 200k tasks), one page of 50:

| state | plan | execution |
|---|---|---|
| as merged here | Seq Scan on session, 49975 rows removed by filter | 6.1 ms |
| plus `CREATE INDEX ON session(user_id) WHERE deleted_at IS NULL` | bitmap index scan | 0.61 ms |
| plus `task(user_id)` as well | bitmap AND | 0.92 ms |

The `session.user_id` index is a one-line migration and is independent of option 2 (the generated column + index for status). Adding `task(user_id)` on top does not help.

## Tests

Adds a Postgres integration test (`TestListTasksForUser`) covering cross-user isolation, the single-session predicate, status and timestamp filters, LIMIT/OFFSET pagination with a stable total, and the empty over-range page. One case covers the owner gate directly: a task owned by another user sitting in the caller's session stays hidden. Existing `task_query_store` unit tests pass unchanged against a fake that mirrors the query semantics.


go/api/database/client.go
```diff
@@ -55,6 +68,7 @@ type Client interface {
 	ListTools(ctx context.Context) ([]Tool, error)
 	ListFeedback(ctx context.Context, userID string) ([]Feedback, error)
 	ListTasksForSession(ctx context.Context, sessionID string, userID string) ([]*a2a.Task, error)
+	ListUserTasks(ctx context.Context, params ListUserTasksParams) (tasks []*a2a.Task, total int, err error)
```
EItanya: Last remaining question, it's not clear to me why you need `total` here and you can't just use `len(tasks)`

QuentinBisson: Hey @EItanya I'm sorry i was on holidays. We need the total from the database (`COUNT(*) OVER()`) because the task list is paginated so len(tasks) is not the total list of tasks as it is based of the offset and limits

QuentinBisson:
I rebased this onto current main and dropped the now-dead legacy-status matching from the v1.0 cutover.
I also fixed a cross-user leak: ListUserTasks was only checking session ownership, not the task's own user_id. Along the way I renamed ListUserTasks/CountUserTasks to ListTasksForUser/CountTasksForUser and made userID a required argument instead of a params-struct field to match the other functions

QuentinBisson:
@EItanya answering here as well, my earlier reply was buried in a review thread.

`total` cannot be `len(tasks)` because pagination now happens in SQL (`LIMIT`/`OFFSET`), so `len(tasks)` is only the size of the page that was returned. The response needs the full filtered count for `TotalSize` and to decide whether to emit a `NextPageToken`. That is what `COUNT(*) OVER()` gives: the count over the whole filtered set, before `LIMIT`. With 4 matching tasks and a page size of 2, `len(tasks)` is 2 and `total` is 4. The old in-memory path got the same number for free as `len(filtered)` since it materialized everything; pushing pagination down is what turns it into an explicit return value.

Also rebased on main (it was conflicting, now a single commit). Post-rebase `TestListTasksForUser` plus the full `core/internal/database` and `core/internal/a2a` packages pass locally against the pgvector test container.

One thing worth deciding before merge: there is still no index on `session.user_id`, so the join drives off a sequential scan of the whole `session` table on every ListTasks call, which scales with the installation rather than with the caller. On a seeded database (2000 users, 50k sessions, 200k tasks), one page of 50:

- as is: `Seq Scan on session`, 49975 rows removed by filter, 6.1 ms
- plus `CREATE INDEX ON session(user_id) WHERE deleted_at IS NULL`: bitmap index scan, 0.61 ms
- adding `task(user_id)` on top: 0.92 ms, so not worth it

That index is one line and independent of the option 2 generated column for status, which I would still leave for later. Happy to add it in this PR or a follow-up, whichever you prefer.

QuentinBisson:
Closing: #2598 removed the session and task tables and `task_query_store.go`. The replacement in `a2agateway` already runs task listing with filtering, keyset pagination and the count in SQL (`ListAgentInstanceTasks` / `CountAgentInstanceTasks`), so there is no all-sessions N+1 left to push down.

PR #2309, opened 2026-07-21
chore: helm template and sample metadata cleanups

mesutoezdil:
Three small cleanups.

The NOTES.txt tools block iterated .Values.tools, which does not exist in values.yaml, so the range never rendered anything. Removed the dead block.

controller-configmap.yaml reads controller.agentImage.pullSecret but the key was missing from values.yaml. Added it with a comment so it is discoverable.

samples/adk/basic was the only python sample without requires-python in its pyproject. Added it as >=3.11, <3.14 to match the workspace bound, since the sample is a uv workspace member and a tighter floor would narrow the root lock. Also dropped its stale standalone uv.lock, no other sample carries one and the workspace lock governs members.


helm/kagent/templates/NOTES.txt
```diff
@@ -28,11 +28,6 @@ ENABLED TOOLS:
 {{- if index .Values "kagent-tools" "enabled" }}
   - Tool Server: Kagent's Built-in tools
 {{- end }}
-{{- range $tool, $config := .Values.tools }}
-{{- if $config.enabled }}
-  - {{ $tool }}
-{{- end }}
-{{- end }}
```
EItanya, line 35: Should we just switch this to `.Values "kagent-tools"` like above?

mesutoezdil: there's no top-level `.Values.tools` so this loop never rendered anything. kept the removal and listed grafana-mcp and querydoc with the same enabled-flag pattern instead

EItanya (approved):

PR #2144, opened 2026-07-03
docs: fix typos, stale content, and port inconsistencies

mesutoezdil:
Fix several doc issues found across the repo.

- `DEVELOPMENT.md`: Go version `v1.26.1+` to `v1.26.3+` (matches go.mod), fix duplicate step numbers 1-5, fix UI port `8001` to `3000`
- `Makefile`: wrong Azure env var `AZUREOPENAI_API_KEY` to `AZURE_OPENAI_API_KEY` in check-api-key and helm-install-provider
- `helm/README.md`: wrong Azure env var names (`AZUREOPENAI_API_KEY`, `AZURE_API_KEY`) to `AZURE_OPENAI_API_KEY`, fix copy-paste placeholder in Azure helm install example
- `examples/substrate-openclaw/README.md`: remove stale `runtime: substrate` field (not in CRD), fix UI port `8001` to `3000`
- `docs/substrate-agentharness-lifecycle.md`: remove draft PR note "This branch should use...", update step 5 to reflect that actors are created on demand by the HTTP gateway, not the controller
- `ui/README.md`: `kagents ui` to `kagent ui`
- `python/README.md`: replace hardcoded `OpenAI API key` prerequisite with provider-agnostic wording

PR #2310, opened 2026-07-21
fix(build): use AZURE_OPENAI_API_KEY env var name consistently

mesutoezdil:
The Makefile checked and passed AZUREOPENAI_API_KEY while the Go and Python code read AZURE_OPENAI_API_KEY, so the helm-install targets silently passed an empty Azure key. This renames the Makefile variable and the two helm README export examples to match the code. Split out of #2144 since it changes Makefile behavior rather than docs.


EItanya (commented):
Good catch, I thought we changed these already a while back

EItanya (approved):

PR #2313, opened 2026-07-22
(otel)(fix): don't double-instrument OpenAI on the agents runtime

krisztianfekete:
### What
This PR fixes OpenAI streaming crashing when OTEL is on, and OTEL mutating runtime behavior even when it's off.

### Why
kagent installs the low-level openllmetry `OpenAIInstrumentor` (patches theOpenAI SDK) *plus* `OpenAIAgentsInstrumentor`. The low-level one chokes on the Agents SDK's streaming Responses path with `AttributeError: 'AsyncStream' object has no attribute 'id'`, task fails. And it was installed unconditionally, so you couldn't disable telemetry to mitigate it.

### Changes
- `configure()`: only instrument when tracing or logging is actually enabled
- kagent-openai: pass `instrument_openai_client=False`, rely on`OpenAIAgentsInstrumentor` alone
- pin traceloop instrumentors `<0.53.0` so versions stop silently drifting

### Notes
We could also open a PR to fix this upstream in Traceloop SDKs.


dhaifley (approved):
LGTM

peterj (approved):

PR #2314, opened 2026-07-22
fix: Go ADK Vertex AI client sends requests to wrong endpoint (missing rawPredict URL rewrite)

### LINKED ISSUE #2315: Go ADK: Vertex AI Anthropic requests fail with 401 CREDENTIALS_MISSING due to SDK option ordering
## Bug Description

The Go ADK's `NewAnthropicVertexAIModelWithLogger` in `go/adk/pkg/models/anthropic.go` applies SDK options in the wrong order, causing all Vertex AI Anthropic requests to fail with `401 CREDENTIALS_MISSING`.

## Root Cause

`vertex.WithGoogleAuth()` is called first, which sets up an OAuth2-authenticated HTTP client, the Vertex AI base URL, and URL-rewriting middleware. Then `option.WithHTTPClient()` is appended after it, overwriting the authenticated client with a plain one. This causes:

1. Requests sent **without OAuth2 credentials** (401 Unauthorized)
2. Requests going to `/v1/messages` instead of the correct Vertex AI `rawPredict` endpoint

## Symptoms

- `POST "https://us-east5-aiplatform.googleapis.com/v1/messages": 401 Unauthorized`
- Error: `CREDENTIALS_MISSING` / `UNAUTHENTICATED`
- `"message": "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential."`

## How to Reproduce

1. Deploy an Agent CR with `runtime: go` and a ModelConfig using `provider: AnthropicVertexAI`
2. Run on GKE with Workload Identity configured (KSA annotated with `iam.gke.io/gcp-service-account`)
3. Send any message to the agent via A2A protocol
4. Observe 401 in agent logs

## Expected Behavior

The Go ADK should construct the correct Vertex AI URL (`https://{region}-aiplatform.googleapis.com/v1/projects/{project}/locations/{region}/publishers/anthropic/models/{model}:rawPredict`) and attach OAuth2 credentials from Application Default Credentials.

## Workaround

Switch the Agent to `runtime: python` — the Python runtime correctly handles Vertex AI URL construction and ADC authentication.

## Fix

PR #2314 reorders the options so `option.WithHTTPClient` (TLS/headers/timeout) is applied first, then `vertex.WithGoogleAuth` is applied last, allowing its internal `WithHTTPClient` with the OAuth2-authenticated transport to take precedence.

**Python runtime is unaffected** — only the Go ADK (`golang-adk` image) has this bug.

jon-waggoner:
Fixes #2315

## Summary

Fix the Go ADK Anthropic Vertex AI client to correctly route requests to the Vertex AI `rawPredict` endpoint instead of the direct Anthropic API path `/v1/messages`.

## Problem

When using Anthropic models via Vertex AI, the Go ADK runtime (`golang-adk` image) sends requests to the wrong URL:

```
POST "https://us-east5-aiplatform.googleapis.com/v1/messages": 401 Unauthorized
```

The correct Vertex AI endpoint should be:

```
POST https://{region}-aiplatform.googleapis.com/v1/projects/{project}/locations/{region}/publishers/anthropic/models/{model}:rawPredict
```

The Python runtime works correctly because the Python Anthropic SDK properly constructs the Vertex AI URL.

## Root Cause

In `NewAnthropicVertexAIModelWithLogger`, the SDK options were applied in the wrong order:

```go
// BEFORE (broken): WithGoogleAuth first, then WithHTTPClient overwrites it
opts := []option.RequestOption{
    vertex.WithGoogleAuth(ctx, region, projectID),  // sets up OAuth2 client + URL rewriting
}
httpClient, _ := BuildHTTPClient(config.TransportConfig)
opts = append(opts, option.WithHTTPClient(httpClient))  // OVERWRITES the OAuth2 client!
```

The `vertex.WithGoogleAuth` function (from `anthropic-sdk-go`) internally does three things via `WithCredentials`:
1. Sets `WithBaseURL` to the Vertex AI endpoint
2. Registers middleware that rewrites `/v1/messages` to the `rawPredict` path
3. Sets `WithHTTPClient` with an OAuth2-authenticated HTTP transport

When `option.WithHTTPClient(httpClient)` was appended **after** `WithGoogleAuth`, it replaced the OAuth2-authenticated client with a plain HTTP client (just TLS + headers + timeout). This meant:
- No OAuth2 token was attached to requests → **401 Unauthorized**
- The Vertex AI URL rewriting middleware was ineffective

## Fix

Reorder the options so `WithHTTPClient` is applied **before** `vertex.WithGoogleAuth`, allowing the Vertex option to take precedence:

```go
// AFTER (fixed): WithHTTPClient first, then WithGoogleAuth overrides with OAuth2 client
var opts []option.RequestOption
httpClient, _ := BuildHTTPClient(config.TransportConfig)
opts = append(opts, option.WithHTTPClient(httpClient))           // plain client first
opts = append(opts, vertex.WithGoogleAuth(ctx, region, projectID)) // OAuth2 client overrides
```

This ensures the Vertex AI SDK's authenticated client, base URL, and URL-rewriting middleware all take effect.


go/adk/pkg/models/anthropic.go
```diff
@@ -92,17 +92,27 @@ func newAnthropicModelFromConfig(config *AnthropicConfig, apiKey string, logger
...
-	}
-
-	// Create HTTP client with timeout, custom headers, TLS, and passthrough
+	var opts []option.RequestOption
+
+	// Create HTTP client with timeout, custom headers, and TLS.
+	// This must be applied BEFORE vertex.WithGoogleAuth so that the Vertex
+	// option (which internally calls WithHTTPClient with an OAuth2-authenticated
+	// transport, sets the Vertex AI base URL, and registers middleware that
+	// rewrites /v1/messages to the rawPredict endpoint) takes precedence.
```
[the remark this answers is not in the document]
jon-waggoner: Good call -- the comment is over-specifying SDK internals that could drift as the anthropic-sdk-go evolves. I'll tighten both comments to focus on the ordering contract (last `WithHTTPClient` wins) rather than describing internal middleware behavior. Thanks for the suggestion.

jon-waggoner:
Note: I considered adding a unit test for the option ordering fix but decided against it. The Anthropic SDK's `option.RequestOption` is opaque — there's no way to inspect which `WithHTTPClient` "won" without actually calling GCP auth, and `vertex.WithGoogleAuth` requires real Application Default Credentials. The existing Go model tests (bedrock, ollama, openai) test conversion logic and response mapping, not constructor option ordering. The inline comments documenting the ordering invariant are the appropriate safeguard here.

go/adk/pkg/models/anthropic.go
```diff
@@ -92,17 +92,22 @@ func newAnthropicModelFromConfig(config *AnthropicConfig, apiKey string, logger
...
+	// be set last to avoid being overwritten.
 	httpClient, err := BuildHTTPClient(config.TransportConfig)
 	if err != nil {
 		return nil, err
 	}
 	opts = append(opts, option.WithHTTPClient(httpClient))
 
+	// Must be last: last WithHTTPClient wins, so Vertex auth must come after
+	// any custom HTTP client.
+	opts = append(opts, vertex.WithGoogleAuth(ctx, region, projectID))
```
mesutoezdil: heads up, vertex.WithGoogleAuth builds its own http client n calls withhttpclient itself, so it wipes buildhttpclient's custom tls/headers client entirely not just the oauth part, so custom ca cert or headers on a vertex cfg would silently stop applying

jon-waggoner:
@mesutoezdil You were right — reversing the order swapped one bug for another: `vertex.WithGoogleAuth` installs its own `option.WithHTTPClient`, which wiped the custom TLS/headers/timeout stack from `BuildHTTPClient`. Fixed in 5db14c57 by composing them: resolve Google ADC ourselves, wrap the caller's `*http.Client` Transport with `oauth2.Transport`, then apply `vertex.WithCredentials` first (registers baseURL + URL-rewrite middleware) and `option.WithHTTPClient(composed)` last (overrides only the HTTPClient field; baseURL and middleware are stored separately and middleware is additive). Added unit tests around a new `composeVertexHTTPClient` helper covering OAuth2 header attachment, custom header preservation, base transport preservation, and timeout preservation.

go/adk/pkg/models/anthropic.go
jon-waggoner: see comment below. thanks for callout. https://github.com/kagent-dev/kagent/pull/2314#issuecomment-5144848240

PR #2317, opened 2026-07-22
fix: key STS token cache by acting subject, not session alone

### LINKED ISSUE #2181: STS token cache keyed by session ID collapses per-user identity in shared sessions
## Summary

The STS token-propagation plugin caches the exchanged (delegated) token **keyed by session ID**, so a session shared by multiple human users runs every user's tool calls under the **first** user's delegated identity.

## Detail

`go/adk/pkg/sts/plugin.go`:
- `tokenCache map[string]*TokenCacheEntry // keyed by session ID`
- `BeforeRunCallback` computes `sessionID = ctx.Session().ID()`, and on a cache hit returns the cached token **without re-exchanging the current request's bearer**:

```go
if entry, ok := p.getCachedToken(sessionID); ok {
    return nil, nil // uses cached token
}
// ... otherwise extract bearer, exchange, setCachedToken(sessionID, ...)
```

A single ADK/A2A session (`contextID`) can legitimately receive messages from different subjects: `UserIDCallInterceptor` sets `CallContext.User` per request from `x-user-id`, and `withBearerToken` reads the `Authorization` bearer per request. But once the first request seeds `tokenCache[sessionID]`, later requests with a different subject reuse that entry, so their MCP tool calls (via `headerRoundTripper` / STS `headerProvider`) execute under the first subject's exchanged token. On cache expiry the next request re-seeds it, so attribution can also drift nondeterministically.

## Proposed fix

Key the exchanged-token cache by `sessionID + subject` (the incoming bearer's/`x-user-id`'s subject), or bypass the cache when the incoming subject differs from the cached entry's. That lets a shared session multiplex per-user identity and lets downstreams drop the consent workaround.

QuentinBisson:
## What

The STS token-propagation plugin cached the exchanged (delegated) token keyed
by session ID alone. A session that carries messages from more than one subject
therefore reused whichever subject's token seeded the cache first, so every
later caller's MCP tool calls ran under that first subject's delegated identity.
On cache expiry the next request re-seeded the entry, so the acting identity
could also drift nondeterministically.

This keys the cache by `(sessionID, subject)`, where `subject` is the acting
bearer's issuer-scoped `sub` claim (`iss` + `sub`), falling back to a hash of
the raw token for opaque or sub-less tokens. Each caller in a shared session now
gets and reuses its own exchanged token.

## Why

Shared sessions already deliver per-caller identity everywhere except this
cache. The session-sharing feature (https://github.com/kagent-dev/kagent/pull/1935)
pins `X-User-Id` to the session owner for history, but deliberately leaves
`Authorization` as the caller's own token and records the real caller as
`initiated_by`. So in a shared session a visitor's tool calls are meant to run
under the visitor's own delegated token. The `sessionID`-only cache defeated
that: the owner's (or first visitor's) exchanged token was reused for everyone,
collapsing per-caller identity and leaking one user's delegated authority to
another.

The condition is reproducible entirely within the in-tree share feature: two
users in one shared session, each authenticated with their own token, produce
tool calls that all execute under the first-seen subject.

## Changes

- Cache keyed by `cacheKey{sessionID, subject}`. `subjectKey()` derives the
  discriminator from the bearer as `iss` + `sub` (`sub` is only unique within an
  issuer), or a hash of the raw token when there is no `sub`. The token is parsed
  unverified and used only to partition the cache, never to gate a security
  decision; it is validated server-side on exchange.
- `BeforeRunCallback` and `HeaderProvider` both recover the acting bearer through
  `actingBearer()`, which prefers the value the executor stored
  (`models.BearerTokenKey`) and falls back to the A2A CallContext `Authorization`
  header. The fallback is the same source the round-tripper's `propagateToken`
  path reads, so the acting subject is recoverable at the MCP transport layer
  even when `BearerTokenKey` is not threaded onto the MCP request context;
  without it a lookup could silently miss and inject no token.
- `AfterRunCallback` sweeps all expired entries rather than a single session key.
- Drops `GetTokenForSession`, which had no callers and could not gain one: the
  subject half of the key is derived by the unexported `subjectKey`.

## Behavior change

For existing read-write shared sessions, a visitor's tool calls move from the
owner's exchanged token to the visitor's own token. This corrects an identity
collapse (one user's authority was borrowed by others), but it is externally
observable for anyone who relied on the prior behavior.

Second, bearer recovery now happens before the cache lookup instead of after it,
so a request carrying no bearer no longer reuses a token the session cached
earlier. It skips propagation instead. That is the fail-closed half of keying by
subject (an unidentified caller must not inherit an identified one's token), but
it means any path that fails to thread the bearer loses token injection silently
rather than falling back. `actingBearer`'s CallContext fallback exists to keep
that from happening on the MCP transport path.

## Open question (does not block this fix)

This fix makes the cache honor whichever acting token arrives; it does not
choose an identity model. Whether a shared session should run tools per-caller
or under a single pinned identity (owner/initiator) is orthogonal, and is
decided by which token the fronting proxy forwards, not by this cache.

Worth confirming the intended direction: single-identity tool pinning cannot be
provided in-cluster for UI shares, because no component holds a non-speaking
principal's live token to exchange or refresh. It is only realizable where a
credential-forwarding layer exists (for example an A2A gateway that can forward
the pinned principal's token every turn). The controller already owns the
history half via the share feature. If you agree, kagent stays mode-neutral here
and this fix stands on correctness alone.

Fixes https://github.com/kagent-dev/kagent/issues/2181

### DIFF
--- a/go/adk/pkg/sts/plugin.go
+++ b/go/adk/pkg/sts/plugin.go
@@ -5,11 +5,14 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
+	"github.com/a2aproject/a2a-go/a2asrv"
 	"github.com/go-logr/logr"
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/kagent-dev/kagent/go/adk/pkg/constants"
 	"github.com/kagent-dev/kagent/go/adk/pkg/models"
 	"google.golang.org/adk/v2/agent"
 	adkplugin "google.golang.org/adk/v2/plugin"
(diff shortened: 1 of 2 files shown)


mesutoezdil:
lgtm

PR #2321, opened 2026-07-23
feat: add envFrom support to agent deployment spec

uncycler:
## Summary

Add an `envFrom` field to the agent deployment spec so users can populate
container environment variables from entire ConfigMaps/Secrets (Kubernetes'
native `envFrom`), instead of enumerating each key via `env`.

## Fix

Add EnvFrom []corev1.EnvFromSource to SharedDeploymentSpec and wire it
through the translator to the pod container, exactly parallel to the existing

## Testing

New unit tests assert EnvFrom propagation for both BYO and declarative specs.


go/api/v1alpha2/agent_types.go
```diff
@@ -453,6 +453,10 @@ type SharedDeploymentSpec struct {
 	// Env are additional environment variables set on the agent container.
 	// +optional
 	Env []corev1.EnvVar `json:"env,omitempty"`
+	// EnvFrom are sources (ConfigMaps/Secrets) used to populate environment variables
+	// on the agent container. Values defined in Env with a duplicate key take precedence.
+	// +optional
+	EnvFrom []corev1.EnvFromSource `json:"envFrom,omitempty"`
```
EItanya, line 459: Why is this necessary? `Env` above already has `ValueFrom` support

EItanya (changes requested):

go/api/v1alpha2/agent_types.go
uncycler: These solve different problems. env + valueFrom binds a single ConfigMap/Secret key to a single env var, so every key has to be listed explicitly. envFrom imports all keys from a ConfigMap/Secret at once, using each key as the variable name.

The use case is a ConfigMap/Secret whose key set isn't fixed (populated from an external system, like external secret operator), where enumerating keys in the spec isn't practical. Same reason upstream PodSpec exposes both.

mesutoezdil:
looks good

EItanya (approved):

PR #2323, opened 2026-07-23
feat: add native AWS Bedrock Guardrails support to ModelConfig

### LINKED ISSUE #2320: [FEATURE] Native AWS Bedrock Guardrails support in ModelConfig
### 📋 Prerequisites

### 📝 Feature Summary

Add an optional `guardrail` block to the Bedrock provider in ModelConfig so the controller can pass guardrailIdentifier and guardrailVersion to the Converse API natively, without requiring an external proxy.

The Bedrock provider in ModelConfig only exposes `region` today. 

If you want to apply AWS Bedrock Guardrails (content filtering, topic denial, PII redaction) to your agents, you're forced to deploy a separate OpenAI-compatible proxy that injects the guardrail config at the gateway level. This adds an extra service to maintain and forces you to use `provider: "OpenAI"` with a custom baseUrl instead of the native Bedrock path. 

The Bedrock Converse API already accepts guardrailIdentifier and guardrailVersion in the request body — kagent just doesn't expose them.

### 💡 Proposed Solution

Add an optional `guardrail` block to the `bedrock` spec in ModelConfig:

```yaml
bedrock:
  region: "eu-central-1"
  guardrail:
    identifier: "abc123def456"  # guardrail ID or ARN
    version: "1"                 # guardrail version
    trace: "enabled"            # optional, defaults to disabled
```

The controller should pass these values as GuardrailConfiguration in the Converse API call (or as headers for InvokeModel). When a guardrail triggers INTERVENED, surface the blocked reason to the agent/user.

Ref: https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-converse-api.html

### 🔄 Alternatives Considered

- External proxy (what we do today): works but adds operational complexity and breaks native Bedrock provider usage
- Annotation-based config on the ModelConfig resource: not type-safe, hard to validate
- Separate GuardrailPolicy CRD referenced by ModelConfig: over-engineered for what's essentially two fields

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

Tested on kagent v0.9.0 with provider: Bedrock. The BedrockConfig struct currently only has `Region` — no guardrail-related fields exist in the CRD.

The AWS SDK Go v2 already exposes the relevant types in `github.com/aws/aws-sdk-go-v2/service/bedrockruntime/types`:
- GuardrailConfiguration{GuardrailIdentifier, GuardrailVersion, Trace}

So it's mostly wiring: add the field to the CRD, read it in the controller, pass it to the SDK call.

We're running kagent in production on AWS and would be happy to validate any implementation against real Bedrock Guardrails in our environment.

sanskar-singh-2403:
## What this PR does

Adds an optional `guardrail` block to the Bedrock provider in `ModelConfig`, so the
controller passes the guardrail config to the Bedrock **Converse** and **ConverseStream**
APIs natively. This removes the need to stand up a separate OpenAI-compatible proxy just
to apply AWS Bedrock Guardrails (content filtering, denied topics, PII redaction) to an
agent.

Example config:

```yaml
provider: Bedrock
bedrock:
  region: "eu-central-1"
  guardrail:
    identifier: "abc123def456"   # guardrail ID or ARN
    version: "1"
    trace: "enabled"             # disabled (default) | enabled | enabled_full
```

## Changes

- **`api/v1alpha2`**: new `BedrockGuardrailConfig` type on `BedrockConfig`. `identifier`
  and `version` are required within the block (`MinLength=1`); the whole block is optional.
  `trace` is a typed enum (`disabled` default / `enabled` / `enabled_full`) mapping to the
  SDK's `GuardrailTrace`.
- **`api/adk`**: matching `BedrockGuardrail` transport struct on `Bedrock`, `omitempty`.
- **Controller translator + adk agent**: wire the field through `v1alpha2` -> adk ->
  runtime `BedrockConfig`, nil-guarded so existing configs are unaffected.
- **`adk/pkg/models/bedrock.go`**: build the guardrail config and attach it to both API
  calls. Note the SDK uses two different types here  `Converse` takes
  `types.GuardrailConfiguration`, while `ConverseStream` takes
  `types.GuardrailStreamConfiguration`. The stream path uses `StreamProcessingMode: sync`
  so guardrail interventions are applied **before** content is streamed to the caller,
  rather than leaking blocked content. Per the issue discussion, interventions surface in
  the response content (not as a hard error) so the agent loop isn't broken.
- Regenerated `zz_generated.deepcopy.go` and the `ModelConfig` CRD (controller-gen
  v0.19.0), and synced the `helm/kagent-crds` copy.

## Tests

- Unit tests for both guardrail builders (`bedrockGuardrailConfig`,
  `bedrockGuardrailStreamConfig`) covering the nil / set / empty-trace / stream-mode cases.
- ADK JSON marshalling tests for the `guardrail` block (present and omitted-when-nil).
- `go build`, `go vet`, and `go test` pass for `api/adk`, `adk/pkg/models`, and the
  controller translator package; `gofmt` clean.

## Notes for reviewers

- Scoped to the Converse / ConverseStream path only the current Bedrock provider doesn't
  use `InvokeModel`, so it was intentionally left out (matches the issue scope).
- @GiGi-cloud-76 offered to validate against real Bedrock guardrail configurations
  (content filters, topic denial, PII redaction) in their environment happy to
  coordinate on that.
- cc @peterj

Fixes #2320


sanskar-singh-2403:
@peterj @dmitri-d @davidkarlsen can we review this? Thanks!

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -294,6 +294,24 @@ type BedrockConfig struct {
...
 	// +kubebuilder:default="5m"
 	CacheTTL string `json:"cacheTTL,omitempty"`
+
+	// +optional
+	Guardrail *BedrockGuardrailConfig `json:"guardrail,omitempty"`
+}
+
+type BedrockGuardrailConfig struct {
+	// +required
+	// +kubebuilder:validation:MinLength=1
```
peterj, line 329: nit: add a max length too, since it's set on the aws config.

peterj: (i wonder if we'd need the actual regex for the full arn); but i see it's either a string OR arn, so that might look too complex

sanskar-singh-2403: hey @peterj 

Good call, added `MaxLength` to both fields:

- `identifier`: `MaxLength=2048` (AWS's cap on `guardrailIdentifier`)
- `version`: `MaxLength=8` (numeric version or `DRAFT`)

Regenerated the CRD manifests and synced the helm copy.

On the ARN regex: you are right I left it out on purpose. AWS's real pattern is a
bare-ID branch OR a full-ARN branch:

`^(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail/[a-z0-9]+))$`

Baking that into a kubebuilder `Pattern` would reject at admission, but
AWS already validates the identifier on the Converse call and returns a
clear error. A strict regex mostly adds risk of false rejections (GovCloud
/China partitions, future ARN shapes) for little gain, and as you noted it
reads as noise. The `MinLength=1` / `MaxLength=2048` bound catches the
empty/garbage cases cheaply and defers real validation to AWS.

sanskar-singh-2403: can we re-run the CI?

sanskar-singh-2403:
@peterj should we merge? If everything seems fine 😸

EItanya (changes requested):
Hey there, to be honest I don't think guardrails belong inside of kagent. If we start adding guardrails the scope may blow up. I think this belongs at the network level. i would look into https://agentgateway.dev/

sanskar-singh-2403:
> Hey there, to be honest I don't think guardrails belong inside of kagent. If we start adding guardrails the scope may blow up. I think this belongs at the network level. i would look into https://agentgateway.dev/

hey @EItanya thanks for taking a look, and totally fair concern. I want to make sure I am not
arguing past your point, so let me reframe what this PR actually does and then
defer to your call.

This isn't adding a guardrails engine or any filtering logic to kagent. It exposes
two fields (`identifier` + `version`) that the Bedrock Converse / ConverseStream
APIs already accept in the request body. kagent does zero enforcement here, it just
forwards those values to AWS and AWS does all the filtering. In that sense it is the
same category as `region`, `cacheTTL`, and `promptCaching`, which already live on
`BedrockConfig` as native Bedrock pass-through params.

So my worry with pushing this to the network layer only: for someone who is on the
native Bedrock provider and not fronting it with agentgateway, there is currently no
way to turn on a guardrail they have already configured in their AWS account,
without standing up an OpenAI-compatible proxy just for that (which is what the
original issue #2320 was asking to avoid).

That said, I hear you on keeping traffic-plane concerns in agentgateway, and I don't
want to open the door to per-provider guardrail schemas creeping in. If the
pass-through framing doesn't change your view, I am happy to close this and point
the issue at agentgateway instead. Your call, just let me know.

go/adk/pkg/models/bedrock.go
```diff
@@ -87,7 +87,33 @@ type BedrockConfig struct {
...
+
+func bedrockGuardrailStreamConfig(c *BedrockConfig) *types.GuardrailStreamConfiguration {
+	if c.GuardrailIdentifier == "" {
+		return nil
+	}
+	return &types.GuardrailStreamConfiguration{
+		GuardrailIdentifier:  aws.String(c.GuardrailIdentifier),
+		GuardrailVersion:     aws.String(c.GuardrailVersion),
+		Trace:                types.GuardrailTrace(c.GuardrailTrace),
+		StreamProcessingMode: types.GuardrailStreamProcessingModeSync,
```
mesutoezdil, line 115: when the guardrail actually blocks smth bedrock returns stopreason=guardrail_intervened (also content_filtered) 
but bedrockStopReasonToGenai has no case for either 
so they fall into the default stop, 
so nothing downstream like the ui or logs can tell a response was actually filtered vs a normal completion

sanskar-singh-2403: good catch, thanks. you're right that both fell into default -> STOP so the filtered state was invisible downstream.

fixed: bedrockStopReasonToGenai now has an explicit case for guardrail_intervened and content_filtered, both mapping to genai.FinishReasonSafety (SAFETY). that way the UI/logs can distinguish a filtered response from a normal completion. added test cases for both in bedrock_test.go.

sanskar-singh-2403:
> > Hey there, to be honest I don't think guardrails belong inside of kagent. If we start adding guardrails the scope may blow up. I think this belongs at the network level. i would look into https://agentgateway.dev/
> 
> hey @EItanya thanks for taking a look, and totally fair concern. I want to make sure I am not arguing past your point, so let me reframe what this PR actually does and then defer to your call.
> 
> This isn't adding a guardrails engine or any filtering logic to kagent. It exposes two fields (`identifier` + `version`) that the Bedrock Converse / ConverseStream APIs already accept in the request body. kagent does zero enforcement here, it just forwards those values to AWS and AWS does all the filtering. In that sense it is the same category as `region`, `cacheTTL`, and `promptCaching`, which already live on `BedrockConfig` as native Bedrock pass-through params.
> 
> So my worry with pushing this to the network layer only: for someone who is on the native Bedrock provider and not fronting it with agentgateway, there is currently no way to turn on a guardrail they have already configured in their AWS account, without standing up an OpenAI-compatible proxy just for that (which is what the original issue #2320 was asking to avoid).
> 
> That said, I hear you on keeping traffic-plane concerns in agentgateway, and I don't want to open the door to per-provider guardrail schemas creeping in. If the pass-through framing doesn't change your view, I am happy to close this and point the issue at agentgateway instead. Your call, just let me know.

hey @EItanya any thoughts on this?

GiGi-cloud-76:
Hey @EItanya, chipping in as the original requester here.

I think the key distinction is: this PR doesn't implement any guardrail logic in kagent. It passes two fields (identifier + version) that the Bedrock Converse API already accepts in its request payload — exactly like region, promptCaching, and cacheTTL that are already on BedrockConfig today.

For users who call Bedrock directly (no agentgateway in front), there's currently no way to activate an AWS-managed guardrail without deploying a proxy just to inject those two fields. That's a lot of operational overhead for what amounts to a pass-through parameter.

Requiring users to deploy and maintain agentgateway — an entire network-level component — just to forward two extra fields in a request body feels disproportionate. It's like requiring a service mesh to add a single HTTP header. The complexity doesn't match the ask.

If kagent routes through agentgateway → agreed, guardrails belong at the gateway level. But when the provider is Bedrock and the call goes straight to the Converse API, omitting these fields means AWS guardrails are simply unusable from kagent — even though they're a native, first-class part of the very API being called.

So the ask is narrow: same category as any other Bedrock API parameter that kagent already passes through, not a guardrail engine inside kagent. No new dependencies, no filtering logic, no scope creep — just two optional fields on an existing struct.

Happy to validate on our production environment once this lands.

EItanya (approved):

PR #2327, opened 2026-07-25
fix(controller): keep user env vars over controller defaults

mesutoezdil:
Agent env was built as user env then shared env appended after it. Same name in both means the shared value wins on the pod, since k8s takes the last entry for duplicate names. A user setting OTEL_SERVICE_NAME or KAGENT_NAME on their own agent got silently overwritten by the controller default, no error or warning.

Now shared entries are only added if the user did not already set that name, so the user's own value always wins.

Added a unit test covering the collision case.


0809android (commented):
Please keep the controller-owned identity variables authoritative. `sharedEnv` also contains `KAGENT_NAME`, `KAGENT_NAMESPACE`, and `KAGENT_URL`, so this merge lets `spec.deployment.env` replace them. `KAgentConfig.app_name` uses the first two as `agent_ref`; `HandleCreateSession` then requires that ref to match a registered Agent. For example, `KAGENT_NAME=custom` on `kagent/my-agent` produces `kagent__NS__custom` and session creation fails as an invalid agent ref. I reproduced that merge on this head with a focused Go test. Could overrides be limited to propagated `OTEL_*` values while preserving controller-owned values?

mesutoezdil:
closing. go/core/internal/controller/translator/agent is gone on main, the whole v1 translator moved to v2. the fix this covers is already there: compiler.go appends harness.Spec.Env after provider defaults and dedupeEnv keeps the later entry, so user env already wins.

PR #2328, opened 2026-07-25
fix(ui): stop stale session load from overwriting the active chat

mesutoezdil:
ChatInterface is not remounted when the session id prop changes, so the message-load effect for the old session keeps running after a new session is picked. If the old fetch resolves after the new one (plausible under normal network jitter), its response silently overwrites the new session's messages on screen, no error shown, wrong conversation visible.

Added a cancelled flag to the effect, same pattern already used by the neighboring effect in this file, so a stale load exits before touching state.

Added a test that reproduces the race with out of order responses, confirmed it fails against the old code and passes with the fix.


mesutoezdil:
closing. ChatInterface.tsx is gone on main, the whole ui was rewritten on vite with a new conversation state machine (#2569). the race this fixes was in a component that no longer exists.

PR #2329, opened 2026-07-25
fix(adk): keep full session state when num_recent_events is set

mesutoezdil:
get_session only fetched the last N events when num_recent_events was set, then built session.state by replaying just those N events. Any state_delta set by an older event outside that window never made it into state, silently, no error.

Now the full event history is always fetched to build state, and num_recent_events only trims the events list on the returned session, after state is already correct.

Added a test with a mock server that respects the limit query param, confirmed it fails against the old code and passes with the fix.


python/packages/kagent-adk/src/kagent/adk/_session_service.py
```diff
@@ -110,11 +102,8 @@ async def get_session(
             session_data = data["data"]["session"]
 
             events_data = data["data"]["events"]
-            if config and config.num_recent_events is not None:
-                if config.num_recent_events == 0:
-                    events_data = []
-                else:
-                    events_data.reverse()
+            if config and config.num_recent_events == 0:
+                events_data = []
```
supreme-gg-gg: For `num_recent_events == 0`, setting events to empty should also happen after we've reconstructed the session in the following loop (like num_recent_events > 0 case), so that the state is rebuilt before trimming events (which is the purpose of the PR and is consistent with how ADK's own implementations like in memory session store interprets this parameter).

supreme-gg-gg (commented):
Looks good overall, one minor comment and it's good to merge.

FYI `GetSessionConfig` is not used in Kagent ADK for now, but the problem of state reconstruction you identified is valid. In the long run we might want to serialize and store session state in the DB instead of always fetching all events and reconstructing it by applying the delta (which I think is what ADK's own db persistence class does). Otherwise, we will need to do the same for other filters like `config.after_timestamp`. On large sessions this would defeat the purpose of the provided "limit". 

However this is a good enough solution for now since there're some event persistence changes planned in v1 so we can leave that for later.

mesutoezdil:
Fixed

python/packages/kagent-adk/src/kagent/adk/_session_service.py
mesutoezdil: it has the same problem since it is still filtered server side. Filed #2353 to track it rather than widening this PR

supreme-gg-gg (approved):

python/packages/kagent-adk/src/kagent/adk/_session_service.py
supreme-gg-gg: Ok, thanks for filing the issue!

PR #2330, opened 2026-07-25
fix(adk): keep token refresh loop alive after a failed read

mesutoezdil:
_refresh_token only handled a token read returning None. Any other exception from the read (anything other than the OSError already handled inside read_token) escaped the while loop and killed the background refresh task for the rest of the process, no restart, no backoff. The token would never refresh again after that, easy to miss since the task is never awaited elsewhere.

Now a failed read is logged and the loop keeps going on its next 60s cycle instead of dying.

Added a test that fails the first read, then succeeds on the second, and checks the loop is still alive and the token still updates. Confirmed it fails against the old code (loop dies on the first exception) and passes with the fix.


python/packages/kagent-adk/src/kagent/adk/_token.py
```diff
@@ -57,10 +57,15 @@ async def _read_kagent_token(self) -> str | None:
...
-                    self.token = token
+            try:
+                token = await self._read_kagent_token()
+                if token is not None and token != self.token:
+                    async with self.update_lock:
+                        self.token = token
+            except Exception:
+                # A single failed read must not kill the loop, or the token
+                # would never refresh again for the life of the process.
+                logger.exception("Error refreshing kagent token, will retry next cycle")
```
EItanya, line 68: Does this log the exception/error? If it starts erroring the user may want to be aware of what the underlying issue is

mesutoezdil: yeas, logger.exception logs the full message and traceback, so the underlying error is visible in the logs on each failed cycle

EItanya (approved):

Issue #2277, opened 2026-07-17
Interrupted turn permanently corrupts session history for Anthropic models (dangling tool_use leads to 400 on every later turn)


QuentinBisson:
### 📋 Prerequisites

- [x] I have searched the [existing issues](https://github.com/kagent-dev/kagent/issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

When a streaming turn is interrupted after the agent has emitted a tool call but before the tool result is recorded (client disconnect, pod restart, the pre-#2113 3-minute A2A client timeout, OOM, ...), the session event log is persisted with a `function_call` event that has no matching `function_response`.

On every subsequent turn, the stored history is replayed verbatim to the model. The Anthropic API strictly requires each `tool_use` block to be followed by a `tool_result` block in the next message, so it rejects the entire conversation:

```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': 'messages.90: `tool_use` ids were found without `tool_result` blocks
immediately after: toolu_01...'}}
```

From that point the session is permanently unusable: every message fails with the same 400 until the session is deleted. Callers (e.g. an A2A client driving a chat surface) receive an opaque JSON-RPC error and have no way to repair it.

### Why this belongs in kagent

#2113 removed the most common interrupter (the hardcoded 3-minute a2a-go client timeout), but interruptions can never be fully prevented: pod restarts, OOM kills, network resets, and client-side cancellations all leave the same partial event log. The session store is kagent's and the history replay is kagent's, so the repair belongs here too. `AnthropicLlm` (google-adk, subclassed by kagent-adk) converts events to provider messages with no pairing validation. The same failure class is widely reported against other frameworks that replay history naively (vercel/ai#8516, warpdotdev/warp#7775, langchain4j#4900).

### Proposed fix

Repair at message-build time (source-agnostic, no migration needed): when converting session events to Anthropic messages, detect any `tool_use` whose id has no `tool_result` in the immediately following message and synthesize one, e.g.

```json
{"type": "tool_result", "tool_use_id": "...", "content": "[tool execution was interrupted; no result]", "is_error": true}
```

Optionally add the same guard for the mirrored case (an orphaned `tool_result`). This keeps the session usable and preserves its context instead of forcing a session delete.

[...]

PR #2331, opened 2026-07-25
fix: allow session id reuse after delete

### LINKED ISSUE #2279: POST /api/sessions with a soft-deleted id writes an invisible row, then 500s
### 📋 Prerequisites

### Describe the bug

`SessionService.CreateSession` with the id of a soft-deleted session writes a row that no read path can see, damages the tombstone on its way, and then fails with an internal error. The write is not rolled back, and every retry behaves the same.

- `SoftDeleteSession` sets `deleted_at` on the session row, which keeps its `(id, user_id)` primary key.
- `UpsertSession` upserts with `ON CONFLICT (id, user_id) DO UPDATE SET name, agent_id, source, updated_at`, so `deleted_at` stays set and the tombstone's `name`, `agent_id` and `source` are overwritten with the new request's values.
- Every read (`GetSession`, `ListSessions`, ...) filters `deleted_at IS NULL`.

So the create path's own reload cannot see the row it just wrote, and the RPC fails `codes.Internal` "Failed to load created session". The caller is told the request failed, while a write did happen, to the row an audit of the deleted session would read. A request should either create a visible session or change nothing.

### Proposed fix

Reject the create with `codes.AlreadyExists` when the `(id, user_id)` belongs to a soft-deleted session, and do not write. Excluding soft-deleted rows from the `ON CONFLICT` update is enough and keeps it atomic; a pre-check in the service leaves a window between the read and the write. The tombstone, its events and its tasks are left exactly as they are.

This is deliberately not a proposal to reuse the id: see #2331, where that was rejected, correctly. A distinct code also lets a caller tell "this id is retired, use another" from "kagent is unhealthy, retry", which an internal error does not.

### Steps to reproduce

1. `CreateSession` with an explicit `id`.
2. `DeleteSession` with that id.
3. `CreateSession` with the same id again: `codes.Internal` "Failed to load created session", with a row written that no read can see and a tombstone whose `name` is now the second request's. Every retry behaves the same.

Reproducible directly against the store: `StoreSession`, `DeleteSession`, `StoreSession` returns no error, the following `GetSession` errors, and `SELECT name FROM session` shows the second value.

### Observed on

kagent 0.9.9; code paths verified identical on v0.10.0-beta7 and reproduced against `main` at 30e056d1.

QuentinBisson:
Fixes #2279.

## What

`DeleteSession` soft-deletes, but `UpsertSession`'s `ON CONFLICT` update never cleared `deleted_at`, so recreating a previously deleted session id upserted an invisible row: the create handler's own reload failed with 500 "Failed to load created session", forever. This breaks any client that derives session IDs deterministically from an external key.

`UpsertSession` now resurrects a soft-deleted (id, user_id) as a fresh incarnation, in one statement:

- clears `deleted_at` and restarts `created_at`
- soft-deletes the previous incarnation's events and owned tasks, and drops its shares, so the recreated session starts empty instead of inheriting the old history

NULL-owned tasks are left alone: the `created_at` restart already hides them from the new incarnation via the ownership bound in tasks.sql, and they may still resolve to another user's same-id session. Upserts of live sessions are unchanged (no purge, `created_at` preserved).

## Tests

`TestSessionRecreateAfterDelete` (recreate succeeds; events/tasks/shares empty; another user's same-id session untouched) and `TestUpsertLiveSessionKeepsHistory` (no purge / no `created_at` reset on live upsert).

### DIFF
--- a/go/core/internal/database/client_test.go
+++ b/go/core/internal/database/client_test.go
@@ -240,6 +240,100 @@ func TestStoreSessionIdempotence(t *testing.T) {
 	require.Error(t, err, "another user's session must not be readable")
 }
 
+// TestSessionRecreateAfterDelete verifies that a session id can be reused
+// after DeleteSession: the upsert resurrects the soft-deleted row and the
+// recreated session starts empty (no events, tasks, or shares from the
+// previous incarnation).
+func TestSessionRecreateAfterDelete(t *testing.T) {
+	db := setupTestDB(t)
+	client := NewClient(db)
+	ctx := context.Background()
+
+	const sessionID = "s-reborn"
+	const userID = "alice"
+
+	require.NoError(t, client.StoreSession(ctx, &dbpkg.Session{ID: sessionID, UserID: userID}))
+	require.NoError(t, client.StoreEvents(ctx,
+		&dbpkg.Event{ID: "e-old-1", SessionID: sessionID, UserID: userID, Data: "{}"},
+		&dbpkg.Event{ID: "e-old-2", SessionID: sessionID, UserID: userID, Data: "{}"},
+	))
+	require.NoError(t, client.StoreTask(ctx, &a2a.Task{ID: "t-old", ContextID: sessionID}, userID))
+	_, err := client.CreateSessionShare(ctx, &dbpkg.SessionShare{Token: "share-old", SessionID: sessionID, UserID: userID})
+	require.NoError(t, err)
+
+	// An unrelated user's session under the same id must survive alice's
+	// delete/recreate cycle untouched.
+	require.NoError(t, client.StoreSession(ctx, &dbpkg.Session{ID: sessionID, UserID: "bob"}))
+	require.NoError(t, client.StoreEvents(ctx,
+		&dbpkg.Event{ID: "e-bob", SessionID: sessionID, UserID: "bob", Data: "{}"},
+	))
+
+	require.NoError(t, client.DeleteSession(ctx, sessionID, userID))
+	_, err = client.GetSession(ctx, sessionID, userID)
+	require.Error(t, err, "deleted session must not be readable")
+
+	require.NoError(t, client.StoreSession(ctx, &dbpkg.Session{ID: sessionID, UserID: userID}),
+		"recreating a deleted session id must succeed")
+
+	recreated, err := client.GetSession(ctx, sessionID, userID)
+	require.NoError(t, err, "recreated session must be readable")
+	require.Equal(t, sessionID, recreated.ID)
+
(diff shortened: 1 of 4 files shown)


iplay88keys:
Discussed a similar case  in [this thread](https://github.com/kagent-dev/kagent/pull/2265#discussion_r3625604512), related to resurrecting tasks.

QuentinBisson:
Thanks for the pointer, hadn't seen that thread. Having read it, I think sessions warrant the opposite choice from tasks.

Tasks have generated ids and an id-only PK, so permanently deleting an id costs nothing and closed a real cross-user takeover race. Sessions are the inverse: #2279 exists precisely because clients derive session ids deterministically from an external key (contextID), so an id that stays deleted forever would permanently brick recreate for that key. And the conflict target here is (id, user_id), so a resurrect only ever touches the caller's own row; the cross-user surprise from the tasks thread can't happen.

Given that, I see two reasonable designs:

1. **Resurrect with purge** (this PR): clear `deleted_at`, restart `created_at`, and soft-delete the old incarnation's events/tasks/shares in the same statement. The recreated session starts empty, so for the caller it's indistinguishable from a fresh session. Keeps soft-delete semantics for everything else.
2. **Hard delete sessions**: if nothing actually reads soft-deleted session rows back, the soft delete is pure liability and the fix collapses to a plain DELETE, recreate becomes a normal insert. Happy to switch to this if that's the case, it's the simpler change.

I went with 1 because it doesn't change delete semantics and I didn't want to assume the retention question. Are you okay with this solution?

iplay88keys:
In my mind, soft-deletion is an auditing tool to be able to see historical data -- even if that's not visible to most users or requires reading from the db directly. It seems that being able to re-use session ids defeats that purpose and the soft-deletion of previous events/tasks makes it difficult/impossible to know which events/tasks belonged to which version of a session if that process occurred multiple times.

Can you provide more info about your use case and why unique session ids doesn't work for you?

QuentinBisson:
Hey @iplay88keys, sorry for the delay I was on holidays. Thinking about this, you're right, and I'm closing this.

To answer the question directly: our client is a Slack gateway that keys the kagent session on the Slack thread id, so a session id is derived rather than generated. But that framing led me to the wrong fix. The reason we were recreating ids at all is that we delete a session to escape an interrupted turn (#2277): once a `function_call` is persisted without its `function_response`, every later turn on that session fails with a 400 from the model API, and deleting was the only recovery available to a client. Recreating the same id was working around that, not a genuine requirement.

Thanks for pushing back on this one, the pointer to the #2265 thread was the right place to start.

PR #2338, opened 2026-07-27
fix(acp-shim): drain child stdout during terminate to avoid teardown deadlock

tomsen02:
When a WebSocket client stalls (never reads), `pump()` blocks in `conn.WriteMessage` and stops draining the child's `out` channel. The stdout reader goroutine then blocks sending on the full channel (cap 64) and can never reach `cmd.Wait()`/`close(done)`. As a result `child.terminate()` — called from `Server.Shutdown` and the reconnect-grace teardown — hangs forever on `<-c.done` even after it has already SIGKILLed the process group, so the shim can never shut down.

Fix: `terminate()` now drains `c.out` while waiting for the child to exit, which keeps the reader goroutine moving. Dropping lines here is fine — the child is being torn down and nobody is listening.

Covered by two new tests: a unit test on `child.terminate` with an undrained `out` channel, and an end-to-end test through the real `Server` with a stalled WebSocket client. Both hang without this fix.


EItanya (changes requested):
nit: Why are the tests all in separate files?

tomsen02:
Thanks, good point. I consolidated the tests in 3acf812d:

- renamed `child_deadlock_test.go` to `child_test.go` for the child-level regression test
- moved the server-level regression test into the existing `server_test.go` and removed `server_deadlock_test.go`

Validation: `go test -race -count=1 -v ./pkg/acpshim`, `go test -count=3 ./pkg/acpshim`, and `make -C go lint` (`0 issues`).

@EItanya, could you take another look?

EItanya (approved):

PR #2346, opened 2026-07-27
honor event timestamp filters when loading sessions

peterj:
Fix incremental session loading in `KAgentSessionService` by forwarding `GetSessionConfig.after_timestamp` to the sessions API.

Previously, the timestamp was silently ignored, causing session loads to return events that had already been consumed. This could replay events and duplicate state transitions when resuming an ADK run.


EItanya (dismissed):
Does this also work in go?

peterj:
need to  add to go too

EItanya (approved):

PR #2348, opened 2026-07-28
fix: reject nodeSelector on sandbox agents instead of silently dropping it

### LINKED ISSUE #2306: [BUG] Sandbox agents silently drop configured nodeSelector (not applied to ActorTemplate)
### Prerequisites
- [x] I have searched existing issues to avoid creating a duplicate
- [x] I agree to follow the Code of Conduct
- [x] I am using the latest version of the software (verified against `main`)

### Affected Service(s)
Controller Service

### Impact/Severity
Minor inconvenience

### Bug Description

For declarative agents that run in a sandbox, a configured node selector is silently dropped. The translator resolves `dep.NodeSelector` and places it in the shared `PodTemplateSpec` (`manifest_builder.go`), and normal Deployments honor it — but the substrate sandbox backend builds its `ActorTemplateSpec` only from the kagent container and worker pool and never reads `podTemplate.Spec.NodeSelector`. So a sandbox workload can be scheduled onto an unintended node, violating placement requirements, with no error to signal the config was ignored.

This is a pre-existing gap in the sandbox backend (it predates and is independent of the `--default-agent-node-selector` work), but it becomes more visible now that a node selector can be configured globally.

`NodeSelector` does not appear anywhere in `go/core/pkg/sandboxbackend/` (confirmed by a repo-wide code search), which is the root of the drop.

### Steps To Reproduce
1. Configure a node selector (globally or per-agent).
2. Create one normal declarative Agent and one sandbox Agent with the same selector.
3. Reconcile both so the translator resolves their deployment selectors.
4. Inspect the normal Deployment's PodSpec (selector present) vs. the sandbox `ActorTemplate` (selector absent).

### Expected Behavior
A configured `nodeSelector` is applied to sandbox workloads too, or the controller rejects an unsupported sandbox scheduling config instead of silently discarding it.

### Actual Behavior
The sandbox `ActorTemplateSpec` is built without `NodeSelector`; the setting is silently dropped and the sandbox pod schedules anywhere the scheduler chooses.

### Additional Context
Relevant code:
- Selector set into the shared template: https://github.com/kagent-dev/kagent/blob/main/go/core/internal/controller/translator/agent/manifest_builder.go#L558
- Template passed to `BuildSandbox`: https://github.com/kagent-dev/kagent/blob/main/go/core/internal/controller/translator/agent/manifest_builder.go#L572-L576
- Sandbox `ActorTemplateSpec` built without reading `NodeSelector`: https://github.com/kagent-dev/kagent/blob/main/go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go#L79-L120

Suggested fix: map the pod template's `NodeSelector` into the sandbox workload type/backend, or explicitly reject unsupported sandbox scheduling configuration rather than silently dropping it.

### Are you willing to contribute?

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/c48021a6-c217-4c72-bd4c-f426ab98829e.

younsl:
## Summary

Fixes #2306

For sandbox agents, a configured `deployment.nodeSelector` was silently dropped: the translator resolves it into the shared `PodTemplateSpec`, but the substrate backend builds its `ActorTemplateSpec` only from the kagent container and worker pool and never reads `podTemplate.Spec.NodeSelector`. The sandbox workload could therefore be scheduled onto an unintended node with no signal that the config was ignored.

`ActorTemplateSpec` has no node-placement field at all — substrate schedules actors onto **WorkerPool** workers, and node placement is governed by the WorkerPool's own `nodeSelector`. So a per-agent selector cannot be honored by mapping it through; the issue's second suggested option (explicit rejection) is implemented instead:

- `ValidateSubstrateSandboxAgentSpec` now rejects a non-empty `deployment.nodeSelector` (declarative and BYO) with an error pointing at the WorkerPool's `nodeSelector` as the supported alternative.
- Since this validation runs in the HTTP handler, the reconciler, and the compiler, the misconfiguration surfaces at creation time and in the agent's status conditions instead of being silently discarded.

### Notes for reviewers

- The global `--default-agent-node-selector` flag is intentionally **not** rejected here: it is not part of the agent spec, and hard-failing every sandbox agent in clusters that set a global default would be a breaking regression. The sandbox backend simply does not consume it (placement is the WorkerPool's job).
- `tolerations`/`affinity` are also silently ignored by the sandbox backend today; left out of scope to keep this PR focused on the reported bug — can follow up if maintainers want the same treatment.

## Testing

- Added table-driven cases to `TestValidateSubstrateSandboxAgentSpec`: declarative selector rejected, BYO selector rejected, empty selector map still allowed.
- `go test ./...` in `go/api`, plus `go/core` controller/translator/sandboxbackend packages — pass.
- `make -C go lint` — 0 issues.

### DIFF
--- a/go/api/config/crd/bases/kagent.dev_sandboxagents.yaml
+++ b/go/api/config/crd/bases/kagent.dev_sandboxagents.yaml
@@ -11369,6 +11369,17 @@ spec:
             x-kubernetes-validations:
             - message: spec.skills is not supported for sandbox agents
               rule: '!has(self.skills)'
+            - message: 'deployment.nodeSelector is not supported for sandbox agents:
+                substrate schedules actors onto WorkerPool workers, so set the WorkerPool''s
+                nodeSelector instead'
+              rule: '!has(self.declarative) || !has(self.declarative.deployment) ||
+                !has(self.declarative.deployment.nodeSelector) || size(self.declarative.deployment.nodeSelector)
+                == 0'
+            - message: 'deployment.nodeSelector is not supported for sandbox agents:
+                substrate schedules actors onto WorkerPool workers, so set the WorkerPool''s
+                nodeSelector instead'
+              rule: '!has(self.byo) || !has(self.byo.deployment) || !has(self.byo.deployment.nodeSelector)
+                || size(self.byo.deployment.nodeSelector) == 0'
             - message: type must be specified
               rule: has(self.type)
             - message: type must be either Declarative or BYO
--- a/go/api/v1alpha2/agent_spec_validation.go
+++ b/go/api/v1alpha2/agent_spec_validation.go
@@ -26,7 +26,9 @@ func AgentSpecHasSkills(spec *AgentSpec) bool {
 // command because substrate copies the container Command verbatim with no image-entrypoint
 // fallback. A per-agent deployment.nodeSelector is rejected: substrate ActorTemplates carry
 // no node placement (actors run on WorkerPool workers), so the selector would otherwise be
-// silently dropped.
+// silently dropped. The skills and nodeSelector checks are also enforced at admission by CEL
+// rules on SandboxAgentSpec; this function keeps them effective for objects created before
+// those rules shipped and for callers that bypass the API server.
 func ValidateSubstrateSandboxAgentSpec(agent *SandboxAgent) error {
 	if agent == nil {
 		return nil
--- a/go/api/v1alpha2/agent_spec_validation.go
+++ b/go/api/v1alpha2/agent_spec_validation.go
@@ -6,8 +6,9 @@ import (
 )
 
 const (
-	substrateSandboxSkillsUnsupportedMsg = "spec.skills is not supported for sandbox agents"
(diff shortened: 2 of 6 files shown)


EItanya (commented):
What about using CEL to reject the whole deployment spec on `SandboxAgent`

younsl:
@EItanya Agreed, CEL is the right place for this. Done in cfa89067.

I scoped the CEL rules to deployment.nodeSelector instead of rejecting the whole deployment spec. BYO agents require deployment.image and deployment.cmd, and substrate honors deployment.env when rendering ActorTemplates, so rejecting the entire block would break supported configuration.

The Go-side check is kept alongside the CEL rules, following the existing spec.skills pattern, so objects created before the rules shipped are still caught at reconcile time. Also added an envtest backed test that pins the rules against the shipped CRD YAML.

younsl:
@EItanya CI is clean now. PTAL

younsl:
Not stale

PR #2369, opened 2026-07-30
feat: support responses api in kagent harness

supreme-gg-gg:
Users can now use the Responses API either directly with OpenAI or via a gateway like AgentGateway. Here's an example with agent gateway.

```yaml
kind: ModelConfig
metadata:
  name: openai-responses
  namespace: kagent
spec:
  provider: OpenAI
  model: gpt-5.4-mini
  apiKeyPassthrough: true
  openAI:
    baseUrl: http://agentgateway-proxy.agentgateway-system.svc/openai
    apiFormat: responses # NEW! <- If not provided, uses chat completions API
```

Note that certain features from Responses API like native tools and stateful chaining are not supported for now since they're only available when using with OpenAI's servers. For those interested, it can be a follow up to add those features in kagent harness.

### DIFF
--- a/go/adk/pkg/models/openai.go
+++ b/go/adk/pkg/models/openai.go
@@ -16,6 +16,12 @@ import (
 	"github.com/openai/openai-go/v3/option"
 )
 
+// OpenAI API format values (ModelConfig openAI.apiFormat).
+const (
+	OpenAIAPIFormatChatCompletions = "chatCompletions"
+	OpenAIAPIFormatResponses       = "responses"
+)
+
 // OpenAIConfig holds OpenAI configuration
 type OpenAIConfig struct {
 	TransportConfig
@@ -30,6 +36,8 @@ type OpenAIConfig struct {
 	Seed                *int
 	Temperature         *float64
 	TopP                *float64
+	// APIFormat selects chatCompletions (default) or responses.
+	APIFormat string
 }
 
 // AzureOpenAIConfig holds Azure OpenAI configuration
--- a/go/adk/pkg/models/openai_adk.go
+++ b/go/adk/pkg/models/openai_adk.go
@@ -128,7 +128,6 @@ func (m *OpenAIModel) Name() string {
 // GenerateContent implements model.LLM. Uses only ADK/genai types.
 func (m *OpenAIModel) GenerateContent(ctx context.Context, req *model.LLMRequest, stream bool) iter.Seq2[*model.LLMResponse, error] {
 	return func(yield func(*model.LLMResponse, error) bool) {
-		messages, systemInstruction := genaiContentsToOpenAIMessages(req.Contents, req.Config)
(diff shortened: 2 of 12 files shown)


go/adk/pkg/models/openai.go
```diff
@@ -16,6 +16,12 @@ import (
 	"github.com/openai/openai-go/v3/option"
 )
 
+// OpenAI API format values (ModelConfig openAI.apiFormat).
+const (
+	OpenAIAPIFormatChatCompletions = "chatCompletions"
```
EItanya, line 21: Can we just call this `completions`?

go/api/v1alpha2/modelconfig_types.go
```diff
@@ -199,12 +199,29 @@ type OpenAIConfig struct {
...
 	// +optional
 	TokenExchange *TokenExchangeConfig `json:"tokenExchange,omitempty"`
 }
 
+// OpenAIAPIFormat selects the OpenAI HTTP API shape used by the Go ADK runtime.
+// +kubebuilder:validation:Enum=chatCompletions;responses
+type OpenAIAPIFormat string
+
+const (
+	OpenAIAPIFormatChatCompletions OpenAIAPIFormat = "chatCompletions"
```
EItanya, line 221: Same comment

go/adk/pkg/models/openai_adk.go
```diff
@@ -128,7 +128,6 @@ func (m *OpenAIModel) Name() string {
 // GenerateContent implements model.LLM. Uses only ADK/genai types.
 func (m *OpenAIModel) GenerateContent(ctx context.Context, req *model.LLMRequest, stream bool) iter.Seq2[*model.LLMResponse, error] {
```
EItanya, line 136: Should these be separate structs since they don't really share much?

go/adk/pkg/models/openai.go
supreme-gg-gg: Unfortunately OpenAI has a [deprecated API called "completions"](https://developers.openai.com/api/docs/guides/completions) which is the legacy version of "chat completions", so it might be a bit confusing

EItanya: Ok cool, I didn’t know that

go/adk/pkg/models/openai_adk.go
supreme-gg-gg: the `OpenAIModel` struct contains shared config, client, logging setup (in `openai.go`), and this structure represents a "model provider" so imo there should only be one of it, and it would handle the response formats internally. I've made a change to make the api format handling cleaner in https://github.com/kagent-dev/kagent/pull/2369/changes/b36dd6f12ff63da16c0815e36ba5e8eaafdbaba5

EItanya (approved):

PR #2372, opened 2026-07-31
fix(adk): preserve empty MCP arrays in task history

erauner12:
## Summary

Preserve nested non-nil empty JSON arrays in MCP structured results as `[]` when they pass through A2A task-history storage.

## Root cause

The task status update path deep-copies stored history with gob. Interface-held zero-length JSON arrays become typed nil slices during that copy, so A2A JSON changes `[]` to `null`.

## Fix

Protect empty JSON arrays at the shared A2A executor seam after MCP conversion and before task-history persistence. Explicit nulls and non-empty arrays remain unchanged.

## Validation

- Focused ADK A2A, MCP, task-store, and compatibility packages pass.
- Regression tests cover nested empty arrays, non-empty arrays, repeated gob copies, and JSON round trips.
- An isolated Kind/Garden run confirmed raw MCP `[]` values remain arrays in untouched A2A `tasks/get` history.

This targets the `release/v0.9.x` line and adds no dependency or schema change.

### DIFF
--- a/go/adk/pkg/a2a/converter.go
+++ b/go/adk/pkg/a2a/converter.go
@@ -2,6 +2,7 @@ package a2a
 
 import (
 	"context"
+	"encoding/gob"
 	"encoding/json"
 	"maps"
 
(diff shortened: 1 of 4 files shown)


supreme-gg-gg (commented):
Is this a bug specific to v0.9.x? Otherwise, why is it not opened against main or v0.10.x, since 0.9.x will be legacy soon as we release 0.10.x

PR #2385, opened 2026-08-03
A2A: Various usage cleanups in Kagent harness

supreme-gg-gg:
This PR changes A2A output to be now artifact-based rather than assistant-message/history-based. Consumers should render `Task.Artifacts`; transient progress remains in task status updates.

## Motivation

This refactor aligns kagent’s A2A usage with the protocol’s separation between task state and task output, as A2A [protocol specification](https://a2a-protocol.org/latest/specification/#37-messages-and-artifacts) noted:

> Messages SHOULD NOT be used to deliver task outputs. Results SHOULD BE returned using Artifacts associated with a Task. This separation allows for a clear distinction between communication (Messages) and data output (Artifacts).

Previously, streamed assistant text and tool data were carried in `WORKING` status messages, with an artifact emitted only at the end of the task. That made status updates serve two roles and required clients to reconstruct the response from transient state. Task output is now emitted as `TaskArtifactUpdateEvents` throughout execution. Status updates communicate only task lifecycle, HITL, failures, etc. A terminal status closes the task, while `lastChunk` closes an individual artifact.

**This is a breaking wire-level change for consumers that read assistant output from status messages or task history.** Consumers should instead render `Task.Artifacts` and process artifact updates as they arrive.

## Change surfaces

ADK/BYO:

- Go ADK: replaced the custom executor flow with the upstream ADK A2A executor, configured for `OutputArtifactPerEvent`, so ADK events become A2A artifact updates. Subsequently, cleanup all kagent specific partial event handling logic since upstream supports it now. Partial events and emitted as artifacts with `append=true`.
- Python ADK: Similar refactor as Go. Python still uses custom executor, but in a follow up PR we will bump python-adk to 2.0 which allows upstreaming executor
- Applies the same A2A change to BYO harness and improves them (CrewAI, OpenAI Agents SDK) by making their converters / event handlers return appropriate tool calling / agent delegation data parts

UI/CLI:

- UI: use task status updates for working/progress and HITL instead of canonical assistant response, which is now moved to artifact part updates, and adds relevant buffering and flushing mechanisms for handling streaming cases (similarly for CLI, except that the Go TUI does not support streaming messages)
- UI cleanup: subagent session IDs are now taken from function-response data, cleanup following #2153


python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
supreme-gg-gg, line 1: Python executor should also be upstreamed to Python-ADK once we're on 2.0 (go is already on 2.0), but that should be a follow up PR

EItanya (approved):

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
supreme-gg-gg: This has been done in #2396 (next in the stack 😄)

EItanya (approved):

peterj (approved):

PR #2396, opened 2026-08-05
A2A: Standardize Kagent's human in the loop with A2A Extension

### LINKED ISSUE #2304: Declare the HITL confirmation payload as an A2A AgentExtension with one stable prefix
The human-in-the-loop confirmation flow (paused long-running tool call -> input-required status -> decision reply) uses part metadata and DataPart schemas that are effectively private ADK conventions: `kagent_type`/`adk_type` = `function_call`, `kagent_is_long_running`/`adk_is_long_running`, an `adk_request_confirmation` DataPart with `args.originalFunctionCall` / `toolConfirmation.hint` / `ask_user` questions, and a decision DataPart (`decision_type`, `ask_user_answers`, `rejection_reason`).

Both the `kagent_` and `adk_` prefixes appear in the wild, so consumers sniff both, and the payload schema is undocumented, so every consumer reverse-engineers it and can break on any refactor.

Ask:
1. Pick one prefix and treat the payload schema as a stable contract.
2. Declare it as an A2A `AgentExtension` in the AgentCard (the spec mechanism for exactly this), with the schema documented, so consumers can feature-detect instead of sniffing.

The A2A mechanics themselves (input-required + status message parts) are fine; this is only about the payload contract.

supreme-gg-gg:
Close #2304 

- Move Kagent's human in the loop (HITL) to an A2A Extension that is framework-neutral and standardized. Previously, we relied on ADK-specific parts from its request confirmation flow and a lot of custom metadata.
- Redesigned the HITL payload following the Extension migration since it's framework neutral now. This redesign means UI and extension becomes simpler, and defers more processing logic to the upstream executor and each framework
- Upgrade Python ADK to 2.0 so that we can reuse upstream A2A agent executor to handle request confirmation flow internally and remove duplicated code introduced in the last HITL PR when upstream does not support it back then
- Added e2e test for HITL (ask user tool)

Core changes are documented in `docs/architecture/human-in-the-loop.md`, see the diff for the design changes.

**This is breaking so old clients without the A2A extension negotiation capability will stop receiving HITL events properly. To upgrade your custom client / agents, please refer to changes in the above doc.**

As before, only ADK and Langgraph (BYO) supports HITL. However, this redesign enables users to easily develop their own adapter for other BYO frameworks / harness since HITL over A2A itself is framework-neutral.


ui/src/components/create/ToolsSection.tsx
supreme-gg-gg, line 1: Note for reviewer: Also added UI option to set and show toggle for isolating subagent sessions added in #2153 (`SelectToolsDialog.tsx, ToolsSection.tsx, AgentDetailsSidebar.tsx`)

Image

python/packages/kagent-adk/src/kagent/adk/converters/event_converter.py
```diff
@@ -155,36 +155,10 @@ def _process_long_running_tool(a2a_part: A2APart, event: Event) -> None:
         a2a_part.metadata.update({get_kagent_metadata_key(A2A_DATA_PART_METADATA_IS_LONG_RUNNING_KEY): True})
 
 
-def _process_subagent_session_id(a2a_part: A2APart, subagent_session_ids: Dict[str, str]) -> None:
```
supreme-gg-gg, line 158: Note for reviewer: cleaning up these unnecessary event converters and utils for stamping subagent session ID as #2153 removed it for Go

python/packages/kagent-adk/src/kagent/adk/sandbox_code_executer.py
supreme-gg-gg, line 1: Note for reviewer: removed sandbox code executor no longer in use

EItanya (dismissed):

peterj (approved):

EItanya (approved):

PR #2399, opened 2026-08-06
Backport of 2407 (pg connection) to 0.10.x release

supreme-gg-gg:
This targets 0.10.x release. 
Verified with `pg_stat_activity` that connection is closed after configured idle timeout + pgx check period

### DIFF
--- a/go/core/pkg/app/app.go
+++ b/go/core/pkg/app/app.go
@@ -136,11 +136,15 @@ type Config struct {
 	// http://host:<port-or-443> so traffic egresses in plaintext to a proxy
 	// that originates TLS upstream. Off by default;
 	MCPEgressPlaintext bool
-	Database           struct {
-		Url            string
-		UrlFile        string
-		VectorEnabled  bool
-		SkipMigrations bool
+	Database struct {
+		Url             string
+		UrlFile         string
+		VectorEnabled   bool
+		SkipMigrations  bool
+		MaxConns        int           // 0 = unset (pgx default)
+		MinConns        int           // -1 = unset (pgx default); 0 is a valid value
+		MaxConnIdleTime time.Duration // 0 = unset (pgx default)
+		MaxConnLifetime  time.Duration // 0 = unset (pgx default)
 	}
 	Substrate struct {
 		AteAPIEndpoint             string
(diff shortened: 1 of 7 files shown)


go/core/pkg/app/app.go
```diff
@@ -230,6 +238,32 @@ func (cfg *Config) SetFlags(commandLine *flag.FlagSet) {
...
 
+// postgresConfigFromApp builds a database.PostgresConfig from app flags.
+// Zero/unset flag values leave the corresponding pool field nil so pgx defaults apply.
+func postgresConfigFromApp(dbURL string, cfg *Config) *database.PostgresConfig {
+	pgCfg := &database.PostgresConfig{
+		URL:           dbURL,
+		VectorEnabled: cfg.Database.VectorEnabled,
+	}
+	if cfg.Database.MaxConns > 0 {
+		v := int32(cfg.Database.MaxConns)
```
iplay88keys, line 249: For each of these, we should probably have an upper bound as well. At least since an int converted to an int32 can truncate. Probably not an issue in practice, though, as that is an incredibly large number.

go/core/pkg/app/app.go
```diff
@@ -230,6 +238,32 @@ func (cfg *Config) SetFlags(commandLine *flag.FlagSet) {
 	commandLine.StringVar(&agent_translator.DefaultAgentBindHost, "default-agent-bind-host", agent_translator.DefaultAgentBindHost, "Default host address for agent pods to bind to. Use '0.0.0.0' for IPv4 only or '::' for dual-stack (IPv4+IPv6).")
 }
 
+// postgresConfigFromApp builds a database.PostgresConfig from app flags.
+// Zero/unset flag values leave the corresponding pool field nil so pgx defaults apply.
+func postgresConfigFromApp(dbURL string, cfg *Config) *database.PostgresConfig {
```
iplay88keys, line 243: We should definitely validate that `minConns` is greater than `maxConns` if `minConns` is set.

go/core/pkg/app/app.go
supreme-gg-gg: I don't think that would be a realistic config though

go/core/pkg/app/app.go
supreme-gg-gg: fair point, added the check

go/core/pkg/app/app.go
iplay88keys: Agreed, just calling it out for completeness.

iplay88keys (approved):
Tested locally, looks good to me!

PR #2400, opened 2026-08-07
fix(otel): stop sending traces to api.openai.com and cleanup

krisztianfekete:
The Agents SDK registers a trace processor that POSTs everything to a hardcoded `https://api.openai.com/v1/traces/ingest` using `OPENAI_API_KEY`. It ignores `OPENAI_API_BASE`, so anyone behind other proxies is shipping their gateway's key to OpenAI and getting 401s back.

`OpenAIAgentsInstrumentor` adds the OTel processor next to that one rather than replacing it, and `set_tracing_disabled(True)` would silence the OTel spans too, so we now drop the built-in processor before instrumenting. Set `KAGENT_OPENAI_AGENTS_NATIVE_TRACING=true` to keep it if you're using a real OpenAI key.

Using `OpenAIAgentsInstrumentor(replace_existing_processors=True)` as suggested in review, with the pin bumped to `>=0.52.3,<0.53.0` since that's when the kwarg landed. Doing it through the kwarg also means a second `build()` can't wipe the OTel processor, since `instrument()` no-ops once instrumented.

Cleaned up a couple of other smaller stuff:
  - `Resource(...)` -> `Resource.create(...)`: the bare constructor ignores `OTEL_RESOURCE_ATTRIBUTES` and drops `telemetry.sdk.*`, so nobody could set `deployment.environment.name` or `service.version` at all. 
  - Same bug in the Go ADK, `resource.New` starts empty so it needs `WithFromEnv()` + `WithTelemetrySDK()`.
  - `HTTPXClientInstrumentor().instrument(excluded_urls=...)` did nothing, cleaned it up.
  - `_a2a.py` read `self.config.kagent_url` but `KAgentConfig` only has `url`, which blows up whenever `KAGENT_URL` isn't set.

Rebased onto main, which dropped the Gemini bullet from this PR since main landed the same fix independently.

### DIFF
--- a/python/packages/kagent-core/src/kagent/core/tracing/_utils.py
+++ b/python/packages/kagent-core/src/kagent/core/tracing/_utils.py
@@ -236,7 +236,10 @@ def configure(
     tracing_enabled = os.getenv("OTEL_TRACING_ENABLED", "false").lower() == "true"
     logging_enabled = os.getenv("OTEL_LOGGING_ENABLED", "false").lower() == "true"
 
-    resource = Resource({"service.name": name, "service.namespace": namespace})
+    # Resource.create merges in OTEL_RESOURCE_ATTRIBUTES and the telemetry.sdk.*
+    # attributes; the bare constructor drops both, so deployment.environment.name,
+    # service.version and friends never reach the backend.
+    resource = Resource.create({"service.name": name, "service.namespace": namespace})
 
     # Configure tracing if enabled
     if tracing_enabled:
@@ -273,9 +276,11 @@ def configure(
 
         # Exclude agent-card endpoint from traces — this is used as a health
         # check endpoint (high-frequency polling requests) and has little
-        # diagnostic value.
+        # diagnostic value. Inbound only: HTTPXClientInstrumentor accepts no
+        # excluded_urls kwarg (newer releases read OTEL_PYTHON_HTTPX_EXCLUDED_URLS
+        # instead), so passing one here was silently dropped.
         _excluded_urls = ".*/\\.well-known/agent-card\\.json"
-        HTTPXClientInstrumentor().instrument(excluded_urls=_excluded_urls)
+        HTTPXClientInstrumentor().instrument()
         if fastapi_app:
             FastAPIInstrumentor().instrument_app(fastapi_app, excluded_urls=_excluded_urls)
             # Pre-response flushing is opt-in (the controller sets this on Agent
--- a/python/packages/kagent-openai/src/kagent/openai/_a2a.py
+++ b/python/packages/kagent-openai/src/kagent/openai/_a2a.py
@@ -16,7 +16,13 @@ from a2a.server.request_handlers import DefaultRequestHandlerV2
 from a2a.server.routes import add_a2a_routes_to_fastapi, create_agent_card_routes, create_jsonrpc_routes
 from a2a.server.tasks import InMemoryTaskStore
 from a2a.types import AgentCard
-from agents import Agent, set_default_openai_api, set_default_openai_client, set_tracing_disabled
+from agents import (
+    Agent,
+    set_default_openai_api,
+    set_default_openai_client,
+    set_trace_processors,
+    set_tracing_disabled,
+)
 from fastapi import FastAPI, Request
 from fastapi.responses import PlainTextResponse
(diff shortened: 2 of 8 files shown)


python/packages/kagent-openai/src/kagent/openai/_a2a.py
```diff
@@ -78,6 +84,35 @@ def _configure_openai_client() -> None:
...
+    would silence the OpenTelemetry spans too, so drop the built-in processor instead.
+
+    KAGENT_OPENAI_AGENTS_NATIVE_TRACING=true keeps it, for real OpenAI platform keys.
+    """
+    if os.getenv("KAGENT_OPENAI_AGENTS_NATIVE_TRACING", "false").strip().lower() == "true":
+        logger.info("Keeping the OpenAI Agents SDK native trace exporter alongside OpenTelemetry")
+    else:
+        # The dropped BatchTraceProcessor starts its worker thread only on first
+        # enqueue, so nothing is ever sent to api.openai.com.
+        set_trace_processors([])
```
dhaifley: Small testing issue:

`BaseInstrumentor.instrument()` returns early with a warning when already instrumented (opentelemetry-instrumentation 0.59b0, `instrumentor.py:102-104`). So on a second call `set_trace_processors([])` clears the OpenTelemetry processor that the first call installed, and nothing puts it back. All agent tracing goes dark silently, with only an "Attempting to instrument while already instrumented" line in the log.

Not reachable through `build()` today since production builds one app per process, but it is a possible trap for tests, `uvicorn --factory` reload, and anything that constructs two KAgentApps. Maybe either use item 2, or add an explicit guard at the top of the helper like:
```python
instrumentor = OpenAIAgentsInstrumentor()
if instrumentor.is_instrumented_by_opentelemetry:
    return
```

python/packages/kagent-openai/src/kagent/openai/_a2a.py
```diff
@@ -173,7 +208,7 @@ def build(self) -> FastAPI:
                 tracing_enabled = os.getenv("OTEL_TRACING_ENABLED", "false").lower() == "true"
                 if tracing_enabled:
                     logger.info("Enabling OpenAI Agents SDK tracing")
-                    OpenAIAgentsInstrumentor().instrument()
+                    _configure_openai_agents_tracing()
```
dhaifley, line 204: If configure_tracing() on L205 throws an exception, this never runs and think the native OpenAI processor is left in place. May not be an actual issue though.

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
```diff
@@ -315,6 +320,9 @@ def configure(
         if instrument_openai_client:
             OpenAIInstrumentor(use_legacy_attributes=False).instrument(event_logger_provider=event_logger_provider)
         _instrument_anthropic(event_logger_provider)
+        # No event-logger variant upstream, but it still has to run here or the
+        # Gemini SDK is left uninstrumented whenever OTEL_LOGGING_ENABLED is true.
+        _instrument_google_generativeai()
```
dhaifley: Wiring this into the logging branch is the right fix, but _instrument_google_generativeai() uses the default use_legacy_attributes=True, so Gemini prompt and completion text goes into span attributes here while OpenAI and Anthropic would emit log events.

The event-logger variant does exist upstream, it is just named logger_provider= instead of event_logger_provider=.

Maybe using as GoogleGenerativeAiInstrumentor(use_legacy_attributes=False).instrument(logger_provider=logger_provider) so OTEL_LOGGING_ENABLED keeps prompt content off spans for all three providers, would work?

dhaifley (approved):
Looks good, but what if we just used the newer version mentioned in the PR description?  `>=0.52.3,<0.53.0`? I'm not familiar enough to know if there is a reason we cannot do that.

Also, does the Go ADK code need a similar fix: `go/adk/pkg/telemetry/tracing.go:81-84`?

krisztianfekete:
Thanks @dhaifley, addressed your comments and fixed a new bug with `KAgentConfig`. Didn't want to upgrade originally, as hopefully shortly we're moving away from the Traceloop libraries, but couldn't find dangerous changes, so went with it this time.

dhaifley (approved):

EItanya (approved):

PR #2401, opened 2026-08-07
feat(helm): add a ServiceMonitor for the controller metrics endpoint

### LINKED ISSUE #1369: [FEATURE] Helm Chart: Add support for enabling metrics endpoint and optional ServiceMonitor
### 📋 Prerequisites

### 📝 Feature Summary

Add Helm values to easily enable the controller metrics endpoint and optionally deploy a Prometheus Operator `ServiceMonitor`.

The kagent controller already supports exposing a metrics endpoint via the flags:

- `--metrics-bind-address`
- `--metrics-secure`

These can also be configured via environment variables, for example:

```bash
export METRICS_BIND_ADDRESS=":9093"
export METRICS_SECURE="false"
```

However, the current Helm chart does not provide a simple, documented way to enable and expose this metrics endpoint.

Even if we manually inject environment variables through values like:

```
controller:
  env:
    - name: "METRICS_BIND_ADDRESS"
      value: ":9093"
    - name: "METRICS_SECURE"
      value: "false"
```

This is still incomplete because:

1. The Helm chart does not automatically expose the metrics port in the controller Service.
1. Users must manually create or modify a Kubernetes Service to reference the new port.
1. Users must manually create a ServiceMonitor (when using Prometheus Operator) to scrape the endpoint.
1. There is no single, cohesive way to “enable metrics” through values.yaml.

This makes observability setup more complex than necessary.

### 💡 Proposed Solution

Introduce Helm values for metrics configuration and optional Prometheus Operator integration.

For example:

```yaml
metrics:
  enabled: true
  bindAddress: ":9093"
  secure: false

  service:
    enabled: true
    port: 9093
    type: ClusterIP

  serviceMonitor:
    enabled: true
    interval: 60s
    scrapeTimeout: 10s
    labels: {}
```

### 🔄 Alternatives Considered

_No response_

### 🎯 Affected Service(s)

Controller Service

### 📚 Additional Context

Many Kubernetes projects provide a built-in metrics and serviceMonitor configuration block in their Helm charts. Adding similar support to kagent would:

1. Improve user experience
2. Reduce configuration drift
3. Encourage consistent observability setups
4. Make kagent easier to integrate into Prometheus-based monitoring stacks

younsl:
## Description

This adds native [ServiceMonitor](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ServiceMonitor) support to the kagent Helm chart so controller metrics are collected by [Prometheus Operator](https://prometheus-operator.dev/) out of the box.

This is the follow up that [#1803](https://github.com/kagent-dev/kagent/pull/1803) deferred. That PR landed the metrics Service and the scrape RBAC and closed [#1369](https://github.com/kagent-dev/kagent/issues/1369), but left the ServiceMonitor out of scope. [#1802](https://github.com/kagent-dev/kagent/pull/1802) carried one and was closed as a duplicate of #1803, so the ServiceMonitor half of #1369 was never reviewed on its own and no template for it exists in the chart today. This PR is additive on top of what #1803 merged and changes none of it.

Today controller.metrics.enabled provisions the metrics Service and the ClusterRoles for authenticated scrapes, but not the resource that makes Prometheus actually scrape them. Every user running [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) has to hand write a ServiceMonitor outside the chart and keep its port name and namespace in sync with the chart by hand. The bundled [kagent-tools](https://github.com/kagent-dev/tools) subchart already ships one, so this closes the gap for the controller.

New values live under controller.metrics.serviceMonitor, disabled by default and gated on controller.metrics.enabled. Interval, scrape timeout, honorLabels, relabelings, metric relabelings, extra labels, annotations and a namespace override are exposed. With secure serving on, the endpoint defaults to https with the ServiceAccount token and insecureSkipVerify, which matches the self signed certificate the controller serves, and both are overridable.

The metrics Service port name moved into a shared helper used by both the Service and the ServiceMonitor endpoint, so the two cannot drift when secure serving is flipped. Rendered output for existing installs is unchanged.

## Changelog

Added an optional Prometheus Operator ServiceMonitor for the controller metrics endpoint, configurable under controller.metrics.serviceMonitor.

## Testing

1. helm unittest helm/kagent passes with 298 tests, 15 of them new in helm/kagent/tests/controller-servicemonitor_test.yaml.
2. Nothing renders unless both controller.metrics.enabled and controller.metrics.serviceMonitor.enabled are set, and the bindAddress disable sentinel still suppresses it.
3. Rendered the secure and plaintext variants with helm template, then validated them against a live cluster with kubectl apply --dry-run=server so the manifest is checked against the real ServiceMonitor CRD rather than only the chart.

## Additional Notes

Rendering the ServiceMonitor does not authorize the scrape on its own. With secure serving, the metrics reader ClusterRole still has to be bound to the Prometheus ServiceAccount, which the values comment calls out.

The chart does not gate on the monitoring.coreos.com API being present, matching what kagent-tools does. The explicit opt in flag is the gate, so a missing CRD fails loudly at install time instead of silently dropping the resource during helm template.


EItanya (commented):
Do you think this can be added as an `extraObject` in you chart instead?

younsl:
Good question, and one I should answer carefully since I am the one who added extraObjects in #2174 and argued there that PrometheusRule belongs in it. I still think that. I would put a ServiceMonitor on the other side of that line, and here is my reasoning.

extraObjects works well when the content of the resource is org-specific and independent of the chart: ExternalSecret, NetworkPolicy, Istio Sidecar, PrometheusRule. This ServiceMonitor is a pointer at a Service the chart itself renders, so its endpoint is not user content but derived state. The port name (https vs http-metrics), scheme, tlsConfig and bearerTokenFile all follow controller.metrics.secureServing, the selector has to match kagent.controller.selectorLabels, and it has to stay unrendered under the METRICS_BIND_ADDRESS=0 sentinel. A copy in extraObjects has to hardcode those, and if someone later flips secureServing the scrape fails quietly: the manifest still applies, Prometheus just drops the target. That is the one thing I could not solve outside the chart, so this PR routes the port name through a shared helper that the Service and the ServiceMonitor both use.

For context on scope, #1803 landed the Service and the RBAC and left this piece for a follow-up, with the same value path it suggested (controller.metrics.serviceMonitor.enabled). This PR is meant to be that follow-up, additive and changing nothing that #1803 shipped. #1369 asked for both halves, so the ServiceMonitor half is the part still outstanding. @danielorbach @Daniel-Vaz, a sanity check from you would help here: is this the shape you had in mind?

Two smaller notes. The kagent-tools chart already has servicemonitor.yaml under tools.metrics.servicemonitor.*, so this would make the controller consistent with it. And there is precedent in the chart for opt-in resources that reference external CRDs, ui-httproute.yaml and openshift-route.yaml, if the Prometheus Operator dependency is the concern. It is disabled by default, so rendered output for existing installs is unchanged and helm unittest covers that.

Happy to go smaller if the values surface is the issue: I can trim to what kagent-tools exposes (enabled, interval, scrapeTimeout, labels) and drop relabelings, metricRelabelings, annotations and the namespace override. And if you would still rather this live outside the chart, I am fine closing it and sending just the shared port-name helper plus a documented ServiceMonitor example in the extraObjects block of values.yaml, so at least the port name and scheme have one reference. Whichever you prefer.

danielorbach:
@younsl I'd advise against `extraObjects` for this one, even though I think it's the right answer for most things people want to bolt on.

It comes down to who owns the contents. An `ExternalSecret` or a `NetworkPolicy` is yours: you write it, you know what belongs in it. A `ServiceMonitor` is really just a pointer at a Service this chart renders, and nearly every field in it is decided by `controller.metrics.secureServing`: the port name, the scheme, whether there's a token, whether TLS verification gets skipped. Hand someone a copy to keep in their values and it's correct the day they paste it. Then somebody flips `secureServing` next quarter, the manifest still applies clean, and Prometheus quietly stops having a target. No error anywhere. A template in the chart just doesn't have that failure mode, because it re-renders on every upgrade.

I'm the one who punted this in #1803, and I was too cautious about dragging Prometheus Operator in. Nothing gets dragged in while the flag defaults to false, and `openshift-route.yaml` and `ui-httproute.yaml` already point at CRDs we don't ship.

In my personal opinion for this PR: `controller.metrics.serviceMonitor.*` is indeed what I now have in mind. Five things I'd want ticked off, and you've already done four:

- [ ] off by default, existing installs render identically
- [ ] the port name defined once and shared by the Service and the endpoint, no stray `https` or `http-metrics` literals left anywhere
- [ ] `secureServing` drives the whole scrape, so flipping it alone gives you a working scrape with nothing else to touch. I didn't spot a test for that pairing, just the secure and plaintext cases on their own
- [ ] `bindAddress: "0"` kills the ServiceMonitor the same way it kills the Service
- [ ] turning it on gives you a scrape that works, not a 403

That last one's the real gap. In #1803 I bound `metrics-auth-role` but left `metrics-reader` unbound becase I had no name to put in it. The auth role goes on the controller's own `ServiceAccount`, which we create, so that one writes itself. The reader role has to go on whatever ServiceAccount Prometheus runs as, and the chart had no idea whether that's `prometheus-k8s` or `kube-prometheus-stack-prometheus` or something homegrown in a namespace we never touch.

Rendering a `ServiceMonitor` changes that, because now the user is telling us they run Prometheus Operator. So ask them for the `ServiceAccount` and render the `ClusterRoleBinding` when they hand you one. Blank means unbound, exactly like today, so nobody's existing setup shifts. Without it, `enabled: true` with `secureServing` on drops a target into Prometheus that can never be scraped, and all anyone sees is a silent 403 behind a manifest that applied fine. A comment in `values.yaml` won't necessarily catch that.

One last thing, and probably outside the scope of this pull-request: we're on track for three spellings of this in one umbrella: `controller.metrics.serviceMonitor.*` here, `tools.metrics.servicemonitor.*` in kagent-tools, and kmcp with no ServiceMonitor at all plus a hardcoded `https` in its metrics Service. If @EItanya and the maintainers all settle on one way of spelling it, I'll be happy to apply it. I'm just not at a position to make that choice, so I try to not invent a fourth.

younsl:
Both CI failures are pre-existing on main, not from these chart changes. The same two jobs fail on [main HEAD](https://github.com/kagent-dev/kagent/actions/runs/31654555861).

younsl:
Not stale

younsl:
@danielorbach thanks for laying out the checklist. Rebased on current main and pushed a second commit that closes the two gaps. Going through the five:

- [x] **Off by default, existing installs render identically.** `helm template` with default values produces no ServiceMonitor and no new binding. Covered by the "should not render by default" cases in both suites.
- [x] **Port name defined once.** `kagent.controller.metricsPortName` in `_helpers.tpl` is the only place `https` / `http-metrics` is spelled. The Service and the ServiceMonitor endpoint both include it.
- [x] **`secureServing` drives the whole scrape.** New suite `controller-metrics-pairing_test.yaml` renders the Service, the ServiceMonitor and both metrics bindings together and asserts they all follow `secureServing` as a set: with it on, the Service port and the endpoint are both `https`, the scheme and bearer token are set, and both bindings render. With it off, both are `http-metrics`, no scheme / token / tlsConfig, and neither binding renders. Flipping that one value is the only change needed.
- [x] **`bindAddress: "0"` kills the ServiceMonitor** the same way it kills the Service, via the shared `kagent.controller.metricsEnabled` helper. Same for the new binding.
- [x] **Turning it on gives a scrape that works, not a 403.** Added `controller.metrics.serviceMonitor.prometheusServiceAccount.{name,namespace}`. When `name` is set and `secureServing` is on, the chart renders a `ClusterRoleBinding` from `<fullname>-metrics-reader` to that ServiceAccount. Empty `name` means unbound, exactly like today, so nobody's existing setup moves. The subject namespace defaults to `serviceMonitor.namespace`, then the release namespace, which matches where kube-prometheus-stack and kube-prometheus put their Prometheus ServiceAccount when the ServiceMonitor is created alongside it.

@EItanya on the `extraObjects` question: I would rather keep it as a template for the reason above. Nearly every field in the ServiceMonitor is derived from `secureServing`, so a copy in `extraObjects` goes stale the day that value changes, and the failure is a silent missing target rather than an error. That said, if you would still prefer it outside the chart I am happy to trim this down to the shared port-name helper plus a documented example in the `extraObjects` block.

On the naming split (`serviceMonitor` here vs `servicemonitor` in kagent-tools), I kept camelCase to match the other keys in this chart. Happy to align to whichever spelling the maintainers pick.

Rebased to main, so the two CI jobs that were failing on main at the time should be green now.

helm/kagent/templates/controller-servicemonitor.yaml
krisztianfekete, line 1: Can you please gate this on having the required 3rd party CRDs in place via `.Capabilities.APIVersions.Has`?

younsl: Done in 623d1ee2. The ServiceMonitor now renders only when `.Capabilities.APIVersions.Has "monitoring.coreos.com/v1"` is true, following the same pattern as `openshift-route.yaml`.

The check lives in a `kagent.controller.serviceMonitorEnabled` helper that the metrics-reader ClusterRoleBinding also uses, since that binding only exists to authorize the ServiceMonitor scrape. The metrics Service and the metrics-auth binding are unaffected.

Tests declare the API version in every case and add one per suite asserting nothing renders when it is absent.

EItanya (approved):

PR #2403, opened 2026-08-07
Backport 2409 (db session cleanup) to 0.10.x release

supreme-gg-gg:
This targets the 0.10.x release. To use, e.g. set `database.posetgres.sessionRetentionDays: 30`

The following tables are eligible for cleanup: session, shares, events (ADK), push notifications (A2A), tasks (A2A), checkpoints (LangGraph), memory and flow states (CrewAI)

This shares the same cleanup runner as long term memory.

### DIFF
--- a/go/core/internal/database/client_postgres.go
+++ b/go/core/internal/database/client_postgres.go
@@ -836,6 +836,37 @@ func (c *postgresClient) PruneExpiredMemories(ctx context.Context) error {
 	})
 }
 
+const sessionRetentionBatchSize int32 = 1000
+
+func (c *postgresClient) PruneExpiredSessions(ctx context.Context, retentionDays int) (int64, error) {
+	if retentionDays <= 0 {
+		return 0, nil
+	}
+	var total int64
+	for {
+		var n int64
+		err := c.withTx(ctx, func(q *dbgen.Queries) error {
+			var err error
+			n, err = q.DeleteExpiredSessionsBatch(ctx, dbgen.DeleteExpiredSessionsBatchParams{
+				RetentionDays: int32(retentionDays),
+				BatchSize:     sessionRetentionBatchSize,
+			})
+			if err != nil {
+				return fmt.Errorf("failed to delete expired sessions batch: %w", err)
+			}
+			return nil
+		})
+		if err != nil {
+			return total, err
+		}
+		total += n
+		if n == 0 {
+			break
+		}
+	}
+	return total, nil
+}
+
 // ── Conversion helpers ────────────────────────────────────────────────────────
 
 func toAgent(r dbgen.Agent) *dbpkg.Agent {
--- a/go/core/internal/database/queries/sessions.sql
+++ b/go/core/internal/database/queries/sessions.sql
@@ -47,6 +47,7 @@ WHERE id = $1 AND user_id = $2 AND deleted_at IS NULL;
 
 -- DeleteExpiredSessionsBatch hard-deletes up to batch_size idle sessions whose
 -- updated_at is older than retention_days, plus cascaded conversation state.
+-- Soft-deleted sessions are included so tombstones still reclaim disk.
 -- name: DeleteExpiredSessionsBatch :one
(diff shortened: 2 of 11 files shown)


go/core/pkg/app/app.go
```diff
@@ -777,9 +779,9 @@ func Start(getExtensionConfig GetExtensionConfig, extraSources []migrations.Sour
 		os.Exit(1)
 	}
 
-	// Memory TTL cleanup runs only on the leader to avoid duplicate deletes.
-	if err := mgr.Add(httpserver.NewMemoryCleanupRunnable(dbClient, 0)); err != nil {
-		setupLog.Error(err, "unable to set up memory cleanup runnable")
+	// DB TTL cleanup (memory + sessions) runs only on the leader to avoid duplicate deletes.
+	if err := mgr.Add(httpserver.NewDbCleanupRunnable(dbClient, 1*time.Minute, cfg.Database.SessionRetentionDays)); err != nil {
```
iplay88keys: `NewDbCleanupRunnable` mentions the default is every 24 hours, but this runs every minute by default and is hardcoded that way. Should we make the frequency configurable?

go/core/internal/database/queries/sessions.sql
```diff
@@ -44,3 +44,64 @@ ON CONFLICT (id, user_id) DO UPDATE SET
 -- name: SoftDeleteSession :exec
 UPDATE session SET deleted_at = NOW()
 WHERE id = $1 AND user_id = $2 AND deleted_at IS NULL;
+
+-- DeleteExpiredSessionsBatch hard-deletes up to batch_size idle sessions whose
+-- updated_at is older than retention_days, plus cascaded conversation state.
+-- name: DeleteExpiredSessionsBatch :one
```
iplay88keys, line 51: There is a potential for a race condition with this. The expiry CTE snapshots updated_at, but the final delete only rechecks id/user_id. So you can have an event update the session, followed by cleanup deleting the session, leaving the new event orphaned.

go/core/internal/database/client_postgres.go
```diff
@@ -836,6 +836,37 @@ func (c *postgresClient) PruneExpiredMemories(ctx context.Context) error {
...
+	if retentionDays <= 0 {
+		return 0, nil
+	}
+	var total int64
+	for {
+		var n int64
+		err := c.withTx(ctx, func(q *dbgen.Queries) error {
+			var err error
+			n, err = q.DeleteExpiredSessionsBatch(ctx, dbgen.DeleteExpiredSessionsBatchParams{
+				RetentionDays: int32(retentionDays),
```
iplay88keys, line 851: Is there a reason to convert the int to an int32 here? Should we just take in an int32? Probably not a problem based on the expected value ranges, but an int could be truncated by this conversion.

go/core/pkg/app/app.go
supreme-gg-gg: Oops, I changed it to every minute for testing but forgot to change it back... I also thought about making this configurable. However, for now I don't find it too useful and once every day would be a reasonable choice for most users. We will need more complicated logic like check cleanup period < TTL, handle different time units (minutes, hours, days) etc and it doesn't bring much benefit.

go/core/internal/database/client_postgres.go
supreme-gg-gg: `flag.IntVar` takes in an `*int` and sqlc generates `int32` for `integer` type, so we will need this conversion somewhere unfortunately

go/core/internal/database/queries/sessions.sql
supreme-gg-gg: I've changed the CTE to use a `FOR UPDATE` locking clause, it also skips sessions locked by other transactions to avoid waiting (since a session actively being updated will not expire), good catch, thanks!

iplay88keys (dismissed):

EItanya (dismissed):

iplay88keys (approved):

PR #2424, opened 2026-08-11
feat(models): Adding support for claude/anthropic models on Azure Foundry

marosset:
Follow up to https://github.com/kagent-dev/kagent/pull/2091 

Adding support for configuring Anthropic chat models on Azure Foundry.
These changes re-use the same Azure authentication logic and workflows that was implemented in 2091
These changes are currently **go only** (just like the previous work on Foundry).
I will follow up with python support for Azure Foundry models in the future.


go/adk/pkg/models/foundry.go
```diff
@@ -105,3 +99,82 @@ func NewFoundryModelWithLogger(ctx context.Context, config *FoundryConfig, logge
...
+//
+// cred is the Azure credential for the Workload Identity path; like the region /
+// projectID arguments on the Vertex and Bedrock constructors it is passed
+// explicitly rather than stored on the shared AnthropicConfig. It is nil in
+// production (resolved to DefaultAzureCredential — Workload Identity in-cluster,
+// or the az CLI in local development) and set by tests to inject a fake.
+//
+// It returns an *AnthropicModel so the shared genai<->Messages translation in
+// anthropic_adk.go is reused unchanged.
+func NewFoundryAnthropicModelWithLogger(ctx context.Context, config *AnthropicConfig, endpoint, deployment string, cred azureai.TokenCredential, logger logr.Logger) (*AnthropicModel, error) {
```
EItanya, line 122: Can't we put the logger in the context? Why do we need to call it out in the constructor specifically?

go/api/v1alpha2/modelconfig_types.go
EItanya, line 1: There's no need to update v1alpha2 types

go/adk/pkg/models/foundry.go
marosset: I was following precident
 https://github.com/kagent-dev/kagent/blob/2393ef715b8b20599981cb406fff67aa66ef70ee/go/adk/pkg/models/bedrock.go#L151
 https://github.com/kagent-dev/kagent/blob/2393ef715b8b20599981cb406fff67aa66ef70ee/go/adk/pkg/models/openai.go#L75
 https://github.com/kagent-dev/kagent/blob/2393ef715b8b20599981cb406fff67aa66ef70ee/go/adk/pkg/models/sapaicore.go#L37
 
 Not all of the model providers take in a context but those that do (Ex: bedrock) still take in logger in the factory method.
 
 Let me know if you would like me to update things the behavior

go/api/v1alpha2/modelconfig_types.go
marosset: I think we need to update v1alpha2 as well as long as the storage version is `v1alpha2` otherwise the field gets pruned and never gets set. @EItanya are there plans to update the storage version to `v1alpha3` soon?

EItanya (approved):

PR #2429, opened 2026-08-12
feat: support file upload in agent chat

supreme-gg-gg:
- Forward non-image file parts through Go model adapters instead of silently dropping them
- Add chat UI file attach (paperclip, drag/drop, history chips) for Go declarative agents only
- Demo video is attached in a comment below

---
Current support matrix:

Provider | Native images | Native documents
-- | -- | --
OpenAI (CC / Azure / Foundry) | image/* | PDF only, text extracted from text files
OpenAI (Responses) | image/* | PDF, text, markdown, csv, html, json, word, pptx, etc.
Anthropic | image/* | PDF, text/plain, text/markdown
Bedrock | png/jpeg/gif/webp | pdf, txt, md, csv, html, doc/docx, xls/xlsx
Ollama | image/* only | —

### DIFF
--- a/go/adk/pkg/models/bedrock.go
+++ b/go/adk/pkg/models/bedrock.go
@@ -753,12 +753,35 @@ func convertGenaiContentsToBedrockMessages(contents []*genai.Content, nameMap ma
 			}
 		}
 
+		// Bedrock Converse: a document ContentBlock requires a text ContentBlock
+		// in the same message (https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Message.html).
+		contentBlocks = ensureBedrockDocumentHasText(contentBlocks)
+
 		if len(contentBlocks) > 0 {
 			messages = append(messages, types.Message{Role: role, Content: contentBlocks})
 		}
 	}
 
 	return messages, mergeSystemInstructionFromConfig(systemInstruction, config)
+}
+
+// ensureBedrockDocumentHasText appends a short prompt when a message has documents but no text
+func ensureBedrockDocumentHasText(blocks []types.ContentBlock) []types.ContentBlock {
+	hasDoc, hasText := false, false
+	for _, b := range blocks {
+		switch b.(type) {
+		case *types.ContentBlockMemberDocument:
+			hasDoc = true
+		case *types.ContentBlockMemberText:
+			hasText = true
+		}
+	}
+	if hasDoc && !hasText {
+		blocks = append(blocks, &types.ContentBlockMemberText{
+			Value: "Please review the attached document.",
+		})
+	}
+	return blocks
 }
 
 // convertGenaiToolsToBedrock converts genai.Tool to Bedrock Tool format.
--- a/go/adk/pkg/models/bedrock.go
+++ b/go/adk/pkg/models/bedrock.go
@@ -667,6 +667,40 @@ func convertGenaiContentsToBedrockMessages(contents []*genai.Content, nameMap ma
 				continue
 			}
 
+			if part.InlineData != nil {
+				mime := part.InlineData.MIMEType
+				name := blobName(part.InlineData)
+				if fmtStr := bedrockImageFormat(mime); fmtStr != "" {
+					contentBlocks = append(contentBlocks, &types.ContentBlockMemberImage{
+						Value: types.ImageBlock{
+							Format: types.ImageFormat(fmtStr),
+							Source: &types.ImageSourceMemberBytes{Value: part.InlineData.Data},
+						},
+					})
+				} else if docFmt := bedrockDocumentFormat(mime, name); docFmt != "" {
+					contentBlocks = append(contentBlocks, &types.ContentBlockMemberDocument{
+						Value: types.DocumentBlock{
+							Format: types.DocumentFormat(docFmt),
+							Name:   aws.String(bedrockSafeDocName(name)),
+							Source: &types.DocumentSourceMemberBytes{Value: part.InlineData.Data},
+						},
+					})
+				} else {
+					contentBlocks = append(contentBlocks, &types.ContentBlockMemberText{
+						Value: unsupportedFileNote(name, mime),
+					})
+				}
+				continue
+			}
+
+			if part.FileData != nil {
+				// Bedrock Converse document/image blocks are bytes-only.
+				contentBlocks = append(contentBlocks, &types.ContentBlockMemberText{
+					Value: unsupportedFileNote(fileDataName(part.FileData), part.FileData.MIMEType),
+				})
+				continue
+			}
+
 			// Handle function call (tool use in Bedrock terminology).
 			// Use the sanitized name from nameMap so Bedrock can correlate the
 			// tool call with the tool spec sent in the same request.
--- a/go/adk/pkg/models/file_parts.go
+++ b/go/adk/pkg/models/file_parts.go
@@ -213,7 +213,8 @@ func bedrockDocumentFormat(mime, name string) string {
 	return ""
 }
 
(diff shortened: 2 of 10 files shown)


supreme-gg-gg:
https://github.com/user-attachments/assets/15b637f2-18fa-4217-8e81-20c7fa665be8

iplay88keys:
Tested locally with a PDF and gpt-4.1-mini:
image

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -304,7 +353,8 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
       rethrowOnError?: boolean;
     } = {},
   ) => {
-    if (!userMessageText.trim() || !selectedAgentName || !selectedNamespace) {
+    const filesToSend = pendingFiles;
```
iplay88keys: `handleMcpAppSendMessage()` also calls `sendChatMessageText`, so an MCP app's `ui/message` will send and clear any files the user has staged.

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -318,6 +368,7 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
 
     if (options.clearInput ?? true) {
       setCurrentInputMessage("");
+      setPendingFiles([]);
```
iplay88keys: If the cross-tab guard below returns "blocked", then the staged files will be cleared without sending. We should only clear after the cross-tab guard succeeds.

go/adk/pkg/models/bedrock.go
```diff
@@ -667,6 +667,40 @@ func convertGenaiContentsToBedrockMessages(contents []*genai.Content, nameMap ma
...
+				mime := part.InlineData.MIMEType
+				name := blobName(part.InlineData)
+				if fmtStr := bedrockImageFormat(mime); fmtStr != "" {
+					contentBlocks = append(contentBlocks, &types.ContentBlockMemberImage{
+						Value: types.ImageBlock{
+							Format: types.ImageFormat(fmtStr),
+							Source: &types.ImageSourceMemberBytes{Value: part.InlineData.Data},
+						},
+					})
+				} else if docFmt := bedrockDocumentFormat(mime, name); docFmt != "" {
```
iplay88keys, line 680: Based on the Bedrock Converse [documentation](https://docs.aws.amazon.com/cli/latest/reference/bedrock-runtime/converse.html), if a message contains a `document` content block, there must also be a `text` content block:
>If you include a ContentBlock with a document field in the array, you must also include a ContentBlock with a text field.

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -45,6 +45,47 @@ import {
   type SessionGuardOptions,
 } from "@/lib/chatSessionGuard";
 
+/** Soft client cap so base64 A2A FileParts stay reasonable. */
+const MAX_CHAT_FILE_BYTES = 10 * 1024 * 1024;
```
iplay88keys: [Bedrock](https://docs.aws.amazon.com/cli/latest/reference/bedrock-runtime/converse.html) has a max size of: 3.75 MB for images and 4.5 MB for documents with a max of 20 images or 5 documents. We should make sure the count/size conform to provider limits before allowing sending.
>- You can include up to 20 images. Each image’s size, height, and width must be no more than 3.75 MB, 8000 px, and 8000 px, respectively.
>- You can include up to five documents. Each document’s size must be no more than 4.5 MB.

go/adk/pkg/models/bedrock.go
supreme-gg-gg: I can think of two ways to enforce this, either we allow the user to only send a document (which is the behaviour right now) and check in adapter if there is no text part, we will add a default message like "review this file", or we will require the user to always send a text part alongside any documents (excluding images, since bedrock only requires this for documents).

I'll go with the first approach since A2A allows `FilePart` without a `TextPart` in messages

ui/src/components/chat/ChatInterface.tsx
supreme-gg-gg: Agreed, I've added the checks and the above 2 UI review comments in 2e28e44

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -474,19 +539,104 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
...
+        toast.error(`You can attach at most ${MAX_CHAT_FILES} files per message`);
+        break;
+      }
+      const isImage = file.type.startsWith("image/");
+      const maxBytes = isImage ? MAX_CHAT_IMAGE_BYTES : MAX_CHAT_DOC_BYTES;
+      if (file.size > maxBytes) {
+        const mb = (maxBytes / (1024 * 1024)).toFixed(2);
+        toast.error(`${file.name} exceeds the ${mb}MB ${isImage ? "image" : "file"} limit`);
+        continue;
+      }
```
iplay88keys, line 575: Bedrock supports images up to a max size of 8000 x 8000 px. Do we want to validate that here as well?

go/adk/pkg/models/file_parts.go
```diff
@@ -0,0 +1,241 @@
...
+				prevSpace = true
+			}
+		}
+	}
+	out := strings.TrimSpace(b.String())
+	if out == "" {
+		return "document"
+	}
+	return out
+}
```
iplay88keys, line 245: Bedrock document names also have a max of 200 characters. It's also worth considering a neutral name for them, though we'd have to distinctly name each of the files.

From the [docs](https://docs.aws.amazon.com/cli/latest/reference/bedrock-runtime/converse.html#output):

>A name for the document. The name can only contain the following characters:
>
>* Alphanumeric characters
>* Whitespace characters (no more than one in a row)
>* Hyphens
>* Parentheses
>* Square brackets
>
>Note
>This field is vulnerable to prompt injections, because the model might inadvertently interpret it as instructions. Therefore, we recommend that you specify a neutral name.
>
>Constraints:
>* min: 1
>* max: 200

ui/src/components/chat/ChatInterface.tsx
```diff
@@ -474,19 +539,104 @@ export default function ChatInterface({ selectedAgentName, selectedNamespace, se
...
       if (options.rethrowOnError) {
         throw error;
       }
     }
   };
 
+  const handleAttachFiles = async (fileList: FileList | null) => {
+    if (!fileList || fileList.length === 0) return;
+    const next: PendingChatFile[] = [];
+    let slots = MAX_CHAT_FILES - pendingFiles.length;
```
iplay88keys: Codex's review mentions:
>`pendingFiles.length` is captured before the asynchronous file reads. Two overlapping picker/drop operations can both calculate the same available slots and ultimately append more than `MAX_CHAT_FILES`. Please reserve slots while reading or enforce the cap atomically in the state update.

iplay88keys (dismissed):
Just some extra validation, but nothing blocking.

go/adk/pkg/models/file_parts.go
supreme-gg-gg: Neutral name sounds like the user's responsibility? It would be non-trivial to detect (and replace) it. Also this might be something that a gateway would be able to do to prevent prompt injection, having it at the agent framework level would be wrong IMO?

iplay88keys: That seems reasonable. I agree that it would be pretty trivial to detect/resovle in this case.

iplay88keys (approved):

PR #2438, opened 2026-08-13
feat(controller): reconcile SandboxAgents when referenced header Secrets change

### LINKED ISSUE #2404: RemoteMCPServer headersFrom secret is baked into agent config and never re-rendered on rotation — agents break permanently with short-lived tokens
## Summary

When a `RemoteMCPServer` authenticates via `headersFrom → Secret`, the controller resolves that Secret **at agent-render time** and writes the literal header value into the per-agent config Secret (`Secret/<agent>`, `config.json` → `http_tools[0].params.headers.Authorization`). The agent pod mounts only its own config Secret, so it uses whatever value was current at the last render.

Nothing re-renders the agent config when the referenced Secret changes. With a short-lived credential the agent works for one token lifetime after each render, then every tool call fails — while a valid token sits unused in the Secret. Agent renders are event-driven and in our cluster occurred 4–17 hours apart, versus a 60-minute token TTL.

## Environment

- controller `0.9.12`
- `RemoteMCPServer` `kagent.dev/v1alpha2`, `protocol: STREAMABLE_HTTP`, remote MCP endpoint over HTTPS
- credential from the external-secrets `GithubAccessToken` generator (60 min token TTL), `ExternalSecret` with `refreshInterval: 45m`, target Secret key `Authorization` holding `Bearer <token>`

## Reproduce

1. Create a `RemoteMCPServer` with `headersFrom[0].valueFrom.type: Secret`, pointing at a Secret that external-secrets refreshes from a short-lived token source.
2. Reference that `RemoteMCPServer` from an Agent's `spec.tools`.
3. Wait past the token TTL without making any change to the Agent spec.
4. All tool calls through that MCP server fail. Restarting the agent pod does **not** help — the config Secret itself is stale, so the new pod receives a byte-identical config (same `kagent.dev/config-hash`).

## Expected vs actual

**Expected:** the agent's rendered config tracks the referenced Secret (watch + re-render), or the credential is resolved at connection time rather than inlined into the config.

**Actual:** the header value is a snapshot taken at render time. Recovery requires some unrelated event that happens to trigger an Agent re-render.

## Evidence

- Three SHA-256-compared values: live auth Secret ≠ agent config Secret = pod-mounted `/config/config.json`. The pod matches its config Secret, so the kubelet mount sync is fine — the config Secret is the stale layer.
- Same pod, unchanged `kagent.dev/config-hash`: a read-only API call with the pod's token returned **200** at time T and **401** at T+16 min, while the freshly rotated token from the auth Secret returned **200** at that same moment. This rules out upstream rate limiting or server-side session state.
- The controller's own MCP loop *does* re-read the Secret — ~720 `successfully registered remote MCP server` log lines in a 3-hour window, with a stable tool catalog. Only the agent render is stale. Consequently `RemoteMCPServer.status` stays `ACCEPTED` and the UI looks healthy while every agent tool call gets a 401: the status reflects the controller's own connection, not the agent's.
- Agent-controller reconciles were rare in the same window (1 occurrence in 3 hours) versus continuous MCP reconciles, which is why the stale window lasts hours.
- The controller ServiceAccount already holds `get,list,watch` on core resources, so watching the referenced Secret needs no new RBAC.

## Secondary finding (happy to split into its own issue)

A 401 from the remote MCP server is swallowed: the agent produces **no** `function_response` at all rather than surfacing the status. The model then invents an explanation — in our case "the repository does not exist / is not accessible" — and reports that to the user as if it were a finding. One session contained 11 consecutive tool calls with zero responses and zero warnings in the agent log, which made the actual cause (an expired credential) very hard to attribute.

Logging the upstream non-2xx status at WARN when an MCP tool call fails would make this diagnosable in seconds.

## Suggested fix

Either:

1. Have the agent reconciler watch Secrets referenced via `headersFrom` and re-render the agent config on change (annotating the agent Deployment so the pod rolls), or
2. Stop inlining the resolved header and resolve it at connection time.

Option 1 keeps the existing "self-contained rendered config" model (the agent pod has no RBAC to read Secrets, which is a reasonable design) while making the implicit assumption — that everything inlined is static — actually hold.

voltagebots:
## What

Adds a `Secret` watch to the SandboxAgent controller so that changes to a Secret referenced by an agent's request headers trigger reconciliation. The watch resolves referencing agents through two paths:
- Direct: a tool's `HeadersFrom` references the Secret.
- Indirect: a tool points at a `RemoteMCPServer` whose `HeadersFrom` references the Secret (namespace-access-checked via `AllowedNamespaces`).

An `UpdateFunc` predicate (`secretDataChangedPredicate`) gates events on actual `Data`/`StringData` changes, and `remoteMCPServerReferencesSecret` is extended to match header-sourced Secret refs.

## Why

Before this change, rotating a Secret used only for request headers left running agents with stale credentials until an unrelated reconcile happened to fire. Watching the Secret closes that gap and makes header-credential rotation take effect deterministically.

Fixes #2404.

## Accepted tradeoff — rollout on rotation

The agent's rendered config hash includes the resolved header value, so a genuine token rotation changes the hash and rolls the pod. This is intended and accepted: a rotated credential *should* propagate to a fresh pod. The predicate ensures only real `Data`/`StringData` changes qualify, so no-op Secret updates do not churn workloads.

## Note — graceful degradation on RemoteMCPServer-list failure

The Secret finder deliberately does not fail closed if the `RemoteMCPServer` list call errors (documented in-code). Failing closed there would also drop the *direct* `HeadersFrom` matches, which never depended on that list. Instead the path degrades to indirect-incomplete — direct matches are always returned; only indirect (via-RemoteMCPServer) matches may be missed on a transient list error, and are recovered on the next reconcile. Covered by an explicit test, including the empty-`apiGroup` RemoteMCPServer reference form.

## Testing

- `go build ./core/internal/controller/...` — passes
- `go test -race -skip 'TestE2E.*' -v ./core/internal/controller/` — all pass, including 7 new `TestSandboxAgentSecretFinder` cases (direct/indirect/cross-namespace/empty-apiGroup/list-failure-degradation) and the data-changed predicate test
- `go vet ./core/internal/controller/...` — clean

Opening as draft per the contribution guide's process for changes >100 lines — happy to discuss approach before this is ready for review.

### DIFF
--- a/go/core/internal/controller/agentobject_helpers.go
+++ b/go/core/internal/controller/agentobject_helpers.go
@@ -78,6 +78,14 @@ func usesModelConfig(agent *v1alpha3.SandboxAgent, obj types.NamespacedName) boo
 		spec.Declarative.ModelConfig == obj.Name
 }
 
+func valueRefsReferenceSecret(refs []v1alpha3.ValueRef, name string) bool {
+	return slices.ContainsFunc(refs, func(ref v1alpha3.ValueRef) bool {
+		return ref.ValueFrom != nil &&
+			ref.ValueFrom.Type == v1alpha3.SecretValueSource &&
+			ref.ValueFrom.Name == name
+	})
+}
+
 func referencesConfigMap(agent *v1alpha3.SandboxAgent, obj types.NamespacedName) bool {
 	spec := agent.GetAgentSpec()
 	if agent.GetNamespace() != obj.Namespace || spec.Type != v1alpha3.AgentType_Declarative || spec.Declarative == nil {
--- a/go/core/internal/controller/remote_mcp_server_controller.go
+++ b/go/core/internal/controller/remote_mcp_server_controller.go
@@ -131,5 +131,10 @@ func remoteMCPServerReferencesSecret(server *v1alpha3.RemoteMCPServer, secretObj
 		return true
 	}
 
+	// check if secret is referenced by an upstream request header
+	if valueRefsReferenceSecret(server.Spec.HeadersFrom, secretObj.Name) {
+		return true
+	}
+
 	return false
 }
--- a/go/core/internal/controller/sandboxagent_controller.go
+++ b/go/core/internal/controller/sandboxagent_controller.go
@@ -119,6 +119,7 @@ func (r *SandboxAgentController) SetupWithManager(mgr ctrl.Manager) error {
 		mcpService:      r.sandboxAgentDependencyFinder("failed to list sandboxagents for Service watch", usesMCPService),
 		configMap:       r.sandboxAgentDependencyFinder("failed to list sandboxagents for ConfigMap watch", referencesConfigMap),
 		mcpServer:       r.sandboxAgentDependencyFinder("failed to list sandboxagents for MCPServer watch", usesMCPServer),
+		secret:          r.sandboxAgentSecretFinder,
 	})
 	if err != nil {
 		return err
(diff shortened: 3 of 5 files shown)


voltagebots:
Still relevant and I am picking this back up.

Recap of the change. A SandboxAgent referencing a Secret through `headersFrom` was not reconciled when that Secret changed, so a rotated credential was not picked up until something else triggered reconciliation. This adds a watch on the referenced Secrets and the helper plumbing to map a Secret back to the SandboxAgents that reference it, with controller tests.

Current state. All checks are green. The branch has fallen behind `main` and now reports a conflict, so I am rebasing it onto current `main` and will push the result shortly. Upstream churn on the three controller files involved has been light, so I expect the rebase to be mechanical rather than a rewrite.

If the approach is not the one you want for Secret rotation I would rather adjust it now than carry it further.

voltagebots:
Closing this. Correcting my previous comment, where I said I was rebasing it.

I attempted the rebase and it failed as modify/delete across all four files, because `26732e86` (#2565, "remove legacy ACP and controller runtime") deleted the controller this PR was built on. SandboxAgent no longer exists as a reconciled type in the current layout.

More to the point, the problem is solved rather than relocated. `go/core/internal/controller/remotemcpserver/reconciler.go` now does exactly what this PR proposed.

```go
Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(r.requestsForDependency)).
Watches(&corev1.ConfigMap{}, handler.EnqueueRequestsFromMapFunc(r.requestsForDependency)).
```

A rotated Secret referenced through `headersFrom` now triggers reconciliation on its own, which was the whole point of the change. Rebasing this would mean reintroducing a controller the project deliberately removed, to fix something that is no longer broken.

Apologies for leaving it open long enough to need the stale bot. Thanks for the restructure, it is a better answer than the one I proposed.

PR #2439, opened 2026-08-13
fix(adk): stop leaking Authorization headers in MCP tool error responses

voltagebots:
## What

Reworks `ConnectionSafeMcpTool` error handling so failure responses and logs never stringify an exception that may carry request metadata:
- `_connection_error_response` no longer interpolates `str(error)` into the model-visible message and no longer passes `exc_info=error` to the logger; it reports only the exception *type*.
- New `_extract_http_status` walks the exception chain and any `BaseExceptionGroup` (via `__cause__`/`__context__`/`.exceptions`, with cycle protection) to find an `httpx.HTTPStatusError` **without** stringifying it, and surfaces a sanitized HTTP-status message.
- `run_async` now handles bare `httpx.HTTPStatusError`, `McpError` wrapping an HTTP status, and `BaseExceptionGroup` wrapping an HTTP status.

## Why

An MCP transport failure could previously place `str(error)` — which may include the outbound `Authorization` header — into both the LLM-visible error text and the logs. This is a credential-leak path. The fix keeps error responses actionable (type name / HTTP status / "do not retry") while removing the leak surface.

## Verified against the pinned transport (`mcp==1.29.0`)

The credential-leak sanitization is proven by tests to hold regardless of how the error is wrapped (bare `HTTPStatusError`, `McpError`-wrapped, `ExceptionGroup`-wrapped, or a raw `ConnectError` carrying an `Authorization` header) — in every case the secret and header name are absent from both the returned dict and the logged output. The connection-error classification itself was checked against the actual pinned `mcp` package source: `streamable_http`'s POST handler runs inside an `anyio` task group, which propagates a non-2xx `raise_for_status()` failure as a `BaseExceptionGroup` wrapping `httpx.HTTPStatusError` — exactly the shape the group-walking branch above handles.

## Testing

`uv run pytest packages/kagent-adk/tests/unittests/test_mcp_connection_error_handling.py` — 16/16 passed (python 3.13.15), including all 4 new tests and all 12 pre-existing tests with zero regressions.


voltagebots:
Still relevant and I would like to take this forward.

Recap of the change. MCP tool errors were surfacing the outbound request headers, which means a configured `Authorization` header reaches an error response and from there a log or a UI surface. The fix redacts the header values in the error path and adds a regression test covering the connection error handling.

Current state. All checks are green, including DCO and both Snyk jobs. The branch is behind `main` but only one upstream commit has touched `_mcp_toolset.py` since this was opened, so it still applies cleanly.

Happy to rebase onto current `main` or split it further if a maintainer prefers a smaller diff. If the direction is wrong I would rather hear that and close it than leave it open.

PR #2449, opened 2026-08-15
fix(golang-adk): resolve user identity from A2A call context

### LINKED ISSUE #2443: [BUG] Chat sessions are not persisted on 0.10.0-rc2 using Golang ADK
### 📋 Prerequisites

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

Chat sessions are not persisted / not reloads after page refresh on 0.10.0-rc2 with using Golang ADK for `Agent`.

### 🔄 Steps To Reproduce

1. Deploy a PromptAgent using Golang ADK version `0.10.0-rc2` 
2. Create a chat session from Kagent UI and have some chat with the agent.
3. Refresh page.
4. The chat history does not reload.

### 🤔 Expected Behavior

Chat history should persist.

### 📱 Actual Behavior

Chat history does not persist.

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

The issue is due to chats in the `task` table being attributed to `admin@kagent.dev` incorrectly via `user_id` column.

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

buravc:
AI Disclosure: This PR was created with an AI agent under my supervision.

Fixes #2443

This PR fixes task/chat history disappearing after 0.10 upgrade.

History:

- #1775 (v0.9.4) added ctx-based user identity propagation (auth.WithUserID), read by TokenRoundTripper to set the outgoing X-User-Id header. a2a-go splits task execution and task persistence into separate goroutines off a shared context. auth.WithUserID only rebinds a local copy in the execution goroutine, so the persistence goroutine never saw it and sent no X-User-Id header on task writes.

- Before #2265, task rows had no owner at all, so this had no visible effect. #2265 added task.user_id and started scoping GetTask/ListTasksForSession/create/delete by owner. Since task writes were still missing X-User-Id (from #1775), the controller's UnsecureAuthenticator fell back to admin@kagent.dev for those writes. This is the same PR where task ownership started being recorded and where it started being recorded wrong.

- Reads now check the real caller's user_id, tasks were saved under admin@kagent.dev, so history was invisible after refresh.

**Fix:** `userIDCallInterceptor.Before` puts the user id on ctx via `auth.WithUserID` before `a2a-go` forks task execution and persistence into separate goroutines, so both inherit it. 

Added a test that builds a CallContext with x-user-id set, calls Before, and asserts the returned ctx carries the user id.

(taskstore needs no changes as `controllerclient.CallContext` already falls back to `auth.UserIDFromContext` when passed an empty user id.)


buravc:
@supreme-gg-gg Really sorry for pinging but this is fixing a broken essential feature at the moment. The bug is reducing the experience during the usage of `Agent` with Golang image.

go/adk/pkg/taskstore/store_test.go
```diff
@@ -124,6 +125,25 @@ func TestGetDecodesCanonicalTask(t *testing.T) {
...
+	service := newTaskStore(t, &taskTestServer{get: func(ctx context.Context, _ *apiv1alpha1.GetTaskRequest) (*apiv1alpha1.GetTaskResponse, error) {
+		values, _ := metadata.FromIncomingContext(ctx)
+		gotUserID = values.Get("x-user-id")
+		return &apiv1alpha1.GetTaskResponse{Task: encoded}, nil
+	}})
+
+	ctx, callCtx := a2asrv.NewCallContext(t.Context(), nil)
+	callCtx.User = a2asrv.NewAuthenticatedUser("call-context-user", nil)
+
+	_, err = service.Get(ctx, a2a.TaskID("task-4"))
```
supreme-gg-gg: nit: The issue here is incorrectly owned task writes but this test only exercises Get, it would match better if this is changed to a Create, though I understand it's the same purpose so it won't be a blocker

go/adk/pkg/a2a/executor.go
```diff
@@ -129,8 +139,8 @@ func (e *KAgentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.ExecutorCon
 		}
 
 		userID := "A2A_USER_" + reqCtx.ContextID
-		if callCtx, ok := a2asrv.CallContextFrom(ctx); ok && callCtx.User != nil && callCtx.User.Name != "" {
-			userID = callCtx.User.Name
+		if id := CallerUserID(ctx); id != "" {
```
supreme-gg-gg: IIUC the root cause is that `auth.WithUserID(ctx, userID)` is applied inside the executor, so the sibling task-store goroutine does not receive the derived context. 

Since we already have `userIDInterceptor.Before()` running before the path fork, could we apply `auth.WithUserID` there and return the derived context? Both goroutines would then inherit the user ID, allowing us to reuse the existing context-based propagation logic and keep identity extraction centralized in the interceptor

supreme-gg-gg (commented):
hi @buravc thanks for reporting the issue and the PR. One main comment, and note that we will need to back port this to `release/0.10.x` in order for this to work for the next 0.10.x release

buravc:
@supreme-gg-gg addressed your comments and also backported changes to `release/v0.10.x` branch with #2507.

supreme-gg-gg (approved):
lgtm, thanks!

buravc:
Do you know what needs to be fixed here? The actions are failing and there should be no changes affecting them AFAIK.

supreme-gg-gg:
Closing this since the fix was shipped in #2514 already to main. Thanks for the contribution @buravc your PR to 0.10.x has been merged

PR #2456, opened 2026-08-17
docs: fix broken star history chart in README

FaintFlower:
The star history chart in the README is currently broken due to GitHub stargazer API restrictions on the upstream provider. This change switches it to the maintained star-history.dera.page mirror so the chart renders correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,11 +129,11 @@ Thanks to all contributors who are helping to make kagent better.
 
 ### Star History
 
-<a href="https://www.star-history.com/#kagent-dev/kagent&Date">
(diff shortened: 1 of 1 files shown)


peterj:
We'll have to do a proper fix for this as explained [here](https://www.star-history.com/blog/how-to-use-github-star-history#how-to-embed-the-chart-in-your-readme), I don't know what dera[dot]page is.

FaintFlower:
The proper fix has to be on your own infrastructure then. It would be so much worse to give a write-enabled token to third-parties

PR #2457, opened 2026-08-17
fix: ensure API key passthrough works for OpenAI embedding models

onematchfox:
At present, one can set `apiKeyPassthrough: true` on an embedding model, however the field is completely ignored. This PR ensures that the field is honored in line with the implementation for chat completions calls.


onematchfox:
CI failures seem unrelated to this work. Linting issues were introduced in #2442. Unit tests + upgrade tests also failed on that PRs [merge commit](https://github.com/kagent-dev/kagent/actions/runs/31814333576) (not sure when they started failing though).

supreme-gg-gg (approved):
this lgtm, the CI failures are due to other works on api v2

EItanya:
> CI failures seem unrelated to this work. Linting issues were introduced in #2442. Unit tests + upgrade tests also failed on that PRs [merge commit](https://github.com/kagent-dev/kagent/actions/runs/31814333576) (not sure when they started failing though).

Hey there, we discussed this during the community meeting and in my recent blog, but `main` has officially deviated from `0.10.x` completely in preparation for the new API + using Substrate as the core runtime.

Changes which specifically target the old release should be opened against `release/v0.10.x`. If it's something which will impact both then we can either backport or forward port depending on what's easier for the situation.

EItanya (approved):

PR #2459, opened 2026-08-17
fix: key STS token cache by acting subject, not session alone

### LINKED ISSUE #2181: STS token cache keyed by session ID collapses per-user identity in shared sessions

QuentinBisson:
Supersedes #2317 (auto-closed by stale-bot; a rebase onto main conflicted with #2106, so GitHub would not let it be reopened).

Closes #2181. The Python counterpart is #2460.

## Problem

The STS token cache was keyed by session ID alone. When several subjects share one session (for example several users on the same A2A session), every caller runs its tool calls with whichever exchanged token was cached first.

## Change

Key the cache by `(sessionID, subject)`, where `subject` is a SHA-256 hash of the caller's bearer token. `BeforeRunCallback` and `HeaderProvider` both resolve the caller's bearer before the cache lookup, so each caller gets its own exchanged token.

The bearer resolution moves to `models.BearerTokenFromContext`. It prefers the value the executor stored under `BearerTokenKey` and falls back to the A2A call context, which is what reaches the MCP transport layer when `BearerTokenKey` was not threaded into the request context. `a2a/executor.go` now calls the same helper, so there is one copy of the parser instead of two.

`GetSubjectTokenFunc` documents that it must be a pure function of the bearer. The hash of the bearer stands in for the subject token only because the hook derives one from the other; an implementation that mints or fetches a token would have its first result served for the entry's lifetime.

## Why the key is a token hash and not iss+sub

A cache hit returns a delegated token without performing an exchange, so the key decides who receives whose authority. `iss` and `sub` are not verified anywhere on this path: a forged, unsigned token carrying a victim's claims would select the victim's entry and never reach the STS that would have rejected it. Hashing the raw token makes a forged token a cache miss, so it goes to the STS and fails there.

The cost is one extra exchange when a caller's bearer rotates mid-session.

An empty subject identifies no principal, so it yields no key at all. An entry stored under it would be shared by every credential-less caller in the session.

## Cache lifetime

An entry never outlives the credential that keyed it. The expiry is the earliest of the exchange's own lifetime (`expires_in`, or the exchanged token's `exp`), the subject token's `exp`, and the caller bearer's `exp`. Without that cap, a caller replaying an expired credential keeps hitting the cached delegated token instead of reaching the STS that would reject it.

When no token states an `exp` there is no expiry, and nothing may invent one: `BeforeRunCallback` runs once per run, so a run outliving a made-up deadline would lose its credential part way through. The entry carries a second field, `evictAfter`, which bounds memory alone. `Expiry` decides whether the token may still be injected; `evictAfter` decides only when the entry may be dropped, and it comes due 5 minutes after the last use when nothing states an expiry.

A cache hit renews `evictAfter` and records a use sequence, so neither the sweep nor either capacity bound can drop an entry a run is still using, and a bound drops the least recently used entries first. `getCachedToken` therefore takes the write lock, and returns a copy so that no caller reads a cached entry's fields once the lock is released. `AfterRunCallback` sweeps every entry whose `evictAfter` has come due.

## Cache capacity

The cache is bounded per session (`maxEntriesPerSession`) and across sessions (`maxCacheEntries`). The per-session bound is what matters: without it, one session presenting many distinct credentials evicts the entries of unrelated sessions, and a victim run that is between tool calls has nothing more recent to show. `BeforeRunCallback` has already run by then, so that run would reach its backends with no delegated identity. The whole-cache bound backstops it so that many sessions cannot grow the cache without limit.

Unlike the sweep, a capacity eviction drops entries whose credentials are still valid, so it is reported above debug.

## API change

`GetTokenForSession` is removed. A session no longer has a single token, so the signature cannot answer the question it asks. It had no callers outside the plugin.

## Out of scope

#2181 also describes session state being keyed by `userID`. A shared session still works: `userID` defaults to `A2A_USER_<contextID>`, so callers on one context share a session and this fix gives each of them its own exchanged token. What remains is that `userID` is part of the session address, so sending a per-sender `x-user-id` resolves those senders onto separate sessions instead of one conversation. That affects attribution, not authorization, and is untouched here.

## Header path

`getCachedToken` checks the entry's expiry, and both `BeforeRunCallback` and `HeaderProvider` go through it, so an entry past the caller credential's expiry is never injected and a rejected exchange leaves nothing usable behind. Two regression tests pin both, matching the cases the Python plugin needed a fix for.

A request whose caller is identified but has no usable entry is reported above debug. It loses its delegated identity, and nothing else says so.

### DIFF
--- a/go/adk/pkg/sts/plugin.go
+++ b/go/adk/pkg/sts/plugin.go
@@ -49,7 +49,7 @@ type TokenPropagationPlugin struct {
 	mu              sync.RWMutex
 	logger          *slog.Logger
 	bufferSeconds   int64
-	earliestExpiry  int64    // earliest Expiry in tokenCache; 0 when nothing is evictable
+	earliestExpiry  int64    // lower bound on the earliest Expiry in tokenCache; 0 when nothing is evictable
 	resource        []string // RFC 8707 resource indicators sent on the STS exchange; empty omits them
 	audience        []string // RFC 8693 audiences sent on the STS exchange; empty omits them
 }
@@ -70,8 +70,8 @@ func NewTokenPropagationPlugin(integration *STSIntegration, logger *slog.Logger,
 }
 
 // parseUnverifiedClaims parses a JWT's claims WITHOUT signature or time
-// validation. It is used only for cache partitioning and TTL, never for a
-// security decision; tokens are validated server-side during STS exchange.
+// validation. It is used only for cache TTL; tokens are validated server-side
+// during STS exchange.
 func parseUnverifiedClaims(token string) (jwt.MapClaims, bool) {
 	if token == "" {
 		return nil, false
@@ -83,27 +83,25 @@ func parseUnverifiedClaims(token string) (jwt.MapClaims, bool) {
 	return claims, true
 }
 
-// subjectKey derives a stable per-principal cache discriminator from a bearer
-// token: the issuer-scoped "sub" claim when present, otherwise a hash of the
-// raw token so opaque or sub-less tokens still partition per principal. "sub"
-// is only unique within an issuer, so it is combined with "iss" to avoid two
-// principals from different issuers colliding onto one cache entry.
+// subjectKey derives a per-principal cache discriminator from a bearer token: a
+// hash of the raw token.
+//
+// A cache hit hands the caller a delegated token without performing an exchange,
+// so the key decides who receives someone else's authority. Deriving it from
+// unverified "iss"/"sub" claims would let a forged, unsigned token select a
+// victim's entry and never reach the STS that would have rejected it. Hashing
+// the raw token instead makes a forged token a cache miss, so it goes to the
+// STS and fails there.
+//
+// The cost is a re-exchange when a principal's bearer rotates mid-session, which
+// is wanted anyway: the delegated token's lifetime tracks the subject token it
+// was exchanged from.
 func subjectKey(token string) string {
 	if token == "" {
 		return ""
 	}
-	if claims, ok := parseUnverifiedClaims(token); ok {
-		if sub, _ := claims["sub"].(string); sub != "" {
-			// An absent "iss" leaves "sub" unqualified, which would collide
-			// across two issuers that both omit it, so those fall back to the
-			// token hash rather than a half-formed key.
-			if iss, _ := claims["iss"].(string); iss != "" {
-				return iss + "\x00" + sub
-			}
-		}
-	}
 	sum := sha256.Sum256([]byte(token))
-	return "h:" + hex.EncodeToString(sum[:])
+	return hex.EncodeToString(sum[:])
 }
 
 // actingCredential returns the credential this request authenticates with, which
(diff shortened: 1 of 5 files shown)


QuentinBisson:
@supreme-gg-gg rebased onto main after the slog migration (#2705), lint and race tests pass. Ready for review.

go/adk/pkg/sts/plugin.go
```diff
@@ -281,22 +476,12 @@ func sessionIDFromContext(ctx context.Context) string {
...
-func (p *TokenPropagationPlugin) GetTokenForSession(sessionID string) string {
-	entry, ok := p.getCachedToken(sessionID)
-	if !ok {
-		return ""
-	}
-	return entry.Token
-}
-
 // ClearCache clears all cached tokens.
 func (p *TokenPropagationPlugin) ClearCache() {
```
mesutoezdil, line 480: getTokenforsession removed here, #2803 asserts this type still implements it

EItanya (approved):

EItanya:
🤖 AI-generated review.

**Follow-up, non-blocking for this PR: preserve valid delegated credentials for active invocations.**

In [the idle sweep](https://github.com/kagent-dev/kagent/blob/216b32ab9fa4786e8e4b4594c0651be21317d005/go/adk/pkg/sts/plugin.go#L411-L419), an invocation waiting more than five minutes between tool calls can lose its entry when another invocation finishes, even though its credential is still valid. This applies when the bearer, subject, and exchanged token provide no expiry. The session/global capacity limits can also evict an unfinished invocation's entry, as the capacity-eviction comment already acknowledges.

`BeforeRunCallback` runs only once per invocation, and `HeaderProvider` returns nil on a cache miss, so the affected invocation cannot recover its delegated token on its next tool call. Recent cache access does not establish whether a run is still active.

The existing in-flight tests refresh the entry immediately before the sweep or before each capacity insertion. A follow-up should retain the active invocation's credential independently of the evictable reuse cache, continue enforcing its actual expiry, and test eviction **before** the invocation's next header lookup.

Validation at `216b32ab`: the ADK race suite and repository Go lint passed. Three isolated, intentionally failing probes reproduced the loss for the idle sweep, session capacity, and global capacity. These exercised the plugin callbacks and header lookup with a local test STS; they were not full transport E2E tests.

PR #2461, opened 2026-08-17
feat(a2a): emit aggregated turn token usage on the terminal status update

QuentinBisson:
Supersedes #2332, which the stale bot closed and GitHub refuses to reopen. This is a reimplementation, not a rebase: both executors were rewritten to delegate to the ADK library's own A2A executor, so the manual event loop the original PR hooked into no longer exists.

## What this adds

Every terminal status update (`completed`, `input-required`, `failed`) carries a `kagent_usage_total` metadata entry with the token usage summed across the task:

```json
{
  "promptTokenCount": 300,
  "candidatesTokenCount": 50,
  "totalTokenCount": 350,
  "modelVersion": "gpt-4o-2024-11-20"
}
```

Partial (streaming chunk) events are skipped, since each LLM call reports its usage on the final non-partial event. Resumed tasks seed the accumulator from `Task.metadata`, so a task that spans several executions (HITL input-required cycles, follow-up messages) reports a task-lifetime total instead of the last segment only. The value is a running total re-emitted on each terminal update, so consumers take the latest value rather than summing across executions.

The value is serialized with the same genai `UsageMetadata` mapping as the existing per-event `adk_usage_metadata` / `kagent_usage_metadata` entries, plus `modelVersion`, so consumers can share one parser (with the caveat that `modelVersion` is not a field of the genai type). Tests in both languages assert that equality directly.

A call that reports no `totalTokenCount` gets one derived from `prompt + candidates + thoughts`, at the moment it is added rather than at the end. The Anthropic models report per-call input and output counts without a total, and a zero total next to non-zero counts reads as "no tokens were used". Deriving per call keeps the total correct for a task that mixes calls with and without a reported total, and for a resumed task whose persisted total was itself derived.

`a2a-go` and `a2a-python` both merge terminal status-update metadata into the stored task, which is how the total is persisted and how the next execution reads it back.

## How it differs from #2332

- Aggregation runs from an ADK runner plugin in both languages (`plugin.Config.OnEventCallback` in Go, `on_event_callback` in Python) instead of the A2A after-event callback. That callback only fires for ADK events the converter turns into at least one A2A event, and the event whose long-running function call is stripped before conversion produces none. That event carries the usage of the call that paused the task. The plugin sees every ADK event the runner produces, before conversion.
- Go stamps the aggregate from the new `AfterExecuteCallback` hook, which covers all three terminal states in one place.
- Test intent is ported rather than the test code; the Go tests now run end-to-end through the executor (including a full HITL pause/resume cycle) instead of unit-testing the accumulator in isolation.

## Known limits

- A cancelled task, or a client that disconnects mid-stream, does not reach the terminal status write, so the tokens of that in-flight execution are not added to the total. Earlier executions of the task keep theirs.
- `toolUsePromptTokenCount` is not aggregated. It is inside `totalTokenCount`, so the total stays correct, but the breakdown fields do not sum to it on Vertex.
- `modelVersion` is last-writer-wins over summed counts, so a task that uses more than one model reports a total that cannot be split back into per-model cost.

## Verification

- `go build ./...`, `go test ./adk/...`, `go vet`, `gofmt`
- `uv run pytest packages/kagent-adk/tests/unittests` (414 passed, 1 skipped)

### DIFF
--- a/go/adk/pkg/a2a/executor.go
+++ b/go/adk/pkg/a2a/executor.go
@@ -55,7 +55,7 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
 		runnerConfig.SessionService = cfg.SessionService
 	}
 	builtin := adka2a.NewExecutor(adka2a.ExecutorConfig{
-		RunnerConfig:       runnerConfig,
+		RunnerProvider:     usageObservingRunnerProvider(runnerConfig),
 		RunConfig:          runConfig,
 		A2APartConverter:   a2aPartConverter,
 		GenAIPartConverter: genAIPartConverter,
@@ -63,6 +63,13 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
 			if event.InvocationID != "" {
 				trace.SpanFromContext(ctx).SetAttributes(attribute.String("gcp.vertex.agent.invocation_id", event.InvocationID))
 			}
+			return nil
+		},
+		// The aggregate rides on the terminal status update (completed,
+		// input-required or failed), whose metadata a2a-go merges into the
+		// stored task.
+		AfterExecuteCallback: func(ctx adka2a.ExecutorContext, finalEvent *a2atype.TaskStatusUpdateEvent, _ error) error {
+			turnUsageFrom(ctx).stampEvent(finalEvent)
 			return nil
 		},
 		OutputMode: adka2a.OutputArtifactPerEvent,
@@ -121,6 +128,12 @@ func (e *KAgentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.ExecutorCon
 
 		ctx = withBearerToken(ctx)
 		ctx = auth.WithUserID(ctx, userID)
+
+		// Resumed tasks (HITL cycles, follow-up messages) carry the previously
+		// persisted total, so kagent_usage_total stays a task-lifetime sum.
+		usage := &turnUsage{}
+		usage.seedFromTask(reqCtx.StoredTask)
+		ctx = withTurnUsage(ctx, usage)
 		spanAttributes := map[string]string{
 			"kagent.user_id":         userID,
 			"gen_ai.task.id":         string(reqCtx.TaskID),
--- a/go/adk/pkg/a2a/executor.go
+++ b/go/adk/pkg/a2a/executor.go
@@ -15,6 +15,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	adkagent "google.golang.org/adk/v2/agent"
+	"google.golang.org/adk/v2/plugin"
 	"google.golang.org/adk/v2/runner"
 	"google.golang.org/adk/v2/server/adka2a/v2"
 	adksession "google.golang.org/adk/v2/session"
@@ -54,8 +55,14 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
 	if cfg.SessionService != nil {
 		runnerConfig.SessionService = cfg.SessionService
 	}
+	logger := cfg.Logger.WithName("kagent-executor")
+	if usagePlugin, err := newTurnUsagePlugin(); err != nil {
+		logger.Error(err, "token usage aggregation is disabled")
+	} else {
+		runnerConfig.PluginConfig.Plugins = append([]*plugin.Plugin{usagePlugin}, runnerConfig.PluginConfig.Plugins...)
+	}
 	builtin := adka2a.NewExecutor(adka2a.ExecutorConfig{
-		RunnerProvider:     usageObservingRunnerProvider(runnerConfig),
+		RunnerConfig:       runnerConfig,
 		RunConfig:          runConfig,
 		A2APartConverter:   a2aPartConverter,
 		GenAIPartConverter: genAIPartConverter,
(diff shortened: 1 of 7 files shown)


QuentinBisson:
Rebased onto `main` (bc92b685). Re-verified locally after the rebase: `go build`, `go vet`, and the touched Go/Python test packages pass.

The three remaining red checks are not from this branch, they reproduce identically on `main` at the same commit (run [32034286380](https://github.com/kagent-dev/kagent/actions/runs/32034286380)):

- `go-unit-tests`: the `go/api/v1alpha2` and `go/api/v1alpha3` CEL tests fail in setup, not in assertions. `setup-envtest` cannot write the 1.36.2-linux-amd64 assets ("unable to create file kubectl/etcd from archive to disk"), so the control plane never starts. Looks like parallel test packages racing on the same `go/bin/k8s` store.
- `upgrade-tests (prev-stable)` / `rolling-upgrade-tests (prev-stable)`: `UPGRADE FAILED: context deadline exceeded` in `TestUpgrade/upgrade_with_helm`.

QuentinBisson:
@supreme-gg-gg this is clean against main and green. Could you take a look when you have a moment?

EItanya (commented):
🤖 AI-generated review.

Reviewed `013356fc`. I found one correctness issue and a simpler way to integrate this with ADK:

1. **Derive missing totals per call, before accumulation.** In both [Go](https://github.com/kagent-dev/kagent/blob/013356fc20bfb1a3c4b072f1da73306c942c8412/go/adk/pkg/a2a/usage.go#L167) and [Python](https://github.com/kagent-dev/kagent/blob/013356fc20bfb1a3c4b072f1da73306c942c8412/python/packages/kagent-adk/src/kagent/adk/_turn_usage.py#L83), the fallback only runs when the accumulated total is zero. An Anthropic-style execution derives and persists 120 tokens; after resuming and spending another 230 without a reported total, prompt/output counts reach 350 but `totalTokenCount` stays at 120. Mixing calls with and without reported totals also undercounts. Reproduced both cases in both languages.

2. **Use Go's existing `plugin.Config.OnEventCallback`.** It observes ADK events before A2A conversion, including the paused tool-call event. Registering it through `RunnerConfig.PluginConfig` lets us retain the default provider and remove the custom runner/provider and reflection test. I tested that replacement: the Go A2A package passes, including the HITL pause/resume test. The Python implementation already uses this approach.

3. **Upstream direction:** both ADKs expose per-call usage/hooks; current Python main also has private telemetry aggregation, though it is absent from this PR's locked version and isn't a public A2A task-total API. A public invocation-usage summary consumed by ADK's A2A adapters would be a useful upstream contribution, leaving a thin kagent metadata mapping. No need to wait for that to use the existing plugin hooks here.

[Reproduction sources, exact plugin experiment, commands, results and coverage limits](https://gist.github.com/EItanya/ce677b4ae7fba0c8cc480d850735b85c). Baseline Go package tests and 17 focused Python tests passed; four added regression probes fail as described. No live-model or database persistence tests were run.

EItanya:
Could we expose this through an optional, versioned A2A usage extension?

A2A doesn't currently define a built-in token-usage schema. Upstream discussed this exact question in https://github.com/a2aproject/A2A/issues/1155#issuecomment-3418586151 and recommended an opt-in extension. There's also a newer cost/budget proposal (https://github.com/a2aproject/A2A/issues/2121), but it remains a proposal rather than an adopted standard.

The existing `metadata["kagent_usage_total"]` is a valid use of A2A metadata. An extension would still use metadata, while making the schema and behavior discoverable and versioned. I'd suggest:

- Define a versioned kagent usage extension, advertise it in `AgentCard.capabilities.extensions` with `required: false`, and support A2A's client opt-in mechanism.
- Document the cumulative task-lifetime semantics: consumers replace earlier totals with the latest value, plus the behavior on resume, failure, and cancellation.
- Prefer provider-neutral fields such as `inputTokens` and `outputTokens` over making Google's `UsageMetadata` the public contract. If model attribution is included, associate counts with each model rather than attaching the last model's name to totals across multiple models.

The aggregation work here would still be useful; this would formalize its public output. A2A extension guidance: https://a2a-protocol.org/latest/topics/extensions/

QuentinBisson:
Agreed, let's make it an extension, following the same pattern as the HITL one: a versioned `kagent.dev` URI, advertised on the Agent Card as optional, and a doc page describing the schema and the cumulative task-lifetime semantics. Provider-neutral field names, and counts attributed per model rather than the last model's name on a combined total.

There are two ways to do it, and I'd like to confirm which you want.

The first is a data-only extension. The schema becomes discoverable and versioned, the value keeps being emitted as it is today, and no existing consumer breaks. This covers what you asked for.

The second also gates emission on client activation. That is what the upstream thread had in mind (https://github.com/a2aproject/A2A/issues/1155#issuecomment-3418586151): an agent that does not want to share its token counts can decline. The catch is that it only means something if the per-event `kagent_usage_metadata` keys move inside the extension too, since they carry strictly more information than the aggregate. That gives one way to read token usage instead of three, which is the better end state, but it moves every current reader at once: the UI, remote A2A sub-agent calls, and anything reading the 0.x `adk_usage_metadata` name.

I'd go with the first for now. Nobody has asked to withhold token counts, and if that need appears, or if an upstream usage extension gets adopted, gating can come as a v2 of the URI. To keep that door open cheaply I'd have our own clients send the activation header from the start, so a later gate is a no-op inside kagent and only affects third parties. The v1 spec would state plainly that the extension is data-only and always emitted.

One detail either way: the aggregate has to keep being computed regardless of activation, because a resumed task reads the running total back from the stored task.

Happy to do the second instead if you'd rather take the break now while the API version is already breaking.

PR #2472, opened 2026-08-18
fix(hitl): compose the same pause text in every runtime and always declare the extension

QuentinBisson:
The HITL extension is optional, and the architecture doc states what a client that does
not opt in should still get:

> a client that does not opt in can still receive an ordinary `input-required` task with
> human-readable text

Today that is one constant sentence in every runtime, so the tool name, the tool's hint
and the `ask_user` question are all lost. On the Go path an agent could also serve a card
that never declares the extension, leaving a client no way to discover it.

### The pause text

One function per runtime builds the text from the pending tools and the hints they
supplied:

| Pause | Text |
|---|---|
| `ask_user` asking "Which namespace?" | `Which namespace?` |
| `delete_file` with a tool hint | `Deleting this file requires approval (delete_file)` |
| `delete_file` with no hint | `Approval is required for tool(s): delete_file` |
| `delete_file` and `restart_pod`, one hint | `Deleting this file requires approval (delete_file, restart_pod)` |
| no confirmation call at all | `Human input is required before the agent can continue.` |

Every hint is kept rather than one of them winning: Go used to keep the last, Python the
first, so the same pause read differently per runtime. Tool names always accompany a
hint, so a pause on several tools no longer hides the ones the hint does not mention. The
text is identical whether or not the extension is activated, and an activated payload
repeats it in `hint`.

Two payload-visible changes: `tool_approval_request.hint` is now always populated, and a
LangGraph question interrupt emits an `ask_user_request` instead of an approval for a tool
named `ask_user`, which the resume side already handled.

Malformed input degrades instead of failing the task. A confirmation part with no usable
tool call, or a LangGraph `action_request` without a name or an id, is skipped, and a
pause left with no usable request stays text-only.

### The card declaration

`app.New` declares the extension whether or not an ADK agent was supplied for skill
derivation, matching what both Python builders already do in `build()`. `EnrichAgentCard`
keeps declaring it, so a caller using that exported function directly is unaffected.

The Go entrypoint replaced `Capabilities` wholesale to set `Streaming`, discarding the
extensions the control plane renders into `agent-card.json`. It now sets the one field.

### Not included

`kagent-crewai` and `kagent-openai` do not declare the extension. Neither has a pause
path: their executors only emit `submitted`, `working`, `completed` and `failed`, so
CrewAI's `human_input` and the Agents SDK's tool interruptions never reach A2A. Wiring
those runtimes up is worth doing separately.


supreme-gg-gg (commented):
Skimmed over the changes here and they make sense to me. When this is ready for review, I'll take another look. I think that Go ADK wasn't originally intended to be used as BYO, but it's good to fix the sharp edges 

EDIT: It would be helpful if you could rewrite the PR description a bit. I read it at first as “Go agents fail to declare the HITL extension,” which was confusing. What you actually mean is that Go only declares it when `cfg.Agent` is set for skill derivation, and you want that declaration to happen unconditionally, the same way Python does.

EItanya:
> ADK wasn't originally intended to be used as BYO,

This isn't true, we DEFINITELY want to support go for BYO

go/adk/pkg/a2a/hitl.go
```diff
@@ -429,14 +432,68 @@ func (tool confirmationTool) asHitlTool() HitlTool {
...
+				questions = append(questions, question)
+			}
+		}
+		return questions
+	}
+	return nil
+}
+
+// askUserQuestionText joins the questions an ask_user call is waiting on.
+func askUserQuestionText(tool HitlTool) string {
```
supreme-gg-gg, line 453: So this fixes direct `ask_user`, but nested-subagent path will still lose question since `RemoteHitlHint` reduces the child request to the tool name before `BuidlHITLStatusMesasge` runs. There's already another PR that contains fixes for this in #2475 but it has gone stale, can you add similar changes here?

QuentinBisson: You're right, the nested path still loses it. `RemoteHitlHint` builds the hint before `RequestConfirmation`, so by the time `BuildHITLStatusMessage` runs the question is already reduced to `ask_user`. The parent's confirmation tool is the subagent tool, not `ask_user`, so the branch here never fires.

#2475 looks ready rather than stale now: approved by you and EItanya, mergeable, DCO passing since Sep 2. Rather than duplicate it here, merging it first seems cleaner, then this PR picks up the nested question with no extra code. One thing to watch either way: both branches add a function named `askUserQuestionText` to package `a2a` with different signatures, so whichever merges second needs a rebase or the package fails to compile.

I can rebase on top of #2475, fold the two helpers into one, and add a test that covers the nested `ask_user` path.

supreme-gg-gg: #2475 has merged so this should be good to go, I'll take another look

python/packages/kagent-langgraph/src/kagent/langgraph/_executor.py
```diff
@@ -69,6 +71,14 @@ class LangGraphAgentExecutorConfig(BaseModel):
     enable_streaming: bool = True
 
 
+def _hitl_request(tools: list[HitlTool], text: str) -> AskUserRequest | ToolApprovalRequest:
+    """An ask_user interrupt becomes a question; anything else becomes an approval."""
+    if len(tools) == 1 and tools[0].name == "ask_user":
+        questions = tools[0].args.get("questions")
+        return AskUserRequest(id=tools[0].id, questions=questions if isinstance(questions, list) else [])
```
supreme-gg-gg: I think we cannot return an empty list of questions here, since `require_ask_user_response` in the resume path will reject an empty list of answers, or it will fail the answer count equals question count check, so effectively every resume becomes rejected. Maybe we should either skip empty questions and don't raise a `AskUserRequest` or just raise an error?

supreme-gg-gg (commented):
Hey @QuentinBisson this PR is in good shape to be merged, there's a minor issue in the langgraph code

QuentinBisson:
You're right, thanks for the review. I skipped when the list is empty to avoid turning a harmless case into a failure :)

PR #2474, opened 2026-08-18
fix: pair every tool call with a result before the model request

### LINKED ISSUE #2277: Interrupted turn permanently corrupts session history for Anthropic models (dangling tool_use leads to 400 on every later turn)

QuentinBisson:
## Summary

- pair every tool call with a result in a `before_model_callback`, in both the Go and Python runtimes, so an interrupted turn no longer leaves a `function_call` without a `function_response`
- describe a call ADK holds open for a human approval, `ask_user`, or a long-running tool as awaiting a response, and any other unanswered call as having no recorded result
- repair the request rather than the store, so recorded history stays intact, broken sessions heal on their next turn with no migration, and one implementation covers every session service and provider
- log when a request had to be repaired

Without this, the dangling call is replayed on every later turn and providers that require strict pairing reject the whole conversation (`tool_use ids were found without tool_result blocks immediately after`), leaving the session unusable until it is deleted.

The two placeholders differ because the difference changes what the model does next: told a tool returned nothing, it reissues the call or proceeds without it; told the call is still awaiting a response, it can wait. `long_running_tool_ids` lives on the event and does not survive the conversion to contents, so it is read from the session.

Pairing is checked positionally, against the immediately following content, because that is the invariant the provider enforces.

## Validation

- `uv run pytest packages/kagent-adk/tests/unittests` (395 passed)
- `go test ./adk/pkg/agent/ -count=1`
- `go build ./adk/...`, `go vet ./adk/pkg/agent/ ./adk/pkg/models/`, `gofmt -l`
- `uv run ruff check packages/kagent-adk`, `uv run ruff format --check`

19 Python and 13 Go tests cover the interrupted turn, the partially answered turn, the pending-approval case in both directions, calls with no id, and the no-op property on healthy history. The Python acceptance test converts repaired contents through google-adk's `content_to_message_param` and asserts the invariant the API enforces; both suites pair it with a guard test asserting the same check rejects unrepaired input.

## Notes

- Python's Anthropic and Bedrock paths and Go's `bedrock.go` had no pairing repair; the other converters already have their own copy and are left in place, since they find a result present once contents are paired.
- ADK removes an orphaned `function_response` before this callback runs and recovers the compaction case itself, so orphan handling is deliberately not duplicated here.
- ADK's per-request content copies are shallow in Python, so this code only assigns `Content.parts` or appends new `Part` objects. Recorded in the module docstring.

Fixes #2277


supreme-gg-gg:
I haven't looked at the code yet, but this seems like an upstream responsibility at the framework level (at least in the long run). Have you checked if there's any discussion / issue / PR relating to this in either python or go ADK?

QuentinBisson:
@supreme-gg-gg you're right.

I asked my agent to take a look and this has been raised upstream several times. google/adk-python#5856 is the same failure (orphaned function_call after interruption, 400 from Claude's strict pairing); an ADK maintainer replied that the recommended approach is to handle it in the callback feature, then the issue was stale-closed. Generic healing PRs in core were closed unmerged (google/adk-python#4055, #4056, #6587), and related pairing fixes are still churning (#6733, #6752, #6764). The Go side has had its own rearrangement bugs (google/adk-go#761).

If you prefer that I close this PR in favor or fixing it the the adk directly that's totally fine with me :)

QuentinBisson:
@supreme-gg-gg fixed upstream in both runtimes, at the level where the pairing invariant belongs:

- google/adk-python#6914 — pairs the calls in `_get_contents`, next to the existing `_drop_orphaned_function_responses` (the mirror case)
- google/adk-go#1405 — same in `buildContentsDefault`

Both drafts, and both drop the callback in favour of a single step in contents assembly, so it covers every session service, every provider and both entry points (`include_contents` default and none).

Keeping this PR open until an ADK release carries the fix. Happy to close it then, or now if you prefer to wait upstream.

QuentinBisson:
@supreme-gg-gg Thank you for your time reviewing this PR. adk-python is taking care of fixing this https://github.com/google/adk-python/pull/6914#issuecomment-5594451733 and I asked the adk-go maintainers if we can do the same so keeping this in kagent does not make much sense. I think closing this PR is the best way forward :)

PR #2354, opened 2026-07-28
A2A: Migrate to v1.0 across all integration surfaces

supreme-gg-gg:
## Summary

Close #1941 

This PR combines the changes of #1929 and #1936 and migrates all Kagent's A2A integration surfaces (UI a2a-js client, CLI a2a-go client, DB and REST API interfaces, declarative ADK in go and python, BYO frameworks in python) from 0.3 and 1.0 dual-compatible (introduced in #1921) to strictly v1.0. This paves the way for future work to consolidate our A2A integration and develop the Kagent v1 APIs.

**This migration is not zero downtime and is not backward compatible, and is intended to go into the next major version.** There is no compatibility matrix for this. Your UI/CLI/custom client and all agents must **ALL** be in v1.0 to work properly.

## Testing

Unit, integration, and E2E tests have been updated and ran successfully.
I have manually tested: using UI and CLI with Go declarative, Python declarative, and BYO agents on K8s and Substrate.

For context, the A2A migration doc is [here](https://github.com/a2aproject/a2a-python/tree/main/docs/migrations/v1_0) (written for python sdk but generally applicable to go and js)

PR #2475, opened 2026-08-18
fix(go-adk): surface the sub-agent's ask_user question in RemoteHitlHint

### LINKED ISSUE #2473: [BUG] RemoteHitlHint drops the sub-agent's actual ask_user question, only names the tool
### 📋 Prerequisites

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

When a delegated sub-agent (via a Go declarative `Agent` tool, `remote_a2a_tool.go`) pauses on its own `ask_user` call, the parent agent correctly bubbles this up as its own `input-required` confirmation (via `ctx.RequestConfirmation`), so a downstream A2A client (e.g. a chat bridge) does get a real `adk_request_confirmation` card to show the human. However, the **hint text on that confirmation never includes the sub-agent's actual question** — only a generic "requires approval for tool(s): ask_user" message. The human sees a card with no way to know what's actually being asked.

**Root cause**: `go/adk/pkg/a2a/hitl.go`, `RemoteHitlHint()`:

```go
func RemoteHitlHint(state *RemoteHitlState) string {
	if state == nil {
		return "Remote agent requires human input before continuing."
	}
	tools := VisibleTools(state.ToolApprovalRequest, state.AskUserRequest)
	names := make([]string, 0, len(tools))
	for _, tool := range tools {
		if tool.Name != "" {
			names = append(names, tool.Name)
		}
	}
	if len(names) > 0 {
		return fmt.Sprintf("Remote agent '%s' requires approval for tool(s): %s",
			state.SubagentName, strings.Join(names, ", "))
	}
	return fmt.Sprintf("Remote agent '%s' requires human input before continuing.", state.SubagentName)
}
```

`RemoteHitlState.AskUserRequest` (populated by `BuildRemoteHitlState` via `GetAskUserRequest`) already carries the real question data in `AskUserRequest.Questions []map[string]any` — confirmed directly from the sub-agent's own ADK session (`GET /api/sessions/<subagent_session_id>`), which shows the actual `adk_request_confirmation` payload with the real question text sitting right there. But `RemoteHitlHint()` only walks `VisibleTools(...)` to collect tool *names* — it never reads `AskUserRequest.Questions` — so the hint is structurally incapable of surfacing the question, regardless of what the sub-agent asked.

The tool's own return value passed back into the parent LLM's context (`remoteA2AResponse{Status: "pending", WaitingFor: "subagent_approval", ...}` in `handleInputRequired`, `remote_a2a_tool.go`) is equally opaque — no hint, no question, just a session id. So even the parent LLM has no way to relay the real question in its own words; it can only reference a generic "the subagent needs approval" state.

### 🔄 Steps To Reproduce

1. Set up a coordinator agent (Go declarative runtime) with a sub-agent as an `Agent` tool, where the sub-agent has the built-in `ask_user` tool available.
2. Ask the coordinator to delegate a task to the sub-agent that requires it to ask a clarifying question (e.g. delegate a GitHub search where the target repo can't be found and the sub-agent needs to ask for the exact owner/org).
3. Observe the confirmation card/message surfaced to the human via A2A `input-required` (or in kagent-ui).

### 🤔 Expected Behavior

The confirmation hint (and ideally the tool's return payload) should include the sub-agent's actual `ask_user` question text, e.g. `"github_agent asks: What is the GitHub username or organization that owns the 'sre-infrastructure' repo?"` — not just the tool name.

### 📱 Actual Behavior

Hint reads only: `"Remote agent 'kagent__NS__github_agent' requires approval for tool(s): ask_user"`. The real question ("What is the GitHub username or organization that owns the 'sre-infrastructure' repo? ...") is discarded before it ever reaches the parent agent or any downstream A2A client — confirmed only by directly querying the sub-agent's own session via the controller's internal REST API, bypassing the normal A2A flow entirely.

This reproduces the spirit of #1475 ("Subagents get stuck when using human in the loop") in a narrower form: propagation now happens, but the payload is stripped of the one piece of information a human actually needs to answer it.

### 💻 Environment

- Go declarative runtime (`go/adk`), observed via `k8s-a2a-agent` → `github_agent` delegation (`kagent__NS__github_agent` Agent tool) with the built-in `ask_user` tool on the sub-agent.
- Discovered while building an external Slack ↔ A2A bridge that correctly renders every other HITL flow (tool approval, direct `ask_user`) from the same session — this is specifically about the nested-sub-agent case losing question content in `RemoteHitlHint`.

### 🔍 Additional Context

Suggested fix: in `RemoteHitlHint()`, when `state.AskUserRequest` is set, include the actual question text(s) from `state.AskUserRequest.Questions` in the hint instead of (or in addition to) the tool name. A fix along these lines has been implemented and submitted as #2475, verified against the full `go/adk` test suite.

vramahandry:
## Summary

`RemoteHitlHint()` builds the hint text shown to a human when a sub-agent's own HITL pause bubbles up to the parent agent. Today it only lists the paused tool's *name* (e.g. `ask_user`) — never the actual question — even though `AskUserRequest.Questions` (already threaded through `VisibleTools()` into `HitlTool.Args["questions"]`) has the real content right there.

- Before: `"Remote agent 'github_agent' requires approval for tool(s): ask_user"`
- After: `"Remote agent 'github_agent' asks: What is the GitHub owner/org for the repo?"`

Real tool-approval hints (the non-`ask_user` case) are unaffected — same wording as before, existing `TestBuildRemoteHitlStateAndHint` still passes unchanged.

Discovered while building an external A2A chat bridge that correctly renders every other kagent HITL flow (direct tool approval, direct `ask_user`) from the same session — this was the one case where the confirmation card reaching a downstream A2A client carried no actionable information, because the hint itself never had it. Confirmed by inspecting the sub-agent's own ADK session directly (`GET /api/sessions/<id>`), where the real question is visible in `adk_request_confirmation` — it's just discarded before reaching `RemoteHitlHint`.

Fixes #2473

## Changes

- `go/adk/pkg/a2a/hitl.go`: new `askUserQuestionText()` helper extracts question text from an `ask_user` `HitlTool`'s args; `RemoteHitlHint()` uses it when present, falling back to the existing tool-name-only wording otherwise.
- `go/adk/pkg/a2a/hitl_test.go`: new `TestBuildRemoteHitlStateAndHintAskUser` covering the fixed behavior.

## Test plan

- [x] `go build ./adk/...` — clean
- [x] `go test ./adk/...` — all packages pass (a2a, agent, app, tools, etc.)
- [x] `gofmt -l` / `go vet ./adk/pkg/a2a/...` — clean
- [x] Existing `TestBuildRemoteHitlStateAndHint` (tool-approval case) passes unchanged

### DIFF
--- a/go/adk/pkg/a2a/hitl.go
+++ b/go/adk/pkg/a2a/hitl.go
@@ -361,10 +361,7 @@ func VisibleTools(approval *ToolApprovalRequest, ask *AskUserRequest) []HitlTool
 	return nil
 }
 
-// askUserQuestionText extracts the human-readable question text(s) from an
-// ask_user request's typed Questions field, or "" if none carry text. This is
-// the content RemoteHitlHint would otherwise discard, surfacing only the
-// tool name ("ask_user") instead of what is actually being asked.
+// askUserQuestionText joins the question text from a typed Questions field, or "" if none carry text.
 func askUserQuestionText(questions []map[string]any) string {
 	texts := make([]string, 0, len(questions))
 	for _, q := range questions {
@@ -379,12 +376,8 @@ func RemoteHitlHint(state *RemoteHitlState) string {
 	if state == nil {
 		return "Remote agent requires human input before continuing."
 	}
-	// AskUserRequest.Questions carries the real question text whether or not
-	// Nested is set (see BuildHITLStatusMessage), so read it directly rather
-	// than re-deriving from VisibleTools()'s nested HitlTool.Args: those args
-	// round-trip through JSON into a plain map[string]any, which decodes
-	// "questions" as []any rather than []map[string]any, silently losing the
-	// question in the nested case.
+	// Read Questions directly: VisibleTools()'s nested Args round-trip through
+	// JSON to []any, silently losing the question text in the nested case.
 	if state.AskUserRequest != nil {
 		if q := askUserQuestionText(state.AskUserRequest.Questions); q != "" {
 			return fmt.Sprintf("Remote agent '%s' asks: %s", state.SubagentName, q)
--- a/go/adk/pkg/a2a/hitl.go
+++ b/go/adk/pkg/a2a/hitl.go
@@ -362,14 +362,12 @@ func VisibleTools(approval *ToolApprovalRequest, ask *AskUserRequest) []HitlTool
 }
 
 // askUserQuestionText extracts the human-readable question text(s) from an
-// ask_user HitlTool's args (see VisibleTools), or "" if the shape is
-// unexpected. This is the content RemoteHitlHint would otherwise discard,
-// surfacing only the tool name ("ask_user") instead of what is actually
-// being asked.
-func askUserQuestionText(args map[string]any) string {
-	raw, _ := args["questions"].([]map[string]any)
-	texts := make([]string, 0, len(raw))
-	for _, q := range raw {
+// ask_user request's typed Questions field, or "" if none carry text. This is
+// the content RemoteHitlHint would otherwise discard, surfacing only the
+// tool name ("ask_user") instead of what is actually being asked.
+func askUserQuestionText(questions []map[string]any) string {
+	texts := make([]string, 0, len(questions))
+	for _, q := range questions {
 		if text, ok := q["question"].(string); ok && text != "" {
 			texts = append(texts, text)
 		}
@@ -381,25 +379,23 @@ func RemoteHitlHint(state *RemoteHitlState) string {
 	if state == nil {
 		return "Remote agent requires human input before continuing."
 	}
-	tools := VisibleTools(state.ToolApprovalRequest, state.AskUserRequest)
-	names := make([]string, 0, len(tools))
-	questions := make([]string, 0, len(tools))
-	for _, tool := range tools {
-		if tool.Name == "" {
-			continue
-		}
-		names = append(names, tool.Name)
-		if tool.Name == "ask_user" {
-			if q := askUserQuestionText(tool.Args); q != "" {
-				questions = append(questions, q)
-			}
+	// AskUserRequest.Questions carries the real question text whether or not
+	// Nested is set (see BuildHITLStatusMessage), so read it directly rather
+	// than re-deriving from VisibleTools()'s nested HitlTool.Args: those args
+	// round-trip through JSON into a plain map[string]any, which decodes
+	// "questions" as []any rather than []map[string]any, silently losing the
+	// question in the nested case.
+	if state.AskUserRequest != nil {
+		if q := askUserQuestionText(state.AskUserRequest.Questions); q != "" {
+			return fmt.Sprintf("Remote agent '%s' asks: %s", state.SubagentName, q)
 		}
 	}
-	// A real ask_user question is more useful to the human than the bare
-	// tool name — surface it directly rather than "requires approval for
-	// tool(s): ask_user", which carries no actionable information.
-	if len(questions) > 0 {
-		return fmt.Sprintf("Remote agent '%s' asks: %s", state.SubagentName, strings.Join(questions, " "))
+	tools := VisibleTools(state.ToolApprovalRequest, state.AskUserRequest)
+	names := make([]string, 0, len(tools))
+	for _, tool := range tools {
+		if tool.Name != "" {
+			names = append(names, tool.Name)
+		}
 	}
 	if len(names) > 0 {
 		return fmt.Sprintf("Remote agent '%s' requires approval for tool(s): %s",
--- a/go/adk/pkg/a2a/hitl.go
+++ b/go/adk/pkg/a2a/hitl.go
@@ -361,16 +361,45 @@ func VisibleTools(approval *ToolApprovalRequest, ask *AskUserRequest) []HitlTool
 	return nil
 }
 
+// askUserQuestionText extracts the human-readable question text(s) from an
+// ask_user HitlTool's args (see VisibleTools), or "" if the shape is
+// unexpected. This is the content RemoteHitlHint would otherwise discard,
+// surfacing only the tool name ("ask_user") instead of what is actually
+// being asked.
+func askUserQuestionText(args map[string]any) string {
+	raw, _ := args["questions"].([]map[string]any)
+	texts := make([]string, 0, len(raw))
+	for _, q := range raw {
+		if text, ok := q["question"].(string); ok && text != "" {
+			texts = append(texts, text)
+		}
+	}
+	return strings.Join(texts, " ")
+}
+
 func RemoteHitlHint(state *RemoteHitlState) string {
 	if state == nil {
 		return "Remote agent requires human input before continuing."
 	}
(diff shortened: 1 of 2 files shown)


go/adk/pkg/a2a/hitl.go
```diff
@@ -361,17 +361,46 @@ func VisibleTools(approval *ToolApprovalRequest, ask *AskUserRequest) []HitlTool
 	return nil
 }
 
+// askUserQuestionText extracts the human-readable question text(s) from an
+// ask_user HitlTool's args (see VisibleTools), or "" if the shape is
+// unexpected. This is the content RemoteHitlHint would otherwise discard,
+// surfacing only the tool name ("ask_user") instead of what is actually
+// being asked.
+func askUserQuestionText(args map[string]any) string {
+	raw, _ := args["questions"].([]map[string]any)
```
supreme-gg-gg: Instead of type-asserting `[]maps[string]any` here, can you read `state.AskUserRequest.Questions` which already has the questions text and is always typed, including when `Nested` is set?  These args can become `[]any` so a nested ask_user would be missed (see how `BuildHITLStatusMesasge` handles it both cases)

go/adk/pkg/a2a/hitl_test.go
```diff
@@ -260,3 +260,29 @@ func TestBuildRemoteHitlStateAndHint(t *testing.T) {
 		t.Fatalf("hint = %q", got)
 	}
 }
+
+// A sub-agent's ask_user pause should surface the actual question in the
+// hint, not just "requires approval for tool(s): ask_user" — a human can't
+// act on a tool name alone.
+func TestBuildRemoteHitlStateAndHintAskUser(t *testing.T) {
```
supreme-gg-gg, line 265: Also add a test where `AskUserReqest.Nested` is populated so we can verify two-level paths also work

supreme-gg-gg (changes requested):
This fix makes sense to me, two comments and this is ready to go

go/adk/pkg/a2a/hitl.go
vramahandry: Done in 9c3ac9e — RemoteHitlHint now reads state.AskUserRequest.Questions directly (typed correctly in both the direct and nested case per BuildHITLStatusMessage) instead of type-asserting the nested tool's Args.

go/adk/pkg/a2a/hitl_test.go
vramahandry: Added TestBuildRemoteHitlStateAndHintAskUserNested in 9c3ac9e, which populates AskUserRequest.Nested and asserts the nested tool's Args["questions"] actually decodes as []any (not []map[string]any) after the JSON round-trip, then checks the hint still surfaces the real question text.

go/adk/pkg/a2a/hitl.go
```diff
@@ -361,10 +361,35 @@ func VisibleTools(approval *ToolApprovalRequest, ask *AskUserRequest) []HitlTool
...
+		}
+	}
+	return strings.Join(texts, " ")
+}
+
 func RemoteHitlHint(state *RemoteHitlState) string {
 	if state == nil {
 		return "Remote agent requires human input before continuing."
 	}
+	// AskUserRequest.Questions carries the real question text whether or not
```
supreme-gg-gg: nit: this inline comment seems a bit too long (but not worth blocking on this)

supreme-gg-gg (dismissed):
lgtm, thanks!

go/adk/pkg/a2a/hitl.go
vramahandry: Good catch, done in 9c3ac9e9 — `RemoteHitlHint` now reads `state.AskUserRequest.Questions` directly instead of re-deriving from `VisibleTools()`'s nested `HitlTool.Args`, so the `[]any` round-trip issue you described no longer applies.

go/adk/pkg/a2a/hitl_test.go
vramahandry: Added in 9c3ac9e9 — `TestBuildRemoteHitlStateAndHintAskUserNested` covers a two-level nested `ask_user` pause (grandchild agent relayed through the child) and asserts the hint still surfaces the real question.

go/adk/pkg/a2a/hitl.go
vramahandry: Trimmed in 40bec6ba, along with the similarly verbose comments added alongside it.

supreme-gg-gg (approved):

supreme-gg-gg:
@vramahandry this commit is not properly signed off: 40bec6ba35e5a4eda888aa9ade999067976a5360 you need to do that for DCO to pass

vramahandry:
Hello @supreme-gg-gg , sorry I forgot the DCO.
Now it seems ok. Thank you !

EItanya (approved):

vramahandry:
Hello @supreme-gg-gg is there any issue of not being merged ?

vramahandry:
Hello @EItanya do you know when it will be released to a tagged version ?

EItanya:
> Hello @EItanya do you know when it will be released to a tagged version ?

If you need this on 0.10 you will need to backport it. Hopefully we'll be starting beta releases on main early next week

vramahandry:
> > Hello @EItanya do you know when it will be released to a tagged version ?
> 
> If you need this on 0.10 you will need to backport it. Hopefully we'll be starting beta releases on main early next week

How I can backport myself ? Do I create a new tag on my fork from the latest tag v0.10.1 ?

EItanya:
> > > Hello @EItanya do you know when it will be released to a tagged version ?
> > 
> > 
> > If you need this on 0.10 you will need to backport it. Hopefully we'll be starting beta releases on main early next week
> 
> How I can backport myself ? Do I create a new tag on my fork from the latest tag v0.10.1 ?

Open a PR against `release-v0.10.x`

vramahandry:
Hello @EItanya,
Looked into backporting this (and #2495) to `release/v0.10.x`. Not straightforward, unfortunately: both fixes sit on top of #2396 (A2A HITL extension rewrite), which itself depends on #2354 (A2A v1.0 migration) — neither of which was ever backported to `release/v0.10.x`. That branch still runs the pre-extension HITL model (`adk_request_confirmation` DataParts), so `_hitl.py` / the current `hitl.go` don't exist there in a form these commits can apply to.

Tried cherry-picking #2396 alone onto `release/v0.10.x` as a test — 20+ conflicting files across Go, Python, and the UI (agent executor, remote A2A tool, event converter, the whole chat UI). It's a real architecture migration, not a mechanical backport.

If this is needed on 0.10.x, the options are either porting the A2A v1.0 + HITL extension migration itself (multi-day effort, real risk to the release branch), or a small hand-written fix against the old `adk_request_confirmation` hint logic that gets the same user-visible result without the migration.

vramahandry:
Hello @supreme-gg-gg @EItanya ,
Can this pr will be on 1.0.0 release ?

Issue #2464, opened 2026-08-17
[BUG] Exact A2A GetTask cannot retrieve controller-persisted tasks


erauner12:
## Scope clarification

This issue now tracks exact A2A GetTask retrieval of controller-persisted tasks.

The separately reproduced runtime identity defect affecting persisted continuation is tracked in:

- [#2465](https://github.com/kagent-dev/kagent/issues/2465) — runtime task-store callbacks lose authenticated user context before task resolution.

A separately suspected controller-hop defect was tracked in #2467, but it was not independently reproduced and is now closed.

The original reproduction evidence below is retained because it exposed the exact-read defect tracked here and the runtime identity defect in #2465.

### 📋 Prerequisites

- [x] I have searched the [existing issues](https://github.com/kagent-dev/kagent/issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

Tasks created and persisted through the kagent controller A2A endpoint are
visible through store-backed `ListTasks` and through an authenticated REST
`GET /api/tasks/{id}`, but they cannot be resolved through exact A2A
`GetTask` or used to resume a persisted `input-required` task.

The failure is independent of the A2A wire version. It reproduced across all
four create/resume combinations:

- A2A 0.3 → A2A 0.3
- A2A 0.3 → A2A 1.0
- A2A 1.0 → A2A 0.3
- A2A 1.0 → A2A 1.0

The apparent cause is an ownership/identity mismatch. The controller persists
the task under the authenticated initiating principal, while the runtime later
performs its REST task lookup using the runtime service account. Because task
access is owner-scoped, that lookup reports that the existing task was not
found.

This prevents A2A clients from safely resuming controller-routed human-in-the-loop
interactions. Re-sending the original request is not a safe workaround because
it creates a new task and could duplicate an already accepted action.

### Scope clarification

The defects are now tracked separately:

- #2465 covers authenticated user-context propagation in the current
  controller-v2/runtime path.
- This issue continues to track exact task retrieval and persisted continuation
  through the legacy controller path, for any maintained release confirmed to
[...]

PR #2483, opened 2026-08-18
fix(adk): preserve A2A user identity before task resolution

### LINKED ISSUE #2465: [BUG] A2A runtime task-store callbacks lose authenticated user context before task resolution
## Scope clarification for `release/v0.10.x`

This issue tracks the same pre-task-resolution identity contract in both maintained v0.10 runtime implementations:

- the Go call interceptor must place the authenticated user in the returned `context.Context` before task-store callbacks run; and
- the Python task store must scope the user from `ServerCallContext` around owner-scoped controller calls without leaking it into a later headerless request.

The original source-level reproduction below describes the Go seam because that was the first implementation in which the defect was isolated. The Python seam is the equivalent runtime contract and belongs in the same focused fix. The suspected controller-hop defect #2467 was not independently reproduced and is closed; exact persistent task reads remain separate in #2464.
### 📋 Prerequisites

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

The Go A2A runtime extracts the authenticated user from inbound `x-user-id`
request metadata and stores it in `CallContext.User`, but the call interceptor
returns the original request context without that identity.

Pre-executor task resolution and task-store persistence callbacks obtain the
user from the returned `context.Context`, not from `CallContext.User`. Those
callbacks can therefore run without the authenticated request identity and fall
back to the runtime service identity for owner-scoped task access.

This is especially significant for persisted `input-required` tasks. A later
A2A continuation is a new request and may be handled by a different process, so
each request must reconstruct and propagate its authenticated identity before
stored-task resolution occurs.

### 🔄 Steps To Reproduce

1. On current `main`, construct an A2A `CallContext` whose service parameters
   contain a non-empty `x-user-id` value.
2. Call `UserIDCallInterceptor().Before(...)` with that context.
3. Observe that `CallContext.User` contains the authenticated user.
4. Inspect the returned `context.Context` with `auth.UserIDFromContext(...)`.
5. Observe that the returned context does not contain the user identity.
6. Source inspection shows that when the returned context is subsequently passed
   to `KAgentTaskStore`, `ControllerClient.CallContext` cannot add `x-user-id` to
   the owner-scoped controller call.

### 🤔 Expected Behavior

The A2A call interceptor should preserve the authenticated request identity in
both `CallContext.User` and the returned `context.Context`.

Pre-executor task resolution and persistence callbacks should therefore access
the authoritative task store using the same authenticated identity as the
current A2A request.

### 📱 Actual Behavior

Only `CallContext.User` is populated. The returned request context is unchanged,
so `auth.UserIDFromContext(returnedCtx)` is empty before executor processing.

Task-store callbacks that run before the executor attaches identity can
therefore use the runtime service identity rather than the authenticated A2A
request identity.

### 💻 Environment

- OS and version: Not environment-specific
- Kubernetes version/provider: Not applicable to the source-level reproduction
- Browser: Not applicable
- Application version: current `main` inspected at `bc92b6850b4e1342fd6b9483961aa163ece94836`
- Affected component: Go ADK A2A runtime and controller-backed task store

### 🔧 CLI Bug Report

Not applicable. The issue is deterministically reproduced at the interceptor
and task-store context boundary.

### 🔍 Additional Context

The existing components already support the required behavior:

- controller-v2 carries upstream authentication headers in A2A service parameters;
- the runtime interceptor reads `x-user-id` and populates `CallContext.User`;
- controller clients emit `x-user-id` when the user identity is present in their
  input context;
- task reads and writes are owner-scoped.

The missing connection is preserving the authenticated user in the context
returned by the call interceptor before task resolution and persistence
callbacks execute. This does not require a new identity source, authentication
mode, header, or trust policy.

This issue isolates the current runtime identity-lifecycle defect from the
legacy controller task-query behavior tracked in #2464.

Focused regression coverage should assert that the interceptor populates both
`CallContext.User` and `auth.UserIDFromContext(returnedCtx)` from the same
inbound authenticated request.

### 📋 Logs

```shell
CallContext.User.Name: <authenticated-user>
auth.UserIDFromContext(returnedCtx): ""
```

### 📷 Screenshots

_No response_

erauner12:
## Summary

Preserve the authenticated A2A user before task-store resolution in the Python runtime on `release/v0.10.x`.

Fixes #2465.

## Problem

A persisted continuation is a new A2A request and may be handled by a different process. Task resolution can run before executor-local identity setup, so task-store callbacks must receive the authenticated request identity at the runtime boundary.

Without that identity, owner-scoped reads can fall back to the runtime service account and report an existing task as not found.

## Change

- Python task-store callbacks derive the user from `ServerCallContext`, scope it around controller calls, restore prior context afterward, and clear stale request identity when a later request has no forwarded user.

## Trust boundary

This introduces no new identity source. The Python runtime consumes the existing authenticated A2A service parameter; it does not trust arbitrary task metadata or invent an owner.

## Tests

- `cd python && uv run pytest packages/kagent-core/tests/test_task_store.py packages/kagent-core/tests/test_request_context.py`

Coverage verifies pre-resolution visibility, scoped restoration after success and failure, that a headerless scoped callback observes no user while an outer `ContextVar` is restored afterward, and no identity leakage from a user-bearing request into a later headerless request.

## Scope and ordering

This PR changes Python runtime identity ingestion only. The superseding Go ADK change is handled by #2507. Persistent exact `GetTask` remains the separate issue tracked in #2464.


erauner12:
### Coordination and forward-port note

#2485 is the controller-side counterpart to this runtime-boundary change and should merge after this PR. #2484 complements both with persistent exact reads but is independently mergeable.

A main-oriented prototype is available in erauner12/kagent#11. A forward-port should preserve this PR's identity-lifecycle invariants while adapting to main's newer boundaries: the Go runtime uses the `a2a-go/v2` interceptor API, and the Python runtime uses controller-client call options rather than the v0.10 HTTP hook. In particular, identity must be visible before task resolution, scoped to the request, restored afterward, and cleared for a later headerless request.

Combined validation on a compatible v0.10 build completed a fresh-process `input-required` resume on the retained task/context without resending the original request.

python/packages/kagent-core/src/kagent/core/a2a/_context.py
```diff
@@ -15,3 +19,25 @@ def set_request_user_id(user_id: str | None) -> None:
...
+        return None
+    headers = context.state.get("headers", {})
+    user_id = headers.get("x-user-id")
+    return user_id if isinstance(user_id, str) and user_id else None
+
+
+@contextmanager
+def scoped_request_user_id(user_id: str | None) -> Iterator[None]:
+    """Temporarily expose a scoped user to controller HTTP request hooks."""
+    if not user_id:
```
supreme-gg-gg: Why does it preserve the existing context var when the current server call context has no user? I would expect the call context to be authoritative

supreme-gg-gg (commented):
Thanks for reporting this issue and opening this fix, this sounds reasonable to me, left one small comment. We also have a PR doing this for main branch

python/packages/kagent-core/src/kagent/core/a2a/_context.py
erauner12: Agreed, the current `ServerCallContext` should be the source of truth inside the task-store callback.

The early return for a missing user leaves the outer `ContextVar` value in place, which could let a headerless request inherit stale identity. I’ll change the scope so it sets `None` for the callback and resets the token afterward.

I’ll also update the regression test to confirm that the headerless callback sees no user, and that the outer value is restored once the callback finishes.

I reviewed #2449 as well. Its Go change enforces the same contract on `main`; this PR is for `release/v0.10.x` and also covers the matching Python runtime boundary. I’ll keep the Go implementation aligned with #2449 while preserving the release-specific and Python coverage.

erauner12 (commented):
left response

python/packages/kagent-core/src/kagent/core/a2a/_context.py
erauner12: Updated in `d75a67f6`.

`scoped_request_user_id` now always installs the current call-context value, including `None`. The regression test verifies that a headerless callback observes no user while the outer `ContextVar` value is restored after the scope exits.

Focused validation passed:

- `pytest packages/kagent-core/tests/test_request_context.py packages/kagent-core/tests/test_task_store.py` — 4 passed
- `go test ./adk/pkg/a2a` — passed

The branch was also updated from the latest `release/v0.10.x` without rewriting the reviewed commits.

supreme-gg-gg (commented):
Hi @erauner12 thanks for the change, this lgtm but can you remove the Go ADK changes given that #2507 is going to merge soon (which adds the Go part to 0.10.x)

erauner12:
@supreme-gg-gg , will do!

mesutoezdil:
description still says go call interceptor changed, and test plan lists go test ./adk/pkg/a2a. last commit here removed the go changes as superseded. diff is python only now. probably just needs the description and test plan updated to match.

erauner12:
@mesutoezdil  Thanks for catching that. I’ve updated the PR description and test plan to reflect the Python-only diff, removed the stale Go references, and noted that the Go work is covered by #2507. Ready for another look when you have a chance!

adalundhe (approved):

onematchfox:
@EItanya See a `v0.10.0` [tag](https://github.com/kagent-dev/kagent/releases/tag/v0.10.0) has been created so I'm guessing the official release is incoming... .FYI: Python agents are pretty broken on `v0.10.x` without this fix.

supreme-gg-gg (approved):

EItanya:
> @EItanya See a `v0.10.0` [tag](https://github.com/kagent-dev/kagent/releases/tag/v0.10.0) has been created so I'm guessing the official release is incoming... .FYI: Python agents are pretty broken on `v0.10.x` without this fix.

I can do a v0.10.1

PR #2484, opened 2026-08-18
fix(a2a): serve exact task reads from the persistent store

### LINKED ISSUE #2464: [BUG] Exact A2A GetTask cannot retrieve controller-persisted tasks

erauner12:
## Summary

Serve exact A2A `GetTask` requests from the controller's authoritative, owner-scoped persistent task store on `release/v0.10.x` without delegating to the managed runtime.

Fixes #2464.

## Problem

The controller already uses persistent storage for `ListTasks`, but exact `GetTask` continued through the passthrough runtime. A persisted task could therefore appear in `ListTasks` while exact retrieval returned `task_not_found`.

## Change

`storeTaskQueryHandler.GetTask` now:

- queries the persistent store with the authenticated caller identity, or an empty identity when none is available;
- returns a shaped copy with the requested recent-history bound and retained artifacts;
- maps caller-scoped `database.ErrNotFound` to A2A `ErrTaskNotFound` without delegation; and
- propagates other storage failures for the A2A handler to return with internal-error semantics.

## Authorization boundary

Exact `GetTask` remains scoped to the authenticated caller. It does not substitute `ShareContext.UserID`: a task-ID-only request cannot prove that the requested task belongs to the shared session. Missing identity and caller-scoped misses return `task_not_found` without invoking the runtime, preserving the store's authorization boundary.

## Why the controller store

The same persistent store already serves task listing and receives runtime persistence callbacks. Making it fully authoritative for exact reads keeps persisted task existence consistent across the A2A query surface and avoids a redundant runtime lookup against the same database.

## Tests

- `cd go && go test ./core/internal/a2a`
- `cd go && go test -race ./core/internal/a2a -run '^(TestGetTask_|TestWire_GetTaskStoreErrorSemantics)'`

Coverage includes persistent hits, absent identity, other-owner and not-found results, backend failures, dual-wire A2A error mapping, requested history shaping, artifact retention, non-mutation, and the share-context boundary above. Every exact-read path asserts that the delegate is not invoked.

## Validation

An independent A2A client against a compatible v0.10 deployment retrieved the same persisted task through exact `GetTask` on both supported 0.3 and 1.0 wires without a `ListTasks` fallback.

This PR fixes exact reads only. Runtime identity propagation is tracked separately in #2465.

### DIFF
--- a/go/core/internal/a2a/task_query_store.go
+++ b/go/core/internal/a2a/task_query_store.go
@@ -4,6 +4,7 @@ import (
 	"cmp"
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"slices"
 	"strconv"
@@ -19,19 +20,19 @@ const (
 	maxTaskPageSize     = 100
 )
 
-// TaskStore is the subset of the persistent store ListTasks reads from.
+// TaskStore is the subset of the persistent store task queries read from.
 // *database.Client satisfies it. GetSession errors (including a missing or
 // other-user session) surface to the caller.
 type TaskStore interface {
+	GetTask(ctx context.Context, taskID, userID string) (*a2atype.Task, error)
 	GetSession(ctx context.Context, sessionID, userID string) (*dbpkg.Session, error)
 	ListSessions(ctx context.Context, userID string) ([]dbpkg.Session, error)
 	ListTasksForSession(ctx context.Context, sessionID, userID string) ([]*a2atype.Task, error)
 }
 
-// storeTaskQueryHandler answers ListTasks from kagent's task store, which is
-// the source of truth for persisted tasks. Every other method (including
-// GetTask, which already resolves to the same store via the passthrough) is
-// delegated to the embedded handler unchanged.
+// storeTaskQueryHandler answers GetTask and ListTasks from kagent's task
+// store, which is the source of truth for persisted tasks. Every other method
+// is delegated to the embedded handler unchanged.
 type storeTaskQueryHandler struct {
 	a2asrv.RequestHandler
 	store TaskStore
@@ -39,6 +40,22 @@ type storeTaskQueryHandler struct {
 
 func newStoreTaskQueryHandler(delegate a2asrv.RequestHandler, store TaskStore) *storeTaskQueryHandler {
 	return &storeTaskQueryHandler{RequestHandler: delegate, store: store}
+}
+
+func (h *storeTaskQueryHandler) GetTask(ctx context.Context, req *a2atype.GetTaskRequest) (*a2atype.Task, error) {
+	userID := callerUserID(ctx)
+	if userID == "" {
+		return h.RequestHandler.GetTask(ctx, req)
+	}
+
+	task, err := h.store.GetTask(ctx, string(req.ID), userID)
+	if err == nil {
+		return shapeTask(task, req.HistoryLength, true), nil
+	}
+	if errors.Is(err, dbpkg.ErrNotFound) {
+		return h.RequestHandler.GetTask(ctx, req)
+	}
+	return nil, err
 }
 
 // callerUserID returns the authenticated principal's user id, or "" when the
--- a/go/core/internal/a2a/task_query_store.go
+++ b/go/core/internal/a2a/task_query_store.go
@@ -31,8 +31,8 @@ type TaskStore interface {
 }
 
 // storeTaskQueryHandler answers GetTask and ListTasks from kagent's task
-// store, which is the source of truth for persisted tasks. Every other method
-// is delegated to the embedded handler unchanged.
+// store, which is the source of truth for persisted tasks. GetTask never falls
+// through to the embedded handler; every other method is delegated unchanged.
 type storeTaskQueryHandler struct {
 	a2asrv.RequestHandler
 	store TaskStore
@@ -43,19 +43,14 @@ func newStoreTaskQueryHandler(delegate a2asrv.RequestHandler, store TaskStore) *
 }
 
 func (h *storeTaskQueryHandler) GetTask(ctx context.Context, req *a2atype.GetTaskRequest) (*a2atype.Task, error) {
-	userID := callerUserID(ctx)
-	if userID == "" {
-		return h.RequestHandler.GetTask(ctx, req)
+	task, err := h.store.GetTask(ctx, string(req.ID), callerUserID(ctx))
+	if err != nil {
+		if errors.Is(err, dbpkg.ErrNotFound) {
+			return nil, a2atype.ErrTaskNotFound
+		}
+		return nil, err
 	}
-
-	task, err := h.store.GetTask(ctx, string(req.ID), userID)
-	if err == nil {
-		return shapeTask(task, req.HistoryLength, true), nil
-	}
-	if errors.Is(err, dbpkg.ErrNotFound) {
-		return h.RequestHandler.GetTask(ctx, req)
-	}
-	return nil, err
+	return shapeTask(task, req.HistoryLength, true), nil
 }
 
 // callerUserID returns the authenticated principal's user id, or "" when the
--- a/go/core/internal/a2a/task_query_store.go
+++ b/go/core/internal/a2a/task_query_store.go
@@ -43,7 +43,12 @@ func newStoreTaskQueryHandler(delegate a2asrv.RequestHandler, store TaskStore) *
 }
 
 func (h *storeTaskQueryHandler) GetTask(ctx context.Context, req *a2atype.GetTaskRequest) (*a2atype.Task, error) {
-	task, err := h.store.GetTask(ctx, string(req.ID), callerUserID(ctx))
+	userID := callerUserID(ctx)
+	if userID == "" {
+		return nil, a2atype.ErrTaskNotFound
+	}
+
+	task, err := h.store.GetTask(ctx, string(req.ID), userID)
 	if err != nil {
 		if errors.Is(err, dbpkg.ErrNotFound) {
 			return nil, a2atype.ErrTaskNotFound
--- a/go/core/internal/a2a/a2a_handler_mux.go
+++ b/go/core/internal/a2a/a2a_handler_mux.go
@@ -58,13 +58,11 @@ func NewA2AHttpMux(agentPathPrefix, sandboxPathPrefix string, authenticator auth
 
 // newTaskQueryHandlers builds the request handler and the legacy (v0) JSON-RPC
 // handler for one agent. kagent persists tasks and is their source of truth,
-// so with a store ListTasks is served from it instead of proxying to the agent
-// runtime (whose legacy 0.3 transport returns ErrUnsupportedOperation for it);
-// every other method, GetTask included, still delegates to the passthrough
-// proxy. v0 has no native tasks/list, so the legacy handler is wrapped to
-// serve that method from the store too (lowercase TaskState). Without a store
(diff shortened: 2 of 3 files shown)


erauner12:
### Related work and forward-port note

#2483 and #2485 repair identity propagation across the managed-runtime path. This PR is independent: authoritative store-backed exact reads remain useful even when runtime forwarding is functioning.

A main-oriented prototype is available in erauner12/kagent#12. The store-first/delegate-on-miss behavior transfers directly, but main's handler-mux structure differs from the v0.10 dual native/legacy JSON-RPC wiring. A forward-port should also retain this PR's authenticated-caller and share-context regression coverage rather than deriving read authority from `ShareContext.UserID`.

go/core/internal/a2a/task_query_store.go
```diff
@@ -41,6 +42,22 @@ func newStoreTaskQueryHandler(delegate a2asrv.RequestHandler, store TaskStore) *
 	return &storeTaskQueryHandler{RequestHandler: delegate, store: store}
 }
 
+func (h *storeTaskQueryHandler) GetTask(ctx context.Context, req *a2atype.GetTaskRequest) (*a2atype.Task, error) {
+	userID := callerUserID(ctx)
+	if userID == "" {
```
supreme-gg-gg, line 47: I don't understand the point of falling back to delegating to the runtime here and below in `ErrNotFound`. Why not treat the DB as authoritative as ListTask method below? The runtime will eventually read from the DB again which gives you the same result, this also weakens authorization boundary

supreme-gg-gg (commented):
Adding the new `GetTask` method makes sense to me, but I left a question about fallback behaviour

go/core/internal/a2a/task_query_store.go
erauner12: Good point. The delegate-on-miss behavior was meant as a compatibility path for tasks not found in the controller store, but in this setup the persistent store is already the source of truth, and the runtime resolves through that same store.

If identity is missing, or if the caller-scoped lookup returns `ErrNotFound`, delegating doesn’t help and can only weaken the authorization boundary.

I’ll make `GetTask` match `ListTasks`: query the persistent store with the authenticated caller and return the store result or error directly, with no delegation. I’ll update the tests, comments, and PR rationale to match.

erauner12 (commented):
left response

go/core/internal/a2a/task_query_store.go
erauner12: Updated in `85333a83`.

`GetTask` now treats the persistent store as authoritative: absent identity and caller-scoped `database.ErrNotFound` no longer delegate to the runtime, and not-found is mapped to A2A `ErrTaskNotFound`. The focused tests assert zero delegate calls for hits, misses, other-owner/share-context reads, absent identity, and backend failures, including 0.3 and 1.0 wire error semantics.

Validation passed:

- `go test ./core/internal/a2a`
- `go test -race ./core/internal/a2a -run '^(TestGetTask_|TestWire_GetTaskStoreErrorSemantics)'`

The PR description, source comments, and #2464 expectations were updated to match the authoritative-store behavior.

go/core/internal/a2a/task_query_store.go
```diff
@@ -41,6 +42,17 @@ func newStoreTaskQueryHandler(delegate a2asrv.RequestHandler, store TaskStore) *
 	return &storeTaskQueryHandler{RequestHandler: delegate, store: store}
 }
 
+func (h *storeTaskQueryHandler) GetTask(ctx context.Context, req *a2atype.GetTaskRequest) (*a2atype.Task, error) {
+	task, err := h.store.GetTask(ctx, string(req.ID), callerUserID(ctx))
```
supreme-gg-gg: can we do the userid check here like in the last version before this commit, so check the `userID == ""` first and return an error immediately instead of calling the DB

erauner12: Updated in `e8b4b1ce`. `GetTask` now checks `callerUserID(ctx)` first and returns `ErrTaskNotFound` immediately when it is empty, before calling the persistent store. The absent-identity test now asserts that neither the store nor delegate is invoked.

Validation passed:
- `go test ./core/internal/a2a`
- `go test -race ./core/internal/a2a -run '^(TestGetTask_|TestWire_GetTaskStoreErrorSemantics)'`

supreme-gg-gg (approved):

PR #2485, opened 2026-08-18
fix(a2a): forward authenticated user to managed agents

### LINKED ISSUE #2467: [BUG] Legacy controller A2A passthrough drops authenticated session principal
## Scope correction for `release/v0.10.x`

This issue tracks forwarding the controller-authenticated **session principal** on the legacy controller-to-managed-runtime A2A hop.

The generic client interceptor must not substitute `ShareContext.UserID`. Although the share token is authenticated, task-ID-only operations do not give this interceptor enough information to prove that the target belongs to `ShareContext.SessionID`. Share-aware forwarding therefore requires operation-specific task/session authorization and is outside this release fix.

The original report below is retained as the discovery record, but its earlier share-owner expectation is superseded by this correction. The safe contract here is: replace any existing outbound `x-user-id` value with the authenticated session principal, and do not elevate it from share context.
### 📋 Prerequisites

- [ ] I have tried to clear cache/cookies or used incognito mode (if UI-related)

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

The legacy controller A2A path authenticates the initiating caller and persists
tasks under that effective owner, but the controller-to-managed-runtime A2A hop
does not forward the effective user identity.

The upstream client interceptor constructs a new outbound request and preserves
provider authentication and tracing metadata, but it does not add `x-user-id`
from the controller's trusted authentication context. The managed runtime
therefore receives the request without the initiating user.

For a persisted `input-required` continuation, the runtime subsequently performs
an owner-scoped task lookup using its service identity rather than the task's
initiating owner. The existing task is then reported as not found, preventing a
standards-based client from continuing the same task safely.

This behavior is independent of whether the compatible legacy controller endpoint
is accessed with the A2A 0.3 or 1.0 wire because both versions use the same
controller-to-runtime hop.

### Scope clarification

This issue covers only the legacy controller's outbound effective-user seam.
Related defects are tracked separately:

- #2464 covers authoritative persistent exact `GetTask` behavior.
- #2465 covers preserving authenticated identity inside runtime task-store
  callbacks before task resolution.

A persisted continuation requires both the controller outbound identity and the
runtime callback identity contracts to hold, but the two seams are independently
testable and reviewable.

### 🔄 Steps To Reproduce

1. Deploy a maintained kagent release using the legacy controller A2A path with
   authenticated, owner-scoped task persistence enabled.
2. Through the controller A2A endpoint, create a task that reaches
   `input-required`; retain its `taskId` and `contextId`.
3. Confirm that the task is persisted under the authenticated initiating
   principal.
4. From a later request authenticated as the same principal, send a continuation
   containing the retained `taskId` and `contextId`.
5. Inspect the controller-to-runtime A2A request metadata and observe that the
   authenticated effective user is not forwarded as `x-user-id`.
6. Observe the runtime's owner-scoped task lookup execute using its service
   identity and return `Task not found` for the existing task.
7. Repeat through the legacy controller endpoint using A2A 0.3 and 1.0; both
   versions reproduce the same missing-identity hop.

The interceptor-level reproduction is deterministic: provide a request context
with an authenticated controller principal, run the legacy upstream client
interceptor, and inspect the resulting service parameters. Provider auth and
trace headers are present, while the effective user is absent.

### 🤔 Expected Behavior

The legacy controller should derive the effective user from trusted controller
authentication context and forward it to the managed runtime as `x-user-id`.

For ordinary requests, the effective user should be the authenticated session
principal. For a validated share context, it should be the share owner. Raw
inbound identity headers must not become an independent trust source.

The runtime should therefore receive the same effective owner used by the
controller's owner-scoped persistence boundary.

### 📱 Actual Behavior

The legacy controller's outbound A2A request contains provider authentication and
trace metadata but omits the effective user identity.

The managed runtime then performs persisted task resolution using its own service
identity. Because the task belongs to the authenticated initiating principal,
the owner-scoped lookup reports `Task not found` and the continuation fails.

### 💻 Environment

- OS and version: Not environment-specific
- Kubernetes version/provider: Not required for the source-level reproduction
- Browser: Not applicable
- Application version: reproduced with kagent `v0.10.0-rc1`
- Affected component: legacy controller A2A client interceptor
- Authentication: authenticated, owner-scoped task persistence
- A2A versions tested through the compatible controller endpoint: `0.3` and `1.0`

The legacy package remains present in current source, while the supported
current-main controller-v2 uses a different gateway path. This report does not
claim that controller-v2 uses the affected legacy outbound interceptor.

### 🔧 CLI Bug Report

Not attached because the live reproduction occurred in an internal Kubernetes
environment and generated diagnostics may contain infrastructure or
authentication details.

The source-level reproduction is deterministic. Reviewed and sanitized runtime
evidence can be provided if maintainers need additional diagnostics.

### 🔍 Additional Context

The required trust inputs already exist in the controller:

- authenticated session context identifies the ordinary initiating principal;
- validated share context identifies the effective share owner;
- the controller-to-runtime request is already an authenticated internal hop;
- runtime A2A handling already recognizes `x-user-id` request metadata.

The missing connection is adding the trusted effective user to the newly
constructed outbound request. The fix should not copy an arbitrary raw inbound
`x-user-id` value.

Focused regression coverage should verify:

- authenticated-principal forwarding;
- validated share-owner replacement of any preexisting value;
- preservation of provider authentication and trace propagation; and
- absence of identity synthesis when trusted controller context has no user.

Before the combined fixes, persisted continuation failed on all four compatible
create/resume wire combinations: `0.3→0.3`, `0.3→1.0`, `1.0→0.3`, and
`1.0→1.0`.

### 📋 Logs

```shell
Controller persisted task owner: <authenticated-initiating-principal>
Controller → runtime service parameters: x-user-id absent
Runtime persisted-task callback identity: <runtime-service-identity>
Owner-scoped lookup: Task not found
```

### 📷 Screenshots

_Not applicable._

erauner12:
## Summary

Forward the controller-authenticated session principal on the legacy controller-to-managed-runtime A2A hop on `release/v0.10.x`.

Fixes #2467.

## Problem

The controller authenticates the initiating caller, but its outbound managed-runtime request did not carry that principal. Runtime task resolution could therefore use the runtime service account instead of the owner established at the controller boundary.

## Change

The existing upstream-auth interceptor now replaces any preexisting `x-user-id` service parameter with exactly one value derived from `session.Principal().User.ID` after the controller's normal authentication step.

Provider authentication and trace propagation remain unchanged.

## Trust boundary

The interceptor does not copy a raw inbound identity header and does not infer share-owner authority. Even when a validated `ShareContext` is present, the generic interceptor forwards only the authenticated session principal. Share-aware forwarding requires operation-specific task/session validation and is intentionally outside this fix.

## Tests

- `cd go && go test ./internal/a2a -run '^TestUpstreamAuthInterceptor' -count=1`
- `cd go && go test -race ./internal/a2a -run '^TestUpstreamAuthInterceptor' -count=1`

Coverage verifies authenticated-principal forwarding, replacement of a preexisting value, non-escalation in the presence of `ShareContext`, and preservation of trace behavior.

## Scope and ordering

This PR changes only controller-side principal forwarding. Runtime identity ingestion is tracked in #2465 and should land first. Persistent exact `GetTask` is tracked separately in #2464.

### DIFF
--- a/go/core/internal/a2a/client_interceptors.go
+++ b/go/core/internal/a2a/client_interceptors.go
@@ -64,6 +64,11 @@ func (u *upstreamAuthInterceptor) Before(ctx context.Context, req *a2aclient.Req
 		if err := u.authProvider.UpstreamAuth(httpReq, session, upstreamPrincipal); err != nil {
 			return ctx, nil, err
 		}
+		authenticatedUserID := session.Principal().User.ID
+		if authenticatedUserID != "" {
+			httpReq.Header.Set("X-User-Id", authenticatedUserID)
+			delete(req.ServiceParams, "x-user-id")
+		}
 	}
 	propagation.TraceContext{}.Inject(ctx, propagation.HeaderCarrier(httpReq.Header))
 	for k, values := range httpReq.Header {
--- a/go/core/internal/a2a/client_interceptors_test.go
+++ b/go/core/internal/a2a/client_interceptors_test.go
@@ -2,13 +2,77 @@ package a2a
 
 import (
 	"context"
+	"net/http"
 	"testing"
 
 	a2aclient "github.com/a2aproject/a2a-go/v2/a2aclient"
+	"github.com/kagent-dev/kagent/go/core/pkg/auth"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 	"k8s.io/apimachinery/pkg/types"
 )
+
+type interceptorTestSession struct {
+	principal auth.Principal
+}
+
+func (s interceptorTestSession) Principal() auth.Principal {
+	return s.principal
+}
+
+type interceptorTestAuthProvider struct {
+	auth.AuthProvider
+	userID string
+}
+
+func (p interceptorTestAuthProvider) UpstreamAuth(r *http.Request, _ auth.Session, _ auth.Principal) error {
+	if p.userID != "" {
+		r.Header.Set("X-User-Id", p.userID)
+	}
+	return nil
+}
+
+func TestUpstreamAuthInterceptor_ForwardsAuthenticatedUser(t *testing.T) {
+	ctx := auth.AuthSessionTo(context.Background(), interceptorTestSession{
+		principal: auth.Principal{User: auth.User{ID: "caller-user"}},
+	})
+	req := &a2aclient.Request{
+		BaseURL:       "http://agent.default:8080",
+		ServiceParams: a2aclient.ServiceParams{},
+	}
+	interceptor := NewUpstreamAuthInterceptor(interceptorTestAuthProvider{}, types.NamespacedName{})
+
+	if _, _, err := interceptor.Before(ctx, req); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if got := req.ServiceParams.Get("X-User-Id"); len(got) != 1 || got[0] != "caller-user" {
+		t.Errorf("X-User-Id: want %q, got %q", "caller-user", got)
+	}
+}
+
+func TestUpstreamAuthInterceptor_IgnoresShareOwner(t *testing.T) {
+	ctx := auth.AuthSessionTo(context.Background(), interceptorTestSession{
+		principal: auth.Principal{User: auth.User{ID: "visiting-user"}},
+	})
+	ctx = auth.ShareContextTo(ctx, &auth.ShareContext{UserID: "owner-user"})
+	req := &a2aclient.Request{
+		BaseURL: "http://agent.default:8080",
+		ServiceParams: a2aclient.ServiceParams{
+			"x-user-id": {"preexisting-user"},
+		},
+	}
+	interceptor := NewUpstreamAuthInterceptor(
+		interceptorTestAuthProvider{userID: "provider-user"},
+		types.NamespacedName{},
+	)
+
+	if _, _, err := interceptor.Before(ctx, req); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if got := req.ServiceParams.Get("X-User-Id"); len(got) != 1 || got[0] != "visiting-user" {
+		t.Errorf("X-User-Id: want exactly [%q], got %q", "visiting-user", got)
+	}
+}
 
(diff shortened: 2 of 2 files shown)


erauner12:
### Dependency and forward-port note

This is the controller-side counterpart to #2483 and should merge after that runtime boundary is present. #2484 is related but independently mergeable.

A main-oriented prototype exists in erauner12/kagent#13, but it predates the authorization-boundary refinement made in this PR. A forward-port must use the authenticated session principal only, replace any caller-supplied `x-user-id` service parameter, and must not elevate the generic passthrough to `ShareContext.UserID`. The non-escalation test in this PR is the authoritative behavior.

supreme-gg-gg (commented):
It seems that `UnsecureAuthenticator` and `ProxyAuthenticator` already sets x-user-id inside the `UpstreamAuth` method above this change. Which auth provider are you using that produced the missing x-user-id?

I could be wrong but I'm suspecting that the symptoms you observed could be caused by #2449 instead (and the behaviour you described seems close to what's in its issue #2443, which is also applicable to 0.10.x). Since even if x-user-id is propagated properly, according to that issue the task store is not storing the owner properly resulting in "task not found" during the next lookup.

Can you take another look at this and lmk which one is the actual root cause in your setup?

EDIT: I noticed you have #2483 open as well, the #2449 I quoted earlier is functionally the same as yours, but my question remains the same

erauner12 (commented):
left response

erauner12:
Closing this after rechecking the shipped authentication paths on `release/v0.10.x`.

Both `UnsecureAuthenticator.UpstreamAuth` and `ProxyAuthenticator.UpstreamAuth` already forward `X-User-Id` from the authenticated session. The test in this PR demonstrates a stronger interceptor-level invariant using a provider stub that intentionally omits the header, but it does not isolate a defect in either shipped provider.

The observed task-resolution failure is accounted for by the runtime-context defect covered by #2483 and the equivalent main work in #2449. Rather than add redundant identity behavior at another layer, I’m withdrawing this PR. If provider-independent interceptor enforcement is desired later, it can be proposed separately as explicit hardening.

Thanks for helping separate these concerns.

PR #2493, opened 2026-08-19
feat(helm): dynamically set ateomImage based on substrate dependency version

### LINKED ISSUE #2491: [BUG] `substrateWorkerPool.ateomImage` ships empty by default, failing any install that enables the WorkerPool
### 📋 Prerequisites

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

`helm/kagent/templates/substrate-workerpool.yaml` requires `substrateWorkerPool.ateomImage` to be set whenever the chart is asked to create the Agent Substrate WorkerPool (`controller.substrate.enabled=true` + `substrateWorkerPool.create=true`):

```yaml
{{- if and .Values.controller.substrate.enabled .Values.substrateWorkerPool.create }}
{{- if not .Values.substrateWorkerPool.ateomImage }}
{{- fail "substrateWorkerPool.ateomImage is required when substrateWorkerPool.create=true" }}
{{- end }}
...
  ateomImage: {{ .Values.substrateWorkerPool.ateomImage | quote }}
```

But `helm/kagent/values.yaml` ships this value empty (`ateomImage: ""`) with no default, even though a matching image is already published and ready: `ghcr.io/kagent-dev/substrate/ateom-gvisor`, tagged to match this chart's own pinned `substrate` dependency version (`helm/kagent/Chart.yaml`: `- name: substrate, version: 0.0.9, repository: oci://ghcr.io/kagent-dev/substrate/helm`; the matching image tag is `v0.0.9`, confirmed by `examples/substrate-openclaw/README.md:60`: `ateomImage=ghcr.io/kagent-dev/substrate/ateom-gvisor:v${SUBSTRATE_VERSION}`).

So every install that opts into the WorkerPool fails immediately with `substrateWorkerPool.ateomImage is required when substrateWorkerPool.create=true`, even though the correct value is fully derivable from information the chart already has.

### 🔄 Steps To Reproduce

1. `helm install kagent ./helm/kagent --set controller.substrate.enabled=true --set substrateWorkerPool.create=true` (without setting `substrateWorkerPool.ateomImage`).
2. Observe the install fails.

### 🤔 Expected Behavior

When `ateomImage` isn't explicitly set, the chart should default it to the `ateom-gvisor` image matching the pinned `substrate` dependency version, only requiring an explicit value in cases where that can't be reliably derived.

### 📱 Actual Behavior

every install of this feature fails until the user manually discovers and supplies the correct, currently-undocumented image reference.

### 💻 Environment

_No response_

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

_No response_

entee28:
This PR closes #2491


EItanya (changes requested):
What do you think about taking it from the `go.mod` version since they match? We can link it at buildtime with an ldflag

entee28:
> What do you think about taking it from the `go.mod` version since they match? We can link it at buildtime with an ldflag

Hi @EItanya , good catch — you're right it should come from go.mod. Couldn't figure out on how to use an ldflag here though, so I went with the build-time route this repo already has instead: `SUBSTRATE_VERSION` now also gets stamped into `Chart.yaml`'s annotations via the same `envsubst` step that sets it for the substrate dependency. 

Kept the fail for the `substrate.enabled=false` case — if `ate-system` is managed separately we don't actually know its version, so still want an explicit value there rather than guessing.

PR #2496, opened 2026-08-19
fix(go-adk): accept standard A2A JSON-RPC method names alongside a2a-go/v2's native ones

### LINKED ISSUE #2497: go/adk A2A server rejects standard A2A JSON-RPC method names (e.g. message/send) sent by every public A2A client SDK
## Summary

`go/adk`'s A2A server serves JSON-RPC directly through `a2a-go/v2`'s native dispatcher (`a2asrv.NewJSONRPCHandler`), which only recognizes its own bare method names — `SendMessage`, `GetTask`, `CancelTask`, `SubscribeToTask`, `GetTaskPushNotificationConfig`, etc.

Every public A2A client SDK sends the standard, slash-namespaced method names from the wider A2A protocol spec instead — `message/send`, `message/stream`, `tasks/get`, `tasks/cancel`, `tasks/resubscribe`, `tasks/pushNotificationConfig/*`, `agent/getAuthenticatedExtendedCard`. Confirmed directly against the reference `@a2a-js/sdk` (v0.3.13) client: its compiled transport hardcodes these exact slash-form names (`node_modules/@a2a-js/sdk/dist/client/index.js`).

The result: any standards-compliant A2A client gets a bare JSON-RPC `-32601 method not found` calling a raw `go/adk` agent port directly, even though the agent is otherwise working correctly.

## Root cause

`a2a-go/v2` ships a compatibility shim for exactly this mismatch — `a2acompat/a2av0`, a complete alternate JSON-RPC handler built from the same `a2asrv.RequestHandler` used by the native dispatcher — but nothing in kagent wires it in anywhere.

This is a real regression introduced when `go/adk` fully migrated off the older, non-versioned `a2a-go` module onto `a2a-go/v2` only. Diffing `go.mod` between the last release tag and current `main` confirms it:

```
v0.10.0-rc2 (released):
  github.com/a2aproject/a2a-go       v0.3.15   ← native dispatch already spoke the standard names
  github.com/a2aproject/a2a-go/v2    v2.3.1

main (current):
  github.com/a2aproject/a2a-go/v2    v2.3.1    ← old module dropped, no compat added
```

`go/adk/pkg/a2a/server/server.go`'s import of `a2asrv` changed from `github.com/a2aproject/a2a-go/a2asrv` (old, standard-compliant names) to `github.com/a2aproject/a2a-go/v2/a2asrv` (new, non-standard bare names) between those two points, with no replacement compatibility layer.

## Discovery context

Found while validating an unrelated fix (#2473 / #2475) against a real external A2A bridge using `@a2a-js/sdk`: sending a message directly to the agent's port with the SDK's standard method names failed outright with `-32601`, while a hand-crafted request using `a2a-go/v2`'s native names succeeded.

## Fix

Filed as #2496 — wires `a2a-go/v2`'s own `a2acompat/a2av0` shim into `go/adk`'s server. The two method-naming conventions never collide (native names are bare identifiers; legacy/standard names are always namespaced with a `/`), so a single peek at the incoming JSON-RPC request's `"method"` field is enough to route to the correct dispatcher. Verified end-to-end against a live orchestrator pod with both a native `SendMessage` call and a `message/send` call shaped exactly like `@a2a-js/sdk`'s wire format — both dispatch correctly and return the standards-shaped response.

vramahandry:
## Summary

`go/adk`'s A2A server serves JSON-RPC directly through `a2a-go/v2`'s native dispatcher (`a2asrv.NewJSONRPCHandler`), which only recognizes its own bare method names — `SendMessage`, `GetTask`, `CancelTask`, `SubscribeToTask`, etc.

Every public A2A client SDK still sends the standard, slash-namespaced method names from the wider A2A protocol spec — `message/send`, `message/stream`, `tasks/get`, `tasks/cancel`, `tasks/resubscribe`, `tasks/pushNotificationConfig/*`, `agent/getAuthenticatedExtendedCard`. I confirmed this directly against the reference `@a2a-js/sdk` (v0.3.13) client, whose compiled transport hardcodes these exact names (`node_modules/@a2a-js/sdk/dist/client/index.js`).

`a2a-go/v2` ships a compatibility shim for exactly this mismatch — `a2acompat/a2av0`, a complete alternate JSON-RPC handler built from the same `a2asrv.RequestHandler` — but nothing in kagent wires it in anywhere. The result: any standards-compliant A2A client gets a bare `-32601 method not found` calling a raw `go/adk` agent port directly, even though the agent is otherwise working correctly.

Discovered while validating an unrelated fix (#2475) against a real external A2A bridge: sending a message directly to the agent's port with the SDK's standard method names failed outright, while a hand-crafted request using `a2a-go/v2`'s native names succeeded. Confirmed this is a real, pre-existing gap on `main` by diffing `go.mod` against the last release tag (`v0.10.0-rc2`), which still depended on the older, non-versioned `a2a-go` module (whose native dispatch already spoke the standard names) — the migration to `a2a-go/v2`-only dropped that compatibility with no replacement.

Fixes #2497

## Changes

- `go/adk/pkg/a2a/server/server.go`: new `newCompatJSONRPCHandler` peeks the incoming JSON-RPC request body's `"method"` field and routes to `a2a-go/v2`'s native handler or the SDK's own `a2acompat/a2av0` handler accordingly. The two conventions never collide — native method names are bare identifiers, legacy ones are always namespaced with a `/` — so this single check is sufficient. Both handlers wrap the same `a2asrv.RequestHandler`, so behavior (including gRPC, which is untouched) stays otherwise identical.
- `go/adk/pkg/a2a/server/server_test.go`: `TestJSONRPCCompatDualDispatch` covers a legacy `message/send` call dispatching successfully, and an unrecognized legacy-shaped method still correctly reporting `-32601`.
- `go/go.mod`, `go/go.sum`: `a2acompat/a2av0` transitively depends on the older `a2a-go` module (already required by `a2a-go/v2`'s own `go.mod`); added via `go mod tidy` after wiring in the import.

## Test plan

- [x] `go build ./...` (both `go/adk` and the full `go/` workspace) — clean
- [x] `go test ./...` in `go/adk` — all packages pass, including the existing `TestA2ARequestSizeLimit` (verified the new body-peeking logic doesn't regress the size-limit error path)
- [x] New `TestJSONRPCCompatDualDispatch` — legacy method dispatches successfully; unknown legacy-shaped method still reports method-not-found
- [x] `gofmt -l` / `go vet ./...` — clean

### DIFF
--- a/go/adk/pkg/a2a/server/server.go
+++ b/go/adk/pkg/a2a/server/server.go
@@ -1,8 +1,11 @@
 package server
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"os"
@@ -13,6 +16,7 @@ import (
 	"time"
 
 	a2atype "github.com/a2aproject/a2a-go/v2/a2a"
+	"github.com/a2aproject/a2a-go/v2/a2acompat/a2av0"
 	a2agrpc "github.com/a2aproject/a2a-go/v2/a2agrpc/v1"
 	a2apb "github.com/a2aproject/a2a-go/v2/a2apb/v1"
 	"github.com/a2aproject/a2a-go/v2/a2asrv"
@@ -51,7 +55,7 @@ type A2AServer struct {
 // NewA2AServer creates a new A2A server using a2asrv.
 func NewA2AServer(agentCard a2atype.AgentCard, executor a2asrv.AgentExecutor, logger logr.Logger, config ServerConfig, handlerOpts ...a2asrv.RequestHandlerOption) (*A2AServer, error) {
 	requestHandler := a2asrv.NewHandler(executor, handlerOpts...)
-	jsonrpcHandler := a2asrv.NewJSONRPCHandler(requestHandler)
+	jsonrpcHandler := newCompatJSONRPCHandler(requestHandler)
 	if maxContentLength := getMaxContentLength(logger); maxContentLength != nil {
 		jsonrpcHandler = withRequestSizeLimit(jsonrpcHandler, *maxContentLength)
 	}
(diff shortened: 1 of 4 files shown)


supreme-gg-gg (commented):
hi @vramahandry, it has been decided earlier that main will only support v1 A2A moving forward. Supporting dual-compatibility has been done in 0.10.x, and adding that to main will require conversion between v0 and v1 tasks and relevant migration in the database layer, which we want to avoid.

supreme-gg-gg:
For your use case, do you think you can update your client library to use v1? a2a-js v1 just released, we're using it for our UI

supreme-gg-gg:
@vramahandry I'm closing this PR as per the above comments. Feel free to reopen if you have another thought on this

PR #2512, opened 2026-08-20
feat(models): adding python support for Azure foundry models (OpenAI + Anthropic types)

marosset:
Part of https://github.com/kagent-dev/kagent/issues/2072

https://github.com/kagent-dev/kagent/pull/2091 added foundry support for the Go runtime. This PR does the same for python.

As part of this change the AzureOpenAI model auth is moved to use unified Azure auth mechanisms (including API_KEY-less / workload identity support)


python/packages/kagent-adk/src/kagent/adk/models/_embedding.py
```diff
@@ -178,6 +194,53 @@ async def _embed_openai(self, texts: List[str]) -> List[List[float]]:
...
+        client = self._build_azure_client(
+            api_version=api_version,
+            endpoint=endpoint,
+            deployment=deployment,
+            api_key=api_key,
+        )
+        response = await client.embeddings.create(
+            model=self.config.model,
+            input=texts,
+            dimensions=self.TARGET_DIMENSION,
```
mesutoezdil: forces 768 dims here too. same bug as _embed_openai above it, a fix for that one is already open in #2542. providers with other dims will fail here as well.

marosset: updated

marosset:
@mesutoezdil, have you had a chance to look at the rest of this PR?
Thanks!

python/packages/kagent-adk/src/kagent/adk/models/_azure.py
```diff
@@ -0,0 +1,179 @@
...
+        if api_key_passthrough:
+            raise ValueError(missing_credential_hint)
+        token_provider = azure_ad_token_provider()
+
+    return AsyncAzureOpenAI(
+        # The sentinel prevents environment-key fallback while the token
+        # provider authenticates each request.
+        api_key=API_KEY_SENTINEL if token_provider is not None else api_key,
+        azure_ad_token_provider=token_provider,
+        api_version=api_version,
```
EItanya, line 145: 1. Prevent `AZURE_OPENAI_AD_TOKEN` from overriding Foundry's selected credential.

`API_KEY_SENTINEL` suppresses the SDK's API-key fallback, but `AsyncAzureOpenAI` still reads `AZURE_OPENAI_AD_TOKEN` and gives that value precedence over both the explicit API key and `azure_ad_token_provider`. With that variable set, I reproduced a Foundry passthrough request sending `Authorization: Bearer unrelated-env-token` and no caller `api-key`; the workload-identity case also sent the environment token instead of calling the selected provider. This can authenticate as the wrong identity or fail against a different Foundry account.

Make the selected authentication mode authoritative by preventing the SDK's AD-token environment fallback, and add real-SDK request tests with `AZURE_OPENAI_AD_TOKEN` populated for passthrough and workload identity.

python/packages/kagent-adk/src/kagent/adk/models/_azure.py
```diff
@@ -0,0 +1,179 @@
...
+_AUTH_HEADER_NAMES = {"authorization", "api-key", "x-api-key"}
+
+AsyncTokenProvider = Callable[[], Awaitable[str]]
+
+
+def azure_ad_token_provider(scope: str = COGNITIVE_SERVICES_SCOPE) -> AsyncTokenProvider:
+    """Return an async bearer-token provider backed by ``DefaultAzureCredential``."""
+    from azure.identity.aio import DefaultAzureCredential, get_bearer_token_provider
+
+    return get_bearer_token_provider(DefaultAzureCredential(), scope)
```
EItanya, line 32: 2. Retain and close the async Azure credential.

This creates `DefaultAzureCredential` inside the token-provider closure without exposing ownership to the caller. Both Azure embedding paths construct a new provider per call and then only `await client.close()`, which closes OpenAI's HTTP client, not Azure Identity's separate HTTP transport. With workload identity or managed identity, repeated memory embeddings therefore leave Azure sessions/connectors unclosed and recreate the token cache each time.

I reproduced this for both `foundry` and `azure_openai` using the real OpenAI client and Azure bearer-token provider with an injected credential: token acquisition ran and the OpenAI HTTP client closed, but `credential.close()` was never awaited. Keep an explicit credential lifecycle and close it when its owning client/operation ends, including failures; cover that alongside the existing client-close tests.

EItanya (commented):
🤖 AI-generated review.

Reviewed the merged result at `5bfc27a878d39e943c0a1642cd0750ebc93d6b98`. The conflict resolution preserves main's Responses API support and the PR's Azure helpers, and fixes the merged import ordering.

I found two credential-handling issues, detailed inline. Both are reproduced with the locked SDKs and mocked HTTP/token boundaries.

Validation: `uv lock --check`, Ruff lint/format on the affected model/config/test files, and the PR diff whitespace check passed. The model/embedding suite had 249 passes and nine TLS failures caused by missing test certificates; after `make generate-test-certs`, the TLS suite passed (11 passed, 1 existing skip). Four separate, intentionally failing review probes reproduced the two findings. No live Azure or Kubernetes validation.

EItanya (approved):

PR #2520, opened 2026-08-21
test: add missing unit tests for adk and core packages

yashrajshuklaaa:
Test-only change adding unit test coverage for -

- `go/adk/pkg/auth/token_test.go`
- `go/adk/pkg/runner/adapter_test.go` (+ `testdata/config_openai.json`, `testdata/mock_openai.json`)
- `go/core/internal/controller/reconciler/status/mcp_server_test.go`
- `go/core/internal/controller/reconciler/utils/reconciler_utils_test.go`
- `go/core/internal/controller/translator/mutate_test.go`
- `go/core/internal/service/agent/service_test.go`
- `go/core/internal/service/agent/validator_test.go`

No production code changes. 
All tests pass locally (`go test ./...`)


go/core/internal/service/agent/service_test.go
```diff
@@ -0,0 +1,299 @@
...
+func TestAuthorizationDenied_ReturnsPermissionDenied(t *testing.T) {
+	kube := fake.NewClientBuilder().WithScheme(newScheme(t)).Build()
+	service := agentsvc.NewService(kube, &stubAuthorizer{err: assertAnError()}, "default")
+
+	_, err := service.List(authedContext("user-a"), agentsvc.ListRequest{})
+	require.Error(t, err)
+	assert.True(t, serviceerrors.IsCode(err, serviceerrors.CodePermissionDenied))
+}
+
+func assertAnError() error {
```
mesutoezdil: name says assert, but this just builds an error. confusing name.

go/core/internal/service/agent/validator_test.go
```diff
@@ -0,0 +1,83 @@
...
+		KubeClient:         kube,
+		DefaultModelConfig: types.NamespacedName{Namespace: "default", Name: "default-model"},
+	})
+
+	agent := &v1alpha3.SandboxAgent{
+		ObjectMeta: metav1.ObjectMeta{Name: "agent-1", Namespace: "default"},
+		Spec: v1alpha3.SandboxAgentSpec{
+			Type: v1alpha3.AgentType_Declarative,
+			Declarative: &v1alpha3.DeclarativeAgentSpec{
+				// ModelConfig deliberately omitted / points at a ModelConfig
```
mesutoezdil: comment says omitted, but code sets it to a missing name. pick one wording.

yashrajshuklaaa: resolved! thanks for the suggestion.

go/core/internal/service/agent/service_test.go
yashrajshuklaaa: fixed :) @mesutoezdil 
PTAL

yashrajshuklaaa:
@EItanya PTAL

yashrajshuklaaa:
@EItanya @supreme-gg-gg the test-e2e failure (TestAgentInstanceActiveTask) looks unrelated to this PR. it's a stream/actor-suspend race in an e2e test and this PR only adds unit tests with no production code changes. Could one of you re-run the job or take a look when you get a chance?

yashrajshuklaaa:
@EItanya Could you please approve the workflow run so CI can complete?

EItanya:
🤖 AI-assisted review.

Thanks for putting time into improving test coverage! I think this PR would be more useful if we narrowed it to tests that protect current behavior against regressions.

- Several of the new core tests reference packages or functions that have since been removed. At the current head (`05d13d19`), the focused test run fails to compile four core packages. Those obsolete cases should be removed.
- The shared-tools tests only assert that an agent is created, but `CreateRunnerConfig` no longer uses `controllerClient` or `ShareTools`. They pass without exercising the behavior their names describe, so those cases can also be removed.
- The runner configuration tests never call the model, so the mock HTTP server and response fixture seem unnecessary.

The auth header assertions and parsing edge cases look worth keeping. Could you trim the PR to focused tests for behavior that still exists, with assertions that would fail if that behavior broke? A smaller suite with clear regression coverage would be a stronger addition.

PR #2523, opened 2026-08-24
fix(substrate): fall back to Secret.StringData in resolvePodEnv

### LINKED ISSUE #2522: [BUG] Substrate SandboxAgent env resolution always fails: resolvePodEnv never checks Secret.StringData
### 📋 Prerequisites

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

`resolvePodEnv` in `go/core/pkg/sandboxbackend/substrate/lifecycle_shared.go` resolves a container's env vars against a `corev1.Secret`. When a requested key is missing from `secret.Data`, the function only checks `ref.Optional` before returning an error — it never checks `secret.StringData`.

Normally this is fine, because when a `Secret` is created or updated through the real Kubernetes API server, the API server converts any `StringData` entries into `Data` (base64-encoded) as part of admission, so `secret.Data` always ends up populated regardless of whether the caller populated `Data` or `StringData`.

However, the substrate `SandboxAgent` config `Secret` is built in-memory by `manifest_builder.go`'s `buildConfigSecretData` and is passed directly into `resolvePodEnv` as the `localSecret` parameter. Looking at the resolution logic:

```go
secret := &corev1.Secret{}
if localSecret != nil && localSecret.Name == ref.Name {
    secret = localSecret
} else if err := kube.Get(ctx, types.NamespacedName{Namespace: namespace, Name: ref.Name}, secret); err != nil {
    ...
}
value, ok := secret.Data[ref.Key]
if !ok {
    if ref.Optional != nil && *ref.Optional {
        resolved[i].ValueFrom = nil
        continue
    }
    return nil, fmt.Errorf("secret %q does not contain key %q", ref.Name, ref.Key)
}
```

When `localSecret.Name == ref.Name`, the in-memory secret is used directly — it is never round-tripped through the API server (no Create/Update call happens for this read path), so the usual `StringData` → `Data` conversion never happens. If `buildConfigSecretData` populates `StringData` (or if `Data` simply doesn't contain the `config.json` key for any other in-memory-only reason), `secret.Data[ref.Key]` will always miss, and — since the env var reference for `config.json` is not marked `Optional` — `resolvePodEnv` deterministically returns `secret "<name>" does not contain key "config.json"` on every reconcile.

### 🔄 Steps To Reproduce

1. Deploy a kagent controller built from `main` with the Agent Substrate backend enabled (`controller.substrate.enabled=true`).
2. Create any declarative `SandboxAgent` custom resource against this controller.
3. Watch the controller reconcile the `SandboxAgent` (e.g. via controller logs or the resource's status/conditions).
4. Observe that reconciliation fails every time with an env-resolution error referencing `config.json`.

### 🤔 Expected Behavior

The `config.json` env var (and any other env var backed by the in-memory config `Secret`) should resolve normally, using either `Data` or `StringData` as the source of truth — matching how the real API server treats a `Secret` object — so declarative `SandboxAgent` reconciliation succeeds.

### 📱 Actual Behavior

Reconciliation fails with:
```
secret "<sandbox-agent-config-secret-name>" does not contain key "config.json"
```
This happens on every reconcile attempt for every declarative `SandboxAgent`, since the failure is deterministic (not transient/flaky).

### 💻 Environment

- Self-managed Kubernetes cluster
- kagent controller built from source (main branch)
- Agent Substrate backend enabled (`controller.substrate.enabled=true`)
- Substrate 0.0.17

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

- Found via live debugging of a Substrate-backed kagent controller deployment.
- Root cause is isolated to `resolvePodEnv` in `go/core/pkg/sandboxbackend/substrate/lifecycle_shared.go`; the in-memory config `Secret` producer is `buildConfigSecretData` in `manifest_builder.go`.

### 📋 Logs

```shell
secret "<name>" does not contain key "config.json"
```

### 📷 Screenshots

_No response_

entee28:
The substrate SandboxAgent config Secret (manifest_builder.go's buildConfigSecretData) is used in-memory and never round-tripped through the API server, so the usual StringData -> Data conversion the apiserver performs on Create/Update never happens for it. resolvePodEnv only checked secret.Data, so every declarative SandboxAgent deterministically failed env resolution with "secret does not contain key" for config.json. Fall back to StringData when the key isn't in Data.

This PR closes #2522

### DIFF
--- a/go/core/pkg/sandboxbackend/substrate/lifecycle_shared.go
+++ b/go/core/pkg/sandboxbackend/substrate/lifecycle_shared.go
@@ -300,6 +300,17 @@ func resolvePodEnv(ctx context.Context, kube client.Reader, namespace string, en
 		}
 		value, ok := secret.Data[ref.Key]
 		if !ok {
+			// localSecret (the substrate SandboxAgent config Secret built by
+			// manifest_builder.go's buildConfigSecretData) is used purely in-memory
+			// here and never round-tripped through the API server, so the usual
+			// StringData -> Data conversion the apiserver performs on Create/Update
+			// never happens. Fall back to StringData so this path doesn't
+			// unconditionally fail for every substrate SandboxAgent.
+			if stringValue, stringOK := secret.StringData[ref.Key]; stringOK {
+				resolved[i].Value = stringValue
+				resolved[i].ValueFrom = nil
+				continue
+			}
 			if ref.Optional != nil && *ref.Optional {
 				resolved[i].ValueFrom = nil
 				continue
--- a/go/core/pkg/sandboxbackend/substrate/lifecycle_shared_test.go
+++ b/go/core/pkg/sandboxbackend/substrate/lifecycle_shared_test.go
@@ -1,0 +1,209 @@
+package substrate
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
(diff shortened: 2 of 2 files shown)


EItanya:
This code will be deleted in the lead-up to the new API, so no need to fix

PR #2532, opened 2026-08-24
fix(ui): preserve post-login redirect target through `oauth2-proxy` sign-in

onematchfox:
When oauth2-proxy intercepts an unauthenticated request it serves its sign-in page carrying the original destination as `.Redirect` (e.g. `/oauth2/sign_in?rd=%2Fagents%2Ffoo`). `sign_in.html` template ignored that and unconditionally redirected to `/login`, and `/login`'s "Sign in with SSO" link was hardcoded to `rd=/` -- so any login, expired- cookie or not, always landed back on the home page instead of the page the user was trying to reach.

Also contains a Helm update that ensures that this change will result in `oauth2-proxy` being redeployed to pick up the updated sign-in page template.

### DIFF
--- a/helm/kagent/values.yaml
+++ b/helm/kagent/values.yaml
@@ -948,6 +948,11 @@ oauth2-proxy:
       mountPath: /templates
       readOnly: true
 
+  extraEnv:
+    # Forces a rollout whenever the sign_in.html ConfigMap's content changes.
+    - name: KAGENT_OAUTH2_PROXY_SIGNIN_TEMPLATE_CHECKSUM
+      value: '{{ include "kagent.oauth2ProxySignInHTML" . | sha256sum }}'
+
   # Mount custom CA certificate for TLS verification (if needed)
   # Add to extraVolumes:
   #   - name: custom-ca-cert
--- a/helm/kagent/values.yaml
+++ b/helm/kagent/values.yaml
@@ -948,11 +948,6 @@ oauth2-proxy:
       mountPath: /templates
       readOnly: true
 
-  extraEnv:
-    # Forces a rollout whenever the sign_in.html ConfigMap's content changes.
-    - name: KAGENT_OAUTH2_PROXY_SIGNIN_TEMPLATE_CHECKSUM
-      value: '{{ include "kagent.oauth2ProxySignInHTML" . | sha256sum }}'
-
   # Mount custom CA certificate for TLS verification (if needed)
   # Add to extraVolumes:
   #   - name: custom-ca-cert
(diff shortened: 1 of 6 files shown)


helm/kagent/values.yaml
```diff
@@ -948,6 +948,11 @@ oauth2-proxy:
       mountPath: /templates
       readOnly: true
 
+  extraEnv:
+    # Forces a rollout whenever the sign_in.html ConfigMap's content changes.
+    - name: KAGENT_OAUTH2_PROXY_SIGNIN_TEMPLATE_CHECKSUM
+      value: '{{ include "kagent.oauth2ProxySignInHTML" . | sha256sum }}'
+
```
jjamroga: I think this is duplicated? extraEnv is also defined here. We should move this to the existing block
https://github.com/kagent-dev/kagent/pull/2532/changes#diff-3b5f638fea13e3428d305315de39c1cca6bf917d0a9c20b4a0d5364537b2c1f9R985-R994

onematchfox: Well caught 🙏 - my bad; this doesn't exist on the branch of the fork that I originally wrote this code against and I completely missed this when cherrypicking the change across to it's own branch. Addressed in https://github.com/kagent-dev/kagent/pull/2532/commits/2b4071a3ca121c8e8a344c33e224989c59503d43

ui/src/lib/loginRedirect.ts
```diff
@@ -0,0 +1,24 @@
...
+ * disguised variant of either (e.g. a backslash or a stripped tab/newline
+ * that the URL Standard normalizes into one of the above) would let a
+ * crafted `/login?rd=...` link send an authenticated session off to an
+ * attacker's site after sign-in.
+ */
+export function sanitizeRedirect(rd: string | undefined): string {
+  if (!rd) return "/";
+  try {
+    const url = new URL(rd, SENTINEL_ORIGIN);
+    if (url.origin !== SENTINEL_ORIGIN) return "/";
```
Charlesthebird:
```suggestion
    if (url.origin !== SENTINEL_ORIGIN || url.pathname.startsWith("//")) {
      return "/";
    }
```

Charlesthebird: `/..//host` and `/.//host` can result in a `"//"` without this.

Charlesthebird:
```suggestion
    if (url.origin !== SENTINEL_ORIGIN || url.pathname.startsWith("//")) {
      return "/";
    }
```

Charlesthebird: `/..//host` and `/.//host` can result in a `"//"` without this.

Charlesthebird (changes requested):
Thanks for adding this! Just had one change request.

ui/src/lib/loginRedirect.ts
onematchfox: Thanks and well spotted. FWIW I believe oauth2-proxy would reject these `rd` values anyway so a user would end up back at `/` rather than the host. But, always good to be on the safe side 😄

Charlesthebird (approved):

PR #2533, opened 2026-08-24
fix(ui): preserve post-login redirect target through `oauth2-proxy` sign-in

onematchfox:
#2532 - against main

### DIFF
--- a/helm/kagent/templates/_helpers.tpl
+++ b/helm/kagent/templates/_helpers.tpl
@@ -294,3 +294,28 @@ kagent.substrate.ateApiEndpoint.
 {{- include "substrate.atenetRouter.url" . -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Body of oauth2-proxy's custom sign_in.html template (see
+templates/oauth2-proxy-templates.yaml). Kept as its own named template, rather
+than inline in that ConfigMap, so oauth2-proxy.extraEnv in values.yaml can hash
+the content.
+
+oauth2-proxy renders this as its own Go html/template (not a Helm template) when
+it shows the sign-in page to an unauthenticated visitor -- e.g. a request to
+/agents/foo is served this page at /oauth2/sign_in?rd=%2Fagents%2Ffoo.
+`Redirect` is oauth2-proxy's template variable carrying that original
+destination (escaped with a Helm string-literal action so Helm emits it for
+oauth2-proxy to evaluate, instead of trying to evaluate it itself). It is
+forwarded to kagent's branded /login page.
+*/}}
+{{- define "kagent.oauth2ProxySignInHTML" -}}
+<!DOCTYPE html>
+<html>
+<head>
+  <meta http-equiv="refresh" content="0;url=/login?rd={{ "{{" }} or .Redirect "/" | urlquery {{ "}}" }}">
+  <script>window.location.href = "/login?rd={{ "{{" }} or .Redirect "/" | urlquery {{ "}}" }}";</script>
+</head>
+<body>Redirecting to login...</body>
+</html>
+{{- end -}}
--- a/helm/kagent/values.yaml
+++ b/helm/kagent/values.yaml
@@ -788,6 +788,9 @@ oauth2-proxy:
   # Cluster-specific OIDC settings - override these per deployment
   # These are set as env vars and referenced in args for easy patching
   extraEnv:
+    # Forces a rollout whenever the sign_in.html ConfigMap's content changes.
+    - name: KAGENT_OAUTH2_PROXY_SIGNIN_TEMPLATE_CHECKSUM
+      value: '{{ include "kagent.oauth2ProxySignInHTML" . | sha256sum }}'
     - name: OIDC_ISSUER_URL
       value: ""
     - name: OIDC_REDIRECT_URL
--- a/ui/src/pages/LoginPage.tsx
+++ b/ui/src/pages/LoginPage.tsx
@@ -1,8 +1,9 @@
 import { Button, Card, Typography } from "antd";
 import { useTheme } from "@emotion/react";
-import { useNavigate } from "react-router-dom";
+import { useNavigate, useSearchParams } from "react-router-dom";
 import { paths } from "@/router/routes";
-import { reauthenticationUrl } from "@/auth/reauthenticate";
+import { sanitizeRedirect } from "@/auth/loginRedirect";
+import { reauthenticationUrl, ssoStartUrl } from "@/auth/reauthenticate";
 import { useAuth } from "@/auth";
 
 const { Title, Paragraph, Text } = Typography;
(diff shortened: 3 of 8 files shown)


ui/src/lib/loginRedirect.ts
```diff
@@ -0,0 +1,24 @@
...
+ * disguised variant of either (e.g. a backslash or a stripped tab/newline
+ * that the URL Standard normalizes into one of the above) would let a
+ * crafted `/login?rd=...` link send an authenticated session off to an
+ * attacker's site after sign-in.
+ */
+export function sanitizeRedirect(rd: string | undefined): string {
+  if (!rd) return "/";
+  try {
+    const url = new URL(rd, SENTINEL_ORIGIN);
+    if (url.origin !== SENTINEL_ORIGIN) return "/";
```
Charlesthebird:
```suggestion
    if (url.origin !== SENTINEL_ORIGIN || url.pathname.startsWith("//")) {
      return "/";
    }
```

Charlesthebird (changes requested):
The UI on main has been updated to Vite + React Router instead of Next JS. So the UI folder changes will need to be moved around, and this will need to get the search params using [useSearchParams](https://reactrouter.com/api/hooks/useSearchParams).

onematchfox:
> The UI on main has been updated to Vite + React Router instead of Next JS. So the UI folder changes will need to be moved around, and this will need to get the search params using [useSearchParams](https://reactrouter.com/api/hooks/useSearchParams).

Done.

ui/src/pages/LoginPage.tsx
```diff
@@ -21,34 +22,48 @@ const { Title, Paragraph, Text } = Typography;
...
+    const rd = searchParams.get("rd");
+
+    // There are two ways to arrive here:
+    // 1. A reader who clicked "Session expired" in the header is still on the
+    //    page they were reading. 
+    // 2. A reader who typed `/agents/foo` while signed out never got there at
+    //    all: oauth2-proxy answered with its `sign_in.html`, which forwards to
+    //    `/login?rd=%2Fagents%2Ffoo`.
+    window.location.assign(
+      rd === null ? reauthenticationUrl(window.location) : ssoStartUrl(sanitizeRedirect(rd)),
```
Charlesthebird, line 45: **Level:** 🟠 Medium · Not Blocking

`AppHeader.tsx:45` does `navigate(paths.login)`, a pushState onto a standalone route, so `startSso` reads `/login` from `window.location` and this branch always produces `rd=%2Flogin`.

`LoginPage.test.tsx` sets `window.location.pathname = "/agents/foo"` while rendering at `/login`, which the router cannot produce, so the test passes on a case that never occurs. `playwright/tests/auth/auth-modes.spec.ts:101` already reaches that button, so the path is live.

The fix belongs in the unchanged `src/components/Structure/AppHeader.tsx:45`, so the destination arrives here as `rd`:

```tsx
onClick={() =>
  navigate(
    `${paths.login}?rd=${encodeURIComponent(location.pathname + location.search + location.hash)}`,
  )
}
```

---
*🤖 written by Claude*

helm/kagent/templates/_helpers.tpl
```diff
@@ -294,3 +294,28 @@ kagent.substrate.ateApiEndpoint.
...
+
+oauth2-proxy renders this as its own Go html/template (not a Helm template) when
+it shows the sign-in page to an unauthenticated visitor -- e.g. a request to
+/agents/foo is served this page at /oauth2/sign_in?rd=%2Fagents%2Ffoo.
+`Redirect` is oauth2-proxy's template variable carrying that original
+destination (escaped with a Helm string-literal action so Helm emits it for
+oauth2-proxy to evaluate, instead of trying to evaluate it itself). It is
+forwarded to kagent's branded /login page.
+*/}}
+{{- define "kagent.oauth2ProxySignInHTML" -}}
```
Charlesthebird, line 312: **Level:** 🟡 Low · Not Blocking

No helm-unittest suite in `helm/kagent/tests/` covers the new ConfigMap, and asserting `sign_in.html` contains `{{ or .Redirect "/" | urlquery }}` verbatim is what catches the `{{ "{{" }}` escaping breaking silently.

---
*🤖 written by Claude*

helm/kagent/values.yaml
```diff
@@ -788,6 +788,9 @@ oauth2-proxy:
   # Cluster-specific OIDC settings - override these per deployment
   # These are set as env vars and referenced in args for easy patching
   extraEnv:
+    # Forces a rollout whenever the sign_in.html ConfigMap's content changes.
```
Charlesthebird, line 791: **Level:** 🟡 Low · Not Blocking

The checksum sits in the list line 788 tells operators to override per deployment, so replacing `extraEnv` drops the rollout trigger without a word.

```suggestion
    # Forces a rollout whenever the sign_in.html ConfigMap's content changes.
    # Keep this entry if you override extraEnv, or template changes will not restart the proxy.
```

---
*🤖 written by Claude*

Charlesthebird (commented):

---

Nothing blocks merge. The core fix works end to end, and both points from the last round are addressed. One Medium: the no-`rd` fallback branch is unreachable in the real app, and the test covering it asserts a state the router cannot produce.

### What I checked

- ✅ `sanitizeRedirect` returns `/` for 17 vectors beyond its own tests — `///host`, `////host`, `\\host`, `\/host`, a leading space, tab or newline before `//host`, `https:host`, `https:/\host`, `/..//host`, `//host\@kagent.invalid/`, and `http://kagent-login-redirect.invalid//evil`.
- ✅ `/%2F%2Fhost` stays percent-encoded through the parser, so it resolves as a path segment rather than a host.
- ✅ Rendered the named template in a minimal chart: `sign_in.html` emits `{{ or .Redirect "/" | urlquery }}` verbatim, and the checksum env var resolves to `41963eed…`.
- ✅ oauth2-proxy 10.7's `deployment.yaml:257` runs `tpl` over `extraEnv`, while `podAnnotations` at line 51 is plain `toYaml` — a `checksum/config` annotation would have shipped the literal `{{ include … }}`.
- ✅ The ConfigMap is volume-mounted at `values.yaml:751-759`, and oauth2-proxy parses `/templates` once at startup.
- ✅ Executed the template through Go's `html/template`: `urlquery` is not double-applied in the `content=` attribute or the `<script>` string, and `/a"><script>alert(1)</script>` emits as `%2Fa%22%3E%3Cscript%3E…`.
- ✅ oauth2-proxy's own `IsValidRedirect` rejects `//` and `/\` independently, so a miss in the UI guard is caught downstream.
- ✅ `AuthProvider.tsx:81` calls `startReauthentication()` while still on the original page, so the automatic path's `rd` was already correct and stays untouched.

---
*🤖 written by Claude*

ui/src/pages/LoginPage.tsx
```diff
@@ -21,34 +22,48 @@ const { Title, Paragraph, Text } = Typography;
...
   };
 
   const copy =
     status === "expired"
       ? {
-          blurb: "Your session has expired. Sign in again to continue.",
-          action: "Sign in with SSO",
-          onClick: startSso,
-        }
+        blurb: "Your session has expired. Sign in again to continue.",
```
Charlesthebird, line 52: [nit]

This ternary's three object literals were reindented away from the surrounding style (compare `AgentTemplateForm.tsx:379` and `mcpServerRequest.ts:233`), which accounts for 30 of the file's 45 changed lines.

---
*🤖 written by Claude*

ui/src/pages/LoginPage.tsx
```diff
@@ -21,34 +22,48 @@ const { Title, Paragraph, Text } = Typography;
...
+  //
+  // The forwarded `rd` is sanitized: it reaches this page through a query string, so a
+  // crafted `/login?rd=https://evil.example.com` link is a URL anybody can send.
   const startSso = () => {
-    window.location.assign(reauthenticationUrl(window.location));
+    const rd = searchParams.get("rd");
+
+    // There are two ways to arrive here:
+    // 1. A reader who clicked "Session expired" in the header is still on the
+    //    page they were reading.
```
Charlesthebird, line 40: [nit]

Trailing whitespace, the only such line in `ui/src`.

```suggestion
    //    page they were reading.
```

---
*🤖 written by Claude*

ui/src/pages/LoginPage.tsx
```diff
@@ -21,34 +22,48 @@ const { Title, Paragraph, Text } = Typography;
...
   const enterApp = () => navigate(paths.dashboard);
   // Carries `rd`, so signing in returns the reader to where they were rather than to
   // whatever the proxy defaults to. Dropping it is how re-authenticating used to cost
   // somebody the page they were reading.
+  //
+  // The forwarded `rd` is sanitized: it reaches this page through a query string, so a
+  // crafted `/login?rd=https://evil.example.com` link is a URL anybody can send.
   const startSso = () => {
-    window.location.assign(reauthenticationUrl(window.location));
+    const rd = searchParams.get("rd");
```
Charlesthebird, line 36: [nit]

`?rd=` yields `""` rather than `null`, so it takes the sanitize branch below and lands on `/` — harmless, though the `rd === null` split reads as if it covers that case.

---
*🤖 written by Claude*

Charlesthebird (approved):
Thanks for updating this! It looks good to me. 

I kicked off the automated review above, and the one medium test note might be worth addressing, but I don't think it's a blocker.

peterj (approved):

PR #2535, opened 2026-08-24
fix(build): pin kube-api-linter plugin to the workspace Go toolchain

sbx0r:


go/Makefile
```diff
@@ -61,6 +61,7 @@ vet: ## Run go vet.
 	go vet ./...
 
 KAL_SO = $(LOCALBIN)/kube-api-linter.so
+GO_TOOLCHAIN := $(shell go env GOVERSION)
```
mesutoezdil, line 64: root makefile already exports GOTOOLCHAIN. why pin it again here?

mesutoezdil (commented):
1question below.

go/Makefile
sbx0r: if you follow the docs ([CLAUDE.md:98](https://github.com/kagent-dev/kagent/blob/2405b6d94aacd162cb091ab1d653831b5f737533/CLAUDE.md#L98), [SKILL.md:57](https://github.com/kagent-dev/kagent/blob/2405b6d94aacd162cb091ab1d653831b5f737533/.claude/skills/kagent-dev/SKILL.md#L57)) you hit this:

```
$ make -C go lint
...
fatal error: runtime: no plugin module data
...

$ go version -m bin/golangci-lint       # go1.27.0
$ go version -m bin/kube-api-linter.so  # go1.26.7
```

[Makefile:79](https://github.com/kagent-dev/kagent/blob/2405b6d94aacd162cb091ab1d653831b5f737533/Makefile#L79) never gets read here, so `GOTOOLCHAIN=auto`. That's fine everywhere except the `$(KAL_SO)`

so one source is lying 😜 lmk which one you want fixed.

EItanya: I think this is fine to unblock, our `Makefile`s probably need cleaning up anyway

EItanya (approved):

PR #2541, opened 2026-08-24
fix(adk): extract text from MCP embedded resource content blocks

mesutoezdil:
extractFunctionResponseContent only reads the top-level text field on each MCP content item. Embedded resource blocks (type: resource) carry their text under resource.text instead, so those blocks were silently dropped and the model never saw that part of the tool result. Falls back to resource.text when no top-level text field is present.


supreme-gg-gg (commented):
Would you mind describing if this fixes a bug in kagent? Codex seems to be telling me that ADK upstream already strips out the embedded resources from MCP responses

https://github.com/google/adk-go/blob/b264039aaec43baedc123e5b9a0cf87681d0bbca/tool/mcptoolset/tool.go#L156-L165

So I'm curious if this will fix anything in Kagent if `mcp.EmbeddedResource` is already dropped in the above code

mesutoezdil:
you're right.
so this fix does not take effect on the current adk-go version. closing this, thanks

PR #2542, opened 2026-08-24
fix(embedding): stop forcing dimensions on OpenAI-compatible embedding calls

mesutoezdil:
The openai, azure_openai, and foundry embedding paths in both the Go and Python runtimes asked for exactly 768 dimensions unconditionally. Providers that only offer other sizes, such as Bedrock Titan behind an OpenAI-compatible gateway, reject the request outright, even though the existing truncate and normalize step already handles anything larger than 768. Drops the dimensions parameter on these paths, since the truncate and normalize step already covers it.


EItanya (approved):

supreme-gg-gg (approved):
Non-blocking note: relying on truncation should work fine for most models, including OpenAI’s embedding models. However, a custom model behind an OpenAI-compatible gateway may accept the request but lose semantic quality after truncation if it was not trained with Matryoshka Representation Learning, as the existing code comment notes. Since a lot of modern embedding models, including many open-source ones, support MRL, I don’t think this should block the PR, but just placing a note here

PR #2543, opened 2026-08-24
refactor: make the v2 stack assemblable outside package main

jjamroga:
## Problem

`core/cmd/controller-v2` is the only place the v2 stack is assembled, and it's `package main`. Two of its dependencies are internal, so nothing outside this module can reproduce it:

- `api/database` exports the `Client` interface but no constructor — `Connect`/`NewClient` are in `core/internal/database`. Every v2 component takes a store, so there's nothing to hand them.
- `core/internal/grpcserver` is internal, so the server that registers those services, including A2A, is unavailable.

The visible consequence: `core/pkg/app` is the only entry point available externally, and it never sets `A2AHandler`. `grpcserver` registers A2A only when that field is non-nil, so a controller built on `app.Start` can create AgentInstances but not talk to them — `Unimplemented: unknown service lf.a2a.v1.A2AService`.

## Why restructure it this way

This is a restructuring, not a feature, and it doesn't add tests. What it could enable is testing the substrate-independent half of the v2 stack without the substrate-dependent machinery.

Today `TestAgentInstanceInteraction` can't run until CI has stood up a kind cluster, installed Substrate, bootstrapped the CA/JWT pools with `kubectl-ate`, brought up a gVisor WorkerPool and built a golden actor. Much of what it then asserts isn't substrate behaviour: AgentInstance lifecycle, the ownership check, A2A routing by instance headers, `DefaultMethodPolicies` on the A2A methods, task persistence.

Those are reachable with a Postgres and two fakes — `a2agateway`'s dialer is one method, `agentinstance`'s actor client is six — leaving the cluster e2e to cover what genuinely needs a cluster. That's also what `a2agateway/gateway.go` already anticipates: *"a standalone gateway can register the same handler on its own server later."*

Two honest caveats. This PR is only the first half: the composition still lives in `func main()`, so a test would have to duplicate that wiring. Extracting it into an exported constructor would finish the job — happy to do that here or as a follow-up. And tests inside this module are unaffected either way; `go/core/test/e2e` can already import `core/internal/...`.

## Change

No behaviour change.

- **`core/pkg/database`** — thin re-export of `Connect`, `NewClient`, `ResolveURL`, `PostgresConfig` aliased. Implementation stays internal.
- **`core/internal/grpcserver` → `core/pkg/grpcserver`** — move plus import updates in `core/pkg/app` and `core/cmd/controller-v2`.

`Config` keeps fields typed on internal packages; that's fine, since omitting a field never requires importing its type and every such field is optional. `SystemService` and `MethodPolicies` self-default, so `DefaultMethodPolicies()` still applies and A2A authorization is unchanged.

## Testing

`go build ./...` and `go vet` clean; `go test ./core/pkg/grpcserver/` passes, with the existing tests moving with the package.

`core/internal/service/{session,task}` stay internal, so `SessionService`/`TaskService` still can't be registered externally — both optional; happy to move them here or in a follow-up.


jjamroga:
Superseded by #2544, which is a much smaller change covering the same need: rather than exporting `grpcserver` and adding a public store constructor, `ExtensionConfig` gains an `A2AHandler` field that an extension can populate with `a2agateway.New`, using the `DbClient` it already receives in `BootstrapConfig`.

The integration-testing argument here still stands on its own — assembling the v2 stack outside `package main` would allow testing AgentInstance lifecycle, ownership checks and A2A routing with fakes for the dialer and actor client, no cluster or Substrate required. Happy to revive this separately if that is wanted; it just should not be bundled with the smaller fix.

PR #2365, opened 2026-07-30
feat(helm): add affinity and topologySpreadConstraints on Kagent controller and ui 

anthonyhaussman:
Adds support for setting `affinity` and `topologySpreadConstraints` on the `controller` and `ui` deployments via `values.yaml`. This allows for more advanced scheduling of these components.

PR #2553, opened 2026-08-25
feat(helm): allow scheduling the bundled PostgreSQL pod

CarlosLanderas:
## Summary

- The `controller` and `ui` deployments expose `nodeSelector`, `tolerations` and
`affinity`, but the bundled PostgreSQL Deployment rendered by
`templates/postgresql.yaml` has no scheduling fields at all
- This makes the bundled database impossible to place. On a cluster with a
dedicated, tainted node pool it cannot follow the rest of the control plane, and
a single-replica PVC-backed database left on an autoscaled pool gets drained
whenever that pool scales in — taking the controller down with it
- Adds `nodeSelector`, `tolerations` and `affinity` to
`database.postgres.bundled`, rendered with the same `{{- with }}` form and in the
same order `controller-deployment.yaml` already uses
- Values documentation copies the wording already used by the `controller` and
`ui` blocks, so the three read consistently
- Defaults stay empty, so rendered output is unchanged for existing installs

Follows the same shape as #2213 (`nodeSelector`/`tolerations` for the
`kagent-tools` subchart) and #2365 (`affinity`/`topologySpreadConstraints` for
controller and ui), which closed the equivalent gap for other components.

## Test plan

- [x] `helm unittest helm/kagent` passes — 274 tests, including 4 new ones in
`tests/postgresql_test.yaml` covering the unset, `nodeSelector`, `tolerations`
and `affinity` cases
- [x] `helm template` with `database.postgres.bundled.nodeSelector` and
`tolerations` set renders both fields on the `kagent-postgresql` Deployment pod
spec
- [x] `helm template` with the values unset renders no scheduling fields on that
pod spec, confirming existing installs are unaffected

### DIFF
--- a/helm/kagent/templates/postgresql.yaml
+++ b/helm/kagent/templates/postgresql.yaml
@@ -57,6 +57,18 @@ spec:
       securityContext:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with $pg.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with $pg.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with $pg.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: postgresql
           image: {{ include "kagent.postgresql.image" . }}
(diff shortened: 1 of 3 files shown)


EItanya (approved):

CarlosLanderas:
@EItanya , just wondering what is necessary for changes to reach a release. 
I saw this was not shipped with 0.10 helm chart.

Thanks!

Issue #2366, opened 2026-07-30
Proposal: split agent configuration into Harness and AgentTemplate CRDs


EItanya:
# Proposal: split agent configuration into Harness and AgentTemplate CRDs

## Summary

This proposal separates authored agent behavior from the runtime that executes
it:

- `Harness` defines a reusable runtime implementation and infrastructure policy.
- `AgentTemplate` defines one logical agent and its ordinary MCP or
  AgentTemplate-backed tools.

The root AgentTemplate is always the public entrypoint. AgentTemplate-backed
tools recursively form the owned composition. A template with no such tools is
the ordinary single-agent case; composition does not require another resource
kind or creation API.

```text
Harness + rooted AgentTemplate tree
                  │
              preparation
                  │
        one prepared runtime plan
                  │
        CreateAgentInstance (gRPC)
                  │
one AgentInstance + one public A2A conversation
```

The public invariant is:

```text
one AgentInstance = one rooted AgentTemplate tree = one public conversation
```

The controller may use one or more internal runtime boundaries to satisfy the
template's isolation requirements. Processes, pods, Actors, and Substrate
topology do not appear in the authored AgentTemplate API.

## Why change the current API?

The current API combines reusable behavior, runtime configuration, compute
lifecycle, and conversation state. The split lets teams reuse behavior across
runtime offerings, validate complete AgentTemplate trees before use, and keep live
conversation state outside Kubernetes reconciliation.

The proposed split gives each concern one owner:

| Concern | Owner |
|---|---|
| Runtime implementation and infrastructure policy | `Harness` CRD |
| Model, prompt, plugins, tools, and skills | `AgentTemplate` CRD |
| Preparation and runtime compilation | Kagent controller |
| Live identity and lifecycle | Kagent gRPC `AgentInstance` API |
| Conversation and Tasks | A2A v1 gRPC |

## `Harness`

A Harness defines a reusable runtime implementation and its infrastructure
policy.

Rules:

- Exactly one typed runtime variant is present.
- The controller derives runtime capabilities; authors do not claim them.
- Image overrides must be digest-pinned.
- Each environment entry contains either a literal or a credential reference.
- Preparation retains credential references but never reads secret values.
- Omitted admission accepts no AgentTemplates.
- Every runtime variant consumes the same internal resolved-bundle contract.
- A Harness creates no compute merely by existing.

## `AgentTemplate`

An AgentTemplate defines portable agent behavior and its available tools. A
[...]

PR #2559, opened 2026-08-25
feat(cli): add API v2 AgentInstance workflows

codyhartsook:
## Summary

- Adds API v2 CLI commands for AgentTemplate discovery and AgentInstance lifecycle and invocation
- Adds typed AgentInstance and upstream A2A clients over the shared gRPC connection
- Connects through configured gRPC or a temporary controller port-forward, with optional TLS
- Removes the legacy client-side agent project workflow while retaining `mcp`

Note that of the 64 files changed, 35 were removals from the legacy CLI project workflow ; no server-side APIs or controllers are removed.

This is the first of three related CLI PRs:

1. [#2559](https://github.com/kagent-dev/kagent/pull/2559) : API v2 CLI foundation: template/instance discovery, create/delete, invoke/streaming, JSON output, and quarantining legacy commands.
2. [#2605](https://github.com/kagent-dev/kagent/pull/2605) : TUI: browse AgentInstances, load history, and chat through the same A2A path as invoke.
3. [#2606](https://github.com/kagent-dev/kagent/pull/2606) : Cleanup: simplify `main.go` and align the command structure with agentregistry.

Suspend and resume are deliberately not exposed by this CLI work.

Related to #2366 and the K13 CLI cutover in `docs/plans/api-v2-execution-plan.md`.

## Test plan

- [x] `make -C go test`
- [x] `go test ./core/cli/...`
- [x] `make -C go lint`
- [x] `make -C go build`
- [x] AgentTemplate and AgentInstance CLI smoke test against a local installation


go/core/cli/internal/tui/dialogs/mcp_server_wizard.go
EItanya, line 1: This API is still technically supported. Is there a way we can retain this functionality?

go/core/cli/internal/cli/agentinstance/lifecycle.go
```diff
@@ -0,0 +1,167 @@
...
+
+	clientSet := cfg.Config.Client()
+	defer func() {
+		err = errors.Join(err, clientSet.Close())
+	}()
+	return deleteAgentInstance(ctx, clientSet.AgentInstance, cfg, format, out)
+}
+
+func prepareCreateCfg(cfg *CreateCfg) error {
+	if cfg.Harness == "" {
```
EItanya: The following just merged: https://github.com/kagent-dev/kagent/commit/76cde41f9d635fda961a898efdf5c9b4b8925a97

We can get this out of clientside code

go/core/cli/internal/config/config.go
EItanya, line 1: Can we please delete this file and viper, it's just complicated.

EItanya (changes requested):
Initial review, I'm still looking for more simplifications, but this is a good start

go/core/cli/internal/cli/agentinstance/lifecycle.go
codyhartsook: Yep, done. I removed the client-side create/delete validation and left that to the shared Protovalidate interceptor. The CLI just generates a request ID when one is not supplied.

go/core/cli/internal/config/config.go
codyhartsook: I removed the internal config package and Viper/ishell from the CLI. Root flags now pass through a small explicit connection options struct. Viper still shows up indirectly through the lint tooling, but the CLI no longer uses it.

codyhartsook: note that the code line count decreased but added some file change churn, fyi to reviewers

go/core/cli/internal/tui/dialogs/mcp_server_wizard.go
codyhartsook: tracing this again, the wizard was only used by the removed add-mcp command and wrote to the legacy kagent.yaml project workflow. From my understanding, I think it should remain deleted, with AgentTemplate manifests as the mcp binding but open to suggestions

go/core/cli/internal/cli/connection/connection.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	if strings.IndexFunc(o.UserID, unicode.IsSpace) >= 0 {
+		return errors.New("caller identity must not contain whitespace")
+	}
+	return nil
+}
+
+type connectionRuntime struct {
+	checkServer    func(context.Context, *client.ClientSet) error
+	commandContext func(context.Context, string, ...string) *exec.Cmd
```
EItanya: I really really hate callbacks unless absolutely necessary. Can you explain this component and why we need these?

go/core/test/e2e/interaction_test.go
```diff
@@ -85,6 +87,101 @@ func TestAgentInstanceAskUserSurvivesSuspension(t *testing.T) {
...
+	}
+	moduleRoot := filepath.Clean(filepath.Join(filepath.Dir(source), "../../.."))
+	binary := filepath.Join(t.TempDir(), "kagent")
+	command := exec.CommandContext(t.Context(), "go", "build", "-o", binary, "./core/cli/cmd/kagent")
+	command.Dir = moduleRoot
+	if output, err := command.CombinedOutput(); err != nil {
+		t.Fatalf("build CLI: %v\n%s", err, output)
+	}
+	return binary
+}
```
EItanya: I'd rather not build from the tests, can we use pre-built?

go/core/test/e2e/interaction_test.go
```diff
@@ -85,6 +87,101 @@ func TestAgentInstanceAskUserSurvivesSuspension(t *testing.T) {
 	}
 }
 
+func TestCLIAgentInstanceDiscoveryAndInvoke(t *testing.T) {
```
EItanya: All e2e tests should be prefixed with `TestE2E`

go/core/test/e2e/interaction_test.go
```diff
@@ -85,6 +87,101 @@ func TestAgentInstanceAskUserSurvivesSuspension(t *testing.T) {
...
+	target := interactionTarget(t)
+	fixture := newInteractionFixture(t, target, startInteractionMock(t))
+	binary := buildKagentCLI(t)
+	baseArgs := []string{
+		"--kagent-grpc-url", target,
+		"--kagent-grpc-tls=false",
```
EItanya: Why do these flags need the `kagent` prefix?

go/core/test/e2e/interaction_test.go
EItanya: Can we group all of the cli tests in the same file

go/core/cli/internal/config/config.go
```diff
@@ -1,97 +0,0 @@
...
-	KAgentURL            string        `mapstructure:"kagent_url"`
-	KAgentGRPCURL        string        `mapstructure:"kagent_grpc_url"`
-	KAgentGRPCTLS        bool          `mapstructure:"kagent_grpc_tls"`
-	KAgentGRPCCAFile     string        `mapstructure:"kagent_grpc_ca_file"`
-	KAgentGRPCServerName string        `mapstructure:"kagent_grpc_server_name"`
-	Namespace            string        `mapstructure:"namespace"`
-	OutputFormat         string        `mapstructure:"output_format"`
-	Verbose              bool          `mapstructure:"verbose"`
-	Timeout              time.Duration `mapstructure:"timeout"`
-}
```
EItanya, line 25: I think we still need this config to be embedded in the other cmd configs.

```
type AgentInstanceConfig struct {
  BaseConfig
}
```

Something like the above

go/core/cli/internal/cli/connection/connection.go
codyhartsook: Yeah, fair point. These callbacks mainly exist for tests and make the production path harder to follow. I’ll remove the runtime wrapper, call the real functions directly.

codyhartsook: Removed these

go/core/test/e2e/interaction_test.go
codyhartsook: Done, the E2E target builds the CLI once and passes it through KAGENT_E2E_CLI. CI uses the same path

go/core/test/e2e/interaction_test.go
codyhartsook: Updated the two E2E tests added by this PR to use TestE2E

go/core/test/e2e/interaction_test.go
codyhartsook: Removed the prefix from the new gRPC flags: --grpc-url, --grpc-tls, --grpc-ca-file, and --grpc-server-name.

go/core/test/e2e/interaction_test.go
codyhartsook: both CLI E2E tests are now in cli_catalog_lifecycle_test.go

go/core/cli/internal/config/config.go
codyhartsook: I ended up addressing this in PR #2606 by reading shared connection and output flags at each command boundary, so BaseConfig would no longer be needed. Would it be okay to leave this as-is here and handle it in that follow-up or prefer we address here?

EItanya (approved):

PR #2562, opened 2026-08-25
refactor: type AgentInstance identifiers as UUIDs

EItanya:
## Summary

- represent AgentInstance, share, checkpoint, and A2A context identifiers as `uuid.UUID` in the database layer
- migrate the corresponding PostgreSQL columns from `TEXT` to native `UUID` columns
- generate UUID-typed SQLC models and query parameters, converting only at protobuf/service boundaries
- preserve empty pagination cursors while comparing non-empty cursors as UUIDs
- update database tests to use valid UUID fixtures

## Testing

- `go test ./...`
- `make -C go lint`


EItanya:
Drafted this as I'm not convinced it's worth it

PR #2569, opened 2026-08-25
feat(ui): rewrite the web interface on Vite, and give conversations a first-class API

Charlesthebird:
---
*🤖 written by Claude (start)*

### Summary

Rebuilds the web interface on **Vite + React 19** — React Router, SWR, antd 6, Emotion. It is a static bundle behind nginx with no server process; settings come from `window.environmentVariables`, rewritten by the container on every start, so one image serves every deployment.

The application API is reached over **gRPC-Web**: `grpcserver.WebHandler` wraps the existing `*grpc.Server`, and the HTTP server routes gRPC-Web requests to it ahead of its middleware chain.

> [!NOTE]
> **Reading the diff.** 616 of the 696 files are `ui/`, which is the rewrite's tree — read it as a new app, not as a diff. The other 80 are three things:
> - **`helm/`** — the UI pod stops running a Next.js server and becomes nginx serving a static bundle, so `nginx.conf`, `supervisord.conf`, `ui-deployment.yaml` and the UI values change together with their tests.
> - **`go/` and `proto/`** — the gRPC-Web seam (`grpcserver/grpcweb.go`, `httpserver/server.go`, `app.go`) and the five `AgentInstance` changes below, plus generated proto and sqlc output.
> - **`CLAUDE.md` and `.nvmrc`** — the repo guide's UI section, and the pinned Node version.

### Testing this PR

One command builds a Kind cluster and installs **this checkout** on it — controller, UI and agent runtime all built from the working tree, over the chart's published images. It ends holding two port-forwards, the UI on 8080 and the controller on 8083, so the last thing it prints is a working URL and `cd ui && yarn dev` needs nothing configured.

```sh
./scripts/setup-cluster/setup-cluster.sh     # ~25 min, mostly image builds
```

**http://localhost:8080**

> [!TIP]
> The script also leaves one agent on the cluster — an `assistant` template on a `kagent` harness — so **Agents** has something in it and you can send a message straight away, without creating anything first.

> [!WARNING]
> `make create-kind-cluster && make helm-install` does **not** work, and fails silently five different ways — including that the chart installs published images, so none of your changes are on the cluster while everything looks healthy. [`scripts/setup-cluster/README.md`](https://github.com/kagent-dev/kagent/blob/charlesthebird/uiRewrite/scripts/setup-cluster/README.md) covers each one, and the dev-server loop for iterating.

### UI Extensions

The app declares vendor extension points anyone can use to add to it or restyle it, all in one configuration object:

| Point | Contributes |
| --- | --- |
| `navItems` | Sidebar entries, positioned by `order` |
| `navOverrides` | Hide, rename, re-order or redirect the app's own entries |
| `routes` | Whole pages, merged into the router |
| `routeHandles` | Shell data attached to the app's own routes |
| `slots` | Components at named points inside existing pages |
| `formFields` | Extra fields on existing forms |
| `tableColumns` | Extra columns on existing tables |
| `api` | Operation and endpoint overrides, payload transforms |
| `providers` | App-level React providers |
| `theme` | Colour, radius and font tokens |
| `shell` | Whole shell regions replaced outright |
| `branding` | Name and mark wherever the shell states its identity |
| `providerIcons` | Provider icons for the model-config form |
| `agentLinks` | Agent destinations |

Installing one is two edits: build a `VendorExtensionConfig`, then point `src/vendorExtensions/activeConfig.ts` at it. Overriding theme tokens restyles the app's own components, not just the extension's.

> [!NOTE]
> 📖 [`ui/docs/vendor-extensions.md`](https://github.com/kagent-dev/kagent/blob/charlesthebird/uiRewrite/ui/docs/vendor-extensions.md) — every extension point and what it receives. Worth reading before reviewing the `vendorExtensions/` tree.

### Substrate

The pages follow the CRDs. An **Agent** is derived, not a resource — a `Harness` × `AgentTemplate` pair read from `AgentTemplate.status.harnesses[]`, so there is no "New agent" button. The landing page explains the four concepts over three tabs; an agent's page lists its conversations, and a conversation is an `AgentInstance`.

Five additive server-side changes, none affecting an existing caller:

| Change | Note |
| --- | --- |
| `AgentInstance` gains a `name` | End to end, via `UpdateAgentInstanceName`. A column, so the write touches only the column. Renamed from the conversations table, the rail's action menu or the details modal; the rail lists newest first. |
| `ListAgentInstances` takes a query | Narrows to one agent, resolved through `prepared_revision`. |
| A2A gateway reads in any state | A suspended conversation still opens. |
| A share resolves to its owner | The token widens what the reader's account may read. |
| `app.go` defaults the A2A gateway | Used when no extension supplies one; an extension's handler still wins. |

### Test Coverage

```sh
cd ui && yarn typecheck && yarn lint && yarn test && yarn test:pw
cd go && go test ./...
make proto-generate && make -C go sqlc-generate   # then confirm `git status` is clean
```

385 unit tests. The browser suite runs in **Chromium and Firefox**, plus a Chromium pass with the example extension installed, and in parallel in CI.

### Follow Ups

- **The browser suite runs against the dev server, not the built image and a real controller.** The `UI Playwright E2E` workflow that did that is removed here, along with the harness it drove (`playwright/scripts/setup.sh`, `playwright/mocks/server.mjs`); restoring it is follow-up work. `yarn test:pw:live` covers a few journeys against a real cluster in the meantime.
- **Storybook and the Chromatic workflow are removed**, since the rewrite ships no stories.
- [`ui/playwright/DEFERRED.md`](https://github.com/kagent-dev/kagent/blob/charlesthebird/uiRewrite/ui/playwright/DEFERRED.md) lists the rest, and the surface each spec waits on. Nothing is committed as a skipped test.

> [!NOTE]
> `go test ./...` fails one pre-existing test on macOS — `TestFetchSourceReusesExistingMaterialization`, `/var` vs `/private/var`, in a package this change does not touch.

### Upgrade Notes

> [!CAUTION]
> **Breaking.** The chart no longer sets `NEXT_PUBLIC_BACKEND_URL`, `BACKEND_INTERNAL_URL` or `BACKEND_GRPC_URL`, and drops `ui.backendInternalUrl`, `ui.backendGrpcUrl` and `ui.volumes.nextjsCache`.

The oauth2-proxy `skip-auth-regex` now names `/assets/` and `env-config.js` instead of the Next.js paths, which no longer exist.

**Also included, unrelated to the rewrite:** `helm/tools/grafana-mcp` now passes `-allowed-hosts`. The server rejects any Host header it was not told about, so one reached over the cluster network answered the MCP handshake with `Forbidden` — the `RemoteMCPServer` sat `Accepted=False` and every agent using it failed to build its tool set. Found while testing the tools pages; happy to split it out if preferred.

---
*🤖 written by Claude (end)*


Charlesthebird:
https://github.com/user-attachments/assets/eb7ff280-b569-425e-a07c-30bdd1cae417



---
*🤖 written by Claude*

Charlesthebird:
https://github.com/user-attachments/assets/502ad3ad-4b6c-4085-8159-6b027e599aad



---
*🤖 written by Claude*

go/core/v2/agentinstance/grpc.go
```diff
@@ -48,6 +49,14 @@ func (s *grpcServer) ListAgentInstances(ctx context.Context, request *apiv1alpha
 	}, nil
 }
 
+func (s *grpcServer) RenameAgentInstance(ctx context.Context, request *apiv1alpha1.RenameAgentInstanceRequest) (*apiv1alpha1.RenameAgentInstanceResponse, error) {
```
EItanya: Do we still need this? I think we're doing auto-suspend/resume

go/api/database/client.go
```diff
@@ -124,19 +140,38 @@ type Client interface {
...
+	// digest is stored.
+	GetAgentInstanceShareByTokenHash(context.Context, []byte) (*AgentInstanceShare, error)
 	DeleteAgentInstanceShare(context.Context, string, string, string) error
 	// CreateAgentInstanceTask reserves the instance's single active-task slot.
 	CreateAgentInstanceTask(context.Context, string, []byte, *a2a.Task) (*a2a.Task, bool, error)
 	GetActiveAgentInstanceTask(context.Context, string) (*a2a.Task, error)
 	// InterruptActiveAgentInstanceTask fails the expected task and records an
 	// interruption. It returns false if that task is no longer active.
 	InterruptActiveAgentInstanceTask(context.Context, string, string) (bool, error)
+	// AbandonActiveAgentInstanceTask cancels the expected task, releasing the
```
EItanya: This is A LOT of new queries, do we need all of these just to get the UI off the ground. These are the sort of thing that are harder to roll back

go/core/internal/grpcserver/agenttemplate.go
```diff
@@ -0,0 +1,151 @@
...
+	apiv1alpha1.UnimplementedAgentTemplateServiceServer
+	service         *agenttemplateservice.Service
+	maxMessageBytes int
+}
+
+func newAgentTemplateServer(service *agenttemplateservice.Service, maxMessageBytes int) *agentTemplateServer {
+	return &agentTemplateServer{service: service, maxMessageBytes: maxMessageBytes}
+}
+
+func (s *agentTemplateServer) ListAgentTemplates(ctx context.Context, request *apiv1alpha1.ListAgentTemplatesRequest) (*apiv1alpha1.ListAgentTemplatesResponse, error) {
```
EItanya, line 26: All of the CRUD functionality is identical, can we use generics?

go/core/internal/grpcserver/server.go
```diff
@@ -132,6 +136,15 @@ func New(config Config) (*Server, error) {
 	if config.AgentService != nil {
 		apiv1alpha1.RegisterAgentServiceServer(grpcServer, newAgentServer(config.AgentService, config.MaxMessageBytes))
 	}
+	if config.AgentTemplateService != nil {
```
EItanya, line 139: if `x == nil` is a claudism, this can't be `nil` here

go/core/internal/httpserver/server.go
```diff
@@ -194,3 +203,18 @@ func wsAuthQueryMiddleware(next http.Handler) http.Handler {
...
+// The split happens outside the router's middleware chain on purpose. That chain
+// authenticates, rewrites content types and maps errors for the REST-shaped
+// handlers, none of which a gRPC-Web frame survives — and it does not need any of
+// it, because the gRPC server authenticates in its own interceptors. Routing these
+// past the chain rather than through it keeps one protocol from being handled by
+// the other's conventions.
+func (s *HTTPServer) withGrpcWeb(next http.Handler) http.Handler {
+	if s.config.GrpcWebRouter == nil {
+		return next
+	}
```
EItanya: claudism, no need for this

go/core/internal/httpserver/server.go
```diff
@@ -27,6 +27,15 @@ type ServerConfig struct {
...
 	DbClient      dbpkg.Client
 	Authenticator auth.AuthProvider
+
+	// GrpcWebRouter composes this server's handler so that gRPC-Web calls reach the
+	// gRPC services and everything else reaches the router. Supplied by the gRPC
+	// server (see grpcserver.Server.WebHandlerOr), which owns the rule — there is
+	// more than one binary serving HTTP beside that server, and a second copy of
+	// the rule would drift.
+	//
+	// Optional: left nil, this server behaves exactly as it did before one existed.
```
EItanya: Is it actually `nil` though, don't we pass it every time?

go/core/internal/service/agenttemplate/service.go
EItanya, line 1: This is also identical to `harness.Service`. Can we use generics?

go/core/pkg/auth/share.go
```diff
@@ -5,9 +5,26 @@ import "context"
...
+	// chat path, an instance is the conversation itself — and a single field would
+	// have the A2A gateway matching an id that named a session.
+	AgentInstanceID string
+}
+
+// IsForAgentInstance reports whether this share grants access to the named instance.
+//
+// Asked rather than assumed: a session share reaching the A2A gateway must not be
+// treated as authority over an instance that happens to share its id space.
+func (s *ShareContext) IsForAgentInstance(instanceID string) bool {
```
EItanya, line 26: Sessions are dead, just replace this

go/core/internal/service/agenttemplate/service.go
EItanya: A concrete way to keep this small is the prototype pattern used by skv2: parameterize the service with `T client.Object` and `L client.ObjectList`, pass `&v1alpha3.Harness{}` / `&v1alpha3.HarnessList{}` (or the AgentTemplate equivalents), and allocate with `prototype.DeepCopyObject().(T)`. See https://github.com/solo-io/skv2/blob/main/pkg/client/client.go. That should share the CRUD mechanics without reflection or constructor plumbing; only the resource-specific spec/status transition needs a small callback.

proto/kagent/api/v1alpha1/harnesses.proto
```diff
@@ -0,0 +1,87 @@
...
+// Harness here is a reusable runtime (one of the kagent, codex or claude
+// adapters, plus a Substrate worker pool and snapshot policy) that admits many
+// AgentTemplates through a label selector; the v2 reconciler pairs the two in
+// go/core/v2/controller/collections.go. The names collide but the kinds do not,
+// so this is a separate service rather than more RPCs on AgentService.
+service HarnessService {
+  rpc ListHarnesses(ListHarnessesRequest) returns (ListHarnessesResponse);
+  rpc GetHarness(GetHarnessRequest) returns (GetHarnessResponse);
+  rpc CreateHarness(CreateHarnessRequest) returns (CreateHarnessResponse);
+  rpc UpdateHarness(UpdateHarnessRequest) returns (UpdateHarnessResponse);
```
EItanya: Can we omit GetHarness and UpdateHarness from this PR? The rewritten UI calls list, create, and delete but has no consumer for either RPC. They pull later CLI and apply API surface into the UI cutover and add proto, generated, handler, service, policy, and test code speculatively. Add them with the consumer that needs them.

go/core/internal/service/system/substratecache.go
```diff
@@ -0,0 +1,149 @@
...
+
+// cachedAnswer is a computed result and the instant it was computed at.
+type cachedAnswer struct {
+	value      any
+	computedAt time.Time
+}
+
+// substrateCache memoises the substrate reads for substrateCacheTTL.
+//
+// The singleflight group is the other half of the point: without it, the three
```
EItanya: This cache does not collapse the three reads described here: summary, actors, and workers deliberately use different keys, and singleflight only coalesces equal keys. Its 400ms TTL also expires before normal polling, while the UI client already deduplicates identical reads. Can we delete this cache and compute the requested result directly until measurements show duplicate same-key calls are a real production cost? That removes the any cache, locking, eviction, and about 280 lines including tests.

go/core/internal/service/system/substrate.go
```diff
@@ -0,0 +1,720 @@
+package system
```
EItanya: Can we split the Substrate inventory scalability work into a separate PR? The new summary/paged RPCs, streaming selector, cache, proto surface, and tests are roughly two thousand handwritten/test lines plus generated output. This is a substantial independent backend feature and makes the UI transport/conversation changes much harder to review atomically.

proto/kagent/api/v1alpha1/agent_instances.proto
```diff
@@ -75,6 +79,10 @@ message CreateAgentInstanceRequest {
     min_len: 1
     max_len: 128
   }];
+  // Optional display name. Omit it to create an unnamed conversation. Unvalidated
+  // because empty is the ordinary case: a conversation is usually named later, or
+  // never.
+  string name = 5;
```
EItanya: This request-intrinsic validation belongs in the proto, per the shared Protovalidate interceptor. max_len handles the 200-character bound and CEL can reject surrounding whitespace and control characters while allowing empty names. That lets us delete validateName; its UTF-8 check is redundant because protobuf strings are already valid UTF-8. The new template and harness filter validation should move here as well rather than being duplicated in the service.

go/api/database/client.go
```diff
@@ -124,19 +140,38 @@ type Client interface {
...
+	// instance's active-task slot for a turn that was parked awaiting the reader.
+	// It returns false if that task is no longer active.
+	AbandonActiveAgentInstanceTask(context.Context, string, string) (bool, error)
+	// ClaimParkedAgentInstanceTask moves a task waiting on the reader into a
+	// working state so a reply can be delivered, returning the task as it was
+	// parked and whether this call claimed it. A second caller is refused, which
+	// is what stops a duplicate reply being delivered twice.
+	ClaimParkedAgentInstanceTask(context.Context, string, string) (*a2a.Task, bool, error)
```
EItanya: ClaimParkedAgentInstanceTask and RestoreParkedAgentInstanceTask are added and extensively tested, but no production caller uses either one. prepareReply still does GetAgentInstanceTask followed by StoreAgentInstanceTaskEvent, so two concurrent replies can both observe the parked state and both dispatch. Please either make prepareReply use the claim and restore replay guard, or remove the unused interface, SQL, implementation, and tests; the current version pays for both approaches while retaining the race.

EItanya (approved):

EItanya (approved):

Issue #1734, opened 2026-04-23
[FEATURE] Propagate A2A message metadata as OTEL span attribute


karma-king-350:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Add support for propagating custom metadata from A2A 'message' payloads as attributes on OTEL spans, enabling external callers to attach contextual data (e.g. approver identity) to traces.

### ❓ Problem Statement / Motivation

  - Current limitation: kagent's OTEL instrumentation only emits spans for its own internal operations (invoke_agent, execute_tool,  
  call_llm.. etc). Any metadata included in the A2A `message/send` payload (at `params.message.metadata` or within                          
  `params.message.parts[].data`) is silently dropped and never surfaces in any OTEL span.                  
  - Who is affected: Anyone building human-in-the-loop or external-trigger workflows on top of kagent where the caller needs to attach
   contextual attributes to the trace (e.g. who approved an action, which system triggered the request, audit metadata).              
  - Why it's needed: In approval workflows (e.g. a Slack bot that gates agent actions), the approver's identity is captured outside  
  kagent. There is currently no way to record that identity in OTEL traces, making audit and observability impossible without running
  a completely separate OTEL pipeline from the external service.     

### 💡 Proposed Solution

When kagent processes an incoming `message/send` request, any key-value pairs found in `params.message.metadata` should be set as  
  attributes on the active OTEL span (e.g. under a `a2a.message.metadata.*` prefix).                                              
                                                                                                                                     
  Example: if the caller sends `metadata: { "approver_email": "[user@example.com](mailto:user@example.com)" }`, the span should include the attribute
  `a2a.message.metadata.approver_email = [user@example.com](mail to:user@example.com)`.                                                              
                                                                                                                                     
  Requirements:
  - Only scalar values (string, number, bool) should be set as span attributes                                                        
  - The prefix `a2a.message.metadata.` should be used to avoid collisions with kagent's own span attributes
  - Nested objects in metadata can be skipped or flattened (up to implementation discretion)

### 🔄 Alternatives Considered

  - External OTEL instrumentation: The calling service (e.g. Slack handler) emits its own spans. This works but requires trace context
   (traceparent) to be propagated through the A2A payload to stitch spans into the same trace
[...]

PR #2575, opened 2026-08-26
feat: promote allowlisted caller context onto every agent span

rtemperini:
## Motivation

Agent spans record *what an agent did* but not *who asked for it*. Once a request
enters kagent the caller's identity and context are gone, so traces cannot be
filtered or grouped by the calling user, the conversation thread, or the ticket
that triggered the run.

Two common deployment shapes hit this:

- An authenticating proxy in front of kagent knows the signed-in user's OIDC
  `sub`. That opaque identifier should appear on every downstream operation as
  `user.id` — tool calls, A2A delegations, MCP calls, model calls.
- A chat integration or other programmatic A2A caller knows the invoking user,
  the thread, and the channel. None of it reaches the trace.

[#1734](https://github.com/kagent-dev/kagent/issues/1734) raised the second case
and was closed by [#1737](https://github.com/kagent-dev/kagent/pull/1737), which
promotes A2A `message.metadata` into `a2a.message.metadata.*` attributes. That
left two gaps:

1. **Go only.** The Python runtimes ignore inbound `message.metadata` entirely.
2. **Root span only.** `SetMessageMetadataAttributes` writes to the span current
   at A2A entry, so descendant spans inherit nothing.

Gap 2 is the one that breaks the use case: Langfuse and comparable backends
resolve trace-level filters against the attributes present on *each* span, so an
attribute on the invocation span alone leaves most views unfilterable.

## What this changes

An operator names the context keys they want traced. Both runtimes read those
keys from **W3C Baggage** and from **A2A `message.metadata`**, sanitise them, and
merge them into the **request-scoped attribute bag** that
`KagentAttributesSpanProcessor` / `kagentAttributesSpanProcessor` already stamps
onto every span of the request.

```yaml
otel:
  tracing:
    enabled: true
    contextKeys:
      - {from: sub, to: user.id}
      - {from: thread_id, to: kagent.thread_id}
      - channel
```

```
baggage: sub=opaque-subject           →  user.id                   = "opaque-subject"
metadata: {"thread_id": "1717171.42"} →  kagent.context.thread_id  = "1717171.42"
metadata: {"channel": "C0AB1"}        →  kagent.context.channel    = "C0AB1"
```

Registry names (`user.*`, `enduser.*`, `session.id`) are left unprefixed so
operators can use the semantic convention names. Names already in the `kagent.`
namespace are left as-is. Everything else is emitted as `kagent.context.<name>`.

Adding a new traced value is a configuration change, not a code change.

### Why baggage

Baggage is the vendor-neutral OTel answer to this problem, and the plumbing
already exists: the controller, both runtimes, and every instrumented HTTP client
run a composite `tracecontext + baggage` propagator. A value set once at the edge
survives controller → agent → sub-agent → tool without kagent adding any
hop-specific mechanism, and it requires no kagent-specific knowledge from the
caller — any OTel SDK or proxy can set it.

A2A `message.metadata` remains supported as the complement, for callers that can
set message fields but not transport headers. Because it is scoped to one message
it is the more specific source, so it wins on conflict.

Alternatives considered and rejected: a fixed set of `user`/`thread`/`channel`
fields (not extensible, needs a code change per new field); custom HTTP headers
(reinvents baggage, does not cross hops); stamping only the root span (does not
satisfy the per-span requirement above).

### Why the attributes go in the request-scoped bag

The bag is the only place where a value is applied by the span processor at
`OnStart` for *every* span, including spans created by upstream ADK, the MCP
client, and the model instrumentation — code kagent does not own and cannot
instrument individually.

### Why one knob instead of an enable flag plus a list

An `enabled: true` with an empty allowlist is a state that does nothing but looks
like it should. Making the allowlist itself the switch removes that state: empty
means off, non-empty means on, and a contradiction cannot be expressed.

## Feature flag

| | |
|---|---|
| **Name** | `KAGENT_TRACE_CONTEXT_KEYS` (env) / `otel.tracing.contextKeys` (Helm) |
| **Type** | Comma-separated keys, or a JSON array of strings and `{from, to, hash}` objects |
| **Default** | Empty — **promotion disabled** |
| **HMAC key** | `KAGENT_TRACE_CONTEXT_HASH_KEY` / `otel.tracing.contextHashKeySecret` for `hash: hmac-sha256` |

The controller forwards both variables to the agents it creates. They need
explicit forwarding because `collectOtelEnvFromProcess` carries only `OTEL_`
prefixed names.

Which caller data reaches a trace backend is cluster-wide operator policy, so the
value is applied after the `Harness` environment *and* any inherited entry of the
same name is dropped first. Without that second step a `Harness` could enable
promotion whenever the operator had configured nothing at all; there is a test
for exactly that. The HMAC key is the same class of policy: a `Harness` cannot
set or replace it.

## Security considerations

Caller-supplied context is untrusted input on both paths, so promotion is
constrained on every axis:

| Risk | Control |
|---|---|
| Cardinality explosion | Only allowlisted keys are read; the allowlist is capped at 32 entries |
| Oversized spans | Values truncated to 256 characters, keys to 64 (rune-based on both runtimes) |
| Log / trace injection | Control characters stripped from values |
| Shadowing semantic conventions | Custom keys are namespaced under `kagent.context.`; only `user.*`, `enduser.*`, and `session.id` pass through unprefixed |
| Secret leakage | Nothing is promoted unless an operator names the key; hashed entries are omitted when the HMAC key is unset rather than falling back to plaintext |
| Malformed attribute names | Allowlist entries containing whitespace or control characters are dropped |
| A tenant widening the allowlist | Operator-level configuration only; a same-named `Harness` env entry is dropped, not inherited |

Non-scalar metadata (objects, arrays) is skipped: unbounded in size, meaningless
as an attribute value.

Prefer an opaque identifier such as an OIDC `sub` for `user.id`. Do not put
names or email addresses on spans
([OTel guidance](https://opentelemetry.io/docs/security/handling-sensitive-data/)).
If a stable identifier must be derived from a value that should not appear on a
span, map it with `hash: hmac-sha256` onto `user.hash`. Hashing at promotion time
only affects the span: baggage travels on HTTP headers, so a value placed in
baggage is still visible to every downstream hop, including model providers and
HTTP MCP servers.

The docs state plainly that anything allowlisted is visible to everyone with
access to the trace backend, and that callers control the values.

## Backwards compatibility

Fully backwards compatible.

- Default is off. With no `contextKeys` set, the ConfigMap key is absent, the env
  var is unset, the helper returns immediately, and not a single span attribute
  changes.
- No existing behaviour is modified or removed. The `a2a.message.metadata.*`
  attributes from #1737 are untouched and still unconditional in the Go runtime.
- No new dependencies. Baggage comes from `go.opentelemetry.io/otel` and
  `opentelemetry-api`, both already required.
- No API, CRD, or protobuf changes.

## Runtime parity

The Go and Python implementations share the same allowlist parsing, the same
baggage-then-metadata precedence, the same limits, the same control-character
definition (Python's `_is_control` deliberately matches Go's
`unicode.IsControl`), rune-based rather than byte-based truncation on both sides,
and the same prefix rules. Both are covered by equivalent test cases so the two
cannot drift silently.

The ADK, LangGraph, and CrewAI Python executors all promote context. Reading a
`Message`'s protobuf `Struct` metadata was extracted into
`kagent.core.a2a.read_message_metadata` rather than repeated three times.

## Testing

Added, in both runtimes:

- empty allowlist → no behaviour change (asserted against exported spans, not just the
  helper's return value)
- flag on, baggage path
- flag on, A2A `message.metadata` path
- metadata overrides baggage on conflict
- keys outside the allowlist ignored
- scalar rendering (string / bool / int / float, including the protobuf `Struct`
  float-integer case) and non-scalar skipping
- control-character stripping, value truncation, key-length and duplicate
  rejection, allowlist cap
- `service.name` cannot be shadowed
- `user.*`, `enduser.*`, and `session.id` stay unprefixed; other `session.*` keys do not
- `{from, to}` mappings onto registry and `kagent.` names
- `hash: hmac-sha256` emits `user.hash` and never the original; missing key or
  unknown algorithm emit nothing
- **attribute-on-every-span**: a root → tool → model span tree, asserting the
  attribute is present on all three

Plus: Helm unittest for ConfigMap rendering of string keys and JSON mappings, and
for injecting the HMAC key from a Secret; Go tests that the controller forwards
both variables and that a `Harness` can neither widen nor enable the allowlist
nor supply the HMAC key; and a regression test that the OTel Python SDK's
default propagator still carries baggage, since the Python runtime relies on that
default rather than configuring a propagator.

## Not in scope

- **E2E coverage.** The behaviour is span-attribute shaping with no CRD, API, or
  lifecycle surface, and it is fully covered by unit tests against a real
  `TracerProvider`. Happy to add an E2E case if maintainers would prefer one.
- **Controller-side baggage injection.** The controller already propagates
  baggage it receives. Deriving baggage from an OIDC token at the edge is the
  gateway's job, not kagent's.

## Docs

New [`docs/architecture/trace-context.md`](docs/architecture/trace-context.md),
linked from the architecture index. Covers configuration, why baggage, the
per-span guarantee, the prefix rules, hashing, the baggage-on-HTTP-headers
warning, the safety properties, and the Collector rename recipe.

## Commits

| | |
|---|---|
| `feat(adk)` | Go runtime: allowlist, sanitisation, A2A executor wiring |
| `feat(python)` | Python runtimes: parity across ADK, LangGraph, CrewAI |
| `feat(core)` | Helm value, ConfigMap, controller forwarding |
| `docs(architecture)` | Documentation |
| `fix` | Review: unprefixed registry names, `{from, to, hash}` mappings, HMAC-SHA256 |

All commits are DCO signed off.

This follows on from #1734 / #1737 rather than starting a new discussion, but
happy to write it up as an enhancement proposal under `design/` first if that is
preferred for a change of this size.


docs/architecture/trace-context.md
```diff
@@ -0,0 +1,136 @@
+# Caller Context in Traces
+
+Agent spans describe *what the agent did*, but they say nothing about *who asked
+for it*. Kagent can promote a configurable allowlist of caller-supplied values —
+the signed-in user's email, a Slack thread, a support ticket ID — onto every
+span of a request, so traces can be filtered and grouped by the caller in
+Langfuse, Jaeger, Grafana Tempo, or any other OTLP backend.
```
krisztianfekete: Totally agree with the goal, but as per the OTel's recommendations: https://opentelemetry.io/docs/security/handling-sensitive-data/, email addresses and names should never be attributes at all.

OIDC already hands you an opaque sub, which is what `user.id` should use. Could we make sub the example instead of the email?

helm/kagent/values.yaml
```diff
@@ -836,6 +836,11 @@ oauth2-proxy:
 otel:
   tracing:
     enabled: false
+    # Allowlist of caller-supplied context keys promoted onto every agent span
+    # as kagent.context.<key>. Values are read from W3C baggage and A2A message
+    # metadata. Empty (the default) disables promotion.
+    # e.g. ["user.email", "user.name", "thread_id", "channel"]
+    contextKeys: []
```
krisztianfekete, line 861: Let's use subs everywhere as exampels as mentioned in the md file above.

docs/architecture/trace-context.md
```diff
@@ -0,0 +1,136 @@
...
+
+---
+
+## Where values land
+
+Every promoted value becomes a span attribute named `kagent.context.<key>`:
+
+```
+baggage: user.email=ada@example.com   →   kagent.context.user.email = "ada@example.com"
+metadata: {"thread_id": "1717171.42"} →   kagent.context.thread_id  = "1717171.42"
```
krisztianfekete: Same as above, let's not document it like this as it's an anti-pattern.

docs/architecture/trace-context.md
```diff
@@ -0,0 +1,136 @@
...
+Caller-supplied context is untrusted input, so promotion is constrained on every
+axis:
+
+| Risk | Control |
+|---|---|
+| Attribute explosion / cardinality | Only allowlisted keys are read; the allowlist itself is capped at 32 entries |
+| Oversized spans | Values are truncated to 256 characters, keys to 64 |
+| Log or trace injection | Control characters are stripped from values |
+| Shadowing semantic conventions | Every attribute is namespaced under `kagent.context.`, so `service.name` and friends cannot be overwritten even if an operator allowlists them |
+| Leaking secrets into a trace backend | Nothing is promoted unless an operator names the key; raw values are never logged |
```
krisztianfekete: Again, emails and username should not be here as per the guidance because it's a compliance question. Operators should also not do this. Also baggage is on HTTP headers, and today this data goes to api.openai.com and every HTTP MCP server too.

go/core/pkg/env/otel.go
```diff
@@ -37,4 +37,13 @@ var (
...
 	)
+
+	KagentTraceContextKeys = RegisterStringVar(
+		"KAGENT_TRACE_CONTEXT_KEYS",
+		"",
+		"Comma-separated allowlist of caller-supplied context keys promoted onto every agent span as "+
+			"kagent.context.<key>. Values are read from W3C baggage and A2A message metadata. "+
+			"Empty (the default) disables promotion.",
+		ComponentAgentRuntime,
+	)
```
krisztianfekete, line 58: If you don't want to lose the OIDC format the recommendation is to hash it, and `user.hash` is a registry attribute that exists for exactly that. Something like:
```
contextKeys:
  - {from: sub,       to: user.id}
  - {from: email,     to: user.hash, hash: hmac-sha256}
  - {from: thread_id, to: kagent.thread_id}
```

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,140 @@
...
+const (
+	// traceContextKeysEnvVar holds a comma-separated allowlist of caller-supplied
+	// context keys to promote onto agent spans. Unset or empty (the default)
+	// disables promotion entirely.
+	traceContextKeysEnvVar = "KAGENT_TRACE_CONTEXT_KEYS"
+
+	// contextAttributePrefix namespaces every promoted value. Because the prefix
+	// is applied unconditionally, caller-supplied data cannot shadow a semantic
+	// convention attribute such as service.name.
+	contextAttributePrefix = "kagent.context."
```
krisztianfekete: The prefix is right for custom keys e.g. channel, but it also blocks the names that do exist in the registry. `user.id`, `enduser.id`, `session.id` are all real attributes, and we should use the semconv names before inventing new ones. 

Could we let a small fixed set through unprefixed (`user.*`, `enduser.*`, `session.id`) and prefix everything else?

krisztianfekete: Also reflect this in tests.

krisztianfekete (commented):
Thanks, added a few initial comments!

docs/architecture/trace-context.md
rtemperini: Thanks for the review, I agree, examples now use an OIDC `sub` mapped to `user.id` rather than an email. Docs, Helm values, and the PR description all follow that. Locked by `TestCallerContextAttributes/maps_source_keys_onto_registry_and_kagent_names` (Go) and `test_maps_source_keys_onto_registry_and_kagent_names` (Python).

helm/kagent/values.yaml
rtemperini: Updated. `values.yaml` now shows `{from: sub, to: user.id}` (and `thread_id` / `channel`), not names or addresses. Helm unittest `should join otel.tracing.contextKeys` / `should encode contextKeys mappings as JSON` lock the rendered forms.

docs/architecture/trace-context.md
rtemperini: Removed. The landing example is now `sub → user.id`. The "Sensitive values" section points at the OTel guidance and tells operators not to put names or addresses on spans.

docs/architecture/trace-context.md
rtemperini: Documented, and the implementation follows that. Hashing at promotion time only changes the span attribute — baggage is still an HTTP header, so a value placed there still reaches model providers and HTTP MCP servers. The docs say to hash or replace at the edge before the request enters the cluster, and `hash: hmac-sha256` never falls back to writing the original onto the span if the HMAC key is missing (`TestCallerContextAttributes_HashWithoutKeyEmitsNothing` / `test_hash_without_key_emits_nothing`).

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Done. `user.*`, `enduser.*`, and `session.id` pass through unprefixed; names already in the `kagent.` namespace stay as-is; everything else is still `kagent.context.<name>`, so `service.name` cannot be shadowed.

Tests: `TestCallerContextAttributes/registry_attributes_stay_unprefixed`, `session.id is unprefixed but session.foo is not`, `CannotShadowSemanticConventions`, and the matching Python cases.

go/core/pkg/env/otel.go
rtemperini: Implemented this mapping shape: `{from, to, hash}`. `hash: hmac-sha256` writes `user.hash` using `KAGENT_TRACE_CONTEXT_HASH_KEY` (Helm `contextHashKeySecret`). If the key is unset or the algorithm is unknown, the attribute is omitted rather than emitted in plaintext. A Harness cannot supply or replace the key.

Tests: `TestCallerContextAttributes_HashesWithHMACSHA256`, `TestCallerContextAttributes_HashWithoutKeyEmitsNothing`, `TestCallerContextAttributes_UnknownHashEmitsNothing`, and the matching Python cases. Tenant-cannot-set is `TestCompileAgentTemplateForwardsTraceContextPolicy/harness_cannot_supply_the_HMAC_key`.

rtemperini:
> Thanks, added a few initial comments!

Thanks for the review @krisztianfekete , just updated with your comments, let me know if anything else required

go/adk/pkg/a2a/executor.go
```diff
@@ -129,6 +130,9 @@ func (e *KAgentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.ExecutorCon
 		if e.appName != "" {
 			spanAttributes["kagent.app_name"] = e.appName
 		}
+		// Allowlisted caller context joins the request-scoped bag rather than a
+		// single span, so tool, sub-agent, and model spans all carry it.
+		maps.Copy(spanAttributes, telemetry.CallerContextAttributes(ctx, reqCtx.Message.Metadata))
```
krisztianfekete: The registry passthrough is great, but `kagent.` names now pass through too and `maps.Copy` lets caller context override them.

It'ss the exact thing the prefix was protecting. Could caller context only fill keys that aren't already set, instead of copying over them?

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+// use the semantic convention names. Names already in the kagent. namespace
+// are left as-is. Everything else is placed under kagent.context. so a
+// caller-supplied service.name cannot shadow the real one.
+func spanAttributeName(name string) string {
+	if isRegistryAttribute(name) || strings.HasPrefix(name, "kagent.") {
+		return name
+	}
+	return contextAttributePrefix + name
+}
+
```
krisztianfekete, line 298: If the merge above in executor gets fixed this should be fine, but specifying the actual attrs `kagent.*` has can make it more robust.

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+func isRegistryAttribute(name string) bool {
+	return strings.HasPrefix(name, "user.") ||
+		strings.HasPrefix(name, "enduser.") ||
+		name == "session.id"
+}
```
krisztianfekete: If we just match the `user.` prefix then `user.asdasd` goes through unprefixed.

Can we use an explicit set of allowed names here?

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+	bag := baggage.FromContext(ctx)
+	attrs := make(map[string]string, len(mappings))
+	for _, mapping := range mappings {
+		value := sanitizeContextValue(bag.Member(mapping.source).Value())
+		if scalar, ok := scalarString(metadata[mapping.source]); ok {
+			value = sanitizeContextValue(scalar)
+		}
+		if value == "" {
+			continue
+		}
```
krisztianfekete, line 167: Let's only override when the sanitised scalar is non-empty.

python/packages/kagent-core/src/kagent/core/tracing/_context_attributes.py
```diff
@@ -0,0 +1,251 @@
...
+    bag = baggage.get_all(context)
+    attributes: dict[str, str] = {}
+    for mapping in mappings:
+        value = _sanitize_context_value(bag.get(mapping.source))
+        if metadata is not None:
+            scalar = _scalar_string(metadata.get(mapping.source))
+            if scalar is not None:
+                value = _sanitize_context_value(scalar)
+        if not value:
+            continue
```
krisztianfekete, line 106: Same as https://github.com/kagent-dev/kagent/pull/2575/changes#r3880978791

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+	case float64:
+		return strconv.FormatFloat(v, 'g', -1, 64), true
+	case int:
+		return strconv.Itoa(v), true
+	case int64:
+		return strconv.FormatInt(v, 10), true
+	default:
+		return "", false
+	}
+}
```
krisztianfekete, line 321: `'g'` flips to scientific notation and `int()` doesn't in the matching logic on Python side, so these disagree on ordinary values.

`strconv.FormatFloat(v, 'f', -1, 64)` matches what Python already does. Also worth noting the int/int64 cases below are unreachable.

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+		to = from
+	}
+	if len(to) > maxContextKeyLength || !isAttributeKey(to) {
+		return contextMapping{}, false
+	}
+	if hash != "" && hash != hashHMACSHA256 {
+		return contextMapping{}, false
+	}
+	return contextMapping{source: from, attribute: to, hash: hash}, true
+}
```
krisztianfekete, line 267: `len()` is bytes here and code points in `_new_context_mapping`, so a 40-character CJK key (120 bytes) is dropped by Go and kept by Python. 

```pyt
len([]rune(from))
```
 on both from and to should fix it.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
```diff
@@ -131,6 +133,10 @@ async def execute(self, context: RequestContext, event_queue: EventQueue) -> Non
                 "gen_ai.task.id": context.task_id,
                 "gen_ai.conversation.id": run_request.session_id,
             }
+            # Allowlisted caller context joins the request-scoped bag rather
+            # than a single span, so tool, sub-agent, and model spans all
+            # carry it.
+            span_attributes.update(caller_context_attributes(read_message_metadata(context.message)))
```
krisztianfekete: Let's pass the `Message` and decode lazily after the allowlist check to avoid running these when this is off. 
This should be also fixed for crewai and langgraph as well.

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+func allowedContextMappings() []contextMapping {
+	raw := strings.TrimSpace(os.Getenv(traceContextKeysEnvVar))
+	if raw == "" {
+		return nil
+	}
+	if strings.HasPrefix(raw, "[") {
+		return capMappings(parseJSONAllowlist(raw))
+	}
+	return capMappings(parseCommaAllowlist(raw))
+}
```
krisztianfekete: This reparses on every request now that there's JSON in it. Wrap it in `sync.OnceValue`? The allowlist can't change without a restart.

go/adk/pkg/telemetry/context_attributes.go
```diff
@@ -0,0 +1,272 @@
...
+	// promotion entirely.
+	//
+	// Accepts a comma-separated list of source keys, or a JSON array of strings
+	// and {from, to, hash} objects. See allowedContextMappings.
+	traceContextKeysEnvVar = "KAGENT_TRACE_CONTEXT_KEYS"
+
+	// traceContextHashKeyEnvVar is the HMAC key used when a mapping sets
+	// hash: hmac-sha256. Required for those entries; without it the hashed
+	// attribute is skipped rather than emitted in plaintext.
+	traceContextHashKeyEnvVar = "KAGENT_TRACE_CONTEXT_HASH_KEY"
```
krisztianfekete: `These names are also declared in `go/core/pkg/env/otel.go`.
Can we use `env.KagentTraceContextKeys.Name()` here so there's one source of truth?

krisztianfekete (changes requested):

go/adk/pkg/a2a/executor.go
rtemperini: Caller context now fills only keys that are unset, via MergeCallerContextAttributes. Go also skips allowlisted source keys in SetMessageMetadataAttributes, so hashed or omitted values cannot leak as plaintext on a2a.message.metadata.<from>.

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Dropped the kagent.* passthrough. A mapping with to: kagent.user_id now lands as kagent.context.kagent.user_id.

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Unprefixed names are an explicit set: user.id, user.hash, enduser.id, and session.id. Any other stay prefixed on the `kagent.context.` scope. Let me know if this is clear and is the approach you want to follow. Thanks

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Empty or whitespace metadata does not wipe baggage. Only a sanitised non-empty scalar overrides.

python/packages/kagent-core/src/kagent/core/tracing/_context_attributes.py
rtemperini: Updated, implementing same as Go: empty metadata leaves baggage in place. Non-string hash or to drops the mapping so Python cannot emit plaintext.

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Go formats floats with FormatFloat using the f verb and precision matching Python. Tests cover 1e20 and 1e-20 in both languages.

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Both from and to now use rune length. A 40 character CJK key is kept; 65 is dropped.

python/packages/kagent-adk/src/kagent/adk/_agent_executor.py
rtemperini: caller_context_attributes now accepts the Message and decodes metadata only after the allowlist check. Wired in ADK, LangGraph, CrewAI, and the OpenAI executor.

go/adk/pkg/telemetry/context_attributes.go
rtemperini: Go caches the parse with sync.OnceValue; Python uses functools.cache. Tests reset both.

go/adk/pkg/telemetry/context_attributes.go
rtemperini: updated, now uses env.KagentTraceContextKeys.Name() and env.KagentTraceContextHashKey.Name() so the names live in one place.

rtemperini:
@krisztianfekete thanks for the reviews, I've added a few updates, let me know if should I take a look on anything else

rtemperini:
@krisztianfekete I merged main to clear the conflict with the new OTEL env forwarding. Could you take another look when you have a moment? The earlier review comments should still be the open questions.

krisztianfekete (changes requested):
Before we go deeper on the details though, I think we can get to the same behaviour with a lot less code by using upstream building blocks.

Both runtimes ship a baggage -> spanattribute processor, each taking a filter/predicate, which is basically what `MergeCallerContextAttributes` + the attributes span processor are doing by hand.

If we then promote allowlisted A2A message.metadata keys into baggage at the executor boundary rather than reading them as a second source with its own precedence rule, we'd have one code path instead of two, and metadata-sourced values would survive subagent hops as well. That would also let us drop most of the policy layer to where OTel already puts it: value truncation is `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`, renaming can be done via a collector transformation, and hashing is really an collector or gw concern.

I'd rather not hardcode am unprefixed set, as `to:` is operator config, not caller input, so prefixing it doesn't really defend against anything, and right now it makes `gen_ai.conversation.id` unreachable even though our own ADK executor already sets it.

rtemperini:
@krisztianfekete I reworked this around the upstream baggage-to-span processors and promote allowlisted metadata into baggage at the executor, so there is one path. The custom hash/prefix/merge layer is gone; leftover hash mappings fail closed. Could you take another look when you have a moment?

rtemperini:
Hi @krisztianfekete I've rebased and fixed conflicts with the latests main rebase, can you take a look on this when you have a minute, will be good to address this to be able to have the proper context attributes propagation. Thanks in advance

PR #2579, opened 2026-08-26
fix(kagent-adk): compact MCP App tool results only when the result carries a UI resource

### LINKED ISSUE #2519: [BUG] MCP Apps compaction blanks tool results for headless agents, even when the result contains no UI resource
## What happens

kagent replaces an MCP tool's result with `MCP_APP_RENDERED_NOTICE` whenever the tool's *definition* declares `_meta.ui.resourceUri`. It does this even when the result itself contains no UI resource and carries the full text payload. The model receives the notice instead of the data.

Classification comes from the tool definition, not from the result.

## Version

`0.10.0-rc3`, the current release.

```
kagent-controller   ghcr.io/kagent-dev/kagent/controller:0.10.0-rc3
agent pod           ghcr.io/kagent-dev/kagent/golang-adk:0.10.0-rc3
```

## Reproduction

1. `RemoteMCPServer` pointing at `https://mcp.atlassian.com/v1/mcp`, Atlassian API token via `headersFrom`.
2. Declarative `Agent`, `runtime: go`, `toolNames: [searchJiraIssuesUsingJql, getJiraIssue]`.
3. Ask it to describe recent issues on a Jira board.

The agent replies that the results are displayed in an interactive format and states no field from them.

### 1. The server returns the data

Using kagent's capability and client name:

```bash
curl -s -X POST https://mcp.atlassian.com/v1/mcp \
  -H "Authorization: Bearer $TOKEN" \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{
        "protocolVersion":"2025-06-18",
        "capabilities":{"extensions":{"io.modelcontextprotocol/ui":{"mimeTypes":["text/html;profile=mcp-app"]}}},
        "clientInfo":{"name":"kagent-adk","version":"1"}}}'
```

`tools/call` on `searchJiraIssuesUsingJql` returns:

```
content block types: ['text']
has structuredContent: True
_meta keys: []
TEXT block: { "issues": [ { "id": "160217", "key": "GF-3687", ... } ] }
```

Text block with the full JSON, `structuredContent` present, no UI resource in the result.

### 2. The tool metadata is capability-gated

Same server, same token, two `initialize` calls.

`capabilities: {}` returns 13 tools, none carrying `_meta`.

`capabilities` containing `io.modelcontextprotocol/ui` returns 13 tools, and `searchJiraIssuesUsingJql` and `getJiraIssue` carry:

```json
"_meta": {"ui": {"resourceUri": "ui://widget/jira-widget.html", "prefersBorder": true, ...}}
```

`mcpUIClientCapabilities()` in `go/adk/pkg/mcp/mcp_ui.go` advertises the extension on every connection. No per-server setting disables it.

### 3. kagent classifies from the definition

`agentVisibleToolFilter` reads `mcpToolKindOf(t.Meta)` during `ListTools` and records matching names in `appToolNames`. Agent log:

```json
{"level":"info","caller":"agent/agent.go:111",
 "msg":"Wiring MCP App model result callback","toolCount":2}
```

Both of the agent's two tools.

### 4. The callback discards the payload

`compact_mcp_app_response`:

```python
compacted = {"content": [{"type": "text", "text": MCP_APP_RENDERED_NOTICE}]}
if "_meta" in response:
    compacted["_meta"] = response["_meta"]
return compacted
```

The result is not inspected. `go/adk/pkg/agent/mcp_apps.go` does the same.

The notice reads: *"The interactive UI for this tool has been rendered to the user and now updates live inside the app. Treat this as complete and do not call this tool again unless the user explicitly asks for it."*

### 5. Session history keeps the real payload

The compaction runs in `before_model` against a deep copy, so `GET /api/sessions/<id>` shows the full tool output. Inspecting a session does not reveal the problem.

## Scope

Any agent invoked over A2A or `kagent invoke` against a server that declares UI tools. There is no widget rendered and no human present, so the model is the only consumer of the tool result.

#2046 scopes MCP Apps to the chat interface. `design/EP-2046-chat-mcp-ui-widgets.md` lists per-server opt-in as an open question.

## Possible fixes

1. Compact only when the result contains a UI resource, rather than when the definition declares one.
2. Keep the text content alongside the notice instead of replacing it.
3. Allow a `RemoteMCPServer` to opt out of the UI capability.

## Checking whether an agent is affected

```bash
kubectl logs -n kagent deploy/<agent> | grep "Wiring MCP App model result callback"
```

A non-zero `toolCount` means the model does not receive those tools' output.

AmirF194:
## Root cause

`compact_mcp_app_response` (`python/packages/kagent-adk/src/kagent/adk/_mcp_apps.py`)
collapses a tool's result into a terminal notice whenever the tool's *name* is in
`app_tool_names`, a set populated once from the tool's `_meta.ui.resourceUri` at
`ListTools` time. A single tool can carry a UI mode without every call rendering it
(the issue's example: an Atlassian search tool with a UI widget), so any plain,
data-only result from such a tool was silently replaced with the notice text for a
headless caller (A2A, `kagent invoke`), with no error.

## Fix

Added `_result_has_ui_resource`, which checks the *result's own* `_meta.ui.resourceUri`
/ `_meta["ui/resourceUri"]`, the same fields `go/adk/pkg/mcp/mcp_ui.go` already parses
to classify a tool's definition, applied here to the per-call result instead.
`compact_mcp_app_response` now passes a result through unchanged unless it declares a
UI resource itself.

Scoped to the Python ADK runtime per the issue; `go/adk/pkg/agent/mcp_apps.go` has the
identical gap and is left as a follow-up, since it is outside this change's surface.

## Verification

- New tests reproduce the issue directly: a data-only result from a UI-capable tool
  (empty `_meta`, and no `_meta` key at all) now passes through unmodified. Both fail
  on `main` (asserting the old collapse) and pass on this branch; confirmed independently
  in a clean container against an unmodified worktree of `main` before rebuilding on the
  branch.
- Full `kagent-adk` unit suite (`uv run pytest packages/kagent-adk/tests/unittests/`):
  402 passed on Python 3.11 and 3.13, both CI matrix extremes.
- `ruff check` and `ruff format --diff` on the two changed files: clean.
- Not verified: a real MCP server whose result carries a UI resource via a content
  block rather than `_meta` (no such server available to test against); the existing
  and new tests all key on `_meta`, matching the only shape this repo's own tests and
  `mcp_ui.go` currently recognize.

Fixes #2519


dimetron:
Looks good. The change correctly moves the compaction decision from tool-level UI capability to the individual CallToolResult, which fixes the data-loss case without changing the existing error path.

Non-blocking: I’d add a direct positive test for _meta["ui/resourceUri"] as well as _meta.ui.resourceUri, since _result_has_ui_resource() intentionally supports both representations.

We should also track the equivalent Go ADK change separately so Python and Go don’t retain different MCP App semantics.

AmirF194:
Added a direct test for both _meta shapes on _result_has_ui_resource, 7/7 pass in a clean container. Left the Go ADK side out of this PR, one concern at a time, but agreed it's worth its own tracking.

Synced onto main to clear the BEHIND state, no conflicts.

AmirF194:
This should be ready for review now, CI's green across all 25 checks and the direct test dimetron flagged for both _meta shapes is in.

aagronow-playon:
@AmirF194 how can I help move this along?

AmirF194:
It's green and just waiting on a maintainer review at this point, nothing outstanding on my end. If you know anyone with merge rights who has bandwidth to take a look, that would help.

aagronow-playon:
@dimetron @EItanya — this one's ready and just needs a maintainer's Approve to unblock.

Context: I filed #2519, the bug this fixes. MCP App compaction blanks out data-only tool results for headless callers (A2A, `kagent invoke`) whenever the tool merely *can* render UI, which silently drops real data with no error.

State of the PR:
- +52/-2 across two files, scoped to `python/packages/kagent-adk`
- All 25 checks green
- @dimetron reviewed the approach on Sep 1 and confirmed it's correct — just never submitted as a formal review
- No maintainer has been formally requested yet (`python/` has no individual CODEOWNER, so nothing auto-routed)

Happy to help however's useful to get it over the line.

PR #2591, opened 2026-08-26
fix: preserve A2A transcript order

### LINKED ISSUE #2584: ListTasks returns a task's messages and artifacts as two unordered lists, so a transcript cannot be replayed in order
### What happens

In a conversation where the agent asks the reader something with `ask_user`, the
transcript renders out of order: every message the reader sent appears above every
message the agent sent, so the answers sit above the questions they answer.

Read back from a real cluster, `ListTasks` returns this for one such task:

```
TASK 01a03efd-bfea   state=COMPLETED   history=3   artifacts=7
  history[0]   user    "ask me a question"
  history[1]   user    "Personal development"      <- 1st answer
  history[2]   user    "none"                      <- 2nd answer
  artifact[0]  agent   ask_user called
  artifact[1]  agent   ask_user result (pending)
  artifact[2]  agent   ask_user result (answered)
  artifact[3]  agent   ask_user called
  artifact[4]  agent   ask_user result (pending)
  artifact[5]  agent   ask_user result (answered)
  artifact[6]  agent   "Thank you for sharing..."
```

Every message the reader sent is in `history`; every message the agent produced is in
`artifacts`. They are two parallel lists, and nothing in the response says how to
interleave them.

### Why a client cannot fix it properly

- **Timestamps do not separate them.** Every message in a task carries the task's
  single `status.timestamp`, so a client sorting by time gets one bucket.
- **Only one side is time-ordered by id.** Artifact ids are UUIDv7 and sort correctly
  (`01a03efd-c565` ... `01a03efe-0482`). Message ids in `history` are minted by the
  client that sent them and are UUIDv4, carrying no time.
- **The correlation ids do not match across the two lists.** An answer in `history`
  carries `{"type":"ask_user_response","id":"adk-c9f53704-..."}` in its metadata under
  `https://kagent.dev/extensions/hitl/v1`. The `ask_user` call it answers carries the
  model's own id, `call_kEdJ3FU3x0W6rHytuDznM732`. Neither list names the other.

So the order can be guessed from position, and that is all.

### What would fix it

Any one of these, in preference order:

1. **Interleave server-side** — return one ordered sequence per task, rather than
   `history` and `artifacts` as separate lists a caller has to reassemble.
2. **A per-entry sequence number** on messages and artifacts, monotonic within a task.
3. **A per-entry timestamp**, distinct from the task's status timestamp.

Any of the three lets a client render a task in the order it happened without
inferring anything.

### Separately: the prose question is never persisted

When the agent parks on `ask_user`, the question also arrives as prose on the status
message, and the reader sees it during the turn. It is in neither `history` nor
`artifacts` afterwards, so it is gone on the next read. That is visible as a question
that sits at the foot of the transcript until a refresh, then vanishes -- and it means
a conversation replayed from `ListTasks` is missing the questions it was about.

### Where the client stands meanwhile

`ui/src/api/chat/a2aGrpcChatClient.ts` (`messagesFromTask`) concatenates
history -> status message -> artifacts, which is what produces the ordering above. It
now interleaves by position as an approximation, and the note there points at this
issue. That approximation should be deleted once any of the three fixes lands.

---
*🤖 written by Claude*

EItanya:
## Summary
- archive input-required status messages before user replies
- publish tool calls and results as A2A status messages instead of artifacts
- remove the UI transcript-order inference and replay protocol history directly

## Testing
- go test ./core/v2/a2agateway ./adk/pkg/a2a
- yarn test src/api/chat/a2aGrpcChatClient.test.ts
- yarn typecheck
- yarn eslint src/api/chat/a2aGrpcChatClient.ts src/api/chat/a2aGrpcChatClient.test.ts

Closes #2584


EItanya:
### Design context: why conversion happens in `KAgentExecutor`

We traced the exact ADK Go v2.2.0 path and audited its callback/plugin surfaces before settling on the executor decorator.

The upstream path is:

`session.Event` -> part conversion -> private `eventToArtifactTransform` -> `TaskArtifactUpdateEvent`

Except for long-running calls that become `input-required`, every content-bearing ADK event is passed through the artifact transform. Both supported output modes produce artifacts.

The available hooks do not provide a persistence-safe event-type replacement:

| Hook | Limitation |
| --- | --- |
| Model callbacks | Modify model requests/responses, not A2A event types |
| Tool callbacks | Modify tool arguments/results only |
| Agent callbacks | Add or replace agent content, which is still converted to an artifact |
| Plugin `OnEventCallback` | Can replace a `session.Event`, but artifact conversion happens afterwards |
| `GenAIPartConverter` | Converts individual parts only |
| A2A `AfterEventCallback` | Receives an already-created `TaskArtifactUpdateEvent` and cannot replace it |
| Custom runner/provider | Still feeds session events into the same artifact-only processor |
| A2A call interceptor | Can replace the outbound streamed event, but runs after the task store processed the original artifact, so replay would remain wrong |

Marking a tool as long-running is only correct when execution must enter `input-required`; using it for ordinary tool calls/results would change task semantics.

Therefore the `AgentExecutor` decorator is the earliest public seam where kagent can replace the concrete A2A event before the A2A SDK persists it. The cleaner long-term solution is upstream support for an event converter that returns a general `a2a.Event` (or an artifact/message/status transform), at which point this local conversion can be removed.

Charlesthebird (approved):
Tested and working 🎉 

Though I needed this change https://github.com/kagent-dev/kagent/pull/2614 to see it in the UI (edit: actually this is fixed in https://github.com/kagent-dev/kagent/pull/2602 ).

Charlesthebird:
There is a follow-up I'm looking into here: https://github.com/kagent-dev/kagent/pull/2616

PR #2597, opened 2026-08-27
Modifying the helm-install-provider target

### LINKED ISSUE #2583: bug: no supported path to install the substrate CRDs
`kagent-crds` gates its `substrate-crds` subchart on `substrate.enabled`, but `helm-install-provider` passes only `kmcp.enabled` to that release:

```make
helm $(HELM_ACTION) kagent-crds helm/kagent-crds \
    ...
    --set kmcp.enabled=$(KMCP_ENABLED)
```

`KAGENT_HELM_EXTRA_ARGS` applies to the `kagent` release only, so the documented flow cannot install `WorkerPool`, `ActorTemplate`, or the other `ate.dev` CRDs. Installing them takes a manual `helm upgrade --install kagent-crds helm/kagent-crds --set substrate.enabled=true`, after which 4 `ate.dev` CRDs appear.

Anything that renders a `WorkerPool` before that runs fails with `no matches for kind "WorkerPool" in version "ate.dev/v1alpha1"`.

Part of #2574.

ekam-walia:
Fixes #2583


ekam-walia:
Hi @EItanya, I think so this approach should work as it will:
 Install both kmcp-crds and substrate-crds
 Make WorkerPool, ActorTemplate, and other ate.dev CRDs available
 No more manual helm upgrade needed

PR #2601, opened 2026-08-27
refactor(ui): rename the extension framework to App Extensions and install an ordered array

Charlesthebird:
---
*🤖 written by Claude (start)*

## Changelog

The UI extension framework is now **App Extensions**, and the app installs an ordered array of extensions (`activeAppExtensions`) rather than a single config. An extension's own declaration is unchanged — installing one alone behaves exactly as before.

## What composes, and how

`AppExtensionConfig` keeps every capability it had. The array only adds a stated rule for reconciling two of them:

| | Rule |
|---|---|
| `navItems`, `routes`, `slots`, `formFields`, `tableColumns`, `providers` | **Additive** — all take effect, in array order. |
| `theme`, `shell`, `branding`, `navOverrides`, `agentLinks`, `providerIcons`, `routeHandles`, `api` | **Merged, later wins**, field by field. |

Worth a reviewer's attention, because they don't follow from the table:

- A slot renders **every** contribution mounted at it, not one, so installing a second extension cannot silently switch off part of the first.
- `theme.supportedModes` is **intersected**, not last-wins — a second install doesn't make the first extension's components legible in the other palette.
- `navOverrides` merges **two** levels, so one extension hiding an entry and another renaming it gives a hidden, renamed entry.
- API transforms compose in order; `operations`/`endpoints`/`baseUrl` beside them are single-valued and last-wins.

Collisions that cannot be reconciled — two extensions claiming a route path or a nav key, or the same extension installed twice — are rejected at startup rather than resolved arbitrarily.

The folds live in `selectors.ts` as pure functions over the install, with `hooks.ts` as a thin React binding, so the rules are testable without rendering and no call site loops over extensions itself.

## Two additions to what an extension is given

- `ExtensionNavItemProps` gains **`collapsed`**. The rail owns that state and flips it on a breakpoint as well as on the control, so no contributed entry could render correctly in a narrow rail without being told.
- **`useThemeMode`** and **`readEnv`** are re-exported from the barrel. The first is the light/dark choice antd's `Menu` takes and no design token stands in for; the second was the documented way to read a setting but wasn't reachable from the documented import path.

Both are additive.

## The bundled example

Rewritten to be self-describing — no invented product domain — and drawn with the application's own components, so it demonstrates fitting in rather than sitting beside. Its nav entry renders the same antd `Menu` the rail uses and is geometrically identical to a core entry (row box and icon column, expanded and collapsed, in both palettes); its dashboard contribution renders the dashboard's own `StatTile`.

It is **not installed by default** and is tree-shaken out of a production build.

## Settings

`VITE_EXAMPLE_EXTENSION=true` is a switch for the bundled example only — it names no extension and cannot, since an extension has to be imported to be in the bundle. An extension's own settings are runtime: named `EXTENSION_*`, set through the `kagent` chart's `ui.env`, read with `readEnv`. Both are documented in `ui/docs/app-extensions.md` and `ui/.env.example`.

## Testing

1. `cd ui && VITE_EXAMPLE_EXTENSION=true yarn dev`
2. On the dashboard, confirm the extension's banner, its "Example" tile sitting flush with the tiles beside it, the sidebar footer, and the floating "Example overlay".
3. Confirm the **Example** nav entry sits between Agents and Models and lines up with the core entries — then collapse the rail and confirm it still does.
4. Open it, and confirm the workspace line (read from a context the extension installed) and the "Example team" field. Pick a value and confirm the payload preview gains the annotation.
5. Restart with plain `yarn dev` and confirm none of it appears.

---
*🤖 written by Claude (end)*


Charlesthebird:
The Example App Extension installed through the extensions array, then two extensions composing at the same points. The second extension in that clip was written only for the recording and is not in this PR.


https://github.com/user-attachments/assets/57cc882d-e9e0-4902-bf25-8266fd2561f6



---
*🤖 written by Claude*

Charlesthebird:
Updated: the example is now self-describing and drawn with the app's own components. Its nav entry is geometrically identical to a core entry — expanded, collapsed, and in both palettes. The second clip shows two extensions composing; that second extension was written only for the recording and is not in this PR.


https://github.com/user-attachments/assets/02bf9342-20ad-4955-bc91-08a4eae3284e



---
*🤖 written by Claude*

EItanya (approved):

PR #2602, opened 2026-08-27
feat(harness): Initial Claude Code Support

supreme-gg-gg:
Basic support for running Claude Code on Substrate and add extensible interfaces for future harness like Codex.

- Add a native Claude harness that runs the pinned Claude Code CLI and supports A2A streaming, cancellation, session resume, and durable state.
- Add Claude template compilation for Anthropic, Amazon Bedrock, and Vertex AI, and shared local subagents
- Add Claude support for remote MCP servers, standalone skills, and plugin-provided skills, with runtime-neutral resource materialization and configuration warnings.
- Wire the Claude harness into the controller and build/release pipelines, including its container image and configurable controller logging.
- Add comprehensive unit and E2E coverage for Claude interactions, persistence, fork and resume, cancellation, built-in tools, subagent routing, and MCP calls.
- Merge adjacent streamed text parts in the UI so Claude responses render as continuous messages.

PR #2614, opened 2026-08-28
fix: wire the system service into controller-v2

### LINKED ISSUE #2613: [BUG] Agents page cannot load: controller-v2 never wires the system service, so ListNamespaces always returns INTERNAL
## Summary

The UI's **Agents** page fails to load on every install, showing `Could not load agents — internal server error`. The page is fine; `SystemService/ListNamespaces` returns `INTERNAL` on every call, and the agents list treats the namespace read as its own failure because templates are read one namespace at a time.

```
$ curl -s -D - -o /dev/null -X POST \
    localhost:8080/api/kagent.api.v1alpha1.SystemService/ListNamespaces \
    -H 'Content-Type: application/grpc-web+proto' --data-binary @empty.grpcweb
Grpc-Message: internal server error
Grpc-Status: 13
```

## Reproduce

1. Stand up a cluster with `scripts/setup-cluster/setup-cluster.sh`.
2. Open **Agents**. The page shows `Could not load agents — internal server error` and lists nothing, although `kubectl get agenttemplates,harnesses -n kagent` returns the harness and template the script created.

## Root cause

`systemservice.WithInventory` is the only thing that gives the system service a Kubernetes client, and it has no caller:

```
core/internal/service/system/service.go:111   func WithInventory(...)   ← declaration only
```

It used to be called from `go/core/pkg/app/app.go`, which `221674d1` — *refactor: remove legacy controller runtime (#2595)* — deleted:

```diff
-	systemService := systemservice.NewService(systemservice.WithInventory(
-		mgr.GetClient(),
-		watchNamespacesList,
-		extensionCfg.Authorizer,
-		substrateAteClient,
-	))
```

`core/cmd/controller-v2/main.go` never carried that over, so `grpcserver.New` falls back to its own default:

```go
if config.SystemService == nil {
    config.SystemService = systemservice.NewService()   // no options → no kube client
}
```

and every namespace read stops at the guard in `ListNamespaces`:

```go
if s.kubeClient == nil {
    return nil, serviceerrors.NewInternal("Failed to list namespaces", fmt.Errorf("kubernetes client is not configured"))
}
```

The silent default is what let this ship: a missing service becomes a runtime `INTERNAL` on first use rather than a failure to start.

## Same deletion, still open

Probed against the running controller, `grpc-status 12 unknown service`:

- `ModelService`, `ToolService`, `PromptTemplateService`, `FeedbackService`, `MemoryService` are not registered at all — `controller-v2` never constructs them. Models, MCP Servers and Prompts are unreachable for the same reason. (Distinct from #2603, which is about the toolserver table having no writer; these RPCs do not exist on the server.)
- `SystemService/GetSubstrateStatus` fails even once the inventory is wired: `managerScheme` in `core/cmd/controller-v2/main.go` registers only `clientgoscheme` and `kagentv1alpha3`, while the deleted `app.go` also registered `atev1alpha1`. Listing worker pools and actor templates has no kind registered for the type.

## Why nothing appears in the controller logs

Nothing in `go/` calls `ctrl.SetLogger`/`log.SetLogger`. controller-runtime prints

```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
```

and then discards every `ctrllog` write — the `RPC completed` lines, and the cause behind each redacted `internal server error`. The failure above is invisible server-side, which is most of why it reads as a broken local setup.

## Version

Observed on `9e246fd3`.

---
*🤖 written by Claude*

Charlesthebird:
---
*🤖 written by Claude (start)*

## Changelog

The Agents page loads again. `controller-v2` now constructs the system service with its Kubernetes client and substrate inventory, so `ListNamespaces` succeeds instead of returning `INTERNAL` on every call.

## Why it was broken

`systemservice.WithInventory` had no caller: the wiring lived in `go/core/pkg/app/app.go`, deleted by #2595, and `controller-v2` never carried it over. `grpcserver.New` then falls back to `systemservice.NewService()` with no options, and every namespace read stops at the `kubeClient == nil` guard. The UI reads templates one namespace at a time, so the agents list reports that as `Could not load agents — internal server error`.

Two gaps from the same deletion are **not** fixed here and are described in #2613: `ModelService`, `ToolService`, `PromptTemplateService`, `FeedbackService` and `MemoryService` are still unregistered, and `managerScheme` still omits `atev1alpha1`, which `GetSubstrateStatus` needs.

## Testing

1. `scripts/setup-cluster/setup-cluster.sh`, then `kubectl -n kagent port-forward svc/kagent-ui 8080:8080`.
2. Open `localhost:8080/agents` — the list renders the harness/template pair the script created, with no error banner.

Closes #2613

---
*🤖 written by Claude (end)*

### DIFF
--- a/go/core/cmd/controller-v2/main.go
+++ b/go/core/cmd/controller-v2/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/kagent-dev/kagent/go/core/internal/grpcserver"
 	authimpl "github.com/kagent-dev/kagent/go/core/internal/httpserver/auth"
 	"github.com/kagent-dev/kagent/go/core/internal/service/kubecrud"
+	systemservice "github.com/kagent-dev/kagent/go/core/internal/service/system"
 	"github.com/kagent-dev/kagent/go/core/pkg/auth"
 	"github.com/kagent-dev/kagent/go/core/pkg/migrations"
 	"github.com/kagent-dev/kagent/go/core/v2/a2agateway"
@@ -142,8 +143,17 @@ func main() {
 		// the only way to author a Harness or an AgentTemplate is kubectl.
 		AgentTemplateService: kubecrud.NewService(manager.GetClient(), authorizer, &kagentv1alpha3.AgentTemplate{}, &kagentv1alpha3.AgentTemplateList{}, "AgentTemplate"),
 		HarnessService:       kubecrud.NewService(manager.GetClient(), authorizer, &kagentv1alpha3.Harness{}, &kagentv1alpha3.HarnessList{}, "Harness"),
-		CheckpointService:    checkpoints,
-		A2AHandler:           gateway,
+		// Namespaces and the substrate inventory. Without the inventory the system
+		// service holds no Kubernetes client, so every namespace read fails as an
+		// internal error — which the agents list reports as a page that cannot load.
+		SystemService: systemservice.NewService(systemservice.WithInventory(
+			manager.GetClient(),
+			namespaces(os.Getenv("WATCH_NAMESPACES")),
+			authorizer,
+			actors,
+		)),
+		CheckpointService: checkpoints,
+		A2AHandler:        gateway,
 	})
 	if err != nil {
 		log.Fatal(err)


Charlesthebird:
Superseded by #2602, which wires the same `systemservice.WithInventory` call, registers the model, tool and prompt-template services, adds `atev1alpha1` to the manager scheme, and calls `ctrl.SetLogger`. Closing in its favour.


---
*🤖 written by Claude*

PR #2616, opened 2026-08-28
fix: archive a parked question before its reply

Charlesthebird:
Before there was a bug where the agent question disappears from the chat:

https://github.com/user-attachments/assets/3e422b25-cb21-4b8b-b21c-52f2e7227e20

After, this is now stable across refreshes:

https://github.com/user-attachments/assets/0ce5a581-7cf4-4693-8276-55949b31a55e

---
*🤖 written by Claude (start)*

## Changelog

An agent's question no longer disappears from the transcript once you answer it: a parked task holds the question in its status, which the reply replaces, so it is now archived as history first.

## Testing

1. Ask an agent to ask you a question, and answer it through the prompt.
2. Reload the page — the question is still there, above your answer.

---
*🤖 written by Claude (end)*

### DIFF
--- a/go/core/v2/a2agateway/gateway_test.go
+++ b/go/core/v2/a2agateway/gateway_test.go
@@ -381,7 +381,8 @@ func TestGatewayContinuesInputRequiredTask(t *testing.T) {
 	if !ok || task.Status.State != a2atype.TaskStateCompleted || !runtime.sent {
 		t.Fatalf("reply result = %#v, runtime sent = %v", result, runtime.sent)
 	}
-	if authorizer.verb != auth.VerbUpdate || reply.ContextID != gatewayTestID || len(store.stored) != 2 {
+	// Three: the question the task was parked on, the reply, and the finished turn.
+	if authorizer.verb != auth.VerbUpdate || reply.ContextID != gatewayTestID || len(store.stored) != 3 {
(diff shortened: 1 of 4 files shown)


go/core/v2/a2agateway/gateway_test.go
```diff
@@ -381,7 +381,8 @@ func TestGatewayContinuesInputRequiredTask(t *testing.T) {
 	if !ok || task.Status.State != a2atype.TaskStateCompleted || !runtime.sent {
 		t.Fatalf("reply result = %#v, runtime sent = %v", result, runtime.sent)
 	}
-	if authorizer.verb != auth.VerbUpdate || reply.ContextID != gatewayTestID || len(store.stored) != 2 {
+	// Three: the question the task was parked on, the reply, and the finished turn.
```
EItanya: What is this comment

EItanya (approved):

go/core/v2/a2agateway/gateway_test.go
Charlesthebird: haha yeah I don't know

Charlesthebird: it might've been some checklist item it was talking about to itself.
but I tested it after your changes and it's still working - thanks for updating it!

PR #2623, opened 2026-08-28
feat: support status for ModelConfig

yuval-k:
2 conditions
Accepted - for translation error
ResolvedRefs - for reference issues (i.e. model config referencing a secret)

### DIFF
--- a/go/core/v2/controller/modelconfig.go
+++ b/go/core/v2/controller/modelconfig.go
@@ -1,0 +1,158 @@
+package controller
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"slices"
+	"strings"
+
+	kagentv1alpha3 "github.com/kagent-dev/kagent/go/api/v1alpha3"
+	v2translator "github.com/kagent-dev/kagent/go/core/v2/translator"
+	"istio.io/istio/pkg/kube/krt"
+	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ModelConfigReconciliation struct {
+	ModelConfigName krt.Named
+	// If Translation is nil, Failure is non-nil and describes why the ModelConfig could not be translated.
+	// note that failure may be not nil even if Translation is non-nil.
+	Translation *v2translator.ResolvedModelConfig
+}
+
+func (r ModelConfigReconciliation) Equals(other ModelConfigReconciliation) bool {
+	if r.ModelConfigName != other.ModelConfigName {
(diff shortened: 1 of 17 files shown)


go/core/v2/translator/reader.go
```diff
@@ -11,4 +13,12 @@ import (
 // KRT supplies a dependency-tracking implementation; tests may use any reader.
 type Reader interface {
 	Get(context.Context, types.NamespacedName, runtime.Object) error
+	GetModelConfigTranslation(context.Context, types.NamespacedName) (*ModelConfigTranslation, error)
+}
+
+type ModelConfigTranslation struct {
+	Model        adk.Model
```
EItanya: All harnesses should consume a validated and resolved ModelConfig, but this shared type is kagent-specific because it embeds adk.Model, and collectionReader constructs it with kagenttranslator.NewModelCompiler. Could this instead be a harness-neutral resolved model containing validated config, resolved non-secret values, references, and provenance, leaving ADK, Codex, and Claude rendering to their adapters?

go/core/v2/controller/modelconfig.go
```diff
@@ -0,0 +1,170 @@
...
+func newModelConfigReconciliations(
+	modelConfigs krt.Collection[*kagentv1alpha3.ModelConfig],
+	configMaps krt.Collection[*corev1.ConfigMap],
+	secrets krt.Collection[*corev1.Secret],
+	opts krt.OptionsBuilder,
+) krt.StatusCollection[*kagentv1alpha3.ModelConfig, kagentv1alpha3.ModelConfigStatus] {
+	statuses, _ := krt.NewStatusCollection(modelConfigs, func(ctx krt.HandlerContext, modelConfig *kagentv1alpha3.ModelConfig) (*kagentv1alpha3.ModelConfigStatus, *ModelConfigReconciliation) {
+		state := &ModelConfigReconciliation{ModelConfigName: krt.Named{Namespace: modelConfig.Namespace, Name: modelConfig.Name}}
+		reader := collectionReader{ctx: ctx, configMaps: configMaps, secrets: secrets, modelConfigs: modelConfigs}
+		translation, translationErr := kagenttranslator.NewModelCompiler(reader).TranslateModel(context.Background(), modelConfig)
```
EItanya: Could we make this shared resolution return typed semantic-vs-reference failures and validate the effective provider references? Right now TranslateModel dereferences SAP/Bedrock Secrets and Foundry ConfigMaps, so missing references also set Accepted=False, while addSecret misses the TLS caCertSecretKey, SAP client_id/client_secret, and Bedrock fixed credential keys. That can make the two conditions inconsistent or report ResolvedRefs=True for unusable credentials.

EItanya (approved):

EItanya (approved):

PR #2634, opened 2026-08-31
fix(adk): read resource-type MCP content in extractFunctionResponseContent

### LINKED ISSUE #2508: [BUG] github remote mcp response getting truncated
### 📋 Prerequisites

### 🎯 Affected Service(s)

Not Sure

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

github remote mcp get_file_contents response getting striped during forwarding of it to an agent.
In UI on tool call in its "Results" content can be seen.
In tool results it looks like:
```
successfully downloaded text file (SHA: some_sha){"resource":{"mimeType":"text/plain; charset=utf-8","text":"ACTUAL CONTENT OF THE FILE"...
```
But agent sees only `successfully downloaded text file (SHA: some_sha)`
Looks like go/adk/pkg/models/base.go:141-169, function extractFunctionResponseContent handles elements of the response in a wrong way.

### 🔄 Steps To Reproduce

1. Configure the GitHub MCP server (https://api.githubcopilot.com/mcp/, streamable HTTP) as a RemoteMCPServer, expose get_file_contents.
2. Ask an agent to fetch a single file via get_file_contents(owner, repo, path).
3. Observe the tool result the model receives is only:
successfully downloaded text file (SHA: <sha>)
3. with no file content, even though the call succeeded and the SHA is correct.
4. Calling get_file_contents on a directory instead of a file works fine — the full JSON listing comes through

### 🤔 Expected Behavior

agents gets full response

### 📱 Actual Behavior

Agents gets truncated response

### 💻 Environment

- Amazon Linux release 2023.9.20251027
- AWS EKS 1.33
- kagent 0.10.0-beta2

### 🔧 CLI Bug Report

_No response_

### 🔍 Additional Context

_No response_

### 📋 Logs

```shell

```

### 📷 Screenshots

Image

HyeonsuPARK8488:
Fixes #2508.

## Problem

`extractFunctionResponseContent` in `go/adk/pkg/models/base.go` converts an MCP tool response's `content` array into a plain string by reading each item's top-level `text` key. MCP's `resource` content type (used alongside `text`/`image`) carries its payload nested under `resource.text` instead, so items of that type are silently skipped.

GitHub's remote MCP server (`https://api.githubcopilot.com/mcp/`) hits this directly: `get_file_contents` on a text file returns two content items — a top-level placeholder `text` ("successfully downloaded text file (SHA: ...)") and a `resource`-type item whose `resource.text` holds the actual file content. Callers only ever see the placeholder; the real file content never reaches the model.

## Change

In the content-array loop, fall back to `resource.text` when an item has no top-level `text` but is `type: "resource"` with a string `resource.text`. Both are appended to the same `parts` slice and joined as before, so a response with both a placeholder and a resource item now includes both lines.

## Testing

Added `TestExtractFunctionResponseContent` in `base_test.go` (no prior test existed for this function) covering: a plain string response, a content array with only top-level `text` items, and the GitHub MCP `get_file_contents` shape (placeholder `text` + `resource.text`).

## Scope

Limited to `resource.text` (string payload). `resource.blob` (binary/base64) is out of scope for this change.


EItanya (approved):

EItanya (approved):

go/adk/pkg/models/base_test.go
```diff
@@ -64,3 +64,52 @@ func TestMergeSystemInstructionFromConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestExtractFunctionResponseContent(t *testing.T) {
```
mesutoezdil, line 68: bedrock_test.go already defines TestExtractFunctionResponseContent, this won't compile

PR #2639, opened 2026-08-31
fix(controller): restore services and RemoteMCPServer discovery for controller-v2

### LINKED ISSUE #2603: [BUG] MCP servers never appear in the list — nothing writes the toolserver table since the legacy reconciler was removed
## Summary

MCP servers never appear in the UI's **MCP Servers** list. Creating one through the UI succeeds and the `RemoteMCPServer` resource is created correctly, but the list stays at `0 of 0 servers · 0 tools` — and so does a fresh install, where the chart's own `kagent-tool-server` and `kagent-grafana-mcp` are also missing.

This is not a UI defect: the UI renders exactly what `ListToolServers` returns, and that RPC returns an empty list.

## Reproduce

1. Install the chart on a fresh cluster.
2. Open **MCP Servers**. The list is empty, although the chart installed `kagent-tool-server` and `kagent-grafana-mcp`.
3. **Add server** → name `probe-server`, URL `mcp.example.com/sse` → **Create server**.
4. `CreateToolServer` returns 200 and the page returns to the list.
5. `kubectl get remotemcpservers.kagent.dev -n kagent` shows `probe-server` with the correct URL.
6. The list still shows `0 of 0 servers · 0 tools`.

## Where it is lost

`ListToolServers` reads the database, not the cluster:

```
ListToolServers (grpcserver/tool.go:53)
  → Service.ListToolServers (service/tool/service.go:113)
    → discoveryStore.ListToolServers   →  the `toolserver` table
```

On the cluster above, four `RemoteMCPServer` resources exist and the table has **zero rows**:

```
$ kubectl get remotemcpservers.kagent.dev -n kagent --no-headers | wc -l
4
$ psql -c 'select count(*) from toolserver;'
0
```

The RPC response body is 26 bytes — an empty list, faithfully rendered.

## Root cause

`StoreToolServer` is the only writer of that table, and it has **no production caller**. Every reference in the module is either its own implementation or its own unit tests:

```
core/internal/database/client_postgres.go:1632   func (c *postgresClient) StoreToolServer(...)
core/internal/database/client_test.go            (tests only)
api/database/client.go:43                        (interface declaration)
```

The caller was removed in `26732e86` — *chore: remove legacy ACP and controller runtime (#2565)* — which deleted `go/core/internal/controller/mcp_server_tool_controller.go` and `.../reconciler/reconciler.go`, containing:

```diff
-	if _, err := a.dbClient.StoreToolServer(ctx, toolServer); err != nil {
```

Nothing took over populating the table, but the read path still expects it to be populated. So the list can only ever be empty.

## Impact

- Every MCP server is invisible in the UI, including the two the chart installs, on every install.
- A server created through the UI looks like it failed, because the only feedback is a list it never joins.
- `ListTools` is affected the same way — the `tool` table is also empty (0 rows).

## Version

Observed on `2a259337ff496344132064ec53f53c820984156d`.

---
*🤖 written by Claude*

supreme-gg-gg:
Close #2603 and #2613 

Some services are removed after #2595 such as ModelService, ToolService, PromptTemplateService, MemoryService, SystemService. They're restored in this PR. Also restores RemoteMCPController which is required by UI and some Harness compiler for tool syncing. Configures required Kubernetes schemas, namespace-scoped caching, and controller-runtime logging.

### DIFF
--- a/go/core/cmd/controller-v2/main.go
+++ b/go/core/cmd/controller-v2/main.go
@@ -28,11 +28,19 @@ import (
 	"syscall"
 	"time"
 
+	atev1alpha1 "github.com/agent-substrate/substrate/pkg/api/v1alpha1"
 	kagentv1alpha3 "github.com/kagent-dev/kagent/go/api/v1alpha3"
+	remotemcpcontroller "github.com/kagent-dev/kagent/go/core/internal/controller/remotemcpserver"
 	"github.com/kagent-dev/kagent/go/core/internal/database"
 	"github.com/kagent-dev/kagent/go/core/internal/grpcserver"
 	authimpl "github.com/kagent-dev/kagent/go/core/internal/httpserver/auth"
+	feedbackservice "github.com/kagent-dev/kagent/go/core/internal/service/feedback"
 	"github.com/kagent-dev/kagent/go/core/internal/service/kubecrud"
+	memoryservice "github.com/kagent-dev/kagent/go/core/internal/service/memory"
+	modelservice "github.com/kagent-dev/kagent/go/core/internal/service/model"
+	prompttemplateservice "github.com/kagent-dev/kagent/go/core/internal/service/prompttemplate"
+	systemservice "github.com/kagent-dev/kagent/go/core/internal/service/system"
+	toolservice "github.com/kagent-dev/kagent/go/core/internal/service/tool"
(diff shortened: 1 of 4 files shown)


Charlesthebird (commented):
Tested and the UI + chat is working 🎉

EItanya (approved):

PR #2640, opened 2026-08-31
fix(controller): derive full runtime image from the tag name

### LINKED ISSUE #2586: [BUG] Declarative agents with skills append `-full` to IMAGE_TAG, producing an invalid reference when the agent image is digest-pinned
### What happened

When the agent runtime image is configured with a digest-pinned tag, every
declarative agent that declares `spec.skills` fails to roll out. The controller
composes the full-variant image by **string-appending `-full` to `IMAGE_TAG`**,
which produces a reference with the suffix *after* the digest:

```
ghcr.io/kagent-dev/kagent/app:0.10.0-rc3@sha256:0c1d336bd8e9c1d6542b002f2309d38c7be9e80c100e869b7811a6af17a55bce-full
```

An OCI digest is `sha256:` followed by hex and nothing may follow it, so this is
not a parseable image reference. The Deployment is created but no pod is ever
produced: the ReplicaSet stays at 0, the Deployment reports `ReplicaFailure=True`
with `FailedCreate` and eventually `ProgressDeadlineExceeded`. Any previously
running pod keeps serving, so the agent can look healthy while being unable to
roll forward.

On clusters with an admission controller that parses image references, the pod
CREATE is rejected outright at admission with `invalid reference format`. Without
such a controller the kubelet rejects it instead (`InvalidImageName`). Either way
the agent never starts.

Note the composed reference is wrong in two ways: besides the invalid syntax, it
pairs the digest of the **non-full** image with a `-full` suffix. The two tags are
genuinely different images — for `0.10.0-rc3`, `app:0.10.0-rc3` is
`sha256:0c1d336bd8e9c1d6542b002f2309d38c7be9e80c100e869b7811a6af17a55bce` and
`app:0.10.0-rc3-full` is
`sha256:2c32b111dddb35093c2df9db353b3b2eca263bd8e850cff13bc52a6a65f3d014`.

### Code path

`go/core/internal/controller/translator/agent/deployments.go` (v0.10.0-rc3):

```go
func resolveRuntimeImage(registry, repository, tag, digest, imageLabel string, full, pinDigest bool) (string, error) {
	if !pinDigest {
		if full {
			tag += "-full"
		}
		return fmt.Sprintf("%s/%s:%s", registry, repository, tag), nil
	}
	if d := normalizeImageDigest(digest); d != "" {
		return fmt.Sprintf("%s/%s@%s", registry, repository, d), nil
	}
	...
}
```

`tag` is `DefaultImageConfig.Tag`, which comes from the chart's `IMAGE_TAG`
(`controller-configmap.yaml`:
`IMAGE_TAG: {{ coalesce .Values.controller.agentImage.tag .Values.tag .Chart.Version | quote }}`).
Both the concatenation and the `%s/%s:%s` format assume `IMAGE_TAG` is a bare tag,
matching `APP_FULL_IMAGE_TAG ?= $(VERSION)-full` in the Makefile. Nothing validates
or rejects a digest-bearing value.

`full` is set in `resolveInlineDeployment`:

```go
full := needsSRTSettings(agent, specRef.Sandbox)
pinDigest := agent.GetWorkloadMode() == v1alpha2.WorkloadModeSandbox
```

and `needsSRTSettings` returns `true` whenever `spec.Skills != nil`, so any
skills-enabled agent takes the `-full` path.

### No supported way to digest-pin declarative agent images

`--app-image-digest` / `--app-full-image-digest` (and the golang-adk equivalents)
are only read in the `pinDigest` branch, and `pinDigest` is true **only** for
sandbox agents. So for declarative agents:

- the digest overrides are ignored, and
- the only way to express a digest is to embed it in `IMAGE_TAG`, which is exactly
  what breaks the `-full` case.

There is also no value of `IMAGE_TAG` that works for both: a normal agent renders
`repo:TAG@sha256:HEX` (valid, and this is why non-skills agents keep working), but
appending `-full` to that same string is invalid by construction. A single global
setting cannot satisfy both, so an operator who needs digest-pinned agent images
currently cannot use skills at all.

### Why this surfaced in 0.10

`-full` selection is not new, but it used to be reachable only from the Go runtime.

| Version | Python runtime takes the `-full` path? |
|---|---|
| 0.9.4 | No — gated on `runtime == Go && needsSRTSettings(...)` |
| 0.9.12 | No — `resolvePythonRuntimeImage` passed `full=false` unconditionally |
| 0.10.0-rc3 | Yes — `full := needsSRTSettings(agent, ...)`, runtime-independent |

Extending full-variant selection to the Python runtime (#2242) made the latent
concatenation bug reachable for Python agents, which is where it started biting.

### Steps to reproduce

1. Install the chart with a digest-pinned agent image, e.g.
   `controller.agentImage.tag=0.10.0-rc3@sha256:0c1d336bd8e9c1d6542b002f2309d38c7be9e80c100e869b7811a6af17a55bce`.
   (Confirm `IMAGE_TAG` in `ConfigMap/kagent-controller` carries the digest.)
2. Apply any declarative `Agent` with a `spec.skills.gitRefs` entry.
3. Inspect the generated Deployment / its conditions.

**Expected:** the agent's container references the published full-variant image
(`…/app:0.10.0-rc3-full`, or its digest).

**Actual:** the container image is
`…/app:0.10.0-rc3@sha256:0c1d…-full`, which is not a valid reference, and the
agent never rolls out.

Agents without `spec.skills` on the same cluster are unaffected, because
`repo:TAG@sha256:HEX` happens to be a valid reference form.

### Environment

- kagent `0.10.0-rc3` (controller and chart)
- Agent `runtime: python`, `type: Declarative`, non-sandbox workload mode
- Agent images pulled from `ghcr.io/kagent-dev/kagent/*`, pinned by digest

### Suggested fix direction

Not prescriptive — a few directions that would each resolve it:

- Parse the configured reference properly (split tag / digest) instead of
  concatenating, and derive the full variant from the **tag** component, dropping
  or replacing a digest rather than appending after it.
- Provide a distinct setting for the full-variant image (e.g. a
  `controller.agentFullImage.tag` / `APP_FULL_IMAGE_TAG` equivalent surfaced in the
  chart) so operators can pin both variants explicitly and independently.
- Alternatively, honour `--app-full-image-digest` / `--golang-adk-full-image-digest`
  for declarative agents too, which would give digest-pinning parity with sandbox
  agents.

At minimum, it would help to fail fast with a clear error when `IMAGE_TAG` contains
a digest and a full-variant image is required, rather than emitting an unparseable
reference that only fails later at admission or on the kubelet.

I'm happy to put up a PR if you can indicate which direction you'd prefer.

rtemperini:
Fixes #2586.

Declarative skills agents built the full runtime image by appending `-full` to `IMAGE_TAG`. A digest-pinned tag such as `0.10.0-rc3@sha256:…` became `…@sha256:…-full`, which is not a valid OCI reference. The slim and full tags are also different images, so reusing the slim digest would be wrong even if the syntax were valid.

## What changed
The controller parses `IMAGE_TAG` as a tag, a `tag@digest`, or a digest-only value. The `-full` suffix is applied only to the tag name (`0.10.0-rc3` becomes `0.10.0-rc3-full`). The slim digest is never reused on the full image.

Released builds always bake an upstream full digest (ldflags). That value is sandbox fallback only. Declarative skills agents pin the full variant only when the operator sets an explicit runtime full digest.

## What operators should set
* Bare tag (`0.10.0-rc3`): the full image is `app:0.10.0-rc3-full`. Nothing extra to set.
* Tag plus digest (`0.10.0-rc3@sha256:…`): reconcile fails unless an explicit full digest is set. Python: `controller.agentImage.fullDigest` (`APP_FULL_IMAGE_DIGEST`). Go: `controller.goAgentImage.fullDigest` (`GOLANG_ADK_FULL_IMAGE_DIGEST`). The baked upstream digest is not used here.
* Digest-only `IMAGE_TAG`: fail closed unless that runtime full digest is set.

## Why this branch
`main` deleted the Agent translator (`26732e86`). The append lives only on `release/v0.10.x`.

## Tests
`go test -race` on `./core/internal/controller/translator/agent/`, including the `0.10.0-rc3@sha256:deadbeef…` regression (must not produce `…deadbeef…-full`) and fail-closed errors that cite the matching Python or Go fullDigest / env flags.


go/core/internal/controller/translator/agent/deployments.go
```diff
@@ -149,23 +150,53 @@ func resolveGoRuntimeImage(registry string, full, pinDigest bool) (string, error
...
+		return "", fmt.Errorf(
+			"cannot derive %s image from digest-only tag %q: set a tag (for example 0.10.0-rc3) so the controller can use the published %s-full tag, or pin the full image via --%s-image-digest",
+			imageLabel, tag, strings.TrimSuffix(imageLabel, "-full"), imageLabel,
+		)
+	}
+
+	// Never reuse the slim digest on the full image.
+	var fullDigest string
+	if embeddedDigest != "" {
+		fullDigest = normalizeImageDigest(digest)
```
iplay88keys, line 198: Can we align the fallback with what the behavior is here? Controller builds always populate the full image digests with the ldflags script, so `digest` will be non-empty even when the Helm `fullDigest` is unset.

For `IMAGE_TAG=tag@digest`, this emits `tag-full@<baked upstream digest>`; the tag-only fallback described above and in the PR body is reachable only in builds without those linker flags.

So this needs a way to distinguish between an explicit runtime override and a baked-in digtest.

rtemperini: Sure, declarative skills agents now pin the full variant only when `fullDigest` (`APP_FULL_IMAGE_DIGEST`) is set at runtime. The baked digest still applies to sandbox agents. `IMAGE_TAG=tag@digest` with no runtime override is now `app:tag-full`, never the slim digest and never the baked upstream digest.

rtemperini: let me know if this covers this case for baked digest without runtime override

iplay88keys: This resolves the baked-versus-explicit digest issue. However, when `IMAGE_TAG` contains a digest and no explicit full digest is provided, this still silently drops digest pinning and uses tag-full. Can we fail reconciliation with a clear error requiring `fullDigest` instead? An operator who explicitly pins the runtime image should not be downgraded to a mutable tag without opting in.

EItanya (dismissed):

go/core/internal/controller/translator/agent/deployments.go
rtemperini: @iplay88keys Agreed. Digest-pinned IMAGE_TAG without an explicit full digest now fails reconcile instead of falling back to a mutable tag-full. Python and Go errors point at their own fullDigest / env flags.

rtemperini: Hi Jeremy (@iplay88keys ) when you have minute, can you take a quick look on this, we've to set a custom kyverno mutation to fix the malformed digest on our side and will be good to address this properly. Thanks in advance!

iplay88keys: Looks like this is resolved now, thanks!

helm/kagent/values.yaml
```diff
@@ -218,6 +218,13 @@ controller:
     registry: ""
     repository: kagent-dev/kagent/app
     tag: "" # Will default to global, then Chart version
+    # -- Explicit manifest digest (sha256:...) of the full (skills/SRT) Python
+    # agent image. Forwarded as APP_FULL_IMAGE_DIGEST. Required to digest-pin
+    # declarative skills agents when agentImage.tag embeds a digest
+    # (tag@sha256:...); the controller's link-time digest is not used for that
+    # path. Sandbox agents use this when set, otherwise the link-time digest.
+    # @default -- "" (declarative: tag-only full image; sandbox: link-time digest)
```
iplay88keys: Can this be updated as well? It still says fullDigest is only required “to digest-pin” and that declarative agents default to a tag-only image. With tag@digest, omission now causes reconciliation to fail.

rtemperini: Updated both agentImage.fullDigest and goAgentImage.fullDigest comments. Bare tags still resolve to tag-full. tag@digest without an explicit fullDigest now documents the reconcile error.

EItanya (approved):

PR #2645, opened 2026-09-01
feat(harness): Initial Codex Support

supreme-gg-gg:
Runs Codex on top of Substrate. Reuses most plumbings from Claude harness, plus some refactoring, and address some follow-up comments on #2602. Feature coverage for Codex is identical to Claude Code right now, and is mostly in parity with Kagent's own ADK harness. The harness driver uses Codex App Server (JSON-RPC over stdio).

## What Changed

- Adds Codex as a supported Harness, including controller compilation, runtime image, and release packaging.
- Runs Codex with approvals set to never and native sandboxing set to danger-full-access; the Substrate Actor is intentionally the security boundary.
- Supports OpenAI Responses and Bedrock gpt-* models, Streamable HTTP MCP servers, selected skills/plugins, and Shared agents; unsupported provider and runtime options are rejected during compilation.
- Adds non-blocking compilation warnings to `AgentTemplate` harness status; Claude now reports ignored MCP TLS, timeout, and terminateOnClose settings there instead of rejecting the template.
- Moves native continuation persistence and process/file security helpers into a shared runtime layer used by both Codex and Claude, while keeping harness-specific protocol behavior separate.

## Validation

- Unit test coverage for Codex
- Cluster E2E coverage for Codex
- Manual verification (see below recording)


supreme-gg-gg:
Demo with Codex:

https://github.com/user-attachments/assets/845aa23b-ba8c-4908-b48c-884833183f73

go/harness/runtime/processgroup/process_group.go
EItanya: Why the runaround, just add `!unix` to the windows one

EItanya (approved):

Issue #2594, opened 2026-08-27
[FEATURE] Add pod annotations support to agent Helm chart values


ewamaslanka01:
### 📋 Prerequisites

- [x] I have searched the [existing issues](https://github.com/kagent-dev/kagent/issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

The Agent CRD supports annotations in spec.declarative.deployment, but the agent Helm charts under helm/agents/* don't expose this through Helm values.

### ❓ Problem Statement / Motivation

When running agent pods alongside tooling that relies on pod annotations, there's currently no way to set these annotations via the Helm chart values for the agents. The global podAnnotations value works for the controller and UI deployments, but is not passed to the Agent CRs, and controller.agentDeployment only exposes podLabels, not an equivalent podAnnotations default.

### 💡 Proposed Solution

- Update each agent's Helm chart under helm/agents/* to support annotations value that maps to spec.declarative.deployment.annotations on the Agent CR.
- Add a global default under controller.agentDeployment.podAnnotations in the parent helm/kagent chart, analogous to the existing controller.agentDeployment.podLabels, so annotations can be applied to all agent pods at once.

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR for this feature

PR #2647, opened 2026-09-01
feat(helm): expose pod annotations in agent Helm charts

ewamaslanka01:
The Agent CRD supports spec.declarative.deployment.annotations, but the charts under helm/agents/* had no way to set it. Each agent chart gains an `annotations` value, rendered into the Agent CR by agent.deploymentSpec.

Refs #2594


EItanya:
Can we please open this against v0.10.x and then backport?

ewamaslanka01:
> Can we please open this against v0.10.x and then backport?

https://github.com/kagent-dev/kagent/pull/2649

PR #2649, opened 2026-09-01
feat(helm): expose pod annotations in agent Helm charts

ewamaslanka01:
The Agent CRD supports spec.declarative.deployment.annotations, but the charts under helm/agents/* had no way to set it. Each agent chart gains an annotations value, rendered into the Agent CR by agent.deploymentSpec.

Refs https://github.com/kagent-dev/kagent/issues/2594


ewamaslanka01:
@EItanya Can you review, please? (https://github.com/kagent-dev/kagent/pull/2647)

PR #2655, opened 2026-09-01
fix(skills-init): skip existing Git skills

### LINKED ISSUE #2646: skills-init: not idempotent across init-container restarts — leftover clone dest bricks the pod until manual deletion
## Summary

When `skills-init` fails partway through its `gitRefs` list (e.g. a transient network error, or an auth failure on a later private repo), every subsequent restart of the init container fails at the **first** already-cloned ref with:

```
fatal: destination path '/skills/demo' already exists and is not an empty directory.
skills-init: clone https://github.com/giantswarm/agent-skills: exit status 128
```

The pod then sits in `Init:CrashLoopBackOff` forever — the retry loop can never succeed, even if the original cause (network blip, rate limit) has long cleared. The only recovery is deleting the pod by hand to get a fresh `emptyDir`.

## Root cause

The skills volume is an `emptyDir`, which persists across *container* restarts within the same pod. The comment on `Run()` in `go/core/internal/skillsinit/runner.go` assumes otherwise:

> successful operations before the failure are left in place on disk (the container restarts and re-runs from scratch)

The re-run is **not** from scratch: earlier successful clones are still on disk, and `CloneGit` does not tolerate an existing non-empty destination. The same applies to the OCI export and S3 fetch stages.

## Reproduction

1. Agent with ≥2 `gitRefs`, where ref 2 fails (easiest: a private repo with an invalid token in `gitAuthSecretRef`).
2. Attempt 1: ref 1 clones fine, ref 2 fails → init container exits.
3. Attempt 2+: ref 1 fails with `destination path already exists`, masking the real error. Pod is stuck permanently.

Observed on `skills-init:0.9.9`.

## Suggested fix

Make each fetch idempotent, e.g. `os.RemoveAll(ref.Dest)` before cloning (or clone into a temp dir and rename into place). Alternatively wipe the skills root at the start of `Run()`. This also keeps the *real* error visible in the logs on every retry instead of the misleading "already exists" failure.

Tethys0:
## Summary

Fixes #2646.

Make Git skill downloads reentrant across `skills-init` restarts: a destination that already contains a real, regular `SKILL.md` is treated as a completed skill and is not downloaded again. This lets a restart continue after another initialization step fails without re-cloning already materialized skills.

Existing destinations are validated before they are accepted: the destination and `SKILL.md` must not be symbolic links, the destination must be a directory, and `SKILL.md` must be a regular file. Invalid existing destinations return a clear error rather than being silently accepted.

## Scope

- Git ref fetches only.
- OCI and S3 destinations are unchanged.
- No new dependencies or API changes.

## Testing

- `go test -race ./core/internal/skillsinit` (40 passed)
- `go vet ./core/internal/skillsinit`
- `gofmt` and `git diff --check`
- Independent final code review: passed with no blocking security or logic findings.

### DIFF
--- a/go/core/internal/skillsinit/git.go
+++ b/go/core/internal/skillsinit/git.go
@@ -10,6 +10,41 @@ import (
 
 var immutableGitCommit = regexp.MustCompile(`^([0-9a-fA-F]{40}|[0-9a-fA-F]{64})$`)
 
+// existingGitSkill reports whether destination already contains a materialized
+// skill. A destination without a regular SKILL.md is not safe to treat as a
+// completed download, because Run would otherwise continue with an unusable
+// skill after a partial initialization.
+func existingGitSkill(destination string) (bool, error) {
+	info, err := os.Lstat(destination)
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, fmt.Errorf("inspect git destination: %w", err)
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		return false, fmt.Errorf("existing git destination is not a skill: destination is a symbolic link")
+	}
+	if !info.IsDir() {
+		return false, fmt.Errorf("existing git destination is not a skill: destination is not a directory")
+	}
+
+	skill, err := os.Lstat(filepath.Join(destination, "SKILL.md"))
+	if os.IsNotExist(err) {
+		return false, fmt.Errorf("existing git destination is not a skill: SKILL.md is required")
+	}
+	if err != nil {
+		return false, fmt.Errorf("inspect existing git skill: %w", err)
+	}
+	if skill.Mode()&os.ModeSymlink != 0 {
+		return false, fmt.Errorf("existing git destination is not a skill: SKILL.md is a symbolic link")
+	}
+	if !skill.Mode().IsRegular() {
+		return false, fmt.Errorf("existing git destination is not a skill: SKILL.md is not a regular file")
+	}
+	return true, nil
+}
+
 // CloneGit fetches a single git ref into ref.Dest. All user-controlled
 // strings (URL, Ref, SubPath) are passed to git as separate argv entries via
 // exec.Command — they never pass through a shell, so metacharacters in any of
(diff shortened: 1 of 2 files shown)


EItanya (commented):
What if we just make the skills download code reentrant safe? We can just check for the presence of an existing skill, and if it exists don't download the new one

Tethys0:
Thanks — I adopted the reentrant-download approach. The updated implementation now skips a destination only when it already contains a real `SKILL.md`; invalid or symlinked destinations are rejected rather than treated as successful downloads.

Added focused coverage for an existing skill, an incomplete destination, and symlinked paths. Verified with `go test -race ./core/internal/skillsinit` and `go vet ./core/internal/skillsinit`.

This force-update also replaces the earlier staging/replacement implementation.

go/core/internal/skillsinit/git.go
```diff
@@ -22,6 +57,14 @@ var immutableGitCommit = regexp.MustCompile(`^([0-9a-fA-F]{40}|[0-9a-fA-F]{64})$
 // SubPath, if set, rewrites the destination so the final layout matches the
 // requested in-repo subdirectory.
 func CloneGit(ref GitRef) error {
+	exists, err := existingGitSkill(ref.Dest)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return nil
```
EItanya: An existing root `SKILL.md` does not prove that the requested skill finished materializing. If a repository has a root `SKILL.md` and `applySubPath` fails, the clone remains at `ref.Dest`. On the next attempt this early return reports success without applying `ref.SubPath`, silently accepting the repository-root skill instead of the requested one.

I reproduced this with a local repository containing a regular root `SKILL.md` and `SubPath: "missing-skill"`: the first `CloneGit` call fails, but the second returns nil. Against the base, both attempts report failure.

Could we stage cloning and subpath processing in a sibling temporary directory and publish the destination only after both succeed, or use an explicit completion marker written after those steps? Please also cover a failed first attempt followed by a retry; the current pre-populated-destination test does not exercise this path.

EItanya (commented):
🤖 AI-generated review.

Found one retry correctness issue below. The existing skillsinit tests with `-race`, `go vet`, and whitespace checks passed. A focused local-Git regression probe reproduces the issue on this head and passes against the base.

PR #2658, opened 2026-09-01
feat(core): make controller startup importable and its auth replaceable

jjamroga:
## What

Controller startup lives in `func main()` in `package main` — nothing can import it — and the authenticator and authorizer are hardcoded at `cmd/controller-v2/main.go:161-162`.

This moves startup into `go/core/pkg/app` behind `Run(ctx, Options)` and makes those two replaceable. `cmd/controller-v2` becomes a signal handler and one call.

## API

| Field | Zero value | Purpose |
|---|---|---|
| `Authenticator` | `UnsecureAuthenticator` | replace the default |
| `Authorizer` | `NoopAuthorizer` | replace the default |
| `SetupWithManager` | no-op | register controllers and scheme types |
| `HTTPMiddleware` | none | wrap the handler ahead of core's auth |
| `Routes` | none | add endpoints alongside core's |
| `ExtraMigrations` | none | apply tracks after the built-ins |

Every field is optional and its zero value is today's behaviour, so the shipped binary is unchanged.

`SetupLogger` is also exported: callers build `Options` before `Run`, and controller-runtime discards everything logged before the first `SetLogger` — so a startup error in that window vanishes. Calling it twice is a no-op; the first caller wins.

**Not included: the A2A handler.** Core builds the gateway itself, including the instance workflow that suspends an instance at end of turn. A caller-supplied one would be less complete.

## Reviewing this

262 of the 532 added lines are a verbatim move of the old `main()` body:

```sh
git show origin/main:go/core/cmd/controller-v2/main.go | sed -n '/^func main() {/,/^}/p' > /tmp/old.txt
sed -n '/^func Run(/,/^}/p' go/core/pkg/app/app.go > /tmp/new.txt
diff -u /tmp/old.txt /tmp/new.txt
```

Only four edits should appear: signal handling stays in `main`, `log.Fatal` becomes a returned error, the auth pair becomes `opts.resolve()`, and `group.Wait()` is returned. The rest is `Options`, `resolve()`, `chain()`, `SetupLogger` and tests.

**Behaviour change:** `log.Fatal` exited the process, so `db.Close()` and `actors.Close()` never ran on a startup failure. Returning the error runs them.

## Testing

6 unit tests in `go/core/pkg/app`; `go build ./...`, `go vet`, `make -C go lint` (0 issues), `go test -skip 'TestE2E.*' ./core/...`.

Validated end to end on a kind cluster.


go/core/pkg/app/app.go
```diff
@@ -0,0 +1,359 @@
...
+	SetupWithManager func(manager.Manager) error
+	// HTTPMiddleware wraps the HTTP handler, outermost first. It runs before
+	// routing, and therefore before the authentication core applies to its own
+	// endpoints -- which is the only order in which middleware can mark a
+	// request that the authenticator then reads.
+	HTTPMiddleware []func(http.Handler) http.Handler
+	// Routes registers additional HTTP endpoints. It runs after core has
+	// registered its own, so a more specific pattern wins over core's "/"
+	// without core having to know the library consumer's paths.
+	Routes func(*http.ServeMux)
```
EItanya: Is this still relevant given that the majority of our endpoints are gRPC now, do we still use this middleware or do we need more?

go/core/internal/grpcserver/policy.go
```diff
@@ -3,17 +3,20 @@ package grpcserver
...
-	AccessPublic AccessMode = "public"
-	AccessRead   AccessMode = "read"
-	AccessCreate AccessMode = "create"
-	AccessUpdate AccessMode = "update"
-	AccessDelete AccessMode = "delete"
+	AccessPublic = auth.AccessPublic
+	AccessRead   = auth.AccessRead
+	AccessCreate = auth.AccessCreate
+	AccessUpdate = auth.AccessUpdate
+	AccessDelete = auth.AccessDelete
```
EItanya: Why do we need to mirror these consts, can't we just use them directly?

EItanya (approved):

PR #2662, opened 2026-09-01
fix(ui): remove empty substrate template metadata

### LINKED ISSUE #2653: [BUG] Harness column is always empty in the Substrate page's Actor Templates table
## 🎯 Affected Service(s)

UI Service (with a controller-side cause)

## 🚦 Impact/Severity

Minor inconvenience

## 🐛 Bug Description

The Harness column in the Actor Templates table on the Substrate page always shows `—`. Two neighboring fields are dead in the same way: the `golden: …` line under each template name, and `managedByKagent`.

Cause is #2638, which moved actor templates from the Kubernetes `ActorTemplate` CRD to ate-api.

- Before: `ActorTemplateForRevision` stamped `kagent.dev/harness` on the template, and `listSubstrateCRs` read it back into `HarnessName`.
- Now: ate-api templates carry no such labels, and `listATEState` never sets `HarnessName`, `GoldenActorID`, or `ManagedByKagent`.
- The fields are still declared on `SubstrateActorTemplate` (`go/core/internal/service/system/service.go`), so nothing fails — the columns just can't ever have a value.

## 🔄 Steps To Reproduce

1. Bring up a cluster with `scripts/setup-cluster/setup-cluster.sh`.
2. Create an AgentTemplate so at least one actor template exists.
3. Open the Substrate page in the UI and find the Actor Templates table.
4. The Harness column reads `—` for every row.

## 🤔 Expected Behavior

Each row names the harness that produced the template.

## 📱 Actual Behavior

`—` on every row.

### Two more regressions on the same table

Both come from the same move to ate-api, and both are cosmetic rather than empty:

- **Sandbox class** reads `SANDBOX_CLASS_GVISOR`. `listATEState` calls `.String()` on the ate-api enum; the CRD field it replaced was a plain string, so the column used to read `gvisor`.
- **Golden snapshot** now reads `atespace/name`, via `objectRefString`. It used to be the bare snapshot name.

Searching the table by harness also stopped matching, since the row's search string includes `harnessName` and `goldenActorId`. That follows from the empty fields above rather than being separate.

### Not issues, for the record

- Phase values changed from CRD strings such as `GoldenSnapshotPending` to a derived `Pending` / `Failed` / `Ready`. The page treats phase as open-ended and the ready count compares lowercased, so both still work.
- With ate-api unreachable the table is empty instead of falling back to Kubernetes CRs. Inherent to the move; `ateApiError` still surfaces on the page.
- The Actors, Workers, and Worker Pools tables are unaffected — every field still has a source.

### Why not derive the harness from the template name

`revisionActorTemplateName` builds `<agentTemplate>-<harness>-<revisionShort>`, so the harness is readable from the Name column, but parsing it back is unreliable. Agent template and harness names may both contain dashes, and `truncateDNS1123To` caps the combined `agentTemplate-harness` at 50 characters before the digest is appended, so long names lose part or all of the harness segment.

### Fix

Carry the harness name — and the golden actor ID — in ate-api template metadata, and populate the fields from there. Alternatively drop the three empty fields and their UI columns, so the table stops showing values it can never fill. Sandbox class needs formatting back to the short form.

No test covers any of the three empty fields now, which is why CI stayed green through #2638.

### Related files

- `go/core/internal/service/system/service.go` — `listATEState`, `SubstrateActorTemplate`, `objectRefString`
- `go/core/v2/substrate/actor_template.go` — `ActorTemplateForRevision`, `revisionActorTemplateName`, `truncateDNS1123To`
- `ui/src/pages/SubstratePage.tsx` — Actor Templates columns and row search string

---
*🤖 written by Claude*

mikemikimike:
## Summary / Problem

Fixes #2653. After Actor Templates moved from Kubernetes CRDs to ate-api in #2638, `goldenActorId`, `harnessName`, and `managedByKagent` are no longer available from the served ate-api template. The Substrate page consequently exposed values that were always empty or dead.

## Changes

- Remove the unavailable metadata from the internal Go inventory model and gRPC serialization path.
- Remove the corresponding TypeScript domain fields, client mapping, mock data, and the Substrate table's golden-actor and Harness projections.
- Keep the existing v1alpha1 wire fields for backward compatibility; the server no longer populates them and the UI no longer depends on them.
- Add a browser regression assertion that the removed columns and golden-actor display do not reappear.

## Tests

- `go test -skip 'TestE2E.*' ./core/internal/service/system ./core/internal/grpcserver` — passed.
- `go vet ./core/internal/service/system ./core/internal/grpcserver` — passed.
- `golangci-lint v2.13.0` with the repository's enabled checks except the Windows-incompatible kube-api-linter plugin — 0 issues.
- `go run github.com/bufbuild/buf/cmd/buf@v1.72.0 lint` — passed.
- `go run github.com/bufbuild/buf/cmd/buf@v1.72.0 generate` — passed with no generated drift.
- `go run github.com/bufbuild/buf/cmd/buf@v1.72.0 breaking proto --against ".git#branch=main,subdir=proto"` — passed.
- `corepack yarn typecheck` — passed.
- `corepack yarn lint` — passed with 0 errors and 7 pre-existing warnings.
- `corepack yarn test` — 34 files and 412 tests passed.
- `corepack yarn build` — passed.
- `corepack yarn playwright test playwright/tests/substrate --project=chromium --project=firefox` — 16 tests passed.
- `git diff --check` — passed.

## Compatibility / Known limitations

The Substrate Actor Templates table no longer displays the three fields that ate-api cannot provide. The protobuf fields remain on the wire so existing clients are not broken, but they are no longer populated by this service. Full Go CI race tests were not completed locally because Windows has `CGO_ENABLED=0` and no cgo compiler; the repository's kube-api-linter plugin is also Unix-only. The full browser suite was attempted: 173/179 passed, with 6 unrelated Firefox concurrency/mock-stream failures; the focused Substrate suite passed in both engines.

## Issue

Issue: #2653 — https://github.com/kagent-dev/kagent/issues/2653

### DIFF
--- a/go/core/internal/grpcserver/system.go
+++ b/go/core/internal/grpcserver/system.go
@@ -77,15 +77,12 @@ func (s *systemServer) GetSubstrateStatus(ctx context.Context, request *apiv1alp
 	}
 	for _, actorTemplate := range result.ActorTemplates {
 		response.ActorTemplates = append(response.ActorTemplates, &apiv1alpha1.SubstrateActorTemplate{
-			Namespace:       actorTemplate.Namespace,
-			Name:            actorTemplate.Name,
-			Phase:           actorTemplate.Phase,
-			GoldenActorId:   actorTemplate.GoldenActorID,
-			GoldenSnapshot:  actorTemplate.GoldenSnapshot,
-			SandboxClass:    actorTemplate.SandboxClass,
-			WorkerSelector:  actorTemplate.WorkerSelector,
-			HarnessName:     actorTemplate.HarnessName,
-			ManagedByKagent: actorTemplate.ManagedByKagent,
+			Namespace:      actorTemplate.Namespace,
+			Name:           actorTemplate.Name,
+			Phase:          actorTemplate.Phase,
+			GoldenSnapshot: actorTemplate.GoldenSnapshot,
+			SandboxClass:   actorTemplate.SandboxClass,
+			WorkerSelector: actorTemplate.WorkerSelector,
 		})
 	}
 	for _, actor := range result.Actors {
--- a/go/core/internal/service/system/service.go
+++ b/go/core/internal/service/system/service.go
@@ -64,15 +64,12 @@ type SubstrateWorkerPool struct {
 }
 
 type SubstrateActorTemplate struct {
-	Namespace       string
-	Name            string
-	Phase           string
-	GoldenActorID   string
-	GoldenSnapshot  string
-	SandboxClass    string
-	WorkerSelector  string
-	HarnessName     string
-	ManagedByKagent bool
+	Namespace      string
+	Name           string
+	Phase          string
+	GoldenSnapshot string
+	SandboxClass   string
+	WorkerSelector string
 }
 
 type SubstrateActor struct {
--- a/ui/playwright/tests/substrate/substrate.spec.ts
+++ b/ui/playwright/tests/substrate/substrate.spec.ts
@@ -65,14 +65,11 @@ test("substrate: the inventory renders, and partial runtime data says so", async
     await expect(templates).toContainText("kagent/coder-template");
     await expect(templates).toContainText("platform/external-template");
 
-    // The golden actor, beneath the name: it is the snapshot every new actor of this
-    // template is cut from, and the one identifier worth carrying beside the name.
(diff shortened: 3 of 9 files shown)


EItanya:
I fixed this on the backend rather than removing the fields

PR #2663, opened 2026-09-01
fix(otel): restore distributed tracing for the controller and agent runtimes

codyhartsook:
## Summary

Restores distributed tracing for the v2 controller and the established kagent runtime.

- Initialize the controller tracer provider before constructing the instrumented gRPC server.
- Forward controller `OTEL_*` variables into kagent runtime revisions; tracing configuration therefore participates in the revision digest.
- Add focused coverage for tracer registration and OTEL environment forwarding.

jaeger-trace

## Verification

- `go test ./core/...`
- `make -C go lint`
- Focused tests: `go test ./core/internal/telemetry ./core/v2/translator ./core/v2/translator/kagent ./core/v2/translator/claude`
- Kind smoke test: a kagent-harness invocation produces one connected Jaeger trace from `kagent-controller` into the actor.

## Deferred

Claude harness tracing needs runtime telemetry initialization and its mock E2E fixture/image from #2602. That work will be handled with the new harness rather than broaden this controller regression fix.

### DIFF
--- a/go/core/cmd/controller-v2/main.go
+++ b/go/core/cmd/controller-v2/main.go
@@ -42,6 +42,8 @@ import (
 	prompttemplateservice "github.com/kagent-dev/kagent/go/core/internal/service/prompttemplate"
 	systemservice "github.com/kagent-dev/kagent/go/core/internal/service/system"
 	toolservice "github.com/kagent-dev/kagent/go/core/internal/service/tool"
+	"github.com/kagent-dev/kagent/go/core/internal/telemetry"
+	"github.com/kagent-dev/kagent/go/core/internal/version"
 	"github.com/kagent-dev/kagent/go/core/pkg/auth"
 	"github.com/kagent-dev/kagent/go/core/pkg/migrations"
 	"github.com/kagent-dev/kagent/go/core/v2/a2agateway"
@@ -68,7 +70,6 @@ import (
 func main() {
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
-
 	logLevel := zapcore.InfoLevel
 	if value := os.Getenv("ZAP_LOG_LEVEL"); value != "" {
 		if err := logLevel.Set(value); err != nil {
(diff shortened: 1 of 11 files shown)


go/core/cmd/controller-v2/main.go
```diff
@@ -77,6 +79,20 @@ func main() {
 	}
 	ctrl.SetLogger(zap.New(zap.Level(logLevel)))
 
+	// otelgrpc snapshots the global TracerProvider and propagator when its handler
+	// is constructed, so tracing has to be registered before any server is built.
```
krisztianfekete, line 81: Great thing to call out, thanks!

go/core/v2/translator/otel.go
```diff
@@ -0,0 +1,30 @@
...
+		if !found {
+			continue
+		}
+		envVars = append(envVars, corev1.EnvVar{Name: name, Value: value})
+	}
+	slices.SortFunc(envVars, func(a, b corev1.EnvVar) int {
+		return strings.Compare(a.Name, b.Name)
+	})
+	return envVars
+}
```
krisztianfekete, line 31: I know that we had a similar logic before it got removed, but wondering if we should go with an explicit allow-list for OTel env vars instead. 

What I mainly want to avoid is silently adding potentially unwanted data as plaintext e.g. when someone decides to propagate `OTEL_EXPORTER_OTLP_HEADERS`. 

There might also be a problem if someone is setting a e.g. `service.name` via an env var as well, as that would falsely label agents after the var is propagated.

go/core/internal/telemetry/tracing_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if otel.GetTracerProvider() != before {
+		t.Fatal("disabled tracing replaced the global TracerProvider")
+	}
+}
+
+func TestInitTracerProviderRegistersGlobals(t *testing.T) {
+	restoreGlobals(t)
+	t.Setenv("OTEL_TRACING_ENABLED", "true")
+	// "none" selects a noop exporter, so the test dials no collector.
+	t.Setenv("OTEL_TRACES_EXPORTER", "none")
```
krisztianfekete, line 47: Nice!

krisztianfekete (commented):
This is looking very good, added some comments on potentially improving what we had before.

go/core/v2/translator/otel.go
codyhartsook: good idea, added the allow list.

EItanya (approved):

PR #2666, opened 2026-09-02
UI rework polish

toreysoloio:
- Substrate tables now have a unified header design and sort behavior - previously the actor/worker columns had a hand-rolled header to avoid antd default sorting (I assume); instead changed it to use antd's sorting controls while using the custom actor/worker sorter logic
- Added * markers on required form fields on `AgentTemplateForm` and `ModelForm` that were missing it
- MCP form stops inventing a server name when switching to "command" tab - previously switching would result in the name being set to "tool-server"/"remote-server"

substrate tables being more uniform:
image

agent template /model form inputs being marked required:
image

image


ui/src/pages/SubstratePage.tsx
```diff
@@ -917,65 +929,52 @@ export function SubstratePage() {
...
+   * own header — the whole cell clickable, the direction in its chevrons, the same as
+   * the two tables above — while leaving the table nothing to reorder. The order is
+   * carried in `sortOrder` from what the server said it applied, and a click goes back
+   * out as the next read. A comparator here would sort the hundred rows on screen and
+   * call it the cluster sorted.
+   *
+   * Each column's `key` is the server's own sort field, which is what lets the change
+   * handler send `columnKey` straight on.
    */
   const actorColumns: ColumnsType<SubstrateActorEntry> = useMemo(
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

`pagedSortChange` casts `columnKey` to the sort field, so nothing catches a key that isn't one (exactly the `key: "pod"` → `"workerPod"` bug fixed at L975).

```suggestion
  const actorColumns: (ColumnsType<SubstrateActorEntry>[number] & {
    key: SubstrateActorSortField;
  })[] = useMemo(
```

`workerColumns` at L994 takes the same shape.

---
*🤖 written by Claude*

Charlesthebird (commented):
Reviewed `db0b844`. All three fixes work.

One blocking concern. Although `sorter: true` is the right choice here (a comparator would only reorder the current page), the comments written around it credit a server round-trip that never happens. `operations.ts` sorts the whole inventory in the browser.

Nothing else blocks.

**Checked**

- ✅ Reverting the hover selector to `> tr…:hover > td` makes the new spec fail in both browsers (`rgb(221, 215, 231)` vs `rgba(0, 0, 0, 0)`).
- ✅ The new `<th>` header carries `tabindex="0"`, `aria-description="sortable"`, `aria-label`, and `aria-sort` once sorted, and Enter activates it.

---
*🤖 written by Claude*

ui/src/components/agent-template-form/AgentTemplateForm.tsx
```diff
@@ -218,6 +218,10 @@ export function AgentTemplateForm({
         {isCreate ? (
           <Form.Item
             label="Name"
+            /* Marked required only while the form authors: read-only is the details
+               page showing a template that already has a name, and an asterisk there
+               would be asking a reader for something. Same for the model below. */
+            required={!readOnly}
```
Charlesthebird: [nit] **Level:** 🟡 Low · Not Blocking

`isCreate` and `readOnly` are exclusive across both call sites (`AgentTemplateNewPage.tsx:127` passes `isCreate` with no `readOnly`, `AgentTemplateDetailsPage.tsx:532` passes `isCreate={false}`), so this is always `true`.

```suggestion
            required
```

The read-only rationale belongs on the Model configuration item below, which is the genuinely conditional one.

---
*🤖 written by Claude*

ui/src/theme/GlobalStyles.tsx
```diff
@@ -48,18 +48,30 @@ export function GlobalStyles() {
...
+         * of both bodies, so one rule covers both.
+         *
+         * A selected row keeps its own hover: that is the checkbox's feedback, not a
+         * claim that the row is a link.
          */
         .ant-table-wrapper
           .ant-table-tbody
-          > tr.ant-table-row:not(.clickable-table-row):hover
-          > td {
+          .ant-table-row:not(.clickable-table-row):not(.ant-table-row-selected)
```
Charlesthebird: [nit] **Level:** 🟡 Low · Not Blocking

The only `rowSelection` in the app is AgentPage's conversations table, where every selectable row already carries `clickable-table-row` (non-openable rows get a disabled checkbox), so this `:not()` can never match.

```suggestion
          .ant-table-row:not(.clickable-table-row)
```

---
*🤖 written by Claude*

ui/playwright/tests/substrate/substrate.spec.ts
```diff
@@ -377,3 +401,44 @@ test("substrate: the actor and worker tables offer no sort, and the inline ones
...
+  ]) {
+    const row = page.getByTestId(testId).locator(".ant-table-row").first();
+    await expect(row).toBeVisible();
+    const cell = row.locator(".ant-table-cell").first();
+
+    const atRest = (await paint(cell)).background;
+    await row.hover();
+    // Waited out rather than polled: the claim is that nothing happens, and there is no
+    // event for a transition that never starts. See `helpers/style`.
+    const hovered = (await settledPaint(cell)).background;
```
Charlesthebird, line 448: **Level:** 🟡 Low · Not Blocking

`settledPaint` is a fixed 350ms wait, making this the one assertion here that can flake on a loaded CI box.

An `await expect(cell).toHaveClass(/ant-table-cell-row-hover/)` after the hover would at least make a missed hover fail loudly rather than pass.

---
*🤖 written by Claude*

ui/src/pages/SubstratePage.tsx
```diff
@@ -383,55 +385,65 @@ function usePageStack(resetKey: string) {
...
- * Not antd's own `sorter`, deliberately. That reorders the rows the table was
- * given, which for one page out of hundreds of thousands looks like sorting and is
- * not — the first row of the sorted cluster is almost certainly not on this page.
- * So the header sends the column and the direction, and the rows come back ordered.
+ * The header is antd's own — the whole cell is the target, and the direction is its
+ * pair of chevrons — because a page where two tables sort by clicking a header and two
+ * more by clicking the words inside one is a page a reader has to learn twice. What is
+ * not antd's is the sorting: the columns below declare `sorter: true`, which is the
+ * form that gives the header its control and no comparator to run, so the table
+ * reorders nothing. The order goes out with the next read, and the rows come back in
```
Charlesthebird: **Level:** 🔴 High · Blocking

No order is sent anywhere: `GetSubstrateStatus` takes only a namespace, and `localPage` (`api/grpc/operations.ts:1419`) sorts the whole inventory in the browser before slicing a page.

So `sorter: true` is right for the opposite reason to the one given here. The read has already ordered every row, so a comparator would re-sort only the hundred on screen.

Users meet the same claim as the literal "Sorted by the server:" at L479.

---
*🤖 written by Claude*

ui/src/pages/SubstratePage.tsx
```diff
@@ -617,16 +629,16 @@ export function SubstratePage() {
...
    */
-  const [actorSort, setActorSort] = useState<{
-    field: SubstrateActorSortField;
-    order: SubstrateSortOrder;
-  }>({ field: "default", order: "asc" });
-  const [workerSort, setWorkerSort] = useState<{
-    field: SubstrateWorkerSortField;
-    order: SubstrateSortOrder;
-  }>({ field: "default", order: "asc" });
+  const [actorSort, setActorSort] = useState<PagedSort<SubstrateActorSortField>>({
```
Charlesthebird, line 642: **Level:** 🟠 Medium · Not Blocking

`sortField`/`sortOrder` are part of the SWR key (`api/hooks/useSubstrate.ts:71`), so every header click refetches the entire inventory to reorder rows the browser already holds.

---
*🤖 written by Claude*

ui/src/pages/SubstratePage.tsx
```diff
@@ -917,65 +929,52 @@ export function SubstratePage() {
...
-          <SortableHeader
-            title="Status"
-            field="status"
-            sort={actorSort}
-            onSort={(field, order) => setActorSort({ field, order })}
-          />
-        ),
+        title: "Status",
         key: "status",
+        sorter: true,
```
Charlesthebird, line 966: **Level:** 🟡 Low · Not Blocking

Actor `status` and worker `pool` fall through to the same sort key as `default` in `localPage` (`api/grpc/operations.ts:1425`, `:1461`), so these two headers move the label and reorder nothing.

---
*🤖 written by Claude*

ui/src/theme/GlobalStyles.tsx
```diff
@@ -78,8 +90,8 @@ export function GlobalStyles() {
          */
         .ant-table-wrapper
           .ant-table-tbody
-          > tr.ant-table-row.clickable-table-row:active
-          > td {
+          .ant-table-row.clickable-table-row:active
+          > .ant-table-cell {
           background: ${theme.color.primary}4D;
         }
```
Charlesthebird, line 99: [nit] **Level:** 🟡 Low · Not Blocking

The expand-icon rule just below at L107 still keys off `tr.clickable-table-row` (the stale selector this PR fixed above it).

---
*🤖 written by Claude*

Charlesthebird (commented):
I tested this locally as well - I think the red required star on the form fields could say "Required" and be a bit larger.

And the "Remote URL"/"Command" tab buttons on the form are a bit too low contrast.

I have some other design notes/ideas for the UI but won't put them all in this review. Thanks for adding this!

ui/src/pages/SubstratePage.tsx
toreysoloio: Both column arrays now type `key` as their sort field, so a key that isn't one fails to compile (`82907bf6`).


---
*🤖 written by Claude*

ui/src/components/agent-template-form/AgentTemplateForm.tsx
toreysoloio: `required` is unconditional now, with the read-only rationale moved to the model configuration item below it (`82907bf6`).


---
*🤖 written by Claude*

ui/src/theme/GlobalStyles.tsx
toreysoloio: Selector is down to `.ant-table-row:not(.clickable-table-row)` in `82907bf6`, noting that a selectable row is always a clickable one.


---
*🤖 written by Claude*

ui/playwright/tests/substrate/substrate.spec.ts
toreysoloio: The cell is asserted to carry `ant-table-cell-row-hover` before the backgrounds are compared, so a hover that never landed now fails (`82907bf6`).


---
*🤖 written by Claude*

ui/src/pages/SubstratePage.tsx
toreysoloio: Corrected in `82907bf6` — the rationale now cites `localPage` ordering the whole inventory rather than a server sort, and the strip reads "Sorted across the whole inventory" (`ServerOrder` → `AppliedOrder`).


---
*🤖 written by Claude*

ui/src/pages/SubstratePage.tsx
toreysoloio: The cost is documented where the key is built (`82907bf6`). Removing it needs the sort and slice moved out of the transport so a reorder is a re-render, which changes what the operation returns — happy to do that as its own PR if you want it in scope.


---
*🤖 written by Claude*

ui/src/pages/SubstratePage.tsx
toreysoloio: Both keys are named explicitly beside `default` in `82907bf6`, with a note that they are the same ordering — ascending is a no-op, descending reverses it.


---
*🤖 written by Claude*

ui/src/theme/GlobalStyles.tsx
toreysoloio: Rewritten against `.ant-table-row.clickable-table-row` in `82907bf6`.


---
*🤖 written by Claude*

toreysoloio:
> I tested this locally as well - I think the red required star on the form fields could say "Required" and be a bit larger.

I think just the red asterisk is fine personally, since it lets us keep it before the label instead of having to move it after it (plus that much red would be distracting)

ui/src/pages/SubstratePage.tsx
toreysoloio: will be addressed in a followup PR

Charlesthebird (approved):

Charlesthebird (approved):

PR #2667, opened 2026-09-02
Switch to using goose for migrations

iplay88keys:
> [!WARNING]
> This is an intentional clean slate for 1.0:
>
> - Existing 0.10.x databases require a new PostgreSQL database; there is no migration bridge.
> - Downgrade from 1.0 to 0.10.x is unsupported.
> - Starting with 1.0, migrations are append-only and future minor releases retain compatibility with the previous release line.

## Summary

Replace golang-migrate with Goose and establish a clean PostgreSQL migration baseline for Kagent 1.0.

- Capture the current core and optional vector schemas as direct version 1 baselines.
- Commit each schema change and migration record atomically.
- Simplify the database migration CLI around Goose.
- Preserve Helm-controlled vector migrations and out-of-band migration management.
- Add upgrade coverage that begins enforcing release-line compatibility after 1.0 is published.

## Why

golang-migrate records a dirty version separately from the migration transaction. A failure between the schema commit and version update can leave committed schema marked dirty and require an operator to force a version.

Goose records both in the same PostgreSQL transaction, so a failed migration rolls back cleanly and startup can retry it. [Substrate PR #1196](https://github.com/agent-substrate/substrate/pull/1196) made the same change; [this review thread](https://github.com/agent-substrate/substrate/pull/1196#discussion_r3905729286) describes the failure mode.

### Migration engine and baseline

- Replace `github.com/golang-migrate/migrate` with `github.com/pressly/goose/v3`.
- Replace the replay of split core and vector migrations with direct final-schema `000001_initial.sql` baselines.
- Require `Up` and `Down` sections and reject `NO TRANSACTION` migrations.
- Validate every source before applying any migration.
- Reject legacy golang-migrate ledgers with a clear fresh-database error.
- Permit non-destructive startup when the database is ahead of the binary for rolling compatibility.
- Update migration immutability CI to allow this one-time cutover, then protect merged Goose migrations.

### Controller and CLI

- `DATABASE_VECTOR_ENABLED=true` includes vector migrations and registers pgvector types on database connections.
- `SKIP_MIGRATIONS=true` verifies migration state without applying migrations.
- Register both settings in the existing environment-variable registry so `kagent env --component database` documents them and the CLI reuses the same vector-setting name.
- Replace the CLI's dirty-state and `force` behavior with Goose-backed `up`, `down`, `goto`, `status`, and `version` commands.
- Reject destructive CLI operations when the database is newer than the CLI's embedded migrations.

### Upgrade coverage

- Update normal and rolling upgrade suites for Goose state, data survival, previous-release behavior against target migrations, clean-install schema equivalence, and application/schema rollback.
- Use the API v2 AgentInstance interaction E2E; document that its second request verifies the Actor wakes after the first terminal task quiesces it.
- Add `adjacent` and `prev-stable` matrix legs. The setup skips 0.10.x because it predates Goose; there will be no 0.11.x. Once a 1.0 prerelease or final tag exists, the previous-stable leg automatically begins enforcing compatibility with 1.0.

## Verification

- Applied current `origin/main` migrations and the Goose baselines to separate fresh databases; normalized schema-only dumps matched exactly after excluding their intentionally different migration ledgers.
- `make -C go sqlc-generate`
- From `go/`: `go test ./core/pkg/migrations ./core/cli/internal/db/migrate ./core/test/upgrade ./core/internal/database ./core/pkg/app ./core/cmd/controller`
- From `go/`: focused short, race, vet, CLI, and E2E compile checks.
- `make -C go lint`
- `bash scripts/version-resolution_test.sh`


go/core/test/e2e/interaction_test.go
```diff
@@ -52,17 +53,47 @@ var interactionMocks embed.FS
...
+	assertAgentInstanceInteraction(t, fixture)
+}
+
+// TestAgentInstanceInteractionAcrossMigration keeps one Substrate-backed agent
+// alive while the upgrade test applies the next release's Goose migrations.
+func TestAgentInstanceInteractionAcrossMigration(t *testing.T) {
+	signalDir := os.Getenv("KAGENT_UPGRADE_SIGNAL_DIR")
+	if signalDir == "" {
+		t.Skip("KAGENT_UPGRADE_SIGNAL_DIR is not set")
+	}
```
EItanya: What is this signal? This is pretty hard to follow

go/core/pkg/app/app.go
```diff
@@ -166,13 +166,18 @@ func Run(ctx context.Context, opts Options) error {
 	if err != nil {
 		return err
 	}
+	vectorEnabled := envBool("DATABASE_VECTOR_ENABLED")
```
EItanya: Can we define these env vars as consts somewhere so they're properly documented, we used to have a env var package somewhere

go/core/test/e2e/interaction_test.go
```diff
@@ -52,17 +52,20 @@ var interactionMocks embed.FS
 func TestAgentInstanceInteraction(t *testing.T) {
 	t.Parallel()
 	fixture := newInteractionFixture(t, interactionTarget(t), startInteractionMock(t))
+	assertAgentInstanceInteraction(t, fixture)
+	assertAgentInstanceInteraction(t, fixture)
```
EItanya: Why this change, you're explicitly running it twice?

go/core/pkg/cli/db/migrate/migrate.go
EItanya: Why is this code in `pkg` and not `cli`?

go/core/test/e2e/interaction_test.go
iplay88keys: The second task verifies the same AgentInstance wakes successfully and handles a new interaction.

iplay88keys: I have updated the test to match `main` and added a comment for clarity.

go/core/pkg/app/app.go
iplay88keys: Yes, that's a great idea. I have updated them accordingly.

EItanya (approved):

PR #2668, opened 2026-09-03
[UI] Prompts edit support added and form polished

toreysoloio:
- The edit button on prompts table now brings you to the edit form as expected
- When creating/editing a prompt, clicking the cancel button now properly confirms if you want to throw away changes if any were made
- Moved the name input above namespace, to be more consistent with other forms
- Fragment delete button is now red
- Delete button is now disabled when only 1 fragment with empty content (since otherwise clicking the button doesn't have any effect so UI makes sure there ia always at least 1)

image

delete disabled when only 1 empty fragment:

image

### DIFF
--- a/ui/playwright/tests/prompts/prompt-editing.spec.ts
+++ b/ui/playwright/tests/prompts/prompt-editing.spec.ts
@@ -53,8 +53,16 @@ test("prompt libraries: the edit form opens from the list and from the library, 
     await expect(page.getByTestId("prompt-fragments-note")).toContainText("is deleted");
     // And the identity is locked, because the ref addresses the ConfigMap: an edit
     // cannot rename a library or move it to another namespace.
-    await expect(page.getByTestId("prompt-name")).toBeDisabled();
-    await expect(page.getByTestId("prompt-namespace")).toBeDisabled();
+    /*
+     * Read-only rather than disabled, and the difference is the point: these two are
+     * shown so the reader is sure which library they are editing, and a disabled field
+     * is dimmed. Both attributes are asserted so a change back to `disabled` fails
+     * here rather than only being noticed as a colour.
+     */
+    for (const field of ["prompt-name", "prompt-namespace"]) {
+      await expect(page.getByTestId(field)).toHaveAttribute("readonly", "");
+      await expect(page.getByTestId(field)).toBeEnabled();
+    }
   });
 
   await test.step("4. leaving with a draft asks before throwing it away", async () => {
@@ -84,6 +92,9 @@ test("prompt libraries: the edit form opens from the list and from the library, 
       await leave.click();
       await expect(page.getByTestId("prompt-discard-body")).toBeVisible();
       await page.getByRole("button", { name: "Keep editing" }).click();
+      // Waited out before the next exit is tried: the dialog's overlay outlives the
+      // click that dismissed it, and swallows whatever is aimed at the page beneath.
+      await expect(page.getByTestId("prompt-discard-body")).toBeHidden();
       await expect(page).toHaveURL(/\/edit$/);
       await expect(fragmentValue(page, 2)).toHaveValue("Edited by the suite.");
     }
@@ -205,8 +216,10 @@ test("prompt libraries: the create form is the same form, and refuses the same t
     // The other half of the shared form: these two are editable when the library
     // does not exist yet, and marked required because nothing can be created
     // without them.
-    await expect(page.getByTestId("prompt-name")).toBeEnabled();
-    await expect(page.getByTestId("prompt-namespace")).toBeEnabled();
+    for (const field of ["prompt-name", "prompt-namespace"]) {
+      await expect(page.getByTestId(field)).toBeEnabled();
+      await expect(page.getByTestId(field)).not.toHaveAttribute("readonly", "");
+    }
 
     await page.getByTestId("prompt-submit").click();
     await expect(page.getByTestId("prompt-form-errors")).toContainText(
--- a/ui/playwright/tests/prompts/prompt-editing.spec.ts
+++ b/ui/playwright/tests/prompts/prompt-editing.spec.ts
@@ -1,0 +1,203 @@
+import type { Page } from "@playwright/test";
+import { test, expect } from "../../fixtures/test";
+import { dataRows, loadPage, rowNamed, routes } from "../../helpers/app";
+
+/**
+ * Editing a prompt library, from the action that used to go nowhere.
+ *
+ * The list's edit action landed on the read-only detail page: it promised an edit and
+ * delivered the same place the library's name already opened, while
+ * `UpdatePromptTemplate` — implemented end to end, from the RPC down to the ConfigMap
+ * write — was reachable from nowhere in the app. So a library could be created and
+ * read and then never changed.
+ *
+ * Driven as journeys rather than as page tests because the interesting parts are the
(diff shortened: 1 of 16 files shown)


cjlg-soloio (approved):
didn't test but code and screenshots lgtm!

ui/playwright/tests/prompts/prompt-editing.spec.ts
Charlesthebird, line 1: I haven't spent any time organizing these tests, but it would be nice to have a single playwright test file for each resources CRUD flow instead of testing all the cases. I added this issue for it: https://github.com/kagent-dev/kagent/issues/2700

Charlesthebird (commented):
Thanks for fixing this! I tested it locally with: 

```sh
./scripts/setup-cluster/setup-cluster.sh
```

I'm kicking off claude to review, but it all looked good during testing. One note is I think that the edit + delete buttons on the prompts table should be on the rightmost side.

Charlesthebird (commented):

---

`yarn lint` fails on this branch, so the `ui-tests` job will go red: 10 `react-hooks/refs` errors, all from the dirty-tracking refs in `PromptForm`. DCO is failing too — three of the six commits carry no `Signed-off-by` (`0015b69a`, `26c9fd99`, `a7f2418c`).

Everything else holds up.

**What I checked**
- ✅ The form's "a fragment removed here is deleted" note matches the controller — `Service.Update` assigns `configMap.Data = cloneStringMap(data)` at `go/core/internal/service/prompttemplate/service.go:141`.
- ✅ "Add at least one fragment key" is the controller's own rule, rejected at `service.go:130`.
- ✅ An update can't rename or move a library — `prompts.update` sends only `ref` + `data` (`ui/src/api/grpc/operations.ts:700`).
- ✅ `promptEdit` is registered after `promptDetail`, and React Router v7 ranks by specificity — the deep-link spec for `/prompts/kagent/not-a-library/edit` reaches the edit page, so nothing shadows it.
- ✅ The fixture now sorts keys the way `summarize` does (`service.go:189`), so the list's keys column can't disagree with a cluster.
- ✅ The ref pattern does work as shipped — `yarn test:pw prompts` is 14/14 on Chromium and Firefox, the save-doesn't-block-itself path included.
- ✅ No existing spec asserted the old edit-icon → detail behaviour, so the `PromptsPage` link change breaks nothing.

---
*🤖 written by Claude*

ui/src/components/prompts/PromptForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+   *
+   * The baseline is captured at mount rather than read from the `initial` prop, so a
+   * created library counts as work too: its baseline is the empty draft, and a typed
+   * name is something to lose.
+   */
+  const contents = (of: PromptDraft) =>
+    JSON.stringify([of.namespace, of.name, of.rows.map((row) => [row.key, row.value])]);
+  const baseline = useRef(contents(draft));
+  const saved = useRef(false);
+  const isDirty = !saved.current && contents(draft) !== baseline.current;
```
Charlesthebird: **Level:** 🔴 High · Blocking

`yarn lint` reports 10 errors from this block (85, 98, 106, 110), all `react-hooks/refs`, and CI runs `yarn lint` in `ui-tests`. State clears the lint and the fragility together: `saved.current = true` re-arms the blocker only because `setSaving(true)` in the same tick forces a render.

```suggestion
  const [baseline] = useState(() => contents(draft));
  const [saved, setSaved] = useState(false);
  const isDirty = !saved && contents(draft) !== baseline;
```

Then `setSaved(true)` / `setSaved(false)` at 128 and 132, and drop `useRef` from the import.

---
*🤖 written by Claude*

ui/src/pages/PromptEditPage.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const missing = library.error !== undefined && isNotFound(library.error);
+
+  /** Where a save or a cancel lands: the library itself, which is what changed. */
+  const detail =
+    namespace && name
+      ? buildPath(paths.promptDetail, { namespace, name })
+      : paths.prompts;
+
+  async function saveLibrary(payload: CreatePromptTemplateRequest): Promise<void> {
+    if (!namespace || !name) return;
```
Charlesthebird: [nit]

A missing param returns without writing, but the form has already marked itself saved — no toast, no navigation, no error. Unreachable through the route, so a `throw` is the honest branch.

---
*🤖 written by Claude*

ui/src/components/prompts/PromptForm.tsx
```diff
@@ -0,0 +1,303 @@
...
+              : checkIdentity && !nameAccepted
+                ? "Lowercase letters, numbers and hyphens only."
+                : undefined
+          }
+        >
+          <Input
+            data-testid="prompt-name"
+            placeholder="team-prompts"
+            value={draft.name}
+            disabled={identityLocked}
```
Charlesthebird: [nit]

`disabled` greys out the two fields the comment above says are shown so the reader is sure which library they're editing; `readOnly` keeps them uneditable at full contrast.

---
*🤖 written by Claude*

ui/playwright/tests/prompts/prompt-editing.spec.ts
toreysoloio: No change here — following #2700 for the per-resource CRUD consolidation, since splitting these files now would collide with whatever shape that lands on.


---
*🤖 written by Claude*

ui/src/components/prompts/PromptForm.tsx
toreysoloio: Both are state now and the guard is derived from them (`e5a0d3fb`) — you're right that it was fragility and not just lint: as refs it re-armed only because `setSaving` forced a render.


---
*🤖 written by Claude*

ui/src/pages/PromptEditPage.tsx
toreysoloio: Throws now (`e5a0d3fb`), so the form's failure path gets it instead of a save that writes nothing and says nothing.


---
*🤖 written by Claude*

ui/src/components/prompts/PromptForm.tsx
toreysoloio: Switched to `readOnly` in `e5a0d3fb`; the spec asserts both attributes so a change back fails there rather than as a colour.


---
*🤖 written by Claude*

Charlesthebird (commented):

---

DCO is the only blocker left — `0015b69a`, `26c9fd99` and `a7f2418c` still carry no `Signed-off-by`. Everything I raised is fixed, and `yarn lint` is back to 0 errors.

**What I checked in the two new commits**
- ✅ `setSaved(true)` forces the render that re-arms the blocker, so the save no longer leans on the `setSaving` beside it.
- ✅ The `readOnly` swap leaves the identity fields uneditable — `promptDraftIssues` still drops their rules under `identityLocked`, and the spec asserts `readonly` *and* enabled, so a slide back to `disabled` fails there rather than only reading as a colour.
- ✅ The `throw` in `saveLibrary` reaches the form's `catch`, which marks the draft unsaved again and renders `SubmitError` with the message.
- ✅ The actions column is last and still `width: 76`, so the fragment tags take the slack instead of trailing off the end.
- ✅ The `toBeHidden()` wait in step 4b closes the overlay race deterministically, with no added timeout.

---
*🤖 written by Claude*

Charlesthebird (approved):

Charlesthebird (approved):

Issue #2458, opened 2026-08-17
[FEATURE] Python runtime never sets a MeterProvider, so google-adk's 7 built-in GenAI metrics are silently discarded


MartinForReal:
**Affected Service(s):** App Service

## 📋 Prerequisites
- [x] I have searched the existing issues to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow the Code of Conduct

## 📝 Feature Summary

The Python agent runtime never installs a `MeterProvider`, so the **seven GenAI metric instruments that `google-adk` already defines and records** are silently discarded. Wiring a `MeterProvider` (gated `OTEL_METRICS_ENABLED`, default-OFF) turns them on with no new instrumentation code.

## ❓ Problem Statement / Motivation

`kagent-adk` pins `google-adk[a2a,db]>=2.6.2,<3` (`python/packages/kagent-adk/pyproject.toml`). Since v2.x, upstream ADK ships a complete metrics layer in `src/google/adk/telemetry/_metrics.py` under meter scope `gcp.vertex.agent`:

| Instrument | Unit | Kind |
|---|---|---|
| `gen_ai.invoke_agent.duration` | `s` | histogram (buckets 0.1 → 409.6) |
| `gen_ai.invoke_workflow.duration` | `s` | histogram |
| `gen_ai.execute_tool.duration` | `s` | histogram (buckets 0.01 → 81.92) |
| `gen_ai.invoke_agent.inference_calls` | `1` | histogram (0 → 64) |
| `gen_ai.invoke_agent.tool_calls` | `1` | histogram |
| `gen_ai.client.operation.duration` | `s` | semconv helper |
| `gen_ai.client.token.usage` | `{token}` | semconv helper |

Dimensioned by `gen_ai.agent.name`, `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.request.model`, `gen_ai.response.model`, `gen_ai.tool.name`, `gen_ai.tool.type`, `gen_ai.token.type`, `gen_ai.workflow.name`. Token accounting is already handled upstream in `telemetry/_token_usage.py` (`input = prompt + tool_use_prompt`, `output = candidates + thoughts`, with `cached_content_token_count` tracked separately).

[...]

PR #2673, opened 2026-09-03
feat(adk): expose gen_ai token usage Prometheus metric

MartinForReal:
## What

Implements the **GenAI token-usage metrics** observability gap from #2148 for the **Go ADK** runtime. The Go ADK (`google.golang.org/adk/v2`) ships spans only, no metric instruments, so kagent supplies the stdlib Prometheus instrument itself, in the shape of the open/dirty #2149 but rebuilt from scratch against today's `go/adk` source and aligned with the Python runtime's attribute set per #2458.

## How

- `pkg/telemetry/metrics.go` — a `gen_ai_client_token_usage` Prometheus histogram (semconv `gen_ai.client.token.usage`, semconv-recommended buckets) on the default registry.
- **Labels** align with what upstream Google ADK Python emits for the same instrument (one dashboard across runtimes): `gen_ai_token_type` (`input`/`output`), `gen_ai_request_model`, `gen_ai_response_model` (falls back to request model), `gen_ai_provider_name` (semconv well-known value via `SemconvProviderName`), `gen_ai_agent_name`, `gen_ai_operation_name` (`chat`), `error_type`.
- **Gating** — recording and the `/metrics` endpoint are **default-OFF** behind `OTEL_METRICS_ENABLED=true`, matching the existing `OTEL_TRACING_ENABLED` / `OTEL_LOGGING_ENABLED` gates.
- `pkg/a2a/server/server.go` — serves `/metrics` on the A2A mux when the gate is on, excluded from request tracing/span flushing.
- `pkg/a2a/executor.go` — records token usage from the ADK `AfterEventCallback`: skips partial (streaming) events so one observation is recorded per LLM call; `input = PromptTokenCount`, `output = CandidatesTokenCount + ThoughtsTokenCount` from `UsageMetadata`.
- `cmd/main.go` / `pkg/config` — model + provider labels resolved at startup; agent name from `AppName`.
- `docs/genai-token-metrics.md` — documents the metric, labels, gating, and scraping.

## Testing

- `go build ./adk/...`, `go vet`, and golangci-lint (repo config) are clean.
- `go test -race` passes for the changed packages (telemetry, a2a, a2a/server, config).
- Two focused test files: `metrics_test.go` (record / skip-zero / disable-gate / response-model fallback / labels / provider-name mapping / /metrics handler) and `executor_metrics_test.go` (partial events not double-counted, per-LLM-call aggregation).

## Notes / decisions

- **Pod scrape annotations**: the annotation piece of #2099 targeted `go/core/internal/controller/translator/agent/manifest_builder.go`, which no longer exists on `main` after the controller/runtime rework. That path was dropped rather than reinvented; operators can add `prometheus.io/*` annotations and this is documented. Flag if you want the annotation automation wired against the current runtime-provisioning location.
- `make -C go lint` could not be run here: `make` and the kube-api-linter plugin are unavailable in this environment. Lint was run with the repo's `.golangci.yaml` minus the plugin (0 issues); kube-apilinter only lints `api/v1alpha` CRD paths, untouched here.

Refs: #2148


MartinForReal:
Hi @supreme-gg-gg, friendly ping for review when you have a moment. This exposes the gen_ai token usage Prometheus metric for the Go ADK runtime. Ready for another look. Thanks!

PR #2675, opened 2026-09-03
feat(go): add gated OTLP metrics, reconcile events, and log bridge (#2148)

MartinForReal:
Implements the Go-side observability gaps from **#2148** as three independent, **default-OFF** additions so the Go runtime/controller expose the same OpenTelemetry signals as the Python runtime.

## 1. Go ADK — GenAI token-usage metrics (`go/adk`)

Configures a MeterProvider with a periodic OTLP metric exporter and emits `gen_ai.client.token.usage` (Int64 histogram, unit `{token}`) per LLM call.

- Attribute set matches the Python runtime so one dashboard covers both runtimes: `gen_ai.token.type` (input/output), `gen_ai.request.model`, `gen_ai.response.model`, `gen_ai.provider.name`, `gen_ai.agent.name`.
- The A2A executor records one input + one output observation per LLM call (prompt tokens; output = candidate + reasoning tokens), skipping streamed `Partial` events so chunked streams are not double-counted.
- Provider/model labels resolved from agent config (incl. `config_usage.ModelName`); response model falls back to the request model.
- Gated behind `OTEL_METRICS_ENABLED` (default OFF) and the standard `OTEL_EXPORTER_OTLP_*` endpoint/protocol resolution shared with traces.
- Tests: `adk/pkg/telemetry/metrics_test.go` (recording via ManualReader, provider-name mapping) and `adk/pkg/a2a/executor_metrics_test.go` (token accounting + partial/nil guard).

## 2. Kubernetes Events on reconcile (`go/core`)

The discovery reconcilers surface lifecycle events via an optional `EventRecorder`, mirroring the agentharness-substrate-controller pattern:

- **RemoteMCPServer** (`internal/controller/remotemcpserver`): `Normal ToolsDiscovered`, `Warning ValidationFailed`, `Warning ReconcileFailed`.
- **MCPServer** (MCPServerTool role, `internal/controller/mcpserver`): `Normal ToolsDiscovered`, `Warning ValidationFailed`, `Warning ReconcileFailed`.

**Note on Agent/ModelConfig:** in current `main` those objects reconcile through the KRT-based status collections in `go/core/v2/controller`, not classic controller-runtime `Reconcile` installers, so they have no EventRecorder path to attach events to. The events gap is therefore implemented on the two remaining classic discovery controllers — the direct successors of the MCPServerTool/RemoteMCPServer roles named in the issue. Emission is a no-op when no recorder is wired (the default).

## 3. Controller log → OTLP bridge (`go/core`)

- `telemetry.InitLoggerProvider` builds an OTLP LoggerProvider via `autoexport` (same resource attributes as traces) and sets it global.
- `telemetry.ControllerZapOpts` additively tees the controller zap logger with an `otelzap` bridge core, preserving stdout.
- Wired into `cmd/controller-v2/main.go` before the controller logger is built.
- Gated behind `OTEL_LOGGING_ENABLED` (already plumbed via the Helm chart `otel.logging.enabled`); when unset both functions are no-ops and logs are byte-identical.

## Validation
- `go build ./...`, `go vet` on all changed packages, `gofmt`/goimports clean.
- golangci-lint on changed packages: **0 issues**.
- `go test` passes for `adk/pkg/a2a`, `adk/pkg/telemetry`, `adk/pkg/config`, `core/internal/telemetry`, and `core/internal/controller/{mcpserver,remotemcpserver}` (incl. the new event tests).
  - `pkg/session` has two pre-existing Windows-only TempDir-cleanup failures (SQLite handle held open during cleanup); unrelated to this change.

## Dependencies
Promotes `otlplog`, `otlpmetric`(grpc/http), `otel/log`, `otel/metric`, `otel/sdk/metric` to direct requires at the same versions already in the graph, and adds `go.opentelemetry.io/contrib/bridges/otelzap v0.19.0`. No version bumps.

Refs: #2148

### DIFF
--- a/go/adk/pkg/telemetry/metrics.go
+++ b/go/adk/pkg/telemetry/metrics.go
@@ -1,0 +1,152 @@
+package telemetry
+
+import (
+	"context"
+	"net/url"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp"
+	"go.opentelemetry.io/otel/metric"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
+)
+
+// GenAI token-usage instrumentation. The metric and attribute names follow
+// the OpenTelemetry GenAI semantic conventions, and the attribute set matches
+// the one upstream google-adk records for gen_ai.client.token.usage, so a
+// single dashboard works across the Go and Python runtimes.
+const (
+	metricGenAIClientTokenUsage = "gen_ai.client.token.usage"
+	genAIMeterScope             = "gcp.vertex.agent"
+
+	attrGenAITokenType     = "gen_ai.token.type"
+	attrGenAIRequestModel  = "gen_ai.request.model"
+	attrGenAIResponseModel = "gen_ai.response.model"
+	attrGenAIProviderName  = "gen_ai.provider.name"
+	attrGenAIAgentName     = "gen_ai.agent.name"
+
+	tokenTypeInput  = "input"
+	tokenTypeOutput = "output"
+)
+
+// tokenUsageHistogram records gen_ai.client.token.usage per LLM call. It is
+// set when the meter provider is initialized (metrics enabled); otherwise it
+// stays nil and recording is a cheap no-op, keeping the gate default-OFF.
+var tokenUsageHistogram metric.Int64Histogram
+
+// TokenUsage carries the per-LLM-call labels and token counts for one
+// recording on the gen_ai.client.token.usage histogram. RequestModel and
+// ProviderName are resolved at startup from the agent config; ResponseModel
+// falls back to RequestModel when a specific response model is unavailable.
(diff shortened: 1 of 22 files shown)


MartinForReal:
Hi @supreme-gg-gg, friendly ping for review when you have a moment. This is the gated Go observability work for #2148 (OTLP metrics, reconcile Events, controller log to OTLP bridge). I have rebased it on latest main and resolved the conflicts, so it is ready for another look. Thanks!

go/adk/pkg/telemetry/metrics.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+
+	"github.com/kagent-dev/kagent/go/pkg/tracing"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
```
mesutoezdil: duplicates #2673's metric using otlp instead, intentional?

MartinForReal:
Addressed: the adk-side OTLP token metric that duplicated #2673 was removed entirely (go/adk/pkg/telemetry/metrics.go and its hooks/tests are gone from this branch). The PR now stays focused on the controller-scoped OTLP metrics, reconcile events, and the controller log bridge; the GenAI token metric lives only in #2673.

PR #2683, opened 2026-09-03
Add scoped authorization for AgentTemplate, Harness, and ModelConfig

iplay88keys:
## Summary

- Add a policy-neutral `CollectionAuthorizer` contract for scoped resource access.
- Enforce scoped access for `AgentTemplate`, `Harness`, and `ModelConfig`.
- Return action capabilities that let the UI disable unavailable controls.
- Keep current OSS access unchanged when no scoped authorizer is installed.

## Authorization contract

The contract supports `ALL`, `NONE`, and composable `ANY_OF` scopes with `IN` predicates.

Kagent supplies trusted `namespace` and `name` attributes. An authorizer defines policies, identities, roles, and catalog keys.

The shared Kubernetes matcher validates each scope. It denies malformed scope output.

Single-resource operations authorize stored or validated resource data. Update operations authorize both the stored and proposed resources.

List operations apply the scope before sorting and response construction. `ScopeNone` returns an empty protected collection.

The no-op authorizer returns `ScopeAll`. This behavior keeps current OSS visibility and write access unchanged.

## Response capabilities

Each protected list response includes `can_create`.

Each returned `AgentTemplate` and `ModelConfig` includes `can_update` and `can_delete`.

Each returned `Harness` includes `can_delete`. The `Harness` service has no update RPC.

The UI uses these fields to disable create, edit, and delete controls. The RPC handlers still enforce every action.

The list response does not include `can_list`. A Boolean cannot represent access to only part of a collection.

The applied list scope defines visible resources. `can_create` reports a separate action that the UI must control.

## Scope

The protected resource types are `AgentTemplate`, `Harness`, and `ModelConfig`.

`AgentHarness`, `SandboxAgent`, `AgentInstance`, and `ModelProviderConfig` remain outside this change.

`ListConfiguredProviders` remains unchanged. It reads `ModelProviderConfig`, not `ModelConfig`.

## Validation

- `make proto-generate`
- `make proto-lint`
- `go test ./internal/grpcserver ./internal/service/kubecrud ./internal/service/model`
- `corepack yarn typecheck`
- `corepack yarn test`


iplay88keys:
Replaced with: https://github.com/kagent-dev/kagent/pull/2710

PR #2692, opened 2026-09-04
feat: per-request pass-through headers on ModelConfig

TarasLykhenko:
## Motivation

kagent supports static `defaultHeaders` on outbound LLM calls, and `apiKeyPassthrough` for forwarding the caller's Bearer credential. There is no way to forward an arbitrary **caller-supplied** header per request — e.g. a guardrail token, tenant token, or request id that a proxy in front of the LLM provider needs. Organizations running LLM traffic through policy gateways need the caller's own token on each outbound call, which static configuration cannot express.

## What this adds

```yaml
apiVersion: kagent.dev/v1alpha3
kind: ModelConfig
spec:
  defaultHeaders:            # unchanged: static values
    x-tenant: acme
  passthroughHeaders:        # NEW: names resolved from each incoming A2A request
    - x-guardrail-token
```

Header names are case-insensitive; values are read per request from the incoming A2A call context. A pass-through value overrides a `defaultHeaders` entry of the same name; headers absent from the request are omitted. `Authorization` is rejected at admission (CEL) — credential forwarding stays on `apiKeyPassthrough` — and both runtimes additionally drop credential carriers (`Proxy-Authorization`, `Cookie`) and RFC 9110 hop-by-hop names at runtime.

## Changes by layer

- **CRD (v1alpha3)**: `passthroughHeaders []string` with bounds and CEL validation; envtest CEL test; manifests + Helm CRDs regenerated.
- **Wire**: `passthrough_headers` on `adk.BaseModel`; `populateHeaderFields` helper (mirrors `populateTLSFields`) in the adkconfig translator, covering all 10 provider branches.
- **Go runtime**: `allowedRequestHeaders` lifted from `pkg/mcp` into a shared `pkg/headers` package (a2a-go v2); `headerTransport` overlays per-request headers from the call context after static ones. 8/10 providers get this via `BuildHTTPClient`; GeminiVertexAI and SAPAICore log a warning — they build clients without the shared transport and already drop `defaultHeaders`/TLS today (pre-existing gap, left as-is here).
- **A2A gateway**: a `callerHeadersInterceptor` on the runtime dialer relays caller-supplied custom headers from the public request onto the private runtime call, excluding credential, hop-by-hop, gRPC pseudo/system, and interceptor-owned (`traceparent`, `x-a2a-extensions`) metadata. Without this hop, neither `passthroughHeaders` nor the existing MCP `allowedHeaders` can see caller headers through the gateway; the runtime-side allowlists remain the actual gate.
- **Python runtime**: `LLMHeaderPassthroughPlugin` resolves session-state headers (case-insensitive, restricted names filtered, empty values omitted) onto `llm_request.config.http_options` for OpenAI/Azure (per-call `extra_headers`) and Gemini (honoured natively); Anthropic uses a cache-invalidating `set_passthrough_headers`, the same mechanism as `set_passthrough_key` (same shared-client concurrency caveat, documented). Bedrock, Ollama, VertexAI, and SAP AI Core log a warning and ignore the field.

## Design points for reviewers

1. **Gateway forwarding scope** — the interceptor forwards *all* non-restricted custom caller headers rather than resolving the instance's configured allowlist at the gateway (the dialer has no ModelConfig access, and the runtime filters by allowlist anyway). If you'd rather plumb the configured names into the resolved bundle instead, happy to rework.
2. **Anthropic delivery** — per-call headers can't reach `messages.create` inside the ADK Claude base class, hence the cache-invalidation setter. A per-request carrier in google-adk would allow removing it.

## Testing

Unit tests at every layer (transport overlay + credential protection, restricted-name filter, wire round-trip, gateway interceptor, CEL admission via envtest, Python plugin + provider behavior), plus a regression asserting a `defaultHeaders`-only config produces byte-identical requests.

### DIFF
--- a/go/adk/pkg/headers/headers.go
+++ b/go/adk/pkg/headers/headers.go
@@ -1,0 +1,92 @@
+// Package headers resolves per-request HTTP headers from the incoming A2A
+// call context. It is shared by the MCP tool transport (allowedHeaders) and
+// the model transport (passthroughHeaders) so both forward caller-supplied
+// headers with identical semantics.
+package headers
+
+import (
+	"context"
+	"strings"
+
+	"github.com/a2aproject/a2a-go/v2/a2asrv"
+)
+
+// restrictedPassthroughHeaders are names that must never be forwarded from a
+// caller onto another hop:
+//   - credential headers (Authorization, Proxy-Authorization, Cookie) would
+//     forward a caller credential onward or clobber the credential the
+//     receiving hop manages itself; apiKeyPassthrough is the supported
+//     mechanism for credential forwarding;
+//   - the rest are hop-by-hop or message-framing headers per RFC 9110, plus
+//     the non-standard Proxy-Connection.
+//
+// Must stay in sync with RESTRICTED_PASSTHROUGH_HEADERS in
+// python/packages/kagent-adk/src/kagent/adk/_llm_header_passthrough_plugin.py.
+var restrictedPassthroughHeaders = map[string]struct{}{
+	"authorization":       {},
+	"connection":          {},
+	"content-length":      {},
+	"cookie":              {},
+	"host":                {},
+	"keep-alive":          {},
+	"proxy-authenticate":  {},
+	"proxy-authorization": {},
+	"proxy-connection":    {},
+	"te":                  {},
+	"trailer":             {},
+	"transfer-encoding":   {},
+	"upgrade":             {},
+}
+
+// IsRestricted reports whether a header name (case-insensitively) must never
+// be forwarded from a caller onto another hop.
+func IsRestricted(name string) bool {
+	_, restricted := restrictedPassthroughHeaders[strings.ToLower(name)]
+	return restricted
+}
+
+// FilterRestricted drops restricted names (case-insensitively) from a
+// configured pass-through header list.
+func FilterRestricted(names []string) []string {
+	var out []string
+	for _, n := range names {
+		if !IsRestricted(n) {
+			out = append(out, n)
+		}
+	}
+	return out
+}
+
+// AllowedRequestHeaders reads the incoming A2A request metadata from ctx and
+// returns only the header key/value pairs whose names appear in allowed.
+// It reads directly from the A2A CallContext that is already present in the Go
+// context, avoiding a redundant copy.
+//
+// Lookup relies on ServiceParams.Get, which does a case-insensitive lookup
+// (NewServiceParams lowercases keys at construction). Keys in the result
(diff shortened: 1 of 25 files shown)


go/adk/pkg/headers/headers.go
```diff
@@ -0,0 +1,92 @@
...
+	"host":                {},
+	"keep-alive":          {},
+	"proxy-authenticate":  {},
+	"proxy-authorization": {},
+	"proxy-connection":    {},
+	"te":                  {},
+	"trailer":             {},
+	"transfer-encoding":   {},
+	"upgrade":             {},
+}
```
EItanya, line 40: Can you please add this list to the k8s API documentation as well, and maybe even add a apiserver side validation if possible.

go/core/internal/a2agateway/runtime.go
```diff
@@ -104,3 +108,49 @@ func (u *upstreamAuthInterceptor) Before(ctx context.Context, req *a2aclient.Req
...
+	}
+	return ctx, nil, nil
+}
+
+// forwardableCallerHeader reports whether a public-request metadata key is a
+// caller-supplied custom header safe to relay to the runtime. gRPC pseudo and
+// grpc-* keys never leave their hop; the named transport headers describe the
+// public request rather than the caller's intent; traceparent/tracestate and
+// x-a2a-extensions are owned by the other interceptors on this client.
+func forwardableCallerHeader(name string) bool {
```
EItanya, line 150: I'm a little worried about this being always on with no configurability. What are the security implications of that?

go/adk/pkg/headers/headers.go
TarasLykhenko: Done in f532c4ea. The passthroughHeaders API doc now enumerates the full restricted set. To keep the CEL rule and the runtime filter from drifting, the envtest CEL test now iterates over `headers.RestrictedNames()` and asserts that every name is also rejected by the API server.

go/core/internal/a2agateway/runtime.go
TarasLykhenko: Yeah, fair point. With the current proposed implementation, we relay everything a caller sends.

What I'm thinking instead, but I wanted to check first, is to forward only what the agent opted into. At compile time we already know every `passthroughHeaders` list in the revision, so we could store that union on `runtime_revision` (like `egress_destinations`) and have the gateway forward just those names. Nothing gets forwarded unless a ModelConfig asks for it, and since the gateway and the runtime read the same list they stay in sync.

Does that sound reasonable, or would you rather go a different way?

PR #2708, opened 2026-09-04
Share PostgreSQL with Substrate

iplay88keys:
> [!IMPORTANT]
> This PR depends on [agent-substrate/substrate#1752](https://github.com/agent-substrate/substrate/pull/1752) and [kagent-dev/substrate#32](https://github.com/kagent-dev/substrate/pull/32). After both merge and release, this branch still needs its Substrate image/chart dependency bumped and locked.

## Summary

- Default embedded Substrate to Kagent's selected PostgreSQL database while keeping Substrate objects in a dedicated `substrate` schema.
- Disable Substrate's bundled PostgreSQL so the default installation deploys only Kagent's bundled database.
- Create a release-scoped Substrate connection Secret from either Kagent's inline external URL or bundled PostgreSQL URL.
- Replace `database.postgres.urlFile` with the Kubernetes-native `database.postgres.secretRef.{name,key}` configuration for Kagent 1.x.
- Mount Kagent's selected Secret as an updatable volume and pass `POSTGRES_DATABASE_URL=@file:...` to the controller.
- Re-read file-backed credentials and TLS material through pgx `BeforeConnect` whenever a new physical connection is opened.
- Wire the existing PostgreSQL pool settings into the controller, including `database.postgres.pool.maxConnLifetime` for bounded credential turnover.
- Keep ownership explicit: Kagent reads `database.postgres.*`, while embedded Substrate reads `substrate.postgres.*`.
- Support bundled sharing, an existing shared Secret, and separate Kagent/Substrate runtime/DDL credentials.

### Bundled database shared by default

```yaml
substrate:
  enabled: true
```

Kagent and Substrate use Kagent's bundled PostgreSQL instance. Kagent uses its normal schema and Substrate uses the dedicated `substrate` schema.

### Existing external Secret shared by Kagent and Substrate

Helm cannot dynamically copy one subchart value into another. Set the same Secret reference on both sides:

```yaml
database:
  postgres:
    bundled:
      enabled: false
    secretRef:
      name: shared-postgres
      key: connectionString
    pool:
      maxConnLifetime: 10m

substrate:
  enabled: true
  postgres:
    enabled: false
    schema: substrate
    connectionStringSecretRef:
      enabled: true
      name: shared-postgres
      key: connectionString
    pool:
      maxConnLifetime: 10m
```

### Separate Kagent, Substrate runtime, and Substrate DDL credentials

```yaml
database:
  postgres:
    bundled:
      enabled: false
    secretRef:
      name: kagent-postgres
      key: connectionString
    pool:
      maxConnLifetime: 10m

substrate:
  enabled: true
  postgres:
    enabled: false
    schema: substrate
    connectionStringSecretRef:
      enabled: true
      name: substrate-postgres
      key: runtimeConnectionString
    ddlConnectionStringSecretRef:
      enabled: true
      name: substrate-postgres
      key: ddlConnectionString
    pool:
      maxConnLifetime: 10m
```

Both Substrate connections may point to the same PostgreSQL server and database as Kagent. The dedicated schema prevents Substrate's runtime grants from covering Kagent tables. The DDL role owns and migrates the Substrate schema and performs outbox partition maintenance; the runtime role receives only the table and sequence permissions ateapi needs.

Omitting the Substrate DDL connection preserves single-connection operation.

## Credential rotation

Kubernetes updates mounted Secret volumes in place. Kagent and Substrate re-read their selected files before opening each new physical pgx connection; an application reload or pod restart is not required.

Existing connections keep their current credentials until retired. Set both `database.postgres.pool.maxConnLifetime` and `substrate.postgres.pool.maxConnLifetime` to bound the rotation window, and keep old and new credentials valid during that overlap.

Rotation may change passwords and TLS material, but not host, port, fallback hosts, database, or user. Identity changes require a restart. Inline URLs remain startup-static.

This supports externally rotated credentials written into Kubernetes Secrets. It does not mint RDS IAM tokens in-process; per-dial token generation remains separate work, although the `BeforeConnect` integration is the required connection hook.

## Migration and DDL credentials

Kagent's existing separation remains available: run `kagent db migrate up --db-url <migrator-url>` out of band, configure restricted runtime credentials, and set `database.postgres.skipMigrations=true`.

Substrate's split is provided by agent-substrate/substrate#1752. Its DDL connection handles migrations and runtime partition maintenance, while its runtime connection handles normal traffic. Both credentials remain in ateapi because partition maintenance is still in-process.

## Breaking change

`database.postgres.urlFile` and `POSTGRES_DATABASE_URL_FILE` are removed for Kagent 1.x. Use `database.postgres.secretRef` in Kubernetes, or pass an `@file:/absolute/path` value through `POSTGRES_DATABASE_URL` when running the binary directly. The chart fails with migration guidance if the removed Helm value is supplied.

## Testing

- `go test -short ./core/internal/database ./core/pkg/app ./core/pkg/env`
- controller Helm unit tests — 65/65 passing
- full Helm unit suite against the updated local Substrate chart — 327/327 passing
- `helm lint helm/kagent` (with the existing chart-version warning)
- rendered shared-Secret integration check for both Kagent and Substrate mounts and `@file:` sources
- `git diff --check`

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -207,13 +207,16 @@ Go code lives in the `go/` workspace (`go/api`, `go/core`, `go/adk`). Run
 
 ### Logging
 
-- Structured logging only (`logr` via controller-runtime); key-value pairs,
-  never `fmt.Sprintf` inside a log call, never `fmt.Printf` for logging.
-- Controllers/HTTP: `ctrllog.FromContext(ctx).WithName("...").WithValues(...)`.
-  ADK: `logr.FromContextOrDiscard(ctx)`.
-- Verbose or per-item output **MUST** be debug level (`log.V(1)` or higher),
-  not `Info`. If the information is already returned to the caller (e.g. an
-  HTTP response), don't also log it at info level.
+- Use standard-library `log/slog`. Binaries write JSON to stderr and read the
+  minimum level from `LOG_LEVEL` (`debug`, `info`, `warn`, or `error`).
+- Carry loggers in `context.Context` with `pkg/logging`; do not add logger
+  parameters, package-global loggers, or `NewXxxWithLogger` constructors.
+- Use `DebugContext`, `InfoContext`, `WarnContext`, or `ErrorContext` whenever
+  a context is in scope. Messages are static and lower-case; keys are
+  snake_case. Record errors under the `error` key and never log secrets,
+  credentials, authorization headers, or request/response payloads.
+- Verbose or per-item output **MUST** be debug level. If information is already
+  returned to the caller, do not also log it at info level.
 
 ### Naming
 
--- a/contrib/cncf/technical-review.md
+++ b/contrib/cncf/technical-review.md
@@ -325,7 +325,7 @@ Default values can be found in [helm/kagent/values.yaml](https://github.com/kage
 **Additional Configurations:**
 For production use, configure:
 
-- External PostgreSQL connection (set `database.postgres.bundled.enabled=false` and set either `database.postgres.url` or `database.postgres.urlFile`)
+- External PostgreSQL connection (set `database.postgres.bundled.enabled=false` and configure `database.postgres.url` or `database.postgres.secretRef`)
 - LLM API keys via Secrets (`providers.openAI.apiKeySecretRef`)
 - TLS for external LLM connections (`modelConfig.tls`)
 - Resource limits based on workload (`agents.*.resources`)
--- a/go/.golangci.yaml
+++ b/go/.golangci.yaml
@@ -12,18 +12,21 @@ linters:
     - copyloopvar
     - depguard
     - dogsled
+    - forbidigo
     - goprintffuncname
     - govet
     - importas
     - ineffassign
     - iotamixing
     - kubeapilinter
+    - loggercheck
     - makezero
     - misspell
     - modernize
     - nakedret
     - nolintlint
     - staticcheck
+    - sloglint
     - unused
     - whitespace
   settings:
@@ -70,11 +73,26 @@ linters:
           deny:
             - pkg: sort
               desc: Should be replaced with slices package
+        structured-logging:
+          files:
+            - $all
+            - "!**/pkg/logging/**"
+          deny:
+            - pkg: github.com/go-logr/logr
+              desc: Use log/slog; adapters belong in pkg/logging
+            - pkg: github.com/go-logr/zapr
+              desc: Use log/slog
+            - pkg: go.uber.org/zap
+              desc: Use log/slog
     govet:
       disable:
         - fieldalignment
         - shadow
       enable-all: true
+    forbidigo:
+      forbid:
+        - pattern: ^log\.(Fatal|Fatalf|Fatalln|Print|Printf|Println|Panic|Panicf|Panicln)$
+          msg: Use log/slog with the shared logging package
     importas:
       alias:
         - pkg: k8s.io/api/core/v1
@@ -88,12 +106,22 @@ linters:
         - pkg: k8s.io/apimachinery/pkg/util/errors
           alias: kerrors
       no-unaliased: true
+    loggercheck:
+      require-string-key: true
+      no-printf-like: true
     modernize:
       disable:
         - omitzero
         - embedlit
         - errorsastype
         - stringscut
+    sloglint:
+      no-mixed-args: true
+      no-global: all
+      context: scope
+      static-msg: true
+      msg-style: lowercased
+      key-naming-case: snake
     revive:
       rules:
         # The following rules are recommended https://github.com/mgechev/revive#recommended-configuration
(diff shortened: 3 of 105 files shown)


smeegoan:
Three gaps after this PR.

--postgres-schema defaults to public, and atepg.go assigns RuntimeParams["search_path"] unconditionally after parsing the DSN. In the shared-Secret setup this PR documents, a connection string carrying search_path=agents is therefore overridden to public on the ate-api-server side, silently, unless the flag is set to match.

Migrations cannot use a separate identity. RunUp runs in-process on the same connection string before the pool opens, so the role serving traffic must also hold DDL rights on its own schema, and every migrated object ends up owned by it. A schema owner distinct from its runtime consumer is not expressible.

Rotation has no input left. POSTGRES_DATABASE_URL is an env var, frozen for the life of the process. urlFile was the only input a rotation system could write to, and passfile is no substitute because ParseConfig resolves that once too. Is rotation meant to work another way?

iplay88keys:
@smeegoan, thanks for taking a look! Just updating this PR and reviewing your comment. Just a heads up that Codex helped put together this response, though I reviewed it.

>--postgres-schema defaults to public, and atepg.go assigns RuntimeParams["search_path"] unconditionally after parsing the DSN. In the shared-Secret setup this PR documents, a connection string carrying search_path=agents is therefore overridden to public on the ate-api-server side, silently, unless the flag is set to match.

This PR defaults `substrate.postgres.schema` to `substrate`, which becomes the explicit `--postgres-schema=substrate` argument. We intentionally do not rely on the DSN’s `search_path`; users can override the Substrate schema value.

>Migrations cannot use a separate identity. RunUp runs in-process on the same connection string before the pool opens, so the role serving traffic must also hold DDL rights on its own schema, and every migrated object ends up owned by it. A schema owner distinct from its runtime consumer is not expressible.

Kagent already supports this operationally. An operator can create the schema/grants and run `kagent db migrate up --db-url <migrator-url>` with privileged credentials, then configure the controller with restricted runtime credentials and `database.postgres.skipMigrations=true`. The chart does not provision those roles or grants. `RunUp` is skipped in that mode.

On the Substrate side, even if we added a way to pre-apply and skip startup migrations, that would not provide complete DDL/DML separation. `ateapi` also performs runtime DDL to create and remove outbox partitions, so its runtime role must retain those privileges.

Operators can configure Substrate with a separate connection Secret and role, even when it points to the same PostgreSQL server and database. This keeps Substrate’s DDL privileges off Kagent’s runtime role, but Substrate currently has no supported mechanism for separating its own migration/DDL and runtime/DML identities. Additional Substrate work would be required for that stricter separation.

>Rotation has no input left. POSTGRES_DATABASE_URL is an env var, frozen for the life of the process. urlFile was the only input a rotation system could write to, and passfile is no substitute because ParseConfig resolves that once too. Is rotation meant to work another way?

`urlFile` was [read only once during startup](https://github.com/kagent-dev/kagent/blob/5e18d9c48a79e47cbe20703c16d10de6935bcc6c/go/core/internal/database/connect.go#L114-L135), so changing it did not rotate an active pool. A Secret-backed environment variable has the same restart requirement. Live rotation would require explicit reload and pool reconnection support.

So the PR does not remove an existing Kagent credential-separation or live-rotation capability. The real follow-up is adding equivalent migration-role separation to Substrate if that is required for shared production databases.

iplay88keys:
@smeegoan, I see now how `urlFile` can be used to rotate credentials. In order to maintain that functionality, while still getting rid of the `urlFile` convention, I have updated this PR (as well as the required upstream Substrate PR) to now mount the secretRef as a volume and use the db url env to instead pass a path to that file.

As such, pgx's `beforeConnect` will re-read file-backed credentials and TLS material before each new physical connection. I also added the `maxConnLifetime` to substrate so the same mechanism for rotation can exist there.

So the updated design preserves a live Kubernetes Secret input and supports credential rotation without a pod restart, while replacing the arbitrary file-path `urlFile` value with explicit Secret references.

smeegoan:
Thank you @iplay88keys, the @file: plus BeforeConnect design closes the rotation gap.

One thing stops us adopting it as is. sameConnectionIdentity treats user as part of connection identity, so a rotation that changes the username fails every new acquire with database connection identity changed; restart required, while existing connections keep serving until maxConnLifetime expires. Our rotation changes the username by design: each cycle creates a new user with the new password and leaves the previous user able to log in until the following cycle, so that two credentials authenticate at once and no consumer fails while the secret store promotes the new version. That is a property of the database's rotation, not something a single consumer opts into. New dials should authenticate as the incoming user while older connections finish on the outgoing one, which is what the pool already does.

Proposal: move User out of the identity comparison and into the set the refresh adopts, next to Password. Host, port, database and fallback host and port stay fenced, since those are the pool's endpoint and changing them mid-life does need a restart. That removes the user comparison rather than adding anything.

iplay88keys:
@smeegoan, thanks for taking another look. I'm glad this approach overall will work for you. I have updated both sides of this (Substrate/Kagent) to allow for rotating the user to allow for the flow you laid out.

PR #2710, opened 2026-09-04
feat(core): add scoped authorization for catalog resources

codyhartsook:
## Summary

- Add policy-neutral collection scopes while keeping the default OSS installation open.
- Enforce scoped access for `AgentTemplate`, `Harness`, and `ModelConfig` using trusted Kubernetes metadata.
- Filter list responses server-side while keeping create, update, and delete authorization authoritative at execution time.
- Keep UI-only authorization hints out of the catalog protobuf API.

Backend half of #2683. The UI may display actions a caller cannot perform; denied requests return `PermissionDenied`.

## Breaking change

`app.Options.Authorizer` moves from `auth.Authorizer` to `auth.CollectionAuthorizer`. It is the documented authorization extension point, so an embedder supplying a `Check`-only authorizer must also provide collection scopes. The default OSS behavior remains unrestricted.

## Testing

```bash
make -C go lint
(cd go && GOCACHE=/tmp/kagent-scoped-auth-go-cache go test -short ./...)
(cd go && GOCACHE=/tmp/kagent-scoped-auth-go-cache go test ./core/pkg/auth ./core/pkg/app ./core/internal/httpserver/auth ./core/internal/service/model ./core/internal/service/prompttemplate ./core/internal/service/system ./core/internal/grpcserver ./core/internal/mcp)
(cd ui && ./node_modules/.bin/tsc --noEmit)
```

The full Go E2E suite also passed against the local Kind cluster.


go/core/pkg/auth/auth.go
```diff
@@ -82,6 +86,34 @@ type Authorizer interface {
...
+
+const (
+	ScopeIn = apiauthorization.ScopeIn
+)
+
+type AuthorizationScope = apiauthorization.AuthorizationScope
+
+type ScopeClause = apiauthorization.ScopeClause
+
+type ScopePredicate = apiauthorization.ScopePredicate
```
EItanya: Do we really need to mirror the types here if they're also public in the API?

go/core/pkg/auth/auth.go
```diff
@@ -82,6 +86,34 @@ type Authorizer interface {
...
+
+	ScopeAll   = apiauthorization.ScopeAll
+	ScopeNone  = apiauthorization.ScopeNone
+	ScopeAnyOf = apiauthorization.ScopeAnyOf
+)
+
+type ScopeOperator = apiauthorization.ScopeOperator
+
+const (
+	ScopeIn = apiauthorization.ScopeIn
```
EItanya: Same goes for all of these

proto/kagent/api/v1alpha1/agent_templates.proto
```diff
@@ -38,6 +38,9 @@ message AgentTemplate {
   // caller may legally pair with this template in CreateAgentInstance, and it
   // is derivable only from the Harness side, so a caller cannot compute it.
   repeated string admitting_harnesses = 5;
+
+  bool can_update = 6;
+  bool can_delete = 7;
```
EItanya: Let's try and remove these

go/core/internal/service/model/service.go
EItanya, line 1: Unclear why this doesn't use `kubecrud`

go/core/pkg/auth/auth.go
codyhartsook: we can use the public API types directly. I removed the aliases, and CollectionAuthorizer.Scope now returns authorization.AuthorizationScope

go/core/pkg/auth/auth.go
codyhartsook: removed the remaining mirrored scope types and constants, callers now reference go/api/authorization directly

proto/kagent/api/v1alpha1/agent_templates.proto
codyhartsook: Removed these fields. As a consequence, list scope still controls resource visibility, but the UI only learns whether create, update, or delete is allowed when the action is attempted. I think this ux change is okay though

go/core/internal/service/model/service.go
codyhartsook: Updated model.Service to use kubecrud for list, get, create, and delete. Update stays local because it also coordinates Secrets, cleanup, and conflict retries

go/core/internal/service/kubeauth/scope.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Resource builds authorization input from trusted Kubernetes metadata.
+func Resource(resourceType string, object metav1.Object) auth.Resource {
+	attributes := make(map[string]string, 2)
+	if object.GetNamespace() != "" {
+		attributes[apiauthorization.AttributeNamespace] = object.GetNamespace()
+	}
+	if object.GetName() != "" {
+		attributes[apiauthorization.AttributeName] = object.GetName()
+	}
```
EItanya: Why is the name in both the attributes and the name field?

go/core/internal/service/kubecrud/service.go
```diff
@@ -42,19 +44,36 @@ func NewService[T Object, L client.ObjectList](
...
 	if namespace == "" {
 		return nil, serviceerrors.NewInvalidArgument("namespace is required", nil)
 	}
+	return s.listWithOptions(ctx, client.InNamespace(namespace))
+}
+
+// ListAll lists resources across namespaces.
+func (s *Service[T, L]) ListAll(ctx context.Context) ([]T, error) {
+	return s.listWithOptions(ctx)
+}
```
EItanya: Typically this is done with namespace == "" rather than a new method

go/core/internal/service/kubecrud/service.go
```diff
@@ -100,21 +123,23 @@ func (s *Service[T, L]) Create(ctx context.Context, object T) (T, error) {
...
-	if err := s.authorize(ctx, auth.VerbUpdate, auth.Resource{Type: s.resource, Name: ref.String()}); err != nil {
+	object, err := s.get(ctx, ref)
+	if err != nil {
+		return zero, err
+	}
+	if err := s.authorize(ctx, auth.VerbUpdate, kubeauth.Resource(s.resource, object)); err != nil {
+		return zero, err
+	}
+	apply(object)
+	if err := s.authorize(ctx, auth.VerbUpdate, kubeauth.Resource(s.resource, object)); err != nil {
```
EItanya: twice?

go/core/internal/service/model/service.go
```diff
@@ -177,13 +160,21 @@ func (s *Service) Update(ctx context.Context, request UpdateRequest) (*v1alpha3.
...
 	oldRefs := referencedSecretNames(modelConfig.Spec)
 	spec := request.Spec
 	if request.APIKey != nil && *request.APIKey != "" && spec.APIKeySecret == "" && spec.Provider != v1alpha3.ModelProviderOllama {
 		spec.APIKeySecret = request.Ref.Name
 		spec.APIKeySecretKey = providerAPIKeySecretKey(spec.Provider)
 	}
+	proposed := modelConfig.DeepCopy()
+	proposed.Spec = spec
+	if err := s.authorize(ctx, auth.VerbUpdate, kubeauth.Resource(modelConfigResource, proposed)); err != nil {
+		return nil, err
```
EItanya: twice? Generally feels a bit off to me that you validate twice, also why doesn't this go into `kubecrud` since it's an update?

go/core/internal/service/kubeauth/scope.go
codyhartsook: removed the duplicate Resource.Name value, attributes only now

go/core/internal/service/kubecrud/service.go
codyhartsook: Updated to  List(ctx, "")

go/core/internal/service/kubecrud/service.go
codyhartsook: the second check was intended to validate the proposed identity, but updates preserve namespace and name, so it was redundant. Removed

go/core/internal/service/model/service.go
codyhartsook: Agreed on the duplicate authorization check, removed it. I kept the ModelConfig update in the model service because it also does owned Secret writes. From my understanding moving that into kubecrud would introduce ModelConfig specific side effects into the generic CRUD helper

go/core/internal/service/kubecrud/service.go
EItanya, line 117: For a follow-up, `GetForUpdate` is leaking DB transaction concepts into k8s, we should eliminate that

EItanya, line 129: https://github.com/kagent-dev/kagent/pull/2710/changes#r4025996541

EItanya (approved):

PR #2711, opened 2026-09-04
ci: support merge queues

Charlesthebird:
---
*🤖 written by Claude (start)*

## Release note

```release-note
NONE
```

## Changes

- CI checks now run on the `merge_group` event, so a merge queue can use them.
- The migration check and upgrade tests learn the base branch from the merge group, since `github.base_ref` is empty there.
- Merge queue runs share one Docker build cache instead of creating a new one each time.

---
*🤖 written by Claude (end)*


Charlesthebird:
## Follow-up: enable the merge queue

Needs repo admin.
Do this after merging.

1. **Settings → Rules → Rulesets**.
2. Open ruleset **default**.
3. Tick **Require merge queue**.
4. Set:

| Setting | Value |
|---|---|
| Merge method | Squash and merge |
| Build concurrency | 5 |
| Minimum group size | 1 |
| Maximum group size | 5 |
| Wait time | 5 minutes |
| Status check timeout | 60 minutes |
| All entries must pass | On |

5. Keep required status checks.
6. Turn off "up to date".
7. Save.
8. Queue one small PR.
9. Confirm it merges alone.

Release branches: merge #2712 and #2713 first.
Then repeat on ruleset **0.x**.
Drop its required **DCO** check.
It never reports on merge groups.

---
*🤖 written by Claude*

.github/workflows/ci.yaml
```diff
@@ -9,6 +9,7 @@ on:
     branches: [main, "release/**"]
     paths-ignore:
       - "**/*.md"
+  merge_group:
```
Charlesthebird, line 12: **Level:** 🟠 Medium · Not Blocking

This covers `main` only. The release branches carry their own `ci.yaml` without `merge_group`, and ruleset `0.x` requires the `DCO` app check, which never reports on a merge group. Backport this PR and reconcile that ruleset before queueing `release/**`.

---
*🤖 written by Claude*

.github/workflows/migration-immutability.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches: [main, "release/**"]
     paths:
       - "go/core/pkg/migrations/**"
+  merge_group:
```
Charlesthebird, line 8: **Level:** 🟡 Low · Not Blocking

`paths` filters are ignored on `merge_group`, so this and `sqlc-generate-check` run on every queued PR. Neither is a required check, so a red run no longer stops a merge once the queue owns the button. Add them to the ruleset if they should gate.

---
*🤖 written by Claude*

Charlesthebird (commented):

---

Nothing blocks merge. Every check the `main` ruleset requires runs on `merge_group`, and the two places that read the PR base branch resolve it inside a queue. The queue itself still has to be switched on; the follow-up comment has the steps.

- ✅ All 11 required contexts in ruleset `default` are job ids in `ci.yaml`, with no `name:` overrides, and the `build` and `python-test` matrices produce the parenthesised names in the list.
- ✅ `ci.yaml` leaves `merge_group:` without `paths-ignore`, so a docs-only group still reports the required checks instead of hanging the queue.
- ✅ The `concurrency` group falls through to `github.ref`, which is unique per group and carries `event_name`, so PR and queue runs never cancel each other.
- ✅ `origin/<base>` exists in a `fetch-depth: 0` checkout because `actions/checkout` fetches `+refs/heads/*:refs/remotes/origin/*`, and the merge base against it is the queued PR's parent on `main`.
- ✅ Both `${{ github.base_ref || github.event.merge_group.base_ref }}` fallbacks return the old value on `pull_request`, `push` and `workflow_dispatch`, since the `refs/heads/` strip is a no-op on a bare branch name.
- ✅ `scripts/prev-stable-version.sh` receives `CURRENT_REF`, so it no longer re-derives the base branch from the empty `GITHUB_BASE_REF`.
- ✅ The push triggers in `ci.yaml` and `image-scan.yaml` filter to `main` and `release/**`, so the queue's `gh-readonly-queue/` branch does not fire them a second time.
- ✅ The two `pull_request_target` labelers, `image-scan`, `tag` and `stalebot` never run on a merge group and are not required checks.
- ✅ A repo-wide grep for `pull_request`, `head_ref`, `base_ref`, `GITHUB_REF` and `github.event` finds no other reader of PR-only context.

</details>

<details><summary>Findings, all fixed on the branch</summary>

- `upgrade-test-setup` derived its base branch from `GITHUB_BASE_REF`, empty in a queue, so release-branch groups ran the wrong upgrade legs. Fixed in [`5cc9a8a`](https://github.com/kagent-dev/kagent/commit/5cc9a8aa).
- `prev-stable-version.sh` re-derived the same value on its own, which undid the action's fix on `release/**`. Fixed in [`fa99e2a`](https://github.com/kagent-dev/kagent/commit/fa99e2a9).
- `BRANCH_CACHE_KEY` used the group's unique ref name, so every group wrote twelve buildx cache scopes nothing reads again. Fixed in [`fa99e2a`](https://github.com/kagent-dev/kagent/commit/fa99e2a9).
- The `base_ref` fallback in `migration-immutability.yaml` had no comment explaining the `refs/heads/` strip. Fixed in [`fa99e2a`](https://github.com/kagent-dev/kagent/commit/fa99e2a9).

</details>

---
*🤖 written by Claude*

.github/workflows/ci.yaml
Charlesthebird: Tracked in the follow-up comment: backport before queueing release branches.


---
*🤖 written by Claude*

.github/workflows/migration-immutability.yaml
Charlesthebird: Repo settings, not this PR. Add them to the ruleset when enabling the queue.


---
*🤖 written by Claude*

EItanya (approved):

PR #2713, opened 2026-09-04
ci: run workflows on Blacksmith runners (backport to release/v0.9.x)

Charlesthebird:
---
*🤖 written by Claude (start)*

## Release note

```release-note
NONE
```

## Changes

Backport of #2659 and #2723: CI, image scan and tag builds run on Blacksmith runners with its Docker builder.

---
*🤖 written by Claude (end)*

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
       - setup
     env:
       VERSION: v0.0.1-test
-    runs-on: ubuntu-latest
+    runs-on: blacksmith-4vcpu-ubuntu-2404
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


Charlesthebird (commented):

---

Nothing blocks merge. The two commits are cherry-picks of #2659 and #2723, with conflicts resolved by keeping the release-branch side and changing only the runner label or the Buildx step.

<details><summary>Findings, none blocking</summary>

- `BUILDX_VERSION` in `ci.yaml` and `image-scan.yaml` has no consumer once the Buildx step is gone, but main carries the same value so it stays.
- The e2e and `build` jobs keep a QEMU step before the Blacksmith builder, which is what main does in its `build` job.

</details>

- ✅ Diff scope: only `runs-on` lines and Buildx→Blacksmith step swaps across the three workflow files, with `git diff --check` clean.
- ✅ Every job's runner label equals main's for the same-named job, and jobs main leaves on `ubuntu-latest` stay there.
- ✅ `useblacksmith/setup-docker-builder@v2` appears in all three places and no `@v1` remains.
- ✅ `BUILDX_BUILDER_NAME` is still defined at the top level and consumed by every `make build` step, as on main.
- ✅ `tag.yaml` keeps its Buildx step with only the runner changed, matching main.
- ✅ Builds still reach the `registry:2` service because `make build-*` runs `buildx-create`, which recreates the `network=host` builder from a Makefile target identical to main's.
- ✅ `git log 50c856b6..origin/main -- .github` shows no Blacksmith follow-up other than #2723, which is included.
- ✅ All three files parse as YAML.

</details>

---
*🤖 written by Claude*

Issue #2688, opened 2026-09-03
Add MCP approval and forwarded-header policy to v2 AgentTemplate bindings


EItanya:
## Problem

The v2 `MCPToolBinding` currently carries only the server reference and selected tool names. It cannot populate the ADK MCP configuration fields `RequireApproval` or `AllowedHeaders`.

Consequences:

- users cannot require human approval for selected MCP tools;
- users cannot explicitly allow selected inbound A2A headers to reach MCP calls;
- `HarnessCapabilities.Approvals` can advertise approval support even though an AgentTemplate cannot configure MCP approval gating.

The legacy `McpServerTool` compatibility wrapper previously exposed these fields, but the v2 compiler never populated them. Removing that wrapper merely made the existing gap visible.

## Scope

- Add typed v2 `AgentTemplate` MCP-binding fields for selected tools requiring approval and explicitly allowed forwarded headers.
- Validate that every approval-required tool is included in the binding tool selection when a selection is present.
- Compile both fields into `adk.HttpMcpServerConfig` and `adk.SseMcpServerConfig`.
- Preserve the existing STS rule: generated authorization credentials take precedence over forwarded `Authorization` headers.
- Add CRD validation, compiler unit coverage, and an end-to-end approval continuation test.
- Ensure harness capability reporting matches the behavior actually available through the public API.

This should use the public A2A HITL continuation semantics; do not introduce a private MCP-only continuation path.

PR #2714, opened 2026-09-04
feat(harness): human-in-the-loop and pause actor on input-required

supreme-gg-gg:
This PR adds end-to-end HITL support for the ADK, Codex, and Claude Code harnesses running on Substrate. When a task enters `input-required`, Kagent pauses the Actor so the worker can be released while preserving the live runtime state needed to continue the same turn.

## Key changes

- Adds binding-scoped MCP tool approval through `AgentTemplate.spec.tools[].mcp.requireApproval`.
- Defines shared, typed A2A HITL payloads and validates approval and ask-user responses before resuming a task.
- Adds a runtime-neutral pending-turn contract for parking, resuming, and canceling live harness processes.
- Pauses Substrate Actors at `input-required` and continues to suspend them at terminal task boundaries.
- Adds Codex MCP approval and native ask-user handling through the App Server protocol.
- Adds Claude MCP approval through an authenticated, Actor-local `permission-prompt-tool`.
- Updates the Go and Python ADK runtimes to apply approval policy to the originating MCP binding rather than matching tool names globally.
- Adds UI controls and durable transcript records for tool approvals, rejection reasons, and ask-user answers.
- E2E test coverage for approval for ADK, Codex, and Claude harness

## Pausing

Input waits use Substrate `Pause` rather than `Suspend`. Pause preserves a `FULL` snapshot, including process memory, which is required for harnesses such as Codex and Claude that retain a live process while awaiting input. The paused snapshot is kept on node-local storage instead of object storage, reducing network overhead while the worker is released.

Terminal task boundaries still use Suspend with `DATA` snapshots.

## `AgentTemplate` API for tool approval

Approval is configured on an MCP binding:

```yaml
spec:
  tools:
    - mcp:
        server:
          kind: RemoteMCPServer
          name: kagent-tool-server
        tools:
          - tool_name
        requireApproval: true
```

`requireApproval` applies to every tool selected by that binding, or every tool on the server when `tools` is omitted. The API represents mixed protected and unprotected tool sets from one server as separate bindings, keeping selection and approval policy together. Partially completes #2688 

ADK supports that split today. Codex and Claude currently reject duplicate bindings to the same `RemoteMCPServer`; Claude also exposes the whole server when partial tool selection cannot be enforced and reports a warning.

## HITL coverage

The following flows are supported and tested:

- MCP tool approval and rejection, including optional rejection reasons where supported by the harness.
- Sequential and parallel MCP tool approval requests.
- Ask-user questions for ADK and Codex, including multiple questions and choices.
- Cancellation while a task is waiting for input.
- Persisted HITL requests and responses in task history and the chat transcript.

Claude ask-user is not supported because the upstream Claude Code tool was removed. I will probably add this as a follow-up.

## Permission mode

Configurable native permission modes for Codex and Claude are out of scope. Both harnesses continue to run with unrestricted native sandbox permissions because Substrate is the sandbox boundary. Built-in tools such as file access, shell, web search, and other MCP tool calls not marked with `requireApproval` will generally be auto-approved.


supreme-gg-gg:
Some demo with Claude (first) and Codex (second), ADK behaviour is similar:

https://github.com/user-attachments/assets/43cbf87a-f2a0-4b58-b96d-97c8d98104bd

https://github.com/user-attachments/assets/b2fa1cf6-083b-4033-8671-246b1da341f0

Charlesthebird (commented):
Thanks for adding this and for the video 🎉 !

The UI looks good, but I'll add some UI feedback here in case you want to polish it some more. These could be moved to a follow-up if we want.


1. The request to run a tool block becomes the tool call approved block. These both look good individually, but because they have different heights, it results in a content layout shift when the blocks are replaced (the chat content above it jumps around). This layout shift might be reduced with an animation, or by solving the next point (e.g. keeping the tool call JSON in the tool call approved block, or otherwise changing this).

Image

Image

2. The actual tool call JSON is also duplicated between the above tool call block and the request to run tool block. It is then not present in the tool call approved block

3. In the first video, there is an "API Error" response right after the tool call approved block. I guess this was an issue with the tool or k8s connection? But then it has the info in the summary? Maybe this was just a tool error.

4. The JSON text in the tool request block is a little low contrast. Maybe the code displayer there should have a slightly lighter background? Or that could be restyled a bit.

5. On the tool approved block, there's an icon at the top next to "Tool access approved", and then a second icon below it for the tool that was approved that looks very similar - it would be nice if they were more visually different. Maybe a bare checkmark, or a filled in circle with a checkmark would be better for the inner icon instead of the outlined circle with a checkmark.

supreme-gg-gg:
Thanks for the comments on the UI:

Regarding (2), the tool call approved block actually has the JSON arguments, but you need to click on "details" to view them

Screenshot 2026-09-08 at 2 10 15 PM

Regarding (3), this message is specific to Claude and it doesn't mean anything is actually wrong with the tool call, but instead the harness (Claude Code itself) thought that something is wrong due to it being suspended and resumed. I can probably add some way to filter out these warnings.

In general, I feel like we can switch back to the old UI style instead, which shows a single tool call approval request block, and once it's approved it switches to a tool call completed / approved block? What do you think? 

Screenshot 2026-09-08 at 2 09 35 PM

Charlesthebird:
> In general, I feel like we can switch back to the old UI style instead, which shows a single tool call approval request block, and once it's approved it switches to a tool call completed / approved block? What do you think?

I like some of the new styles,  (overall the yellow background is nice, and the icons are too) but switching back to that style of having it in one block sounds good to me!

supreme-gg-gg:
@Charlesthebird I've updated the styling of HITL components in the UI regarding your earlier comments

https://github.com/user-attachments/assets/c1a28bee-e436-45a7-a84d-8a4f7b49ea6c

I've also filtered out the Claude connection warnings in the harness adapter to avoid confusing UX (the "API Error" message)

Screenshot 2026-09-09 at 1 22 07 PM

go/harness/codex/internal/driver/events.go
```diff
@@ -160,6 +166,19 @@ func (t *eventTranslator) closeActiveTools(sink runtime.EventSink) error {
 	return nil
 }
 
+func (t *eventTranslator) approvalTool(server string) (string, string, error) {
+	queue := t.approvals[server]
+	for len(queue) != 0 {
+		id := queue[0]
```
EItanya: 1. **Codex can display the wrong tool for approval.** This correlates approvals by per-server FIFO, but parallel calls can request approval in a different order from their `item/started` events. A synthetic protocol probe with read and delete calls starting in that order, followed by the delete approval, labeled the request as `protected.read` with the read call ID. The response still goes to the delete request's RPC ID, so approving the displayed read authorizes the delete. Please use explicit correlation and reject ambiguous matches instead of selecting the first active call. The probe does not launch a real Codex process; the pinned 0.148.0 source emits item-started before asynchronous approval processing.

EItanya (commented):
🤖 AI-generated review.

Three findings from reviewing 2b76a493 are attached inline. Focused Go tests, database/migration tests, and 83 UI tests passed. Additional probes reproduced the issues described in the comments; the Codex correlation probe uses synthetic protocol events, and the cancellation probe forces the overlapping schedule. Live Substrate E2E and Python tests were not run.

go/core/internal/a2agateway/gateway.go
```diff
@@ -561,6 +561,24 @@ func (g *Gateway) prepareReply(ctx context.Context, instance *apiv1alpha1.AgentI
 	if stored.Status.State != a2atype.TaskStateInputRequired && stored.Status.State != a2atype.TaskStateAuthRequired {
 		return nil, a2atype.NewError(a2atype.ErrUnsupportedOperation, "task is not waiting for input")
 	}
+	if pending, parseErr := apia2a.ParseToolApprovalRequest(stored.Status.Message); parseErr != nil {
+		return nil, a2atype.NewError(a2atype.ErrInternalError, "stored tool approval request is invalid")
+	} else if pending != nil {
+		response, responseErr := apia2a.ParseToolApprovalResponse(message)
+		if responseErr != nil || apia2a.ValidateToolApprovalResponse(pending, response) != nil {
```
EItanya, line 568: 2. **Nested tool approvals cannot resume.** `ValidateToolApprovalResponse` checks `pending.Tools`, which contains the parent continuation ID for a nested request. The UI and ADK's `processNestedApproval` instead use the child IDs from `pending.Nested.Tools`. Reproduced with the same response accepted by `BuildResumeHITLMessage` and rejected by `prepareReply` with `tool approval response does not match the pending request`. Please validate against the child IDs when the request is nested.

go/harness/runtime/a2a/executor.go
```diff
@@ -82,29 +112,55 @@ func (e *Executor) Execute(ctx context.Context, reqCtx *a2asrv.ExecutorContext)
...
 				close(active.done)
 			})
+			return finishedCanceled
+		}
+		park := func(pending runtime.PendingTurn) bool {
+			parked := false
+			finishOnce.Do(func() {
+				cancel()
+				parked = e.park(active, pending)
+				close(active.done)
```
EItanya, line 148: 3. **Cancellation completes before process cleanup.** When cancellation wins the parking race, `e.park` clears the execution slot and returns false, but this closes `active.done` before the subsequent `outcome.Pending.Cancel(...)` runs. `Executor.Cancel` can therefore report completion while the live process is still being cleaned up, and the slot is already available for another execution. A channel-controlled probe blocked `PendingTurn.Cancel` and observed `Executor.Cancel` return early in 20/20 runs; this is a forced schedule, not a natural flake-rate estimate. Please finish pending-process cleanup before releasing the slot and signaling completion.

Charlesthebird (approved):
Thanks for the UI updates! The reduced chat content layout shift looks good! 🎉

supreme-gg-gg:
Review comments are addressed in [9911c72](https://github.com/kagent-dev/kagent/pull/2714/commits/9911c72a8101cd25544749b032556b3f3bdab103)

go/core/internal/a2agateway/gateway.go
```diff
@@ -52,6 +52,7 @@ type runtimeDialer interface {
 }
 
 type instanceWorkflow interface {
+	PauseForInput(context.Context, *apiv1alpha1.AgentInstance, func(context.Context) error) error
```
EItanya: Why does this need a callback, these agents love callbacks

EItanya (approved):

PR #2486, opened 2026-08-18
fix(adk): validate ask_user questions

erauner12:
## Summary

Validate built-in `ask_user` inputs in both Go and Python ADK runtimes on `release/v0.10.x` before confirmation handling.

Fixes #2468.

## Problem

Both implementations accepted an empty `questions` list and entries whose question text was empty or whitespace-only. Those invalid calls could pause a task without presenting a usable question.

An invalid model tool call should return an error the model can correct; it should not create an unanswerable `input-required` interaction.

## Change

Both runtimes now:

- require at least one question;
- require nonblank question text for every entry;
- reject invalid input before requesting confirmation;
- tell the model to provide at least one question; and
- preserve existing pending and confirmed behavior for valid questions.

Runtime validation remains authoritative and does not depend on a provider enforcing a generated schema.

## Tests

Go:

- `go test ./adk/pkg/tools`
- `go test -race ./adk/pkg/tools`

Python:

- `uv run pytest packages/kagent-adk/tests/unittests/test_ask_user_tool.py packages/kagent-adk/tests/unittests/test_hitl.py -q`
- Ruff checks on the touched Python files

Coverage verifies empty-list rejection, blank-text rejection, no confirmation call for invalid input, and unchanged valid pending/confirmed behavior in both runtimes.

## Validation

An independent A2A client against a compatible v0.10 deployment completed a valid structured-question round trip: one nonblank question was returned, a later process answered the retained interaction, and the same task reached terminal completion without resending the original request.

This PR changes only built-in `ask_user` validation; it does not change identity propagation, persistence, or A2A transport formats.

Issue #2468, opened 2026-08-18
[BUG] Built-in ask_user accepts empty or blank questions


erauner12:
### 📋 Prerequisites

- [x] I have searched the [existing issues](https://github.com/kagent-dev/kagent/issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [ ] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Multiple services / System-wide issue

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

The built-in `ask_user` tool in both the Go and Python ADK runtimes accepts an empty `questions` list. It also accepts question entries whose `question` text is empty or contains only whitespace. These invalid inputs can request confirmation without presenting a usable question.

### 🔄 Steps To Reproduce

1. Invoke the built-in `ask_user` tool in either runtime with `{"questions": []}`.
2. Observe that a confirmation request is created without a usable question.
3. Invoke it with a single question whose text is empty, for example `{"questions": [{"question": ""}]}`.
4. Repeat with whitespace-only question text and observe the same behavior.

### 🤔 Expected Behavior

The tool should reject an empty question list and any question with empty or whitespace-only text before requesting confirmation. Valid single-question behavior should remain unchanged: the initial invocation should request confirmation, and a confirmed invocation should return the answer normally.

### 📱 Actual Behavior

Both runtimes can request confirmation for an empty question list or blank question text, producing an unusable interaction.

### 💻 Environment

- Current kagent release
- Go and Python ADK runtimes

### 🔧 CLI Bug Report

_Not applicable_

### 🔍 Additional Context

Validation should be consistent in both runtimes and occur before the confirmation API is called.

### 📋 Logs

```shell

```

### 📷 Screenshots

_Not applicable_

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR to fix this issue

erauner12:
A proposed fix covering both runtimes is available in https://github.com/erauner12/kagent/pull/14.

erauner12:
Draft fix: #2486 targets `release/v0.10.x` and adds provider-independent validation for empty or blank `ask_user` questions in both Go and Python runtimes.

[...]

Issue #1683, opened 2026-04-16
[BUG] Function tool schemas not sent to LLM via OpenAI/Anthropic model adapters


renato0307:
## 📋 Prerequisites

- [x] I have searched the existing issues to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our Code of Conduct
- [x] I am using the latest version of the software
- [x] I can consistently reproduce this issue

## 🎯 Affected Service(s)

App Service

## 🚦 Impact/Severity

Blocker

## 🐛 Bug Description

When tools are registered using `functiontool.New` with typed Go structs (e.g., the remote A2A tool, `ask_user`, `save_memory`), the Google ADK sets `FunctionDeclaration.ParametersJsonSchema` to a `*jsonschema.Schema` value (from `github.com/google/jsonschema-go`). Both model adapters attempt to read this with a `map[string]any` type assertion that silently fails:

```go
// openai_adk.go:342, anthropic_adk.go:239
if m, ok := fd.ParametersJsonSchema.(map[string]any); ok {
```

The actual type is `*jsonschema.Schema`, not `map[string]any`. The assertion returns `ok=false`, so the tool schema sent to the LLM has empty properties and no required fields. The LLM calls every function tool with `{}`, which then fails ADK input validation.

Tools registered with `genai.Schema` directly (via `fd.Parameters`) are unaffected — they use a separate code path that already works.

## 🔄 Steps To Reproduce

1. Create a declarative agent with a `type: Agent` tool referencing another agent
2. Send a message to the orchestrator agent
3. The LLM calls the A2A tool with empty args `{}`
4. ADK validation fails: `validating root: required: missing properties: ["request"]`

## 🤔 Expected Behavior

The tool schema (including `properties` and `required` fields) should be sent to the LLM so it can provide the correct arguments.

## 📱 Actual Behavior

All function tools created via `functiontool.New` with typed structs receive empty schemas. Affects `remote_a2a_tool` (`request`), `ask_user` (`questions`), `save_memory` (`content`), and any custom typed tools.

## 💻 Environment

- kagent v0.9.0-beta4
- Go ADK with `google.golang.org/adk` v1.0.0
- Bedrock via OpenAI-compatible adapter
- EKS 1.32

## 🔍 Additional Context

The fix is to JSON-roundtrip the `ParametersJsonSchema` value when it's not already `map[string]any`:

```go
func parametersJsonSchemaToMap(v any) map[string]any {
    if v == nil {
[...]

PR #2727, opened 2026-09-07
fix(adk): return a retryable error from ask_user instead of raising

pboers1988:
## What

`ask_user` aborts the entire A2A request when the model calls it with an unusable `questions` list. The rejection is right; killing the request is not.

ADK re-raises tool exceptions (`google/adk/flows/llm_flows/functions.py:546`, when no `on_tool_error` callback returns a response), so the `ValueError` propagates out through `handle_function_call_list_async` and `_agent_executor` publishes the task as **failed**. The caller receives a well-formed `200` with no content: the user sees an empty reply, and nothing downstream logs an error.

Observed in production on `kagent/app:0.10.0` from an ordinary underspecified prompt ("show me active alarms"). The agent decided to ask a clarifying question, called `ask_user` with `questions: []`, and the turn died silently:

```
kagent_adk.kagent.adk._agent_executor - ERROR - Error handling A2A request: ask_user: at least one question is required
  File ".../google/adk/flows/llm_flows/functions.py", line 1152, in __call_tool_async
    return await tool.run_async(args=args, tool_context=tool_context)
  File ".../kagent/adk/tools/ask_user_tool.py", line 92, in run_async
    raise ValueError("ask_user: at least one question is required")
ValueError: ask_user: at least one question is required
```

That call was **schema-legal**. `_get_declaration` declared `questions` as an unconstrained array, so only the prose description forbade an empty list — `required: ["questions"]` merely forces the key to exist. Confirmed by running the OpenAI converter against the live tool; the model receives no `minItems`.

There is a second, undocumented crash mode on the same path: the per-question loop called `.get()` on every entry, so `{"questions": ["some text"]}` raised `AttributeError` and killed the request identically.

## How

- **Return instead of raise.** `run_async` now returns `{"error": ...}` — the same shape `McpTool.run_async` uses — so the model can correct itself and ask a real question. **The rejection criteria are unchanged**; only the delivery differs. `AskUserTool.is_long_running` is `False`, so an early plain-dict return is an ordinary function response.
- **Fix the `AttributeError`.** `_question_text` pattern-matches the entry shape rather than assuming a dict, so malformed entries are rejected with the existing per-question message instead of crashing.
- **Constrain the declaration.** `min_items=1` so the empty call is not expressible in the first place.

The confirmation hint still passes the model's text through **verbatim** — `test_valid_question_requests_confirmation_without_rewriting_text` covers this, and an earlier draft that trimmed via the helper broke it.

## Tests

`tests/unittests/test_ask_user_tool.py` — the existing parametrized rejection test now asserts the returned error rather than `pytest.raises`, and gains the shapes that previously crashed differently:

| Args | Before | After |
|---|---|---|
| `{"questions": []}` | `ValueError`, request dies | `{"error": ...}` |
| `{}` | `ValueError`, request dies | `{"error": ...}` |
| `{"questions": ["text"]}` | **`AttributeError`**, request dies | `{"error": ...}` |
| `{"questions": None}` / `"not-a-list"` | `AttributeError`/`ValueError` | `{"error": ...}` |
| `{"questions": [{"question": "Which subscription?"}]}` | pending | pending (unchanged) |

```
$ uv run pytest tests/unittests/test_ask_user_tool.py tests/unittests/test_hitl.py -q
36 passed

$ uv run pytest tests/unittests/ -q
418 passed, 9 failed
```

The 9 failures are all `tests/unittests/models/test_tls_e2e.py` (`FileNotFoundError` on cert fixtures) and reproduce identically on a clean checkout of this branch's base — unrelated to this change.

## Notes for reviewers

**Branch targeting.** This targets `release/v0.10.x` because that is where the validation lives. #2486 merged into this branch only — `git log -S "at least one question is required"` on `main` has no hits, and `main` still carries the tolerant original. So the branches have diverged into opposite bugs: `v0.10.x` crashes the request, `main` still pauses the UI with zero questions (#2468). A separate PR is needed for `main`; the patch there is different.

**`min_items` does not currently reach OpenAI/Ollama models.** `kagent/adk/models/_openai.py:301` and `_ollama.py:113` serialize schemas with `prop_schema.model_dump(exclude_none=True)` and no `by_alias=True`, so it emits `"min_items"` rather than the JSON Schema keyword `"minItems"` and is silently ignored. `_bedrock.py:230` already passes `by_alias=True`. Fixing that converter also corrects other aliased keys (`any_of` -> `anyOf`) across every tool schema on those adapters, and looks adjacent to #1683 — so it deserves its own PR and regression run rather than being folded in here. The graceful return in this PR is what actually stops the outage; `min_items` is belt-and-braces for adapters that honour it.

**The Go runtime** (`go/adk/pkg/tools/ask_user.go`) has the same validation but returns `(nil, error)` rather than raising. I have not verified whether its runner degrades gracefully.

Refs #2468, #2486

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/tools/ask_user_tool.py
+++ b/python/packages/kagent-adk/src/kagent/adk/tools/ask_user_tool.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import json
 import logging
-from typing import Any
+from typing import Any, Final
 
 from google.adk.tools.base_tool import BaseTool
 from google.adk.tools.tool_context import ToolContext
@@ -39,6 +39,37 @@ _QUESTION_SCHEMA = types.Schema(
     },
     required=["question"],
 )
+
+
+_RETRY_HINT = (
+    ". Call ask_user again with a 'questions' array holding at least one object whose "
+    "'question' field is non-empty text, or answer the user directly without calling ask_user."
+)
+
+
+def _question_text(question: Any) -> str:
+    """The question's trimmed text, or "" when the entry cannot supply one."""
+    match question:
+        case {"question": str() as text}:
+            return text.strip()
+        case _:
+            return ""
+
+
+def _validation_error(questions: Any) -> str | None:
+    """Why `questions` is unusable, or None when every entry carries real text."""
+    match questions:
+        case list() if questions:
+            return next(
+                (
+                    f"ask_user: question {index} must contain non-whitespace text"
+                    for index, question in enumerate(questions, start=1)
+                    if not _question_text(question)
+                ),
+                None,
+            )
+        case _:
+            return "ask_user: at least one question is required"
 
 
 class AskUserTool(BaseTool):
@@ -73,7 +104,8 @@ class AskUserTool(BaseTool):
                     "questions": types.Schema(
                         type=types.Type.ARRAY,
                         items=_QUESTION_SCHEMA,
-                        description="List of questions to ask the user.",
+                        min_items=1,
+                        description="List of questions to ask the user. Must contain at least one question.",
                     ),
                 },
                 required=["questions"],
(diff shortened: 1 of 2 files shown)


python/packages/kagent-adk/src/kagent/adk/tools/ask_user_tool.py
```diff
@@ -86,18 +118,19 @@ async def run_async(
...
-        for index, question in enumerate(questions, start=1):
-            question_text = question.get("question")
-            if not isinstance(question_text, str) or not question_text.strip():
-                raise ValueError(f"ask_user: question {index} must contain non-whitespace text")
+        # Raising here aborts the whole request: ADK re-raises tool exceptions, so the caller
+        # gets a failed task and the user is shown nothing at all. The call is still rejected,
+        # but as a response the model can act on and retry.
+        if error is not None:
+            logger.warning("%s (received %r)", error, questions)
+            return {"error": error + _RETRY_HINT}
```
EItanya, line 129: The retry instructions never reach OpenAI or Ollama. This returns `{"error": ...}`, but those adapters only extract `content` or `result`. Calling with `questions: []` therefore sends an empty tool response to the model. The exception is avoided, but the model cannot see why validation failed or how to retry.

I reproduced this using the PR's locked dependencies: passing the actual tool result through ADK's `_build_function_response_content` and the OpenAI Chat Completions, OpenAI Responses, and Ollama converters produces `""` in all three cases.

The smallest fix is returning the error as a string, which ADK wraps in `{"result": ...}`. Please add an adapter-level regression check so the test verifies that the retry instructions reach the model.

EItanya (commented):
🤖 AI-generated review.

Found one issue: the retry instructions never reach OpenAI or Ollama; details inline.

Validation on `8a368f6b`: 36 focused ask_user/HITL tests passed, along with Ruff lint, formatting, and diff checks. A separate probe through ADK's response builder and the local adapters reproduced the empty tool response. No live-model testing.

PR #2732, opened 2026-09-07
fix(ui): keep chat sidebars below header

helloxjade:
## Summary

- keep desktop chat sidebars within the chat layout instead of positioning them against the viewport
- prevent both chat sidebars from covering the global header on the 0.9.x UI
- add a regression test for the provider and desktop panel positioning contract

## Root cause

The desktop sidebar panel used `fixed inset-y-0 h-svh`, so it started at the top of the viewport while the chat layout itself starts below the global header. Its opaque background and higher stacking order covered the left side of the header, clipping the kagent wordmark.

The provider now establishes the positioning context and the panel uses `absolute inset-y-0 h-full`, keeping it within the available chat area without hard-coding the header height.

## Testing

- `npm test -- --runInBand` (25 suites, 276 tests)
- `npm run lint` (0 errors; 3 pre-existing warnings)
- `npm run build`
- `npm run test:vitest` (33 files, 134 tests)

### DIFF
--- a/ui/src/components/ui/sidebar.tsx
+++ b/ui/src/components/ui/sidebar.tsx
@@ -147,7 +147,7 @@ const SidebarProvider = React.forwardRef<
               } as React.CSSProperties
             }
             className={cn(
-              "group/sidebar-wrapper relative flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
+              "group/sidebar-wrapper relative flex min-h-svh w-full overflow-hidden has-[[data-variant=inset]]:bg-sidebar",
               className
             )}
             ref={ref}
--- a/ui/src/components/ui/sidebar.tsx
+++ b/ui/src/components/ui/sidebar.tsx
@@ -147,7 +147,7 @@ const SidebarProvider = React.forwardRef<
               } as React.CSSProperties
             }
             className={cn(
-              "group/sidebar-wrapper flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
+              "group/sidebar-wrapper relative flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
               className
             )}
             ref={ref}
@@ -244,7 +244,7 @@ const Sidebar = React.forwardRef<
         />
         <div
           className={cn(
-            "fixed inset-y-0 z-10 hidden h-svh w-[--sidebar-width] transition-[left,right,width] duration-200 ease-linear md:flex",
+            "absolute inset-y-0 z-10 hidden h-full w-[--sidebar-width] transition-[left,right,width] duration-200 ease-linear md:flex",
             side === "left"
               ? "left-0 group-data-[collapsible=offcanvas]:left-[calc(var(--sidebar-width)*-1)]"
               : "right-0 group-data-[collapsible=offcanvas]:right-[calc(var(--sidebar-width)*-1)]",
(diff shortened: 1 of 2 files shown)


Charlesthebird:
Hello! Could you please post a before/after screenshot or video of this?

helloxjade:
> Hello! Could you please post a before/after screenshot or video of this?

Thanks! Here are the before/after screenshots.

**Before — deployed kagent UI 0.9.12:**
kagent-deployed-before


**After — this PR branch, connected to the same 0.9.12 backend:**
kagent-live-after-converted


The sidebar now starts below the global header, so the full kagent logo and navigation remain visible. No cluster upgrade was required for this verification; I ran the PR branch UI locally against the same backend.

Charlesthebird:
Frames for the change request on `ui/src/components/ui/sidebar.tsx`, in order: both sidebars collapsed, then the chat scroll container pushed fully right.

collapsed-pan-before
collapsed-pan-after

---
*🤖 written by Claude*

ui/src/components/ui/sidebar.tsx
```diff
@@ -147,7 +147,7 @@ const SidebarProvider = React.forwardRef<
               } as React.CSSProperties
             }
             className={cn(
-              "group/sidebar-wrapper flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
+              "group/sidebar-wrapper relative flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
```
Charlesthebird: **Level:** 🟠 Medium · Blocking

`fixed` never contributed to an ancestor's scrollable overflow; `absolute` does, so the swap on line 247 puts the off-canvas panel inside the root `<main className="flex-1 overflow-y-scroll">` (`src/app/layout.tsx:34`), whose `overflow-x` computes to `auto` because the other axis is `scroll`.

Collapse the sidebars — ⌘B or the `SidebarRail`, which share one open state with the right-hand panel — and `right-[calc(var(--sidebar-width)*-1)]` becomes 350px of reachable overflow, so the page pans sideways and the hidden Agent Details panel comes into view.

```suggestion
              "group/sidebar-wrapper relative overflow-hidden flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
```

`overflow-x-hidden` clears it too, but leans on the same `visible`→`auto` promotion that caused this; `overflow-hidden` adds no vertical scroll of its own here, measured at 0.

Both sidebars collapsed, then the scroll container pushed fully right — before, `scrollLeft` reaches 0px:

collapsed-pan-before

After, it reaches 350px:

collapsed-pan-after

---
*🤖 written by Claude*

Charlesthebird (changes requested):

---

The header fix is right, but trading `fixed` for `absolute` moves the collapsed off-canvas panel into the page's scroll container, so the chat page can be panned 350px sideways — one class on the provider closes it, inline.

✅ The sidebars stay pinned while a long chat scrolls: `overflow-x-hidden` on `ChatLayoutUI.tsx:111` promotes that column's `overflow-y` to `auto`, so the root scroller reports `scrollHeight === clientHeight` and the panels hold at `top: 56` across 2360px of messages.

✅ `h-full` resolves against a definite height, since the provider is `h-full` inside `flex-1` of an `h-screen` body.

---
*🤖 written by Claude*

Charlesthebird (approved):
Thanks for the fixes!

peterj (approved):

Issue #959, opened 2025-09-27
[FEATURE] vllm modelconfig


poblin-orange:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

add support for vllm modelconfig

### ❓ Problem Statement / Motivation

unable to use kagent with corporate vllm instance.
Would need a specific modelconfig, enabling setting host + api key for vllm

### 💡 Proposed Solution

A new modelconfig type

### 🎯 Affected Service(s)

None

### 🙋 Are you willing to contribute?

- [ ] I am willing to submit a PR for this feature

peterj:
Could you configure it like this: https://kagent.dev/docs/kagent/supported-providers/byo-openai ?

poblin-orange:
@peterj thks for the link ! I was able to create a ModelConfig with our corp vllm.

However, cant define it in the HelmChart values (the baseUrl attribute is ignored by the chart with openAI provider)

PR #2739, opened 2026-09-08
fix(adk): allow a keyless OpenAI-compatible baseUrl

QuentinBisson:
## Problem

A `ModelConfig` with `provider: OpenAI`, a custom `openAI.baseUrl`, and no `apiKeySecret` or `apiKeyPassthrough` reconciles fine, but every Go ADK agent pod that references it crashloops at boot:

```
Failed to create Google ADK Runner config: failed to create agent: failed to create LLM: OPENAI_API_KEY environment variable is not set
```

`NewOpenAIModel` required `OPENAI_API_KEY` whatever the endpoint was. OpenAI-compatible endpoints (vLLM, LiteLLM, corporate proxies) are commonly keyless, which is why `apiKeySecret` is optional in the CRD. Related: #959.

## Change

The key requirement belongs to the OpenAI provider rather than to the client builder, so `resolveOpenAIAPIKey` lives in the `*adk.OpenAI` case of `agent.CreateLLM` and passes its result through the new `OpenAIConfig.APIKey`. It still requires `OPENAI_API_KEY` when no `baseUrl` is set, so the failure for api.openai.com is unchanged; with a `baseUrl` and no key source it logs a warning and returns an empty key. `option.WithAPIKey("")` overrides the SDK's own env default and makes the client send no `Authorization` header, so no placeholder credential is invented. `pkg/models` now only builds the client.

`NewOpenAICompatibleModel` is gone. It had no callers, and once the `"ollama"` placeholder was removed it did nothing that `NewOpenAIModel` does not. `apiKeyPassthrough` is unchanged.

A typo in `apiKeySecret` does not reach this path: the controller already fails reconciliation when a referenced secret is missing.

Scope: Go ADK only. The Python runtime (`OpenAINative` in kagent-adk) passes `api_key=None` to the OpenAI SDK and fails the same way; that needs a separate change. The Go ADK does not consume `openAI.tokenExchange`, so such a config with a `baseUrl` and no secret now starts unauthenticated instead of failing at boot.

The failure still surfaces only in pod logs. Surfacing it as an `Accepted=False` condition on the `ModelConfig`, next to the existing API-key secret check in `ResolveModelConfig`, is worth a follow-up issue.


EItanya (commented):
The spirit of this change makes sense, but it feels like maybe we should pull the API validation logic out of the client config, and have that part be `openai` validation, rather than attaching it to the client in all cases and using a fake key.

Would that be possible?

QuentinBisson:
Done in `edb42485` — the fake key is gone on the keyless path.

`resolveOpenAIAPIKey` now returns an empty key when a custom `baseUrl` is set and no key source is configured, and `newOpenAIModelFromConfig` passes it straight through. `option.WithAPIKey("")` is applied after the SDK's own env defaults, and openai-go only sets the `authorization` header when the key is non-empty, so a keyless endpoint is called with no header rather than with a placeholder. That also let me drop the pre-existing `"ollama"` placeholder in `NewOpenAICompatibleModel`. Added a test that runs a real request against an `httptest` server and asserts no `Authorization` header is sent.

On where the validation should live, I want your call between two options rather than guessing:

1. Keep it in the runtime, but move it up one level, out of `pkg/models` into the `*adk.OpenAI` case of `agent.CreateLLM`. `pkg/models` becomes pure client construction with no policy, and the "a key is required unless `baseUrl` is set" rule sits with the provider dispatch. Small follow-up to what is already pushed.

2. Move it to the controller. `ResolveModelConfig` already switches per provider and has an `addSemanticFailure` path that surfaces as `Accepted=False` on the `ModelConfig`, but the OpenAI case only checks the API-key secret reference today. Adding the rule there means the user sees an actionable condition on the `ModelConfig` instead of an agent pod that crashloops. Bigger change, and it is the one that stops the next person hitting this.

I read your comment as asking for 1. I would rather do 2, or 1 now and 2 as a separate issue, since the current failure mode only shows up in pod logs. Which do you prefer?

I left the `"passthrough"` placeholder alone. It is the same convention in `anthropic.go`, `foundry.go` and `azureai.AuthOptions`, so changing it belongs in its own PR.

go/adk/pkg/models/openai.go
```diff
@@ -60,26 +60,41 @@ type OpenAIModel struct {
...
+		"OPENAI_API_KEY is not set; calling the custom OpenAI base URL without an Authorization header",
+		"base_url", config.BaseUrl)
+	return "", nil
+}
+
 // NewOpenAICompatibleModel creates an OpenAI-compatible model (e.g. LiteLLM, Ollama).
 // baseURL is the API base (e.g. http://localhost:11434/v1 for Ollama). apiKey is optional; if empty,
-// OPENAI_API_KEY is used, then a placeholder for endpoints that do not require a key.
+// OPENAI_API_KEY is used, and if that is also empty the endpoint is called unauthenticated.
 func NewOpenAICompatibleModel(ctx context.Context, baseURL, modelName string, headers map[string]string, apiKey string) (*OpenAIModel, error) {
```
EItanya: Why do we have 2 separate functions here at all. `NewOpenAIModel` has the new resolution logic you added, and this function does not. It's not clear to me why we need both if the above handles both cases

QuentinBisson: It can be removed. The only real difference is that NewOpenAICompatibleModel accepts an API key that overrides the environment but this can be covered by adding APIKey to the OpenAIConfig struct. I'm removing it

EItanya (approved):

QuentinBisson:
Hey @EItanya thanks for approving :) It looks like the go-lint failed due to a transient error. Would you like me to push an empty commit to retrigger CI?

PR #2097, opened 2026-06-29
feat(scheduledrun): add cron-based agent execution

0xLeo258:
Adds the v1alpha2 ScheduledRun CRD, controller scheduler, REST API, metrics, and UI flow for creating, listing, viewing, updating, and manually triggering scheduled agent runs.

ScheduledRun targets resolve through a shared same-namespace Agent/SandboxAgent reference path. Suspended runs do not schedule next executions, but manual triggers are still allowed while suspended.

RunStatus unifies dispatch and outcome into a single enum (DispatchFailed/InProgress/Succeeded/Failed/TimedOut); RunHistoryEntry is StartTime, EndTime, SessionID, Status, Message. The outcome poller is restart-safe: InProgress entries are resumed on Start so a pod restart between dispatch and terminal resolution does not leave entries stuck InProgress.

CRD admission pattern annotations enforce DNS-label constraints, so duplicate handler-side DNS validators are removed. Service creation is enabled for agents.x-k8s.io Sandboxes so SandboxAgent A2A endpoints are reachable in real e2e deployments.

Includes CRD/Helm generation plus focused unit and e2e coverage.

PR #2741, opened 2026-09-08
feat: add scheduled runs to the current agent API

EItanya:
Scheduled runs in #2097 depend on the legacy CRD, REST, and Session APIs. This ports that functionality to the current AgentTemplate/Harness, PostgreSQL, and A2A APIs. Each cron or manual firing creates a fresh AgentInstance and retains its own execution record after the conversation is deleted.

- Add creator-scoped schedule CRUD, etag updates, pause/resume, idempotent manual triggers, and paginated execution history, newest first.
- Persist binary protobuf payloads with indexed coordination columns, SQL-generated timestamps, and transactional reservations. Keep the unreleased schema in migration 1; this requires a fresh database.
- Isolate malformed stored records: remove invalid schedules from the cron queue, preserve their payloads for repair, and skip invalid executions without discarding healthy leases. Authorized deletion still works and returns only trusted tombstone metadata for damaged schedules; Get/List continue to reject malformed payloads.
- Run accepted work through the shared A2A handler under control-plane authority, with SQL leases, restart recovery, deadline cancellation, and snapshot cleanup. An uncertain dispatch is never resent automatically; it may time out and require a new manual firing.
- Reserve cron firings in a leader-elected scheduler, independently of execution reconciliation on every replica using SQL leases. Slow execution batches cannot delay cron reservation.
- Dispatch scheduled work through streaming A2A sends. The existing gateway ingester persists events independently of the reconciler; subscriptions recover interrupted streams. Keep Get/List requests database-only.
- Add the Schedules UI with create/edit/delete, pause/resume, run now, history, and conversation links. Use frequency, day, and time controls with readable summaries; preserve arbitrary cron expressions in advanced mode. Embed IANA timezone data for minimal controller images.

Substrate actor credentials for tool calls and user-auth elicitation remain deferred. Existing process-local gateway coordination is unchanged.

Validation:

- Malformed-record regression tests reproduce cron starvation and discarded execution leases before the fix. PostgreSQL tests now cover isolation and repair; a real gRPC/PostgreSQL test verifies authorized, idempotent deletion and retained history. Database, gRPC, scheduling controller/service tests and scoped Go lint pass. Public request validation remains in Protovalidate; these tests inject invalid durable data through internal boundaries.
- After rebasing onto main `03fe32d3`, UI typechecking, 420 unit tests, and 76 selected Chromium/Firefox tests passed; gateway/controller/gRPC race tests and database/migration tests passed. Live cluster E2Es were compile-checked only in that rebase pass.
- Schedule controls: 22 focused unit tests, eight Chromium/Firefox browser checks, source lint, and production image build passed. The deployed Kubernetes UI also passed schedule create/reload/edit/delete against the real controller and PostgreSQL, including timezone preview defaults.
- Latest SQL timestamp, typed UUID, and shared pagination refactors passed database, migration, service, controller, gRPC, and app tests plus scoped Go lint. Empty/malformed scheduling IDs are rejected by the real gRPC interceptor. The full Go live runs below predate these refactors.
- Relevant Go package tests passed: scheduling semantics/service, database, migrations, gRPC, and A2A gateway. Database/gRPC pagination tests passed after the history-ordering fix.
- Full Go lint, Buf lint/regeneration, sqlc generation, and UI typecheck passed. UI unit tests, build, and source lint passed (seven existing lint warnings).
- Live tests passed on an isolated Kind cluster with Substrate 0.0.26 and fresh PostgreSQL: cron/manual execution, deadline cleanup, and controller restart recovery with exactly one model call. The cron test was rerun after reconnecting a stale port-forward following rollout. Live Chromium schedule create/reload/edit/delete also passed, including `America/New_York` and a fractional timeout.
- Full UI browser suite: **202 passed, one Firefox chat startup failure**. The trace showed `NS_ERROR_FILE_NO_DEVICE_SPACE` loading an application module while `/tmp` was 94% full. The affected test passed on rerun with browser profiles on the workspace filesystem. All scheduling tests passed in Chromium and Firefox.

Original work: [0xLeo258](https://github.com/0xLeo258), #2097. Commits `7e6ad0dd` and `f68a7cc0` are retained as ancestors with their original SHAs and authors. The `ours` merge preserves that history while keeping the completed API port; it changes no files. #2097 now targets the release branch, so this is a successor against main.

**The repository currently allows only squash merges.** Preserving the contributor commits on main requires enabling and using a merge commit. If squash-only policy remains, include this footer in the squash commit to retain credit (the original SHAs remain on this PR branch):

```text
Co-authored-by: 0xLeo258 <noixe0312@gmail.com>
```


Charlesthebird:
Reviewed the UI changes only. Typecheck, lint and unit tests pass, and `yarn test:pw schedules` is 8/8 on Chromium and Firefox. I also ran the page against a real cluster.

The scheduling UI is in good shape. What I checked that came out correct:

- ✅ Client validation mirrors the proto constraints exactly — name 200, cron 256, time zone 253, prompt 32768 **bytes** via `TextEncoder`, timeout 1µs–9223372036s.
- ✅ Idempotency works in both directions: the create request ID survives a failed retry, and the manual-trigger ID is held on failure and cleared on success, with unit tests covering both.
- ✅ Fractional timeouts serialise without floating-point drift, asserted down to exact nanos.
- ✅ A deleted schedule disables its actions and keeps its history.
- ✅ Read failures stay distinct from an empty list, per the repo's own rule.
- ✅ `Checkbox.Group` sorts by option order, so the "Weekdays" shorthand does not depend on click order.

Seven notes, none of them blocking:

| # | Location | Note |
|---|---|---|
| 1 | `ScheduledRunsPage.tsx:54,163` | `loading={runs.isValidating}` disables both pagination buttons on every background poll |
| 2 | `ScheduledRunsPage.tsx:21-29` | `PageButtons` re-implements `PageControls` and drops its documented behaviour |
| 3 | `ScheduledRunsPage.tsx:51,153` | Status tags carry no colour, so Succeeded and Failed render identically |
| 4 | `ScheduledRunsPage.tsx:155` | The "Result" column is `dataIndex: "failureReason"`, so it is blank on success |
| 5 | `ScheduledRunsPage.tsx:137-139` | A single `busy` flag spins Run and Pause together |
| 6 | `ScheduledRunsPage.tsx:43,119` | `a: { color: primaryText }` duplicates what `theme.ts:221-222` already sets |
| 7 | `ScheduledRunsPage.tsx:152` | A trigger oneof with neither arm set is labelled "Manual" |

Two are worth more than a line.

**1** is the one I would fix first. `useApiResource.ts:19` documents `isValidating` as "including a refresh over data already on screen", and the history table polls at 5s, so Prev/Next go dead on a duty cycle. `PageControls` takes `isLoading` for exactly this reason.

**2** — `SubstratePage.tsx:530` carries the rationale for hiding the pager on a single page: "two dead buttons under a five-row table read as a broken control rather than as a complete list." The new pager always renders, so a one-row list shows Previous / Page 1 / Next with both greyed out. It is also now the second hand-rolled token pager in the app, which makes this the moment to share one.


---
*🤖 written by Claude*

Charlesthebird (commented):

---

Scoped to the UI changes only. Nothing here blocks merge — the page is sound and every note is inline.

Two are worth more than a line: the pager disables itself on every background poll, and it re-implements the pager `SubstratePage.tsx` already has, dropping that one's reason for hiding itself on a single page.

<details>
<summary>✅ What I checked</summary>

- Client validation matches every `buf.validate` constraint in `scheduled_runs.proto` — name 200, cron 256, time zone 253, prompt 32768 **bytes** measured with `TextEncoder`, timeout `1µs`–`9223372036s`.
- The create request ID is generated once per editor and reused after a failed save, and the manual-trigger ID is cleared only after the server accepts it, so neither path can queue a duplicate.
- A fractional timeout reaches the wire as exact nanos rather than a float, pinned at `1.001s` → `{seconds: 1n, nanos: 1000000}`.
- A soft-deleted schedule disables Run, Pause, Edit and Delete while still rendering its history.
- `?mock=error` renders the error state and `?mock=empty` the empty one, so a read failure cannot present itself as an empty list.
- `Checkbox.Group` emits its value sorted by option order, so the `1-5` → "Weekdays" shorthand does not depend on the order the boxes were ticked.
- Against a live cluster the page created, listed, paused and read back a five-minute schedule in `America/New_York`.

</details>

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <Table<ScheduledRun> rowKey="id" loading={runs.isLoading} pagination={false} scroll={{ x: 800 }}
+        dataSource={runs.data?.scheduledRuns ?? []} locale={{ emptyText: runs.error ? "Schedules unavailable" : "No schedules yet" }} columns={[
+          { title: "Name", key: "name", render: (_, row) => <Link to={buildPath(paths.scheduledRun, { id: row.id })}>{row.config?.name || row.id}</Link> },
+          { title: "Agent", key: "agent", render: (_, row) => `${row.agentTemplate?.name ?? "—"} on ${row.harness?.name ?? "—"}` },
+          { title: "Schedule", key: "schedule", render: (_, row) => row.config ? scheduleDescription(row.config.schedule) : "—" },
+          { title: "Time zone", key: "zone", render: (_, row) => row.config?.timeZone || "UTC" },
+          { title: "Status", key: "status", render: (_, row) => <Tag>{row.config?.paused ? "Paused" : "Active"}</Tag> },
+          { title: "Next execution (local)", key: "next", render: (_, row) => time(row.nextExecutionTime) },
+        ]} />
+      <PageButtons tokens={tokens} setTokens={setTokens} next={runs.data?.page?.nextPageToken} loading={runs.isValidating || !!runs.error} />
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

`isValidating` is true on background revalidation, so both pagination buttons go dead on every poll — 10s here and 5s on the history table below.

`useApiResource.ts:19` names the distinction: `isValidating` is "a request is in flight, including a refresh over data already on screen", while `isLoading` is "first load, nothing to show yet". `PageControls` in `SubstratePage.tsx` takes `isLoading` for exactly this reason.

```suggestion
      <PageButtons tokens={tokens} setTokens={setTokens} next={runs.data?.page?.nextPageToken} loading={runs.isLoading} />
```

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+function PageButtons({ tokens, setTokens, next, loading }: {
+  tokens: string[]; setTokens: (tokens: string[]) => void; next?: string; loading: boolean;
+}) {
+  return <Space>
+    <Button disabled={loading || tokens.length === 1} onClick={() => setTokens(tokens.slice(0, -1))}>Previous page</Button>
+    <Typography.Text>Page {tokens.length}</Typography.Text>
+    <Button disabled={loading || !next} onClick={() => next && setTokens([...tokens, next])}>Next page</Button>
+  </Space>;
+}
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

This is the second hand-rolled token pager in the app, and it drops a decision the first one documents.

`SubstratePage.tsx:515` has `PageControls`, whose comment explains why it returns `null` on a single page: "two dead buttons under a five-row table read as a broken control rather than as a complete list." Because this one always renders, a one-row list shows Previous / Page 1 / Next with both greyed out. `usePageStack` at `SubstratePage.tsx:360` is the matching token-stack hook, with a derived reset that this component's `setTokens([""])` at line 101 is doing by hand.

Worth lifting both into `components/table/` and using them here rather than keeping two.

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+    <Space orientation="vertical" css={{ display: "flex", a: { color: theme.color.primaryText } }} size="middle">
+      {runs.error && <Alert type="error" showIcon title="Could not load schedules" description={runs.error.message} />}
+      <Table<ScheduledRun> rowKey="id" loading={runs.isLoading} pagination={false} scroll={{ x: 800 }}
+        dataSource={runs.data?.scheduledRuns ?? []} locale={{ emptyText: runs.error ? "Schedules unavailable" : "No schedules yet" }} columns={[
+          { title: "Name", key: "name", render: (_, row) => <Link to={buildPath(paths.scheduledRun, { id: row.id })}>{row.config?.name || row.id}</Link> },
+          { title: "Agent", key: "agent", render: (_, row) => `${row.agentTemplate?.name ?? "—"} on ${row.harness?.name ?? "—"}` },
+          { title: "Schedule", key: "schedule", render: (_, row) => row.config ? scheduleDescription(row.config.schedule) : "—" },
+          { title: "Time zone", key: "zone", render: (_, row) => row.config?.timeZone || "UTC" },
+          { title: "Status", key: "status", render: (_, row) => <Tag>{row.config?.paused ? "Paused" : "Active"}</Tag> },
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

A bare `Tag` renders Paused and Active identically apart from the word. The convention elsewhere is a colour alongside the label — `DashboardPage.tsx:261`, `AgentTemplatesPage.tsx:174`, `AgentTemplateDetailsPage.tsx:312`.

```suggestion
          { title: "Status", key: "status", render: (_, row) => row.config?.paused ? <Tag color="warning">Paused</Tag> : <Tag color="success">Active</Tag> },
```

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+        </Space>
+      </>}
+      <Typography.Title level={3}>Execution history</Typography.Title>
+      <Typography.Text type="secondary">Times are shown in your local time zone. Conversation links may refer to conversations that have since been deleted.</Typography.Text>
+      {history.error && <Alert type="error" showIcon title="Could not load execution history" description={history.error.message} />}
+      <Table<ScheduledRunExecution> rowKey="id" loading={history.isLoading} pagination={false} scroll={{ x: 800 }}
+        dataSource={history.data?.executions ?? []} locale={{ emptyText: history.error ? "History unavailable" : "No executions yet" }} columns={[
+          { title: "Created", key: "created", render: (_, row) => time(row.createdAt) },
+          { title: "Trigger", key: "trigger", render: (_, row) => row.trigger.case === "scheduledTime" ? `Scheduled: ${time(row.trigger.value)}` : "Manual" },
+          { title: "State", key: "state", render: (_, row) => <Tag>{executionState(row.state)}</Tag> },
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

Same here, and it matters more: Succeeded, Failed and Timed out all come out as the same grey pill, so a failed run is not scannable in the history.

The label still carries the state on its own, so colour is a second channel rather than the only one.

```suggestion
          { title: "State", key: "state", render: (_, row) => executionStateTag(row.state) },
```

with `executionState` returning the tag — `success` for Succeeded, `error` for Failed, `warning` for Timed out, `processing` for Running.

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+          { title: "Completed", key: "completed", render: (_, row) => time(row.completedAt) },
+          { title: "Result", dataIndex: "failureReason", key: "result" },
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

A column called "Result" that is bound to `failureReason` is empty on every successful execution, which reads as missing data rather than as nothing to report.

```suggestion
          { title: "Failure reason", key: "failureReason", render: (_, row) => row.failureReason || "—" },
```

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+          { key: "schedule", label: "Schedule", children: scheduleDescription(config.schedule) },
+          { key: "zone", label: "Time zone", children: config.timeZone || "UTC" },
+          { key: "next", label: "Next execution (local)", children: time(schedule.nextExecutionTime) },
+          { key: "timeout", label: "Execution timeout", children: config.executionTimeout ? `${Number(config.executionTimeout.seconds) + config.executionTimeout.nanos / 1e9} seconds` : "15 minutes" },
+          { key: "prompt", label: "Prompt", span: 2, children: <Typography.Paragraph css={{ whiteSpace: "pre-wrap", margin: 0 }}>{config.prompt}</Typography.Paragraph> },
+        ]} />
+        <Space wrap>
+          <Button icon={<Play size={14} />} disabled={disabled} loading={busy} onClick={() => void act("trigger")}>Run</Button>
+          <Button icon={config.paused ? <Play size={14} /> : <Pause size={14} />} disabled={disabled} loading={busy}
+            onClick={() => void act("pause")}>{config.paused ? "Resume" : "Pause"}</Button>
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

One `busy` boolean drives both buttons, so pressing Run puts a spinner on Pause too.

Holding which action is in flight fixes it: `useState<"pause" | "trigger">()` at line 72, `setBusy(operation)` at line 91, `setBusy(undefined)` at line 111, then `loading={busy === "trigger"}` and `loading={busy === "pause"}` here. `disabled` at line 83 becomes `!!busy || ...`.

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [creating, setCreating] = useState(false);
+  const [tokens, setTokens] = useState([""]);
+  const token = tokens[tokens.length - 1];
+  const runs = useApiResource(["scheduledRuns.list", token],
+    () => invoke("scheduledRuns.list", { page: { limit: 25, pageToken: token } }), { refreshInterval: 10000 });
+
+  return <PageFrame title="Schedules" description="Run an agent automatically. Each execution starts a new conversation."
+    actions={<Space><RefreshButton onRefresh={runs.refresh} what="Schedules" loading={runs.isValidating} />
+      <Button type="primary" icon={<Plus size={14} />} onClick={() => setCreating(true)}>New Schedule</Button></Space>}>
+    <Space orientation="vertical" css={{ display: "flex", a: { color: theme.color.primaryText } }} size="middle">
```
Charlesthebird: [nit]

This override is already the default: `theme.ts:221-222` sets both `colorLink` and `colorLinkHover` to `color.primaryText`, and no other page does this. Line 119 has the same one, and dropping both makes `useTheme` unused in each component.

```suggestion
    <Space orientation="vertical" css={{ display: "flex" }} size="middle">
```

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+            onDelete={async () => { await invoke("scheduledRuns.delete", { scheduledRunId: id }); }} onDeleted={refresh} />
+        </Space>
+      </>}
+      <Typography.Title level={3}>Execution history</Typography.Title>
+      <Typography.Text type="secondary">Times are shown in your local time zone. Conversation links may refer to conversations that have since been deleted.</Typography.Text>
+      {history.error && <Alert type="error" showIcon title="Could not load execution history" description={history.error.message} />}
+      <Table<ScheduledRunExecution> rowKey="id" loading={history.isLoading} pagination={false} scroll={{ x: 800 }}
+        dataSource={history.data?.executions ?? []} locale={{ emptyText: history.error ? "History unavailable" : "No executions yet" }} columns={[
+          { title: "Created", key: "created", render: (_, row) => time(row.createdAt) },
+          { title: "Trigger", key: "trigger", render: (_, row) => row.trigger.case === "scheduledTime" ? `Scheduled: ${time(row.trigger.value)}` : "Manual" },
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

A `trigger` oneof with neither arm set falls through to the "Manual" label, so an unset trigger is reported as a deliberate manual run.

```suggestion
          { title: "Trigger", key: "trigger", render: (_, row) => row.trigger.case === "scheduledTime" ? `Scheduled: ${time(row.trigger.value)}` : row.trigger.case === "manualRequestId" ? "Manual" : "Unknown" },
```

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: Withdrawing this one — I had it wrong, and it stays as you wrote it.

I removed the override and measured the result: the link rendered `rgba(21, 25, 39, 0.88)` instead of the `rgb(109, 40, 217)` the same table link gets on `/agents` and `/prompts`. `colorLink` is a ConfigProvider token, so it styles antd's own components and never reaches a bare react-router `<Link>`. Colouring these explicitly is the established pattern — `PromptsPage.tsx:96` even documents why — so the override is doing real work.


---
*🤖 written by Claude*

Charlesthebird:
Pushed fixes for the notes above, in [`5ac671b3`](https://github.com/kagent-dev/kagent/commit/5ac671b3) and [`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c). One of them I got wrong and reverted — the detail is on its thread.

<details>
<summary>What changed against each note</summary>

| Note | Resolution |
|---|---|
| Pager disabled on every poll | Uses the shared controls, which take `isLoading` |
| Pager duplicates `PageControls` | `PageControls` and `usePageStack` moved to `components/table/`, used by both pages; a `reset()` replaces `setTokens([""])` |
| Status tag has no colour | `success` for Active, `warning` for Paused, plain for Deleted |
| Execution states indistinguishable | `success`, `error`, `warning`, `processing` beside the unchanged labels |
| "Result" empty on success | Renamed "Failure reason", em dash when empty |
| One `busy` flag spins both buttons | `busy` holds which action is in flight |
| Redundant link override | **Withdrawn** — the override is load-bearing, reverted |
| Unset trigger reads as "Manual" | Now "Unknown" |

</details>

<details>
<summary>✅ How the fixes were checked</summary>

- The pager is absent on a single-page list, asserted in `schedules.spec.ts` on `schedules-pages` rather than on a timing window.
- Tag contrast measured rather than assumed: Active is 4.57:1 in light and 4.99:1 in dark, both past AA for normal text.
- Table links measured back at `rgb(109, 40, 217)` light and `rgb(167, 139, 250)` dark, matching `/agents` and `/prompts`.
- The substrate suite still passes on both browsers, which is what covers the pager in its original home.
- One Firefox chat spec failed under full-suite load and passes in isolation; it predates this change and touches nothing here, so I left it rather than papering over it with a timeout.

</details>


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: Pagination takes `isLoading` now, through the shared controls ([`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: `PageControls` and `usePageStack` live in `components/table/` and serve both pages ([`5ac671b3`](https://github.com/kagent-dev/kagent/commit/5ac671b3), [`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: Active and Paused carry `success` and `warning` ([`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: Execution states carry a colour beside the unchanged label ([`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: The column is "Failure reason" and shows an em dash when there is none ([`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: `busy` holds which action is in flight, so only that button spins ([`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: An unset trigger reads as Unknown ([`a7340e2c`](https://github.com/kagent-dev/kagent/commit/a7340e2c)).


---
*🤖 written by Claude*

Charlesthebird:
Recorded against a real cluster, not fixtures — one schedule created, read back, renamed and deleted.

https://github.com/user-attachments/assets/18514812-69af-42e9-89ec-fa4f9c64f42c

<details>
<summary>What changed</summary>

- Create and edit moved onto their own routes
- Form left the modal it outgrew
- Edit page reads before mounting the form
- Edit page does not poll, protecting the etag
- Per-row edit and delete on the list
- Detail page gained its own header block
- Status, cadence and zone shown as pills
- Every control moved beside the title
- Run button is now filled
- Conversation links carry an external-link icon
- Agent page lists the schedules running it
- Empty state offers to create one
- Both halves of the pair are matched
- Client-side narrowing, capped at one page
- Failed reads withhold the empty-state wording
- Link colour and press states shared once
- Create, read, update, delete covered in CI

</details>

<details>
<summary>Deliberately left out</summary>

- No cron toggle; Repeat → Custom already covers it
- No server-side agent filter; needs an RPC field
- No polling on the agent section
- CRUD tests use mocks, matching every other spec

</details>

---
*🤖 written by Claude*

Charlesthebird (commented):

---

Second pass, covering the UI work added since the first review — seven commits, about 1300 lines. Nothing here blocks merge. I wrote or integrated most of it, which is the weakest position to review from, so the two findings that matter are backed by a probe rather than by reasoning.

The one worth reading is the first: a row a search reveals cannot be closed again by any means.

<details>
<summary>✅ What I checked</summary>

- The four copy-pasted row-click guards are gone: the selector now appears once, in `rowClick.ts`, and `McpServersPage` and `AgentPage` keep their previous behaviour including the conditionally-clickable rows.
- `expandRowByClick` is still avoided everywhere, so a click on a delete button does not unfold the row behind its own confirmation — the reason the expansion is controlled at all.
- A schedules row and an agent-page schedules row both navigate, while their edit and delete buttons do not, and the row's own `<Link>` still carries keyboard users.
- Search matches what is rendered rather than what is stored, so "Timed out" finds `TIMED_OUT` and a date matches as displayed.
- Three empty states stay distinct: a failed read, a history with nothing in it, and a search that matched nothing.
- Deleting from the detail page returns to the list and sweeps every `scheduledRuns.` key, so the agent page's section stops offering a schedule that is gone; a deleted schedule still opens by address and says so.
- The danger section carries no fill, border or icon of its own, and its copy is `textMuted` — 5.23:1 light and 6.81:1 dark, measured in a browser, where antd's `type="secondary"` had been 2.85:1.
- Search and page position survive a reload through the address, using the writer `useListView` already shares.

</details>

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
```diff
@@ -0,0 +1,319 @@
...
+  const schedule = run.data?.scheduledRun;
+  const config = schedule?.config;
+  const disabled = !!busy || !!run.error || !!schedule?.deletedAt;
+
+  const executions = useMemo(() => history.data?.executions ?? [], [history.data]);
+  const { rows, revealed } = useMemo(() => searchExecutions(executions, query), [executions, query]);
+  /* A row whose only match is inside its panel is opened, so the reader can see why it
+     came back. Merged with what they opened by hand rather than replacing it. */
+  const expandedKeys = useMemo(
+    () => [...new Set([...expanded.keys, ...revealed])], [expanded.keys, revealed]);
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

A row revealed by a search cannot be collapsed again — not by clicking it, not twice, and not with its own expand icon.

`revealed` is recomputed from the query on every render, so it is a permanent member of this union rather than a one-time seed. Clicking the row calls `toggle`, which sees the key is absent from `expanded.keys` and *adds* it, so the first click does nothing visible; the second removes it from `keys`, and the union puts it straight back.

Measured against `?mock=ok`, searching `mock-scheduled-task-3` — a term only the panel holds:

```
after 1st click, panels visible: 1
after 2nd click, panels visible: 1
after expand-icon click, panels visible: 1
```

Seeding `expanded.keys` when the query changes, and rendering `expanded.keys` alone, would keep the reveal and give the row back to the reader.

---
*🤖 written by Claude*

ui/src/components/table/rowClick.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  return {
+    keys,
+    set,
+    toggle: (key: string) => set(key, !keys.includes(key)),
+    expandAll: (visible: string[]) => setKeys(visible),
+    collapseAll: () => setKeys([]),
+    /** Whether every one of `visible` is already open, for a button that says which it will do. */
+    allExpanded: (visible: string[]) =>
```
Charlesthebird, line 97: **Level:** 🟡 Low · Not Blocking

`allExpanded` reads `expanded.keys`, but the table renders the union with `revealed`, so the label contradicts the screen: after a search that reveals a row, every visible row is open and the button still reads "Expand all".

Same probe run: `expand-all label: Expand all`, with the one matching row expanded. Fixing the finding above fixes this too, since the two states stop diverging.

---
*🤖 written by Claude*

ui/src/components/table/rowClick.ts
```diff
@@ -0,0 +1,86 @@
...
+ * `expandAll` takes the keys rather than reading them from anywhere, because only the
+ * caller knows which rows are on screen — a filtered, paginated table's "all" is its
+ * current page's matches, not everything it could ever show.
+ */
+export function useExpandedRows() {
+  const [keys, setKeys] = useState<string[]>([]);
+
+  function set(key: string, open: boolean) {
+    setKeys((current) =>
+      open ? [...current, key] : current.filter((candidate) => candidate !== key),
```
Charlesthebird: [nit]

`set(key, true)` on a key already present appends a duplicate, because the decision is taken from the rendered `keys` while the write is a functional update on `current`.

Nothing reaches it today — `toggle` guards and `ScheduledRunsPage` dedupes downstream through `new Set` — but `McpServersPage` passes `expanded.keys` straight to `expandedRowKeys`, so the guard is the only thing standing between this and a duplicate key.

```suggestion
      open ? [...new Set([...current, key])] : current.filter((candidate) => candidate !== key),
```

---
*🤖 written by Claude*

ui/src/pages/ScheduledRunsPage.tsx
Charlesthebird: `revealed` is a seed now rather than a union, derived the way `usePageStack` resets, so a new query starts from what it revealed and everything after is the reader's ([`125fd91b`](https://github.com/kagent-dev/kagent/commit/125fd91b)). The probe that found this reports 1 revealed, 0 after a click, 1 after clicking again, and re-seeded on a fresh search — now pinned as a spec step.


---
*🤖 written by Claude*

ui/src/components/table/rowClick.ts
Charlesthebird: Fixed by the same change: the keys and the rendered set no longer diverge, so the label reads "Collapse all" while a revealed row is open ([`125fd91b`](https://github.com/kagent-dev/kagent/commit/125fd91b)).


---
*🤖 written by Claude*

ui/src/components/table/rowClick.ts
Charlesthebird: Adding a key twice dedupes now ([`125fd91b`](https://github.com/kagent-dev/kagent/commit/125fd91b)).


---
*🤖 written by Claude*

Charlesthebird:
Backend findings below high. The blockers are line comments.

**Confirmed by reading**

| Where | What |
| --- | --- |
| `scheduled_runs.go:35` | Schedules are UUIDv4. Listing orders by id. Order is arbitrary. |
| `scheduled_runs.go:208` | Advances once per pass. Outages drop the newest miss. |
| `queries/scheduled_runs.sql:82` | Retries every second. No backoff. |
| `queries/scheduled_runs.sql:23` | A normal save un-deletes. Callers pre-check, so safe. |
| `scheduled_runs.proto` | No `total_size` field. History size is unknowable. |

**Reported, not verified.** No cron minimum. No per-creator quota. Unbounded timeout. Create validates after commit. `input-required` polls to deadline. Every tick rewrites the row.

---
*🤖 written by Claude*

go/core/internal/a2agateway/task_run.go
```diff
@@ -56,30 +59,38 @@ func (g *Gateway) startTaskRun(ctx context.Context, instance *apiv1alpha1.AgentI
...
+		_ = r.closeRuntime()
 		close(r.done)
 		_ = r.gateway.events.Destroy(ctx, r.queueID)
 		r.gateway.runs.CompareAndDelete(r.key, r)
 	}()
 
 	for event, eventErr := range events {
 		if eventErr != nil {
-			r.gateway.failAttempt(ctx, &preparedSend{instance: instance, task: task, previous: previous})
 			r.setError(eventErr)
```
Charlesthebird, line 86: **Level:** 🔴 High · Blocking

`failAttempt` is gone and `failTask` has no callers. `setError` is in-memory, so the task never finishes. Chat then refuses the next message until cancelled. If deliberate, this needs a comment and a test.

---
*🤖 written by Claude*

Charlesthebird (commented):

---

Backend findings, as line comments. High only; the rest are in the comment below.

What blocks merge: a dropped runtime stream leaves a task unfinished forever, and either scheduling batch can be stopped for everyone by one undecodable row.

<details>
<summary>✅ What I checked</summary>

- Multi-replica reservation is sound: `FOR UPDATE SKIP LOCKED` plus `UNIQUE (scheduled_run_id, scheduled_time)` makes a double firing unrepresentable.
- No transaction or lock spans a network call; every A2A and Substrate call sits outside the store.
- `schedule.go` embeds tzdata, rejects `Local` and pins `CRON_TZ`, so DST comes from the zone rather than the host.

</details>

---
*🤖 written by Claude*

go/core/internal/database/scheduled_runs.go
```diff
@@ -0,0 +1,483 @@
...
+	token := uuid.New()
+	rows, err := c.q.LeaseScheduledRunExecutionsForUpdate(ctx, dbgen.LeaseScheduledRunExecutionsForUpdateParams{Limit: int32(limit), LeaseToken: token})
+	if err != nil {
+		return nil, fmt.Errorf("failed to lease scheduled executions: %w", err)
+	}
+	leases := make([]LeasedScheduledRunExecution, 0, len(rows))
+	for _, row := range rows {
+		execution, err := toScheduledRunExecution(row)
+		if err != nil {
+			return nil, err
```
Charlesthebird: **Level:** 🔴 High · Blocking

The UPDATE at `:396` runs outside `withTx`, so eight leases are already committed by here. One undecodable row then drops all of them, and re-sorts into the next window. Nothing progresses while it exists. Skip and log the row instead of returning.

---
*🤖 written by Claude*

go/core/internal/database/scheduled_runs.go
```diff
@@ -0,0 +1,483 @@
...
+	err := c.withTx(ctx, func(q *dbgen.Queries) error {
+		rows, err := q.GetDueScheduledRunsForUpdate(ctx, int32(limit))
+		if err != nil {
+			return err
+		}
+		for _, due := range rows {
+			row, now := due.ScheduledRun, due.DbTime
+			schedule, err := toScheduledRun(row)
+			if err != nil {
+				return err
```
Charlesthebird: **Level:** 🔴 High · Blocking

Returning aborts the batch, so nothing is reserved or advanced. The row sorts first again and fails every second, stopping cron for every user. `DeleteScheduledRun` decodes first too, so the API cannot remove it. Reachable through version skew, not through today's API.

---
*🤖 written by Claude*

Charlesthebird (approved):
UI updates should be all done now. I tested this and it looks good 🎉 

https://github.com/user-attachments/assets/92e5ebc4-796e-42e9-9733-02d1298c7b9d

There was an automated backend review that was just posted above.

PR #2747, opened 2026-09-08
feat(ui): fork a conversation, and open the agent rail to extensions

Charlesthebird:
---
*🤖 written by Claude (start)*

Adds a way to fork a conversation from the web interface, and opens the agent rail's navigation to app extensions.

- Fork is offered from the reader's own messages, enabled on the latest one and disabled on the others, because `CreateCheckpoint` takes no cutoff and a checkpoint is always the conversation's latest turn boundary. The disabled item is shown rather than hidden, so the limit is where someone looks for the feature.
- `agentInstances.fork` is one operation over two RPCs, `CreateCheckpoint` then `ForkAgentInstance`, sharing a request id so a retry cannot leave a second checkpoint or fork behind. It titles the fork after its source.
- The rail's two entries become data, and gain the pair of extension points the application sidebar already had: `agentRailItems` contributes entries interleaved by `order`, and `agentRailOverrides` hides, renames, retargets, re-icons or reorders the application's own. Contributing a key the application owns is a startup error naming the override.

Worth knowing while reviewing: `agentLinks.details` is documented as where the rail's Agent Details entry points, and nothing reads it. That predates this branch, and `agentRailOverrides.agentDetails.path` now covers the same intent for a fixed destination.

## Testing

1. `ENABLE_MOCK_UI=true yarn dev`, open a conversation, and use the menu on your latest message to fork it. The fork opens holding the source's transcript.
2. Open the menu on an earlier message of your own; Fork chat is disabled there.
3. `VITE_EXAMPLE_EXTENSION=true yarn dev` and open a conversation: the bundled example contributes a rail entry between Agent Details and New chat.

Videos and screenshots against a real cluster are in the comments.

---
*🤖 written by Claude (end)*


Charlesthebird:
Fork moved off the sidebar chat row and onto the message itself, next to the "You" / "Agent" label. Screenshots are from mock data (`ENABLE_MOCK_UI=true`).

**On your own message.** The button stays hidden until the message is hovered or it takes focus.

02-message-menu-open

**On an agent message**, same place.

03-agent-message-menu

**The sidebar row menu** offers rename and delete only.

04-sidebar-menu-no-fork

The commit on this branch is still the sidebar version; the move is local and not pushed yet.

---
*🤖 written by Claude*

Charlesthebird:
Fork is now on your own messages only, in the same slot beside the "You" label.

Every menu forks the whole conversation at its latest turn boundary, since `CreateCheckpoint` takes no position. So an anchor on an agent reply or a tool-call row promised something it does not do, and a transcript full of tool calls carried a menu per row.

**On your own message.** Hidden until the message is hovered or the button takes focus.

01-fork-on-your-message

**On the agent's, nothing.** The pointer is on the tool-call message here.

02-none-on-agent-message

**The sidebar row menu** offers rename and delete only.

03-sidebar-menu-no-fork

Screenshots are from mock data (`ENABLE_MOCK_UI=true`). The commit on this branch is still the sidebar version; the move is local and not pushed yet.

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** Superseded by [this comment](https://github.com/kagent-dev/kagent/pull/2747#issuecomment-5593189723), which is narrated and shows #2754 in the branch.

<details>
<summary>First recording, harder to follow</summary>

Verified against a real cluster rather than the mock. Five conversations, the same source forked twice at different lengths, each continued on its own.

The observable is the agent's memory, because a fork sharing history would answer with the source's newer fact.

- Source told two facts, codeword ALPHA and number 7, then forked at two turns.
- That fork answers `ALPHA, 7`, so history carries across a fork.
- The fork sets its number to 99. The source sets its own to 42, then is forked again at three turns.
- Asked afterwards: the early fork says 99 and never sees 42, the later fork says 42 and never sees 99.
- A second conversation and its fork repeat it.
- All five reloaded with their message counts and contents unchanged.

https://github.com/user-attachments/assets/ea4345c2-89dd-40b1-bd1f-916a2791ccec

Cluster built by `scripts/setup-cluster/setup-cluster.sh`, on this branch with #2754 merged in. The one thing not shown is forking from a chosen point in a conversation: `CreateCheckpoint` takes no cutoff, so each fork captures the conversation as it stood, which is why the two forks above differ by when they were taken.

</details>

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** Fork moved back to the conversation's row in the rail. Superseded by [this comment](https://github.com/kagent-dev/kagent/pull/2747#issuecomment-5593478011).

<details>
<summary>Earlier video, forked from a message</summary>

New video: narrated, easier to follow, and it opens on git showing #2754 in this branch.

Each fork is told a fact the other never hears, so a leak would be visible. Fork A knows OTTER, Fork B knows TEAL, and neither sees the other's. Five conversations, all reloaded unchanged.

https://github.com/user-attachments/assets/f1f88b8b-eda6-43b8-986a-65e03a845e36

**Forking from a chosen point is still not possible.** `CreateCheckpoint` takes an instance id and a request id and nothing else, and the service picks the latest quiescent turn boundary itself. So a checkpoint is always the conversation as it stands, and a fork inherits all of it. The two forks here differ only in when they were taken, one at two turns and one at three. Putting the menu item on a message promises nothing about cutting history at that message, which is why it reads "Fork chat" and not "Fork from here". A real per-message fork needs a cutoff on the request. The checkpoint row already records a head task and a history sequence to carry one.

</details>

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** Fork moved back onto the reader's own messages, enabled on the latest one only. Superseded by [this comment](https://github.com/kagent-dev/kagent/pull/2747#issuecomment-5593587618).

<details>
<summary>Video of fork on the rail's conversation row</summary>

Fork is back on the conversation's row in the rail. The per-message menu is kept in `ChatMessageItem`, commented out with a TODO, and the extension point beside it is untouched.

New video, driven from the rail. Every conversation is asked the same question, "what is my animal", and each answers differently because each was told its own after branching. Fork A says OTTER, Fork B says BADGER, the Original says HERON. Fork A also never sees the city the Original learned after Fork A branched off. It opens on git showing #2754 in this branch.

https://github.com/user-attachments/assets/93c0c39d-f7be-4095-8870-964fc22085d5

**Why it is not on a message.** `CreateCheckpoint` takes an instance id and a request id and nothing else, and the service checkpoints the latest quiescent boundary itself. So a checkpoint is always the conversation as it stands, and a fork inherits all of it. A control anchored on one message would imply cutting history at that message and would not do it. The boundary is a real concept underneath: the checkpoint row stores a cutoff and events are copied through it. What is missing is a way for the caller to choose which boundary.

</details>

---
*🤖 written by Claude*

Charlesthebird:
Where the rest of the fork work lives, since the commented-out per-message menu invites the question. All of this is in [the API v2 execution plan](https://github.com/kagent-dev/kagent/blob/main/docs/plans/api-v2-execution-plan.md).

- **K17 and K18 describe what this PR uses.** K17 checkpoints a quiescent turn boundary and commits the history head and sequence. K18 forks from that snapshot tag, and says in-place restore is not added.
- **Two plan items are still open.** The `google.longrunning.Operations` contract for checkpoint and fork, from K2, and extending both atomically across a member set.
- **Choosing an earlier boundary is not in the plan.** `CreateCheckpoint` takes an instance id and a request id, and the store pins the boundary to the head of the history, so a fork always copies the whole conversation.

That last one is why fork is enabled on your latest message only. The menu is drawn on earlier messages and disabled, since a control that forked the whole conversation from a message halfway up would imply cutting history there. Nothing tracks it, in the plan or in an issue, so it needs a decision before it becomes work. The awkward part is the runtime snapshot, which is taken from the source Actor's current state and has no equivalent at an earlier boundary.


---
*🤖 written by Claude*

Charlesthebird:
Fork is back on your own messages, and only your latest one can be forked from. Earlier messages show the item disabled rather than hiding it, so the limit is visible instead of something you discover by guessing.

https://github.com/user-attachments/assets/1e937e02-156d-4a25-b035-309b326bb6a1

The agent's replies carry no menu. The extension point beside it is untouched. Picking a message will not truncate history there until `CreateCheckpoint` accepts a cutoff, so the enabled item forks the whole conversation, which is all the latest message can mean today.

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** This showed a page replacing the agent page, not an entry added to the agent rail. Superseded by [this comment](https://github.com/kagent-dev/kagent/pull/2747#issuecomment-5594138084).

<details>
<summary>Screenshot of a route replacing the core agent page</summary>

What a product's own agent page looks like, contributed as an app extension: a route that `replaces` the core `agent` key, so it serves the application's own address and reads the agent from the same params. The shell around it stays the application's.

custom-agent-page

The page in the shot is throwaway and is not in this branch. Two things I found while making it:

- **`agentLinks.details` is documented as "where the rail's Agent Details entry points" and nothing reads it.** True on `main` too, so it is not from this branch. `agentLinks.chat` is wired; `details` is not. `agentRailOverrides.agentDetails.path` now does that job.
- **An override's `path` is a fixed string**, so it cannot express a per-agent destination. Replacing the route, as above, is the way to get a custom page that knows which agent it is for.

</details>

---
*🤖 written by Claude*

Charlesthebird:
A new entry in the **agent rail**, directly under Agent Details, contributed by an extension with `agentRailItems` at order 150. Order 100 is Agent Details and 200 is New chat, so it lands between them rather than being appended.

rail-entry-closeup

In place, beside the application's own entries:

rail-entry-in-context

It opens the extension's own page, which is handed the conversation the rail was drawn beside and shows the agent's name:

rail-entry-page

The entry and that page are throwaway and are not in this branch. The bundled example does contribute a rail entry at the same position, which the extension specs assert the order of.

---
*🤖 written by Claude*

Charlesthebird:
A new entry in the **agent rail**, directly under Agent Details, contributed by an extension with `agentRailItems` at order 150. Agent Details is 100 and New chat is 200, so it lands between them rather than being appended.

rail-entry

In place, with the application's own entries around it:

rail-entry-full

It opens the extension's own page, which is handed the conversation the rail was drawn beside. That entry and page are throwaway and are not in this branch; the bundled example contributes one at the same position, whose order the extension specs assert.

---
*🤖 written by Claude*

Charlesthebird:
A new entry in the **agent rail**, directly under Agent Details, contributed by an extension with `agentRailItems` at order 150. Agent Details is 100 and New chat is 200, so it lands between them rather than being appended.

rail-crop

Open, with the rail still beside it. The page is the extension's own, and it mounts the application's rail rather than keeping a copy, so the entry stays lit while its page is showing:

custom-page-open

That entry and page are throwaway and are not in this branch; the bundled example contributes one at the same position, whose order the extension specs assert.

---
*🤖 written by Claude*

Charlesthebird:
This should be all done now, just waiting on backend changes for checkpointing to enable the rest of the forking behavior for all messages.

EItanya:
> This should be all done now, just waiting on backend changes for checkpointing to enable the rest of the forking behavior for all messages.

What do you mean by this?

Charlesthebird:
> are you 100% sure about the forking not being able to be done per-message?
> ...
> Fair challenge. No, I was wrong, and the code says so plainly.

🤦 

I'll try this out again 😆

EItanya (approved):

PR #2750, opened 2026-09-08
fix(substrate): render BYO extraContainers in the SandboxAgent ActorTemplate

carlochessamp:
## Problem

A BYO SandboxAgent that sets `spec.byo.deployment.extraContainers` gets them rendered into its pod template (the translator appends them after the kagent container, `manifest_builder.go`), and they work fine on the plain Deployment path. On substrate, however, `buildSandboxAgentActorTemplate` picks out only the kagent container via `findKagentContainer` and constructs a **one-element** `Containers` slice — every extra container is silently dropped. The SandboxAgent CRD accepts them, the ActorTemplate CRD supports up to ten containers with per-container env, so nothing fails; the sidecars are just gone at runtime.

Concrete case we hit: an agent plus a credential-brokering proxy sidecar (loopback forward proxy injecting a `secretKeyRef`-sourced bearer). On substrate the proxy never starts, and since the agent is configured to route egress through `127.0.0.1` it loses all outbound API access.

## Fix

Render the pod template's remaining containers as additional ActorTemplate containers (`actorTemplateExtraContainers`), preserving order after the agent container:

- The ActorTemplate container surface is restricted, so unsupported constructs **fail the build loudly** instead of being silently dropped:
  - image must be digest-pinned (reuses `pinImageRef`; snapshots key on the image)
  - command must be explicit — command + args are concatenated, since atelet copies `Command` verbatim into the OCI Process.Args with no shell and no image-entrypoint fallback
  - env supports literal values and `secretKeyRef` only; `envFrom`, `fieldRef`/`configMapKeyRef` env sources, and `volumeMounts` are rejected with a named-container error
  - resources, ports, and probes have no ActorTemplate equivalent and are ignored (documented on the helper)
- The agent container stays `Containers[0]`, so the durableDir session-store mount (`applyDurableDirSessionStore`) remains agent-private.
- The extra containers feed `actorTemplateShapeHash`, so changes to them fan out blue-green to a new template + actors, same as any other spec change.

## Testing

- `TestBuildSandboxAgentActorTemplateExtraContainers`: happy-path conversion (command/args concat, literal + secretKeyRef env, agent stays first with the durableDir mount, sidecar gets no mounts) plus rejection cases for unpinned image, missing command, `envFrom`, `fieldRef` env, and `volumeMounts`.
- Full `go test ./core/pkg/sandboxbackend/...` green; `go vet` clean.

Scoped to the SandboxAgent path; the AgentHarness substrate path has no extra-container field today, so it is untouched. Targeting `release/v0.10.x` since that is the line that carries the substrate backend; happy to rebase onto `main` if preferred (the sandbox backend was restructured there).

### DIFF
--- a/go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
+++ b/go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
@@ -106,6 +106,11 @@ func (p *Lifecycle) buildSandboxAgentActorTemplate(
 			OnCommit: atev1alpha1.SnapshotScopeFull,
 		},
 	}
+	extraContainers, err := actorTemplateExtraContainers(podTemplate.Spec.Containers, kagentContainer)
+	if err != nil {
+		return nil, err
+	}
+	spec.Containers = append(spec.Containers, extraContainers...)
 	applyDurableDirSessionStore(&spec)
 
 	actorTemplateHash, err := actorTemplateShapeHash(spec)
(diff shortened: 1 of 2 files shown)


carlochessamp:
Closing in favor of #2799 — the same change combined with #2691 into a single PR (superset of both: #2691's structure + readiness/Readyz mapping and 10-container cap, #2750's fail-closed env policy). Please review there.

PR #2752, opened 2026-09-08
fix(controller): honor kagent.dev/discovery=disabled for RemoteMCPServer and MCPServer tool discovery

teemow:
## Problem

A `RemoteMCPServer` that authenticates every caller has no credential to offer the controller. Agents forward the caller's token at run time (`KAGENT_PROPAGATE_TOKEN`), but the controller's own discovery request (`initialize` + `tools/list` for `status.discoveredTools` and the catalog) carries no bearer: the server answers 401 and the CR stays `Accepted=False (DiscoveryFailed: … Unauthorized)` for good, although every agent that references it works. Operators see a permanently red resource and anything reading the controller-side inventory sees an empty server.

`spec.headersFrom` is not a way out. Its values are resolved into each agent's rendered config, and the runtime applies those static headers last (`headerRoundTripper.RoundTrip` in `go/adk/pkg/mcp/registry.go`: "static headers … take precedence over all dynamic sources"), so a controller credential there would replace the propagated caller token in every agent.

`main` also honors the `kagent.dev/discovery=disabled` label nowhere for `MCPServer`: #2565 removed `predicates.DiscoveryDisabledPredicate` together with the old MCPServer tool controller, and the catalog reconciler restored in #2644 watches every kmcp `MCPServer` unfiltered. The opt-out the docs describe (label the `MCPServer` so kagent does not discover it while agentgateway fronts it) is a no-op on `main` today.

## Change

- The `kagent.dev/discovery=disabled` label now opts a `RemoteMCPServer` out of tool discovery. The reconciler skips the listing, publishes `Accepted=True` with reason `DiscoveryDisabled` and a message that names the label, clears `status.discoveredTools`, and keeps the server in the catalog with no tools and no `LastConnected`. Any other value, or no label, keeps discovery exactly as it is.
- The `RemoteMCPServer` controller reacts to label changes as well as generation changes (`predicate.Or(GenerationChangedPredicate, LabelChangedPredicate)`), so the opt-out and its removal apply at once instead of on the next 5-minute refresh.
- The `MCPServer` catalog reconciler honors the same label again, with the same projection: a labeled kmcp `MCPServer` is not dialed and stays in the catalog disconnected, with no tools. From the controller's point of view both kinds are discovery targets, so the opt-out looks the same on both.
- The `MCPServer` controller uses the same predicates: generation and label changes, plus a flip of the kmcp `Ready` condition (kmcp reports readiness through status without bumping the generation). The remaining status writes (observed generation, timestamps) no longer trigger a reconcile.
- The label key and value live in `consts` (`DiscoveryLabel`, `DiscoveryDisabled`).

## Testing

- `go test ./core/internal/controller/remotemcpserver/`: new `TestReconcileAcceptsWithoutDiscoveryWhenDisabled` asserts the discoverer is never called, the condition (`True` / `DiscoveryDisabled`), the cleared stale inventory, the catalog row without `LastConnected`, and the unchanged requeue.
- `go test ./core/internal/controller/mcpserver/`: new `TestReconcileKeepsDisconnectedCatalogWhenDiscoveryDisabled` asserts the discoverer is never called, the server row without `LastConnected` and no tools, and no requeue; `TestReconcileDiscoversWhenDiscoveryLabelIsNotDisabled` asserts any other label value keeps discovery on; `TestReadinessChangedPredicate` covers the readiness flips and the ignored status noise.
- `go build ./core/...`, `go vet` on the touched packages.

A backport to `release/v0.10.x` is #2753 (the 0.10 reconciler layout differs). The `MCPServer` part is `main`-only: 0.10.x still filters labeled `MCPServer`s through `predicates.DiscoveryDisabledPredicate`.

## Context

giantswarm/agent-platform#299 — the platform puts a shared MCP gateway (an OAuth resource server that evaluates the toolset per person) behind one `RemoteMCPServer`; every installation shows the red condition on it while all agents work. Happy to follow up with a docs PR to kagent-dev/website (`concepts/tools.md`) once the label semantics are agreed.


EItanya (commented):
Does this also need to be done for `McpServer` on `main`? You're defining the label const

teemow:
Yes, and thanks for the pointer. main currently honors the label nowhere, so the opt-out the docs describe for the agentgateway setup doesn't work. 

I've updated the PR.

go/core/internal/controller/remotemcpserver/reconciler.go
```diff
@@ -76,7 +82,12 @@ func New(client client.Client, discoverer ToolDiscoverer, catalog CatalogStore)
 
 func (r *Reconciler) SetupWithManager(manager ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(manager).
-		For(&v1alpha3.RemoteMCPServer{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		// A label change does not bump the generation, and the discovery opt-out
+		// is a label: watch both so the opt-out (and its removal) applies at once
+		// instead of on the next periodic refresh.
+		For(&v1alpha3.RemoteMCPServer{}, builder.WithPredicates(predicate.Or[client.Object](
+			predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{},
+		))).
```
EItanya, line 90: Do we want this for MCPServer as well, to ignore status changes?

go/core/internal/controller/mcpserver/reconciler.go
```diff
@@ -110,6 +111,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, r.deleteCatalog(ctx, request.String())
 	}
 
+	if discoveryDisabled(server) {
+		// The operator opted this server out of discovery (for example because
+		// agentgateway fronts it and agents reach it through a RemoteMCPServer):
+		// do not connect, and keep it out of the catalog altogether. This
+		// controller filters no events, so adding or removing the label re-enters
+		// Reconcile at once; no requeue is needed.
+		if err := r.deleteCatalog(ctx, request.String()); err != nil {
```
EItanya, line 120: Why does this one delete the catalog, but remotemcpserver just makes the catalog empty?

teemow: Sorry. I am new to kagents codebase.

I was thinking that MCPServer is only there to manage actual instances of mcp servers and RemoteMCPServer is actually there to populate the tools. But then why even reconcile MCPServer here. 

I saw #2565 #2639 #2644 

It looks like there was a bit of back and forth in the codebase. And I wasn't sure if AgentTemplate will support MCPServer or if it just came back because the UI currently needs it. Can you help me understand where this is going?

go/core/internal/controller/remotemcpserver/reconciler.go
teemow: Should this go into KMCP then?

EItanya: No no, kmcp is responsible for spinning up the MCP server deployments, kagent reads the tools but doesn't write anything to the status, it's slightly confusing. All I'm asking is if we should also use the same predicates for the mcpserver reconciler

go/core/internal/controller/mcpserver/reconciler.go
EItanya: They're slightly different. `MCPServer` is a kmcp resource that also manages the k8s lifecycle of the mcp server. kagent treats them as first class citizens and can also do discovery on them. `RemoteMCPServer` is any mcpserver which is running anywhere kagent can reach. So from the kagent controller perspective they're both just discovery targets

go/core/internal/controller/remotemcpserver/reconciler.go
teemow: Done in 9672c1e8: the MCPServer controller now filters on generation, label and a flip of kmcp's Ready condition; other status writes are ignored.

go/core/internal/controller/mcpserver/reconciler.go
teemow: Thanks, that settles it. Same projection for both now (9672c1e8): a labeled MCPServer stays in the catalog disconnected, with no tools.

EItanya (approved):

PR #2753, opened 2026-09-08
fix(controller): accept a RemoteMCPServer labeled kagent.dev/discovery=disabled without listing its tools (backport to release/v0.10.x)

teemow:
Backport of #2752 to `release/v0.10.x` (the 0.10 reconciler layout differs from `main`, so this is a separate patch of the same change).

## Problem

A `RemoteMCPServer` that authenticates every caller has no credential to offer the controller. Agents forward the caller's token at run time (`KAGENT_PROPAGATE_TOKEN`), but the controller's own discovery request (`initialize` + `tools/list`) carries no bearer: the server answers 401 and the CR stays `Accepted=False (ReconcileFailed: … Unauthorized)` for good, although every agent that references it works. `spec.headersFrom` is not a way out: its values are resolved into each agent's config and applied last by the runtime (`headerRoundTripper.RoundTrip`, `go/adk/pkg/mcp/registry.go`), so a controller credential there would replace the propagated caller token in every agent.

## Change

- `kagent.dev/discovery=disabled` — the label `predicates.DiscoveryDisabledPredicate` already reads to keep a Service or `MCPServer` out of auto-discovery — now also opts a `RemoteMCPServer` out of tool discovery: `ReconcileKagentRemoteMCPServer` skips the listing, stores the server in the database with no tools, and reports `Accepted=True` with reason `DiscoveryDisabled` (message names the label), clearing `status.discoveredTools`. Any other value, or no label, keeps discovery as it is.
- The TLS Secret hash is still computed and published with the `DiscoveryDisabled` condition (agents fold it into their rollout hash), and a broken `spec.tls` Secret reference still fails the server.
- The empty inventory is written once. The periodic refresh (60 s) re-enters the opt-out path, and the tool-snapshot cache from #2796 skips the Postgres write while the server and its (empty) tool list are unchanged; a spec change is a new snapshot and is written again.
- The label key and value move to `consts` (`DiscoveryLabel`, `DiscoveryDisabled`); the predicate uses them.
- The `RemoteMCPServer` controller reacts to label changes as well as generation changes (`predicate.Or(GenerationChangedPredicate, LabelChangedPredicate)`), so the opt-out and its removal apply at once instead of on the next periodic refresh.
- `reconcileRemoteMCPServerStatus` is split into the existing error→condition mapping and `setRemoteMCPServerStatus`, which takes the condition explicitly; the unchanged-status short-circuit is kept.

The `MCPServer` part of #2752 is not needed here: `release/v0.10.x` still filters labeled `MCPServer`s through `predicates.DiscoveryDisabledPredicate`, which `main` lost in #2565.

## Testing

- `go test ./core/internal/controller/reconciler/ -run DiscoveryDisabled`: `TestReconcileKagentRemoteMCPServer_DiscoveryDisabled` reconciles a labeled server with an unroutable URL through a recording `database.Client` that panics on anything but the two catalog writes — asserts the condition (`True` / `DiscoveryDisabled`), the cleared stale inventory, the stored server with no tools, and that a second reconcile of the unchanged server writes nothing while a description change is written again; `TestReconcileKagentRemoteMCPServer_DiscoveryDisabledKeepsSecretHash` covers the published hash and the broken-reference failure; `TestRemoteMCPServerDiscoveryDisabled` covers the label matching.
- `go build ./core/...`, `go vet ./core/internal/controller/...`.

Context: giantswarm/agent-platform#299.


go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -667,6 +668,25 @@ func (a *kagentReconciler) ReconcileKagentRemoteMCPServer(ctx context.Context, r
...
+			return fmt.Errorf("failed to store toolServer %s: %w", dbServer.Name, err)
+		}
+		if err := a.dbClient.RefreshToolsForServer(ctx, dbServer.Name, dbServer.GroupKind); err != nil {
+			return fmt.Errorf("failed to clear tools for toolServer %s: %w", dbServer.Name, err)
+		}
+		if err := a.setRemoteMCPServerStatus(ctx, server, nil, "", metav1.ConditionTrue, "DiscoveryDisabled", remoteMCPServerDiscoveryDisabledMessage); err != nil {
+			return fmt.Errorf("failed to reconcile remote mcp server status %s: %w", req.NamespacedName, err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
teemow: Good catch — fixed in 15109743: the TLS Secret hash is computed before the opt-out branch and published with the DiscoveryDisabled condition, and a broken spec.tls Secret reference still fails the server (both covered by TestReconcileKagentRemoteMCPServer_DiscoveryDisabledKeepsSecretHash).

go/core/internal/controller/reconciler/reconciler.go
```diff
@@ -676,6 +677,33 @@ func (a *kagentReconciler) ReconcileKagentRemoteMCPServer(ctx context.Context, r
...
+	if remoteMCPServerDiscoveryDisabled(server) {
+		// The operator opted this server out of discovery (a server that
+		// authenticates every caller has no credential to offer the controller).
+		// Accept it without connecting: the status carries no tools, the DB keeps
+		// the server with an empty inventory, and agents resolve the tool list at
+		// run time with the credentials they carry. The TLS Secret hash is still
+		// published (agents fold it into their rollout hash), and a broken
+		// spec.tls Secret reference still fails the server.
+		l.Info("skipping tool discovery for remote MCP server", "url", server.Spec.URL, "label", consts.DiscoveryLabel)
+		if _, err := a.dbClient.StoreToolServer(ctx, dbServer); err != nil {
```
mesutoezdil, line 689: wouldn't toolSnapshotUnchanged avoid a postgres write every 60s here?

teemow: Yes. Done in d836d6a4: the empty inventory is written once; later refreshes skip the write while the snapshot is unchanged.

PR #2767, opened 2026-09-09
fix(adk): send Host from headersFrom on the MCP request

### LINKED ISSUE #2733: RemoteMCPServer headersFrom: a "Host" header is silently dropped (headerRoundTripper writes it into the header map, net/http ignores it)
## Summary

`RemoteMCPServer.spec.headersFrom` lets you attach static headers to a remote MCP toolset. They flow into the Go ADK's `headerRoundTripper` (`go/adk/pkg/mcp/registry.go`), which applies them with `req.Header.Set(key, value)`:

```go
// Apply static headers last — they take precedence over all dynamic sources.
for key, value := range rt.headers {
    req.Header.Set(key, value)
}
```

Go's `net/http` **does not send a `Host` key from the header map** — the request host is taken from `req.Host` (or derived from the URL). So `headersFrom` with `name: Host` is silently a no-op on the wire.

## Reproduction

```go
package main

import (
	"fmt"
	"net/http"
	"net/http/httptest"
)

func main() {
	var seen string
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		seen = r.Host
	}))
	defer srv.Close()
	req, _ := http.NewRequest("GET", srv.URL, nil)
	req.Header.Set("Host", "my-virtual-host.example") // dropped by net/http
	resp, err := http.DefaultClient.Do(req)
	fmt.Println(resp.StatusCode, err, "server saw Host:", seen)
}
```

The server sees the listener's host, never `my-virtual-host.example`. Same shape as the `headerRoundTripper` loop above (verified empirically 2026-09-06 against the ADK transport).

## Impact

Host-routed upstreams are a legitimate MCP deployment pattern — e.g. a virtual-host gateway that routes `Host: <actor-id>.gateway.internal` to per-tenant MCP servers (the agent-sandbox substrate's router is one). For those, `headersFrom: [{name: Host, value: ...}]` is the natural CR expression, and today it silently fails: the request reaches the default backend instead of the routed one.

## Suggested fix

Special-case `Host` in each of the header-application loops (static headers at minimum; allowed/dynamic headers would be consistent):

```go
for key, value := range rt.headers {
    if strings.EqualFold(key, "Host") {
        req.Host = value
    } else {
        req.Header.Set(key, value)
    }
}
```

`req` is already a clone at that point, so mutating `req.Host` is safe.

## Environment

Observed with kagent 0.10.0-rc5 (vendored chart), and the same code shape is present on current `main` (`go/adk/pkg/mcp/registry.go`, `headerRoundTripper.RoundTrip`).

siddiqueirshad:
## Problem

`RemoteMCPServer.headersFrom` (and the other MCP header sources) can set a `Host` header so a virtual-host MCP gateway receives the intended hostname. On the Go runtime that never left the process: `headerRoundTripper` wrote it with `req.Header.Set`, and Go's `net/http` does not send a `Host` key from the header map. It uses `req.Host` (falling back to `URL.Host`). The configured value was silently dropped.

This is the bug reported in #2733. It is Go-only; Python uses a separate `x-kagent-host` path for gateway routing.

## Root cause

From Go's `net/http` docs: the Host header is taken from `Request.Host`, not from `Request.Header["Host"]`. Every header source in `headerRoundTripper.RoundTrip` used `Header.Set`, so `Host` from:

- static `headersFrom`
- `allowedHeaders`
- `headerProvider`
- `KAGENT_PROPAGATE_TOKEN`

never reached the wire.

## Change

Add `applyRequestHeader`, which sets `req.Host` when the key is `Host` (case-insensitive) and still writes the header map. All four header sources go through that helper.

This is orthogonal to open PR #2556 (static vs dynamic precedence in the same loops). After that PR lands, Host still needs this special case or it remains a silent no-op.

## Testing

- `TestHostHeader_IsSentAsRequestHost` — static `Host` / `host`, `allowedHeaders`, and dynamic `HOST` all land on `r.Host` of an `httptest` server
- `TestStaticHostHeader_OverridesDynamic` — static Host still wins over a dynamic value (current precedence)

```
go test -count=1 ./adk/pkg/mcp/
```

Passes (`ok … 3.080s`). `-race` is not runnable on this Windows host (`cgo` required); CI should still run it.

Fixes #2733

### DIFF
--- a/go/adk/pkg/mcp/registry.go
+++ b/go/adk/pkg/mcp/registry.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"strings"
 	"time"
 
 	"log/slog"
@@ -303,6 +304,15 @@ type headerRoundTripper struct {
 	headerProvider DynamicHeaderProvider
 }
 
+// applyRequestHeader sets an outbound header. Host must also be copied onto
+// req.Host because net/http ignores a Host key in the header map on the wire.
+func applyRequestHeader(req *http.Request, key, value string) {
+	if strings.EqualFold(key, "Host") {
+		req.Host = value
+	}
+	req.Header.Set(key, value)
+}
+
 func (rt *headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	req = req.Clone(req.Context())
 
@@ -312,7 +322,7 @@ func (rt *headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 		if callCtx, ok := a2asrv.CallContextFrom(req.Context()); ok {
 			if meta := callCtx.ServiceParams(); meta != nil {
 				if vals, ok := meta.Get(constants.AuthorizationHeader); ok && len(vals) > 0 && vals[0] != "" {
-					req.Header.Set(constants.AuthorizationHeader, vals[0])
+					applyRequestHeader(req, constants.AuthorizationHeader, vals[0])
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


mesutoezdil (commented):
models/base.go headerTransport has the same host bug

siddiqueirshad:
Thanks for the review. Addressed the requested change:

- Applied the same Host handling to `go/adk/pkg/models/base.go` `headerTransport`: `Host` headers are now also copied onto `req.Host` via `applyRequestHeader`, since `net/http` ignores a `Host` key in the header map.
- Added regression test `TestHeaderTransport_SendsHostAsRequestHost`.

Validation: `go test ./pkg/models/ -run TestHeaderTransport_SendsHostAsRequestHost` PASS; existing `go/adk/pkg/mcp` host tests PASS; `go vet` on both packages clean.

siddiqueirshad:
Hi @QuentinBisson, when you get a chance, could you please take a look at this PR? Thanks!

QuentinBisson:
@siddiqueirshad i'm sorry but i'm not a maintainer

PR #2770, opened 2026-09-09
fix: discover Kubernetes issuer for kubectl-ate

### LINKED ISSUE #2763: setup-cluster.sh and CI hardcode an ate-api issuer that does not match what kind advertises, breaking kubectl-ate
> **Disclosure:** this issue was written by an AI agent (Claude) while working on a cluster
> stood up by `scripts/setup-cluster/setup-cluster.sh`, and reviewed by a human before filing.

## Summary

Both `setup-cluster.sh` and the CI workflow write an `ate-api-authentication` ConfigMap with
`issuer: https://kubernetes.default.svc`, but a kind cluster advertises
`https://kubernetes.default.svc.cluster.local`. Because the issuer does not match, bearer
auth to ate-api fails and `kubectl-ate` cannot be used to inspect Actors or ActorTemplates.

## Evidence

On a cluster created by `setup-cluster.sh` (kagent `b8d53d39`, substrate 0.0.25, Kubernetes 1.35):

```
$ kubectl get --raw /.well-known/openid-configuration | jq -r .issuer
https://kubernetes.default.svc.cluster.local

$ kubectl get configmap ate-api-authentication -n ate-system \
    -o jsonpath='{.data.authentication\.yaml}' | grep issuer
  issuer: https://kubernetes.default.svc
```

Result:

```
$ kubectl-ate get actor-template ...
rpc error: code = Unauthenticated desc = invalid bearer token
```

## Where it is written

- `scripts/setup-cluster/setup-cluster.sh:62`
- `.github/workflows/ci.yaml:104`

Both use the same inline literal:

```
actorIdentityJWTProvider: kubernetes
jwtProviders:
- name: kubernetes
  issuer: https://kubernetes.default.svc
  audiences: [api.ate-system.svc]
  ...
```

## Impact

kagent itself is unaffected — the controller reaches ate-api over mTLS/SPIFFE, and
AgentInstances work normally. What breaks is the one clean way to inspect what a compiled
revision actually delivers to an Actor, which makes debugging substrate-backed behaviour
considerably harder than it needs to be.

## Suggested fix

Read the issuer from the cluster rather than hardcoding it, for example:

```bash
issuer="$(kubectl get --raw /.well-known/openid-configuration | jq -r .issuer)"
```

If the existing entry must stay as-is because actor identity depends on it, adding a second
`jwtProviders` entry carrying the `.cluster.local` issuer also resolves it without changing
the first.

yashrajshuklaaa:
Fixes the hardcoded Kubernetes issuer in setup/CI by discovering the issuer from the cluster’s OIDC config. 
This fixes kubectl-ate auth on kind.

Fixes https://github.com/kagent-dev/kagent/issues/2763

### DIFF
--- a/scripts/setup-cluster/setup-cluster.sh
+++ b/scripts/setup-cluster/setup-cluster.sh
@@ -61,7 +61,16 @@ kubectl create secret generic actor-id-ca-certs -n ate-system \
 
 k8s_issuer="$(kubectl get --raw /.well-known/openid-configuration | jq -r '.issuer')"
 kubectl create configmap ate-api-authentication -n ate-system \
-  --from-literal=authentication.yaml=$'actorIdentityJWTProvider: kubernetes\njwtProviders:\n- name: kubernetes\n  issuer: '"${k8s_issuer}"$'\n  audiences: [api.ate-system.svc]\n  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt\n  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token\n' \
+  --from-literal=authentication.yaml="$(cat <<EOF
+actorIdentityJWTProvider: kubernetes
+jwtProviders:
+- name: kubernetes
+  issuer: ${k8s_issuer}
+  audiences: [api.ate-system.svc]
+  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+EOF
+)" \
   --dry-run=client -o yaml | kubectl apply -f -
 
 helm upgrade substrate "oci://ghcr.io/kagent-dev/substrate/helm/substrate" \
--- a/scripts/setup-cluster/setup-cluster.sh
+++ b/scripts/setup-cluster/setup-cluster.sh
@@ -58,8 +58,10 @@ actor_id_ca_root="$(kubectl get secret actor-id-ca-pool -n ate-system -o jsonpat
   | openssl x509 -inform der -outform pem)"
 kubectl create secret generic actor-id-ca-certs -n ate-system \
   --from-literal=ca.crt="${actor_id_ca_root}" --dry-run=client -o yaml | kubectl apply -f -
+
+k8s_issuer="$(kubectl get --raw /.well-known/openid-configuration | jq -r '.issuer')"
 kubectl create configmap ate-api-authentication -n ate-system \
-  --from-literal=authentication.yaml=$'actorIdentityJWTProvider: kubernetes\njwtProviders:\n- name: kubernetes\n  issuer: https://kubernetes.default.svc\n  audiences: [api.ate-system.svc]\n  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt\n  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token\n' \
+  --from-literal=authentication.yaml=$'actorIdentityJWTProvider: kubernetes\njwtProviders:\n- name: kubernetes\n  issuer: '"${k8s_issuer}"$'\n  audiences: [api.ate-system.svc]\n  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt\n  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token\n' \
   --dry-run=client -o yaml | kubectl apply -f -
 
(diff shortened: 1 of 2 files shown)


supreme-gg-gg (commented):
One minor comment. I ran into this multiple times with `kubectl-ate` locally and fixed it similarly, so this should be the correct fix

scripts/setup-cluster/setup-cluster.sh
```diff
@@ -58,8 +58,10 @@ actor_id_ca_root="$(kubectl get secret actor-id-ca-pool -n ate-system -o jsonpat
   | openssl x509 -inform der -outform pem)"
 kubectl create secret generic actor-id-ca-certs -n ate-system \
   --from-literal=ca.crt="${actor_id_ca_root}" --dry-run=client -o yaml | kubectl apply -f -
+
+k8s_issuer="$(kubectl get --raw /.well-known/openid-configuration | jq -r '.issuer')"
 kubectl create configmap ate-api-authentication -n ate-system \
-  --from-literal=authentication.yaml=$'actorIdentityJWTProvider: kubernetes\njwtProviders:\n- name: kubernetes\n  issuer: https://kubernetes.default.svc\n  audiences: [api.ate-system.svc]\n  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt\n  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token\n' \
+  --from-literal=authentication.yaml=$'actorIdentityJWTProvider: kubernetes\njwtProviders:\n- name: kubernetes\n  issuer: '"${k8s_issuer}"$'\n  audiences: [api.ate-system.svc]\n  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt\n  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token\n' \
```
supreme-gg-gg: This syntax here is brittle, can we use something like heredoc instead to avoid trying to glue strings together with different quotes:

```bash
kubectl create configmap ate-api-authentication -n ate-system \
  --from-literal=authentication.yaml="$(cat <<EOF
actorIdentityJWTProvider: kubernetes
jwtProviders:
- name: kubernetes
  issuer: ${k8s_issuer}
  audiences: [api.ate-system.svc]
  certificateAuthorityFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
  discoveryTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
EOF
)"
```

yashrajshuklaaa: fixed :)

EItanya:
This makes sense if we're doing generic setup, but this is part of a CI setup that always installs into `kind`, so how would this come up?

yashrajshuklaaa:
@EItanya yeah, CI uses kind but that’s actually where the mismatch shows up. 
kind advertises https://kubernetes.default.svc.cluster.local while we were hardcoding https://kubernetes.default.svc. 
Reading the issuer from the cluster avoids making that assumption.

EItanya:
> @EItanya yeah, CI uses kind but that’s actually where the mismatch shows up. kind advertises https://kubernetes.default.svc.cluster.local while we were hardcoding https://kubernetes.default.svc. Reading the issuer from the cluster avoids making that assumption.

But then wouldn't it be failing?

yashrajshuklaaa:
> > @EItanya yeah, CI uses kind but that’s actually where the mismatch shows up. kind advertises https://kubernetes.default.svc.cluster.local while we were hardcoding https://kubernetes.default.svc. Reading the issuer from the cluster avoids making that assumption.
> 
> But then wouldn't it be failing?

@EItanya I checked and CI doesn't actually hit the code path this fixes. admin commands (make-ca-pool, make-jwt-pool) auth through the kubeconfig context directly, not through ate-api's JWT check. So the issuer in the configmap never gets validated during CI setup. mismatch only shows up on stuff like kubectl-ate get actor-template which sends a real bearer token to ate-api for validation. that's the exact command failing in #2763. CI just never runs anything like that so it stays green either way.

Could add a CI step that runs a kubectl-ate get after setup to catch this going forward, if that's useful. Can do it here or separate PR.

EItanya (approved):

EItanya:
CI isn't running on this PR for some reason and I can't approve without that. Can you either: 
1. remove the `ci.yaml` change. (It might only let maintainers edit workflows)
2. add a very small other change just to run CI

yashrajshuklaaa:
@EItanya Removed the ci.yaml change as suggested. PR now only has the setup-cluster.sh issuer discovery fix. 
workflow is waiting for approval to start the required checks.

EItanya (approved):

Issue #2709, opened 2026-09-04
[BUG] Substrate workers never report the actor they are holding


Charlesthebird:
On the substrate page in the UI, the associated actors in the workers lists aren't updating when I send a message, even though the actor shows that it is scheduled on a worker pod on the actors list. It sounds like this is an issue with the worker -> actor mapping in the API response.

---
*🤖 written by Claude (start)*

### 🎯 Affected Service(s)

Controller Service

### 🚦 Impact/Severity

Minor inconvenience

### 🐛 Bug Description

`GetSubstrateStatus` never sets `SubstrateWorker.actor_id`, `actor_template` or `actor_namespace`, so every worker reports holding nothing and the page reads that as an idle pod. The actors half of the same response is correct; only the worker → actor direction is missing.

### 🔄 Steps To Reproduce

1. On a cluster with ate-api configured, send a message so an actor is scheduled.
2. Open the Substrate page. **Actors** names the worker pod; **Workers** shows `idle` for that same pod.

### 🔍 Additional Context

Introduced in #2697 (`a9c209f6`), the substrate v0.0.25 update. That release removed the singular assignment from `WorkerStatus`, so the PR correctly deleted the lines reading it — but nothing replaced them:

```go
-	assignment := worker.GetStatus().GetAssignment()
-		ActorNamespace:  assignment.GetActorTemplateRef().GetAtespace(),
-		ActorTemplate:   assignment.GetActorTemplateRef().GetName(),
-		ActorID:         assignment.GetActor().GetName(),
```

Under v0.0.25 the binding lives only on the actor, as `ActorStatus.worker_assignment`, so the worker side has to be joined from the actor list — in the service, where both whole lists are in hand rather than in the UI, which only ever holds pages.

Worth settling: a v0.0.25 worker holds several actors and `SubstrateWorker` has room for one, so either the pick is made deterministic or `system.proto` gains a repeated field.

---
### 🧭 Fixing it

The binding lives only on the actor under v0.0.25 (`ActorStatus.worker_assignment`), so the worker side has to be joined from the actor list. That belongs in `system.Service.GetSubstrateStatus`, which holds both complete lists — done in the UI it would match only where two paged reads overlap, and the reads that reach a browser are pages.

Two decisions come with it, which is why this is not a UI change:

[...]

PR #2771, opened 2026-09-09
feat(api): page, sort and search the substrate reads server-side

### LINKED ISSUE #2704: [BUG] Substrate actor and worker lists are read whole, not paged, and cannot succeed at scale
*🤖 written by Claude*

---

### 🎯 Affected Service(s)

Multiple services — UI Service and Controller Service.

### 🚦 Impact/Severity

Blocker on a large cluster; no impact on a small one.

### 🐛 Bug Description

The substrate page's actor and worker lists are described, in the code and on
screen, as server-paged and server-ordered reads. Neither is true: every read
fetches the entire inventory and the browser pages and sorts it. On a cluster
holding hundreds of thousands of actors the read cannot succeed at all.

What each layer actually offers:

| Layer | Paging | Sort / filter |
| --- | --- | --- |
| ate-api `ListActors` | `page_size` + `page_token` | none — only `atespace` |
| `go/core/v2/substrate/list.go:12` | follows the token until it drains, returning every actor | — |
| `GetSubstrateStatus` (`proto/kagent/api/v1alpha1/system.proto:41`) | none — takes a namespace, answers with the whole inventory in one message | none |
| UI transport `localPage` (`ui/src/api/grpc/operations.ts`) | slices in the browser | sorts in the browser |

So ate-api pages, and the client deliberately drains that pagination into a
single slice, which `GetSubstrateStatus` then packs into one response.

### 🔄 Steps To Reproduce

1. Point the controller at an ate-api holding enough actors that one
   `GetSubstrateStatusResponse` exceeds the gRPC message limit — the figure
   recorded in `ui/src/pages/SubstratePage.tsx` is 410,110.
2. Open the Substrate page.
3. Watch the actors and workers lists fail with *"trying to send message larger
   than max"* — the paged look of the tables notwithstanding.
4. On a small cluster, where the read does succeed, click any actor or worker
   column header and watch the whole inventory be re-read to reorder rows the
   browser is already holding.

### 🤔 Expected Behavior

A page of actors is a page: one read returns one page in a requested order, and
paging or reordering costs one page rather than the whole inventory.

### 📱 Actual Behavior

1. **The read fails at scale.** `SubstratePage.tsx` records this response as one
   "gRPC refuses to send" at 410,110 actors. The page was split into three reads
   to solve exactly that, but all three call the same whole-inventory RPC, so the
   split is cosmetic.
2. **Every read walks the whole actor list** — roughly 1.6s at that size, per the
   note on `Timed` in `ui/src/api/domain/substrate.ts`.
3. **Reordering costs a full re-read.** The sort is part of the SWR key, so each
   header click re-fetches the entire inventory.

### 🔍 Additional Context

Found while reviewing #2666, which corrects the claims the UI made about this —
its copy and comments said the server sorted and paged these reads. This issue
is the API work behind them.

#### Proposed work

Add paged list RPCs to `SystemService` — `ListSubstrateActors` and
`ListSubstrateWorkers`, taking a namespace plus page size and page token — and
implement them in `go/core/internal/service/system` by passing the token through
to ate-api's own pagination rather than draining it. Counts stay on the summary
path. The UI already has the paged shape; its transport would drop `localPage`
for a real call, and its read key becomes correct as written.

#### The decision this needs first

ate-api exposes no sort or filter, only paging. So "sorted across the whole
inventory" and the list search cannot be honoured server-side without draining
every row anyway — which is the thing being removed. Someone has to choose
between:

- **order and filter within a page**, and say so on screen; or
- **keep whole-inventory ordering**, and accept the drain and its message-size
  ceiling; or
- **push sort and filter upstream into ate-api**, and depend on that landing.

That is a product call rather than a mechanical port, and it decides how much of
the UI work is worth doing.

#### What is not worth doing on its own

A UI-only change — moving the sort and slice out of the transport so a reorder is
a re-render — removes the redundant re-reads and nothing else. The remaining read
is still the whole inventory, and the change would be rewritten by the API work
above. Noted so it is not picked up as the fix.

toreysoloio:
workers table with enough entries to trigger pagination
image

---
*🤖 written by Claude (start)*

## Changelog

Substrate actors and workers load on clusters too large for a single inventory read: each list is paged, sorted and searched server-side, across the whole inventory.

Closes #2704

## Testing

1. `cd ui && VITE_API_MODE=mock yarn dev`, open `/substrate`.
2. Actors and Workers each show one page of 25; Previous and Next turn real pages.
3. Click a column header — the read runs again and the note under the table says which order the server applied.
4. Search `7f21` in Actors: the heading counts every match in the scope, not just the ones on this page.
5. Search for something absent — the empty state says the whole scope was searched.

## Additional Notes

ate-api offers paging and nothing else — no order, no filter, no total — so the controller reads every one of its pages, applies both, then cuts the page it answers with. That costs a walk of the inventory per request, which is the price of a sort and a search that mean the cluster rather than the rows on screen; only a page and some integers cross the wire, which is what `GetSubstrateStatus` could not do at 410,110 actors.

The Workers table loses its Actor column (#2709): ate-api's `Worker` carries no actor reference, so the column read `idle` on every real cluster. How much of the fleet is busy is a tile instead.

Removing the walk needs an upstream change: ate-api stores each row as an opaque `proto bytea` blob, so status, template and pod are invisible to SQL. Promoting them to columns — a migration plus a Go backfill, since Postgres cannot decode the blob — would make the filter a `WHERE`, the sort an `ORDER BY` and the totals a `count(*)`, and the API here already has that shape, so the controller would stop walking and pass the parameters through. Two smaller ones stand alone: exposing the actor reference on `Worker` needs no migration, since the indexed `worker_assignments` table already holds it, and would restore the column above; and a `GetSubstrateActor(id)` call would remove the last caller needing a whole-inventory read, letting `GetSubstrateStatus` be deprecated.

---
*🤖 written by Claude (end)*


Charlesthebird:
~Got feedback on this: we should move some of this sorting + pagination logic to the database layer if possible~
[edit]: since this is Substrate live data, we can't do this - needs to be read live.

EItanya (approved):

PR #2772, opened 2026-09-09
fix: flush buffered OTEL logs

### LINKED ISSUE #2759: ForceFlush never flushes the LoggerProvider, so GenAI audit records are silently dropped when the Actor checkpoints
> **Disclosure:** this issue was written by an AI agent (Claude) while verifying kagent's
> audit-logging documentation, and reviewed by a human before filing.

## Summary

`ForceFlush` flushes the TracerProvider only. The LoggerProvider's batch processor is never
flushed, so GenAI audit log records can be lost when Substrate checkpoints the Actor as the
response closes. The loss is silent — no error, no retry, and the records are dropped rather
than deferred to the next resume.

For a feature whose audience is security and compliance review, silent loss is a
correctness problem rather than a caveat.

## The code

```go
// go/adk/pkg/telemetry/tracing.go:50-61
func ForceFlush(ctx context.Context) {
	type flusher interface{ ForceFlush(context.Context) error }
	fp, ok := otel.GetTracerProvider().(flusher)     // <-- TracerProvider only
	if !ok {
		return
	}
	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
	defer cancel()
	if err := fp.ForceFlush(flushCtx); err != nil {
		otel.Handle(err)
	}
}
```

`newLoggerProvider` wraps its exporter in `sdklog.NewBatchProcessor`, so log records sit in
a batch queue with the same hazard the `ForceFlush` doc comment describes for spans.
`grep -rn 'GetLoggerProvider' go` returns **zero hits** in the repository.

Spans have an opt-in pre-response flush and logs have no equivalent:

- `go/core/internal/translator/kagent/compiler.go:75` sets `KAGENT_PRE_RESPONSE_TRACE_FLUSH=true`
- `go/adk/pkg/a2a/server/server.go:110` honours it
- `grep -rn 'PRE_RESPONSE' go` shows no logging counterpart

## Observed

On a kind cluster with logging enabled via `Harness.spec.env`, two identical single-turn
invokes of the same AgentInstance with the same prompt produced **0** GenAI log records and
then **3**. A second invoke did not release the first turn's missing records, so they were
dropped rather than deferred.

Later in the same session, with a 3-replica WorkerPool, delivery was reliable across four
consecutive invokes. So the loss is load- and timing-dependent rather than constant, which
is consistent with a batch-interval race.

## Expected

`ForceFlush` also flushes the LoggerProvider, or a `KAGENT_PRE_RESPONSE_LOG_FLUSH`-style
opt-in exists that matches the trace behaviour, so a record emitted during a turn is
exported before the Actor is checkpointed.

## Suggested fix

Cast `otel.GetLoggerProvider()` alongside the tracer provider in `ForceFlush` and flush
both, reusing the existing timeout.

yashrajshuklaaa:
## What

Fix `ForceFlush()` so it flushes both the tracer and logger providers.

## Why

GenAI audit logs are buffered by the OTEL batch log processor. When an Agent Substrate actor checkpoints right after the A2A response closes, those buffered logs can be left behind and never exported.

## Changes

* Flush the global `LoggerProvider` in `ForceFlush()`.
* Keep tracer flushing working as before.
* Use the same timeout for both providers.
* Added a regression test to make sure buffered logs are exported, even when the request context is already cancelled.
* Updated the related comment from spans to telemetry.

## Validation

* `go test ./adk/pkg/telemetry ./adk/pkg/a2a/server `
* `go build ./adk/pkg/telemetry ./adk/pkg/a2a/server`

All passed.

Fixes #2759

### DIFF
--- a/go/adk/pkg/telemetry/tracing.go
+++ b/go/adk/pkg/telemetry/tracing.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	logglobal "go.opentelemetry.io/otel/log/global"
 	"go.opentelemetry.io/otel/propagation"
 	sdklog "go.opentelemetry.io/otel/sdk/log"
 	"go.opentelemetry.io/otel/sdk/resource"
@@ -39,24 +40,36 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
 	return otel.Tracer("gcp.vertex.agent").Start(ctx, "invocation")
 }
 
-// ForceFlush exports any spans still buffered in the tracer provider's batch
-// processor. Call it before an A2A response completes when the process may be
-// suspended right afterwards: Agent Substrate checkpoints the actor as soon as
-// the response body closes, so unexported spans stay frozen in the snapshot
-// until the session's next resume (or forever, for a session's last message).
+// ForceFlush exports any telemetry still buffered in the tracer and logger
+// providers' batch processors. Call it before an A2A response completes when
+// the process may be suspended right afterwards: Agent Substrate checkpoints
+// the actor as soon as the response body closes, so unexported telemetry stays
+// frozen in the snapshot until the session's next resume (or forever, for a
+// session's last message).
 // Uses its own detached timeout because the request context is typically
 // already canceled by the time deferred cleanup runs. The timeout defaults to
 // 3s and is configurable via KAGENT_TRACE_FLUSH_TIMEOUT_MS.
 func ForceFlush(ctx context.Context) {
 	type flusher interface{ ForceFlush(context.Context) error }
-	fp, ok := otel.GetTracerProvider().(flusher)
-	if !ok {
+
+	tracerProvider, tracerOK := otel.GetTracerProvider().(flusher)
+	loggerProvider, loggerOK := logglobal.GetLoggerProvider().(flusher)
+	if !tracerOK && !loggerOK {
 		return
 	}
+
 	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
 	defer cancel()
-	if err := fp.ForceFlush(flushCtx); err != nil {
-		otel.Handle(err)
+
+	if tracerOK {
+		if err := tracerProvider.ForceFlush(flushCtx); err != nil {
+			otel.Handle(err)
+		}
+	}
+	if loggerOK {
+		if err := loggerProvider.ForceFlush(flushCtx); err != nil {
+			otel.Handle(err)
+		}
 	}
 }
 
--- a/go/adk/pkg/telemetry/tracing.go
+++ b/go/adk/pkg/telemetry/tracing.go
@@ -48,35 +48,35 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
 // session's last message).
 // Uses its own detached timeout because the request context is typically
 // already canceled by the time deferred cleanup runs. The timeout defaults to
-// 3s and is configurable via KAGENT_TRACE_FLUSH_TIMEOUT_MS.
+// 3s and is configurable via KAGENT_TELEMETRY_FLUSH_TIMEOUT_MS.
 func ForceFlush(ctx context.Context) {
-	type flusher interface{ ForceFlush(context.Context) error }
+        type flusher interface{ ForceFlush(context.Context) error }
 
-	tracerProvider, tracerOK := otel.GetTracerProvider().(flusher)
-	loggerProvider, loggerOK := logglobal.GetLoggerProvider().(flusher)
-	if !tracerOK && !loggerOK {
-		return
-	}
+        tracerProvider, tracerOK := otel.GetTracerProvider().(flusher)
+        loggerProvider, loggerOK := logglobal.GetLoggerProvider().(flusher)
+        if !tracerOK && !loggerOK {
+                return
+        }
 
-	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
-	defer cancel()
+        flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
+        defer cancel()
 
-	if tracerOK {
-		if err := tracerProvider.ForceFlush(flushCtx); err != nil {
-			otel.Handle(err)
-		}
-	}
-	if loggerOK {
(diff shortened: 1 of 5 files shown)


EItanya:
@krisztianfekete

go/adk/pkg/telemetry/tracing.go
```diff
@@ -39,24 +40,36 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
...
-	fp, ok := otel.GetTracerProvider().(flusher)
-	if !ok {
+
+	tracerProvider, tracerOK := otel.GetTracerProvider().(flusher)
+	loggerProvider, loggerOK := logglobal.GetLoggerProvider().(flusher)
+	if !tracerOK && !loggerOK {
 		return
 	}
+
 	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
```
krisztianfekete: Can we give each provider its own budget, or flush logs first?

go/adk/pkg/telemetry/tracing.go
```diff
@@ -39,24 +40,36 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
...
-// until the session's next resume (or forever, for a session's last message).
+// ForceFlush exports any telemetry still buffered in the tracer and logger
+// providers' batch processors. Call it before an A2A response completes when
+// the process may be suspended right afterwards: Agent Substrate checkpoints
+// the actor as soon as the response body closes, so unexported telemetry stays
+// frozen in the snapshot until the session's next resume (or forever, for a
+// session's last message).
 // Uses its own detached timeout because the request context is typically
 // already canceled by the time deferred cleanup runs. The timeout defaults to
 // 3s and is configurable via KAGENT_TRACE_FLUSH_TIMEOUT_MS.
```
krisztianfekete: We probably want to rename this now that it covers both traces and logs.

krisztianfekete (changes requested):
Thanks for the PR! Would you mind also fixing this on the Python side, address the comments and add test coverage for both runtimes and both signals?

yashrajshuklaaa:
@krisztianfekete Thanks for the review. I’ve addressed the comments and updated the PR. fix now covers both Go and Python, flushing logs first while keeping a shared timeout budget. also added test coverage for both logs and traces in both runtimes and renamed the timeout setting to KAGENT_TELEMETRY_FLUSH_TIMEOUT_MS
Validation is passing locally and the changes are pushed to the PR.

go/adk/pkg/telemetry/tracing.go
```diff
@@ -39,31 +40,43 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
...
-	fp, ok := otel.GetTracerProvider().(flusher)
-	if !ok {
-		return
-	}
-	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
-	defer cancel()
-	if err := fp.ForceFlush(flushCtx); err != nil {
-		otel.Handle(err)
-	}
+        type flusher interface{ ForceFlush(context.Context) error }
```
mesutoezdil, line 53: gofmt flags this: tabs, not spaces

go/adk/pkg/telemetry/tracing.go
```diff
@@ -39,31 +40,43 @@ func StartInvocationSpan(ctx context.Context) (context.Context, trace.Span) {
...
+// providers' batch processors. Call it before an A2A response completes when
+// the process may be suspended right afterwards: Agent Substrate checkpoints
+// the actor as soon as the response body closes, so unexported telemetry stays
+// frozen in the snapshot until the session's next resume (or forever, for a
+// session's last message).
 // Uses its own detached timeout because the request context is typically
 // already canceled by the time deferred cleanup runs. The timeout defaults to
-// 3s and is configurable via KAGENT_TRACE_FLUSH_TIMEOUT_MS.
+// 3s and is configurable via KAGENT_TELEMETRY_FLUSH_TIMEOUT_MS.
 func ForceFlush(ctx context.Context) {
```
mesutoezdil, line 52: #2805 rewrites this to tracer-only flush, dropping your fix

python/packages/kagent-core/tests/test_tracing_configure.py
```diff
@@ -245,28 +296,28 @@ def test_force_flush_calls_provider_force_flush(monkeypatch):
...
     calls = []
-    provider = SimpleNamespace(force_flush=lambda timeout: calls.append(timeout))
-    monkeypatch.setattr(_utils.trace, "get_tracer_provider", lambda: provider)
-    monkeypatch.setenv("KAGENT_TRACE_FLUSH_TIMEOUT_MS", raw)
-
-    _utils.force_flush()
-
-    assert calls == [expected]
 
+    logger_provider = SimpleNamespace(
```
mesutoezdil, line 300: ruff wants this joined to one line

yashrajshuklaaa:
@mesutoezdil resolved all comments :) PTAL

yashrajshuklaaa:
@krisztianfekete pls review this

python/packages/kagent-core/src/kagent/core/tracing/_utils.py
krisztianfekete, line 313: This is nested under `if tracing_enabled`, so with `OTEL_LOGGING_ENABLED=true` and tracing off the new log flush is never called. Can you move out of the tracing branch?

go/pkg/tracing/tracing.go
```diff
@@ -89,20 +91,43 @@ func environmentValue(name, fallback string) string {
...
+	loggerProvider, loggerOK := logglobal.GetLoggerProvider().(flusher)
+	if !tracerOK && !loggerOK {
 		return nil
 	}
-	flushCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), flushTimeout())
+
+	flushCtx, cancel := context.WithTimeout(
+		context.WithoutCancel(ctx),
+		flushTimeout(),
+	)
```
krisztianfekete, line 109: Both implementations share this one deadline, while Python splits the budget with `remaining_millis`, can you please match the two?

go/pkg/tracing/tracing.go
```diff
@@ -89,20 +91,43 @@ func environmentValue(name, fallback string) string {
...
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.Join(errs...)
 }
 
 func flushTimeout() time.Duration {
-	if value := strings.TrimSpace(os.Getenv("KAGENT_TRACE_FLUSH_TIMEOUT_MS")); value != "" {
+	if value := strings.TrimSpace(os.Getenv("KAGENT_TELEMETRY_FLUSH_TIMEOUT_MS")); value != "" {
```
krisztianfekete, line 130: The rename drops any existing `KAGENT_TRACE_FLUSH_TIMEOUT_MS` override to the 3s default with no warning. Can we keep the old name as a fallback for a release and log a notice for the rename?

PR #2775, opened 2026-09-09
feat(ui): checkpoint a chat turn, fork from a checkpoint, and duplicate a chat

Charlesthebird:
---
*🤖 written by Claude (start)*

Checkpoint a turn in a chat, fork from a checkpoint, duplicate a chat, and the agent rail work that came out of using it.

The Snapshots page and the `ListCheckpoints` change it needs are stacked on top, in #2798.

## Checkpoints and forks

- The Checkpoint button beside Send saves the conversation's current turn boundary. It belongs on the composer rather than a message because `CreateCheckpoint` takes no cutoff — what it saves is always "as it stands now" — and it is disabled during a turn and once the latest boundary is saved.
- A saved boundary is a line under the turn it was taken at, with `Fork` on it. The line is the promise: everything above it travels into a fork and nothing below it does. Matched by `headTaskId` against the turn each message carries, so the lines survive a reload.
- `Duplicate chat` in each rail row's menu is a checkpoint of the conversation plus a fork of it, and the copy opens.
- A fork refetches its own transcript rather than inheriting the page's, so it never shows its source's boundaries as its own.

## The agent rail

Rows gained the room and the states they were missing: a menu on every row carrying Chat details and Share chat as well as rename, duplicate and delete; checkbox hover and press states; controls that line up whether or not the list scrolls. The three columns fold as the window narrows — the agent panel at 1440, the application sidebar at antd's `xl`, the rail at 1040 — and unfold again. The agent panel's drag handle is gone with `ResizableAside`: its wrapper clipped to a fixed width, so dragging widened an aside inside a box that never grew.

The create buttons on the list pages are title-cased, which is unrelated to the rest and was asked for separately.

## Testing

1. `ENABLE_MOCK_UI=true yarn dev`, open a conversation, and press Checkpoint beside Send. A line appears under the last turn and is still there after a reload.
2. Press `Fork` on an early line. The fork opens holding what was above it and nothing below.
3. Open a rail row's menu and choose `Duplicate chat`. The copy opens with the whole transcript.
4. Narrow the window from wide: the agent panel folds, then the sidebar, then the rail.

The reel in the comments shows this against a real cluster: two checkpoints three turns apart, a fork from each, and the earlier fork knowing only what was above its line. Reels about the Snapshots page moved with it to #2798 and are marked outdated here.

---
*🤖 written by Claude (end)*


Charlesthebird:
Checkpoint, fork and duplicate against a real cluster — five turns, three checkpoints, two forks from different points, and two duplicates that keep their own favourite city.

https://github.com/user-attachments/assets/5df70aec-06db-475c-b6d8-6991251896f4

---
*🤖 written by Claude*

Charlesthebird:
Checkpoint, fork and duplicate against a real cluster — five turns, three checkpoints, two forks from different points, and two duplicates that keep their own favourite city.

https://github.com/user-attachments/assets/76b52fde-5ecf-49ea-a6db-f4c90a05752b

---
*🤖 written by Claude*

Charlesthebird:
Checkpoint, fork and duplicate against a real cluster — five turns, three checkpoint lines that survive a reload, two forks from different lines, and two duplicates that keep their own favourite city.

https://github.com/user-attachments/assets/9dcd7c2a-7b97-4d4a-8465-adad31c0d24f

---
*🤖 written by Claude*

Charlesthebird:
Checkpoint, fork and duplicate against a real cluster — five turns, three checkpoint lines that survive a reload, two forks from different lines, and two duplicates that keep their own favourite city.

https://github.com/user-attachments/assets/252bbca0-ca85-4a02-a30b-29646fbbeead

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** Superseded by the checkpoint-and-fork reel further down this PR.

<details>
<summary>The original comment</summary>

Everything this branch adds, against a real cluster: checkpoints taken from the composer and surviving a reload, forks from two different lines, two duplicates that keep their own favourite city, the agent rail's menus and checkboxes, and the three columns folding as the window narrows.

https://github.com/user-attachments/assets/65e3604e-689f-472b-82c2-f9feb4f27934

</details>

---
*🤖 written by Claude*

Charlesthebird:
Snapshots, against a real cluster: three checkpoints taken in a chat, the new page listing every one across all conversations, and three deleted together — one request each, sequential, under one toast.

https://github.com/user-attachments/assets/ed4b53fe-5ed5-44f3-8e20-84d2ef46cf72

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** Superseded — the listing moved to the controller, in #2798.

<details>
<summary>The original comment</summary>

The Snapshots page here reads every checkpoint into the browser and searches, sorts and pages there. That is deliberate — `ListCheckpoints` requires an `agent_instance_id`, so there is no way to ask the controller for a caller's checkpoints at all — but it costs one request per conversation.

#2783 is the follow-up: an RPC that does pagination, search and multi-column sort server-side. It needs a proto change, so it is out of scope here.

</details>

---
*🤖 written by Claude*

Charlesthebird (commented):
> [!NOTE]
> **Outdated.** This reviewed the branch before the Snapshots page and its listing moved to #2798. The findings still hold, but half of them are against files this PR no longer touches.

<details>
<summary>The original review</summary>

---

I wrote this branch, so this is a review of my own work: everything below was found, fixed and pushed before posting. Nothing here is outstanding.

The one thing a reviewer should weigh rather than take from me is the **Snapshots page reading every checkpoint into the browser**. It is a request per conversation, it is deliberate, and the reasoning is in the description and in #2783. If that trade is wrong, the page should not merge in this shape.

<details><summary>What I fixed after review (12)</summary>

| Level | Where | What was wrong |
| --- | --- | --- |
| 🔴 High | `ChatComposer.tsx` | Checkpoint was gated on the streaming phase alone, so it was live between Send and the first token — where `CreateCheckpoint` refuses with `FailedPrecondition`. Gated on the whole active turn. |
| 🟠 Medium | `ChatTranscript.tsx` | Each group was a fragment keyed by its first message. The grouping changes when the boundaries land, so two groups becoming one retired a key and React remounted the messages under it — every rendered mermaid diagram flashed back to loading. One flat run keyed by message id. |
| 🟠 Medium | `chat.spec.ts` | I had converted a strict assertion into a 20s poll and blamed the checkpoint read. The real cause was the remount above; the poll was hiding it. Strict again. |
| 🟠 Medium | `AgentChatPage.tsx` | A checkpoint that saved but whose list re-read failed reported `Could not checkpoint` — over a line already on screen. The re-read is outside the try. |
| 🟠 Medium | `grpc/operations.ts` | `ListCheckpoints` read one page and dropped `nextPageToken`; the controller defaults to 50. Past that the oldest boundaries silently stopped drawing lines. It follows the token. |
| 🟠 Medium | `CheckpointDivider.tsx` | `role="separator"` puts its children on ARIA's presentational list, which pruned the label and the only fork control in the application. A labelled group instead. |
| 🟠 Medium | `mocks/state.ts` | Fixture checkpoints lived in a module array while transcripts persist in `sessionStorage`, so a checkpoint vanished on reload and the messages it marked did not. The fixture contradicted the controller. Persisted. |
| 🟠 Medium | `mocks/state.ts` | A second seeded boundary shared a `headTaskId` with the first, so one line was drawn for two and its Fork silently targeted one of them. Different turn. |
| 🟠 Medium | `AgentChatPage.tsx` | A fork inherited its source's marks: they are keyed by message id and a fork gets copies of those messages under the same ids. Now held against the conversation they belong to. Guarded by a test I watched fail without the fix. |
| 🟡 Low | `ChatComposer.tsx` | The tooltip sat on a disabled antd button, which fires no pointer events — invisible in the one state where it explains itself. |
| 🟡 Low | `domain/checkpoints.ts` | `canForkFrom` was exported and never called. Deleted. |
| 🟡 Low | rail | Focus rings sliced flat on three controls, each by a different clipping ancestor. |

</details>

- **Fork truncation, on a cluster.** Forking turn 1 of a five-turn chat gives a chat that knows only the first fact; forking turn 4 knows all four. Not inferred from the SQL — driven in the browser and read back from the agent.
- **Duplicates diverge.** Two copies of one fork, given different favourite cities, each keep their own after two further messages.
- **Marks survive a reload.** Three lines saved, reloaded, still three, and the button still refuses a fourth at the same boundary.
- **A fork inherits nothing.** Reproduced the leak first (fork showed 1 line with 0 checkpoints), then fixed it, then reverted the fix to watch the new test fail.
- **Deleting releases the snapshot.** `Create` calls `CreateTag`, `Delete` calls `DeleteTag`; nothing else does. Verified against `service.go`.
- **Both themes, every new control.** Photographed at rest, hover, press and focus.
- **The collapse order.** Measured at seven widths in both directions: agent panel 1440, sidebar `xl`, rail 1040.
- **Control alignment under both scrollbar modes**, including `--disable-features=OverlayScrollbar`, which is the only way the misalignment reproduces.
- **CI green**, including the lint error I had been misreading locally — I was treating "0 errors and 1 warning potentially fixable" as the total.

</details>

## Worth a reviewer's judgement

- **The fan-out.** One request per conversation to draw Snapshots. #2783 has the RPC it needs.
- **`NotFound` for a forked snapshot.** The controller refuses to delete one a chat was forked from and reports it as not found. The page states the condition; correcting the error is a backend change.
- **Toasts sit over the composer.** `position="bottom-right"` puts them on Send and Checkpoint, so a click is intercepted while one shows. Pre-existing chrome, but this branch made it easy to hit.
- **Repeated forks are all named `X (fork)`.** Distinguishable only by age in the rail.

---
*🤖 written by Claude*

</details>

Charlesthebird:
> [!NOTE]
> **Outdated.** Superseded — #2783 is done in #2798.

<details>
<summary>The original comment</summary>

See the last 2 videos for the demos of the new features, and this follow-up that is needed for performance: https://github.com/kagent-dev/kagent/issues/2783

</details>

---
*🤖 written by Claude*

Charlesthebird:
Snapshots page against a real cluster, with the search, the sort and the count all coming from `ListCheckpoints`.

https://github.com/user-attachments/assets/3941b4e1-123b-4673-abbc-ef3ebc57d9d9

- Typing narrows the rows and the total together.
- Clicking `Conversation` re-asks for the page in that order.
- Deleting takes rows out and the total down with them.

---
*🤖 written by Claude*

Charlesthebird:
Snapshots page against a real cluster, 32 snapshots over two pages.

https://github.com/user-attachments/assets/5e63caca-e64a-40fe-a453-d88a483f25d5

- Search `otter` narrows 32 to 2, count and rows together.
- Next and back move between `1–25 of 32` and `26–32 of 32`.
- Sorted A–Z, page 2 picks up where page 1 stopped.
- A row's own delete button takes the total to 31.

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** The Snapshots page moved to #2798.

<details>
<summary>The original comment</summary>

Snapshots page against a real cluster, 31 snapshots over two pages.

https://github.com/user-attachments/assets/10d0fd5d-e3b6-44cb-a528-906d0c833a8f

- Search `otter` narrows 31 to 2, count and rows together.
- Next and back move between `1–25 of 31` and `26–31 of 31`.
- Sorted A–Z, page 2 picks up where page 1 stopped.
- A row's own delete button takes the total to 30.

Turning a page, sorting, and a five-letter search are one `ListCheckpoints` request each. The conversations are read once per visit — two requests for 70 of them — and only decide which names link to a chat that still exists.

</details>

---
*🤖 written by Claude*

Charlesthebird:
> [!NOTE]
> **Outdated.** The Snapshots page moved to #2798.

<details>
<summary>The original comment</summary>

What the review changed, against a real cluster with 34 snapshots.

https://github.com/user-attachments/assets/a34ba52b-a4d5-40ec-a9f8-2f99c0a9c414

- Searching `50%` finds the one row that contains it, not the 11 the old `ILIKE` matched on the wildcard. Same for `user_id`.
- State has no sorter: a listing is READY snapshots only, so it ordered every row by the same value.
- `?page=99` lands back on `1–25 of 34`, and the stale page number leaves the address bar.

The index that would keep this off a sequential scan is not here — migration files are immutable once merged, so it needs its own. Tracked in #2783.

</details>

---
*🤖 written by Claude*

Charlesthebird:
Injection payloads, and the view living in the address bar.

https://github.com/user-attachments/assets/7298efa9-f5c9-4a08-9a1c-ce3530b6d8b6

- `'; DROP TABLE agent_instance_checkpoint; --`, `' OR '1'='1` and `%' OR 1=1 --` each match no snapshot name and run nothing. The 34 rows are still there afterwards.
- `?q=e&sort=conversation&dir=desc&page=2` opens already searched, sorted and on page 2, and survives a reload.
- Clicking the same controls writes that address back.

Every value reaches Postgres as a bind parameter — the statement is one literal with `$1`–`$7` and pgx prepares it, so nothing is interpolated. The sort column is an enum on the wire and is whitelisted against a map before the query runs; the filter is `position(lower($3) IN …)`, so it has no pattern syntax to escape.

---
*🤖 written by Claude*

Charlesthebird:
The Snapshots page moved to #2798, stacked on this branch.

This PR is now the chat side only: checkpoint a turn, fork from a checkpoint, duplicate a chat, and the agent rail work. No proto, no Go — 30 files, all under `ui/`.

#2798 carries the page itself, the `ListCheckpoints` change it needs to page, search and sort across conversations, and the store and service work behind that. It closes #2783.

Merge this one first; GitHub retargets #2798 to `main` once it lands.

The reels above are marked outdated for that reason — the ones about snapshots belong with the other PR now.

---
*🤖 written by Claude*

Charlesthebird:
Checkpoint and fork against a real cluster, with a real agent answering.

https://github.com/user-attachments/assets/ee96c05b-25fc-4416-a244-a9074af5d473

One conversation, two checkpoints three turns apart, then a fork from each:

- After `my favorite color is teal`, Checkpoint. Two more turns — the otter, Kyoto — then Checkpoint again. Two lines at different depths.
- Fork the **first** line: two messages, and asked what it knows it says **teal**. The otter and Kyoto were below the line.
- Fork the **second** line: six messages, and it lists **teal, the otter and Kyoto**.
- Fork the first line again: another two-message copy. A boundary is not spent by forking it.

Neither fork shows the source's lines as its own — a fork reads its own boundaries.

---
*🤖 written by Claude*

peterj (approved):

PR #2776, opened 2026-09-09
fix(ui): keep the transcript pinned while it settles, and wait for the app to boot

### LINKED ISSUE #2773: Two Playwright specs flake under parallel load: filters after reload, and the chat scroll-to-bottom button
Two specs in `ui/playwright/tests/` fail under `fullyParallel` load and pass when run on their own. Both assert state the app restores asynchronously just after a mount or a reload, and neither waits for that restoration — so the assertion's own 5s window is what decides the result.

## 1. `lists/list-filters.spec.ts` — filters not yet re-applied after a reload

`lists: a narrowed view is an address, so it survives a reload`, step 3, on Chromium:

```
Error: expect(locator).toHaveValue(expected) failed
Locator:  getByTestId('models-filters-search')
Expected: "config"
Received: ""
Timeout:  5000ms
  - waiting for getByTestId('models-filters-search')
    13 × locator resolved to <input value="" type="text" data-testid="models-filters-search" …/>
```

The input is present and empty for the whole 5s. The address still carries `q=config`, so this is the control not having been populated from the query yet rather than the parameter being lost.

## 2. `chat/chat.spec.ts` — scroll-to-bottom button present at the foot

`chat: leaving the foot of a conversation offers a way back to it`, step 1, on Firefox:

```
Error: expect(locator).toHaveCount(expected) failed
Locator:  getByTestId('chat-scroll-bottom')
Expected: 0
Received: 1
Timeout:  5000ms
  - waiting for getByTestId('chat-scroll-bottom')
    14 × locator resolved to 1 element
       - unexpected value "1"
```

The spec opens a conversation and immediately asserts the button is absent, on the premise that a freshly opened transcript is scrolled to its foot. The button is there for the whole window, which suggests the intersection observer that decides "at the foot" has not yet reported for the first time — the transcript is still settling its height while the assertion runs.

## Reproducing

Neither reproduces reliably on an idle machine. Both did on a machine under heavy load, which is what makes them worth reporting rather than dismissing: the same conditions exist on a busy CI runner.

- `yarn test:pw` with the machine otherwise loaded: both fail.
- Each spec alone (`-g` on its title): 3/3 pass.
- Platform: macOS. Chromium for the first, Firefox for the second.

## Why one issue

The mechanism looks shared. Both assert a *derived* piece of UI state at the earliest possible moment — a filter control populated from the query string, and a scroll affordance driven by an observer — with a plain `expect` and no wait for the thing that produces it. A fix for one is likely the shape of the fix for the other: wait on the state the app actually settles into, rather than on a default timeout being long enough.

Happy to send a PR if you agree with the approach.

---
*🤖 written by Claude*

Charlesthebird:
Fixes e2e UI test flake

Closes #2773.

---
*🤖 written by Claude (start)*

Two Playwright specs failed only under parallel load, and each had a real cause rather than a timeout that needed raising.

- **The chat scroll-to-bottom button could strand a reader at the foot.** Content landing below the fold moves the foot away without anybody scrolling, and a scroll event arriving mid-growth read that as the reader having left. That unset the pin, and the resize observer only re-pins while the pin holds — so the button appeared and stayed. Growth no longer unpins; only the reader scrolling up does.
- **`expectSettled` waited for a spinner to be absent**, which is also true of a page that has not started rendering. Straight after a `goto` or a `reload` it returned at once, leaving the next assertion's own five seconds to boot the app. It now waits for the app to be on screen first.

### Testing

1. `cd ui && yarn test:pw`.
2. Step 4 of `chat: leaving the foot of a conversation offers a way back to it` covers the transcript bug — it fails on `main` and passes here.
3. For `lists: a narrowed view is an address`, throttle the CPU (devtools → Performance → 20×) before step 3's reload: on `main` the step spends its whole budget waiting for an app that has not booted.

---
*🤖 written by Claude (end)*

### DIFF
--- a/ui/playwright/helpers/app.ts
+++ b/ui/playwright/helpers/app.ts
@@ -156,7 +156,17 @@ export function dataRows(page: Page): Locator {
   return page.locator("tbody tr.ant-table-row");
 }
 
-/** Resolves once no loading indicator is left on the page. */
+/** A navigation-sized budget, for the app booting rather than for what it rendered. */
+const APP_BOOT_TIMEOUT = 15_000;
+
+/**
+ * Resolves once the app is on screen and no loading indicator is left on it.
+ *
+ * Waiting for a spinner to be absent is also true of a page that has not started
+ * rendering, so after a `goto` or a `reload` this returned at once and the next
+ * assertion's own five seconds became the app's boot budget.
+ */
 export async function expectSettled(page: Page): Promise<void> {
+  await expect(page.locator("#root")).not.toBeEmpty({ timeout: APP_BOOT_TIMEOUT });
   await expect(page.locator(".ant-spin-spinning")).toHaveCount(0);
 }
--- a/ui/playwright/tests/chat/chat.spec.ts
+++ b/ui/playwright/tests/chat/chat.spec.ts
@@ -344,6 +344,20 @@ test("chat: leaving the foot of a conversation offers a way back to it", async (
      */
     await expect(button).toHaveCount(0, { timeout: 10_000 });
   });
+
+  await test.step("4. and a transcript still growing does not take it away", async () => {
+    /*
+     * Growth landing after the pin, with the scroll event the pin queued behind it.
+     * Dispatched rather than waited for: on an idle machine the two land in the same
+     * frame, which is why this only ever failed on a loaded one.
+     */
+    await box.evaluate((node) => {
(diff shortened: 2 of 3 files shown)


Charlesthebird:
https://github.com/user-attachments/assets/832f8537-e6e2-4ae9-b5fa-04b06978c854



---
*🤖 written by Claude*

EItanya (approved):

PR #2777, opened 2026-09-09
fix(ui): light an agent rail conversation only on the page it links to

Charlesthebird:
## Description

`ChatEntry`'s active state in the agent rail was `candidate.id === ref.id` — an id match with no reference to the route. That is true on every surface which mounts the rail for an instance, and one of those is the agent's own details page.

The row links to that conversation's **chat**, so on `/agents/:id` the rail highlights a row for a page you are not on. Two entries in the rail can look current at the same time.

## Reproducing

1. Open any agent conversation, then use **Agent Details** in the rail.
2. On `/agents/:id`, that conversation's row in the rail is still highlighted.
3. Its href is `/agents/:id/chat` — a different page from the one you are on.

## The fix

`isActive` is compared against the row's own href, which is how the nav entries above it already decide the same thing:

```diff
-                href={url.chat({ id: candidate.id })}
-                isActive={candidate.id === ref.id}
+                href={href}
+                isActive={location.pathname === href}
```

`href` is hoisted out of the JSX so the comparison and the link cannot drift apart. The reads beneath it keep matching on the id on purpose — which conversation's live state to prefer is a question about the instance, not about the route.

The row also gained `aria-current="page"` when active, as `RailEntry` already has. Without it the highlight was the only thing indicating the current page.

## Testing

`playwright/tests/chat/agent-rail.spec.ts` gains one case, asserting from the details page — the chat is the one place the old behaviour happened to be right. It checks the row is listed but not current there, that nothing in the rail claims to be current, and that opening the conversation is what makes it so.

- `yarn typecheck` — clean
- `yarn lint` — no errors
- `npx playwright test playwright/tests/chat/ --project=chromium` — 31 passed

---
*🤖 written by Claude*

### DIFF
--- a/ui/playwright/tests/chat/agent-rail.spec.ts
+++ b/ui/playwright/tests/chat/agent-rail.spec.ts
@@ -654,3 +654,42 @@ test("agent rail: a conversation can be renamed from inside it, two ways", async
     ).toContainText("Named from the rail");
   });
 });
+
+test("agent rail: only the page you are on is marked as current", async ({ page }) => {
+  /*
+   * A conversation row used to be lit by id alone, so it claimed to be the current
+   * page on every surface that mounts the rail for an instance — the agent's own
+   * details page included, which is a different page from the chat the row links to.
+   * Two rows then carried `aria-current="page"` at once, which is both wrong on its
+   * face and wrong for a screen reader.
+   *
+   * Asserted from the details page rather than the chat, because the chat is the one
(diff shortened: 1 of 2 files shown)


Charlesthebird:
Folded into #2778, which fixes both halves of the rail's active state together — this one and the contributed-entry matcher. Closing in favour of that.

---
*🤖 written by Claude*

PR #2781, opened 2026-09-09
fix: retain fork checkpoints with direct history ancestry

EItanya:
Forks omit inherited checkpoints because listing filters only by the instance that created them. A fork at C2 now lists C1 and C2 alongside its own checkpoints, excluding source-history boundaries after C2 and preserving each checkpoint's original provenance. Listing by a deleted instance's ID returns the same complete checkpoint lineage.

Store immutable parent-history and cutoff fields on `agent_history`, renamed from `a2a_context` to distinguish a history branch from its public A2A context ID. Retain a unique originating instance ID on history so listing can start there directly, without a local-checkpoint fallback. Composite foreign keys enforce instance/history identity and ownership and matching parent/child owner and context. The fork transaction validates that the cutoff identifies an event in the parent history with the expected task/snapshot boundary; failure rolls back the new history and instance. This avoids a circular history/event foreign-key dependency without adding a table. Schema changes remain in migration 1's pre-release baseline.

Forks continue copying bounded events and rebuilding private task projections at terminal task boundaries. The instance's source-checkpoint reference retains fork-request identity and runtime provenance, including on deletion tombstones. Conversation reads remain local; checkpoint listing traverses retained history ancestry and bounds candidates per history before combining the checkpoint-ID page.

Retained histories protect inherited checkpoints and their snapshot Tags even after all related instances are deleted; there is currently no history garbage collection. Future event GC must preserve inherited boundaries and coordinate with fork creation. Fork creation and deletion checks share a history lock. Deletions already in progress can finish or retry. Deterministic PostgreSQL tests cover both operation orders, for both the fork's source checkpoint and an earlier checkpoint inherited by a later fork.

The protobuf contract documents that `Checkpoint.agent_instance_id` remains the originating instance ID. The UI allows viewing and forking inherited checkpoints, while showing rename/delete only for checkpoints created in the current instance, in both the transcript and details modal. Mock listings now retain the inherited prefix and original provenance. The obsolete `(source_instance_id, id)` listing index is removed; the remaining source-instance lookups use the existing CREATING partial index.

Validation:

- PostgreSQL database suite and migration Up/Down tests passed with the race detector, including every inline SQL prepare check, owner/context constraints, and pagination/ownership after instance deletion. Store tests also verify missing/wrong-history cutoffs and mismatched task boundaries roll back the fork, then permit a successful retry after repair.
- Go lint reported zero issues. Protobuf generation, lint, and compatibility checks passed. UI type checking, the ten focused unit tests, and UI lint passed (zero errors; existing warnings only). The twelve checkpoint browser tests passed in Chromium and Firefox using the mock backend, including inherited action gating and local checkpoint controls.
- Backend live E2E passed on the previous head `af7db928`, including checkpoint lineage after deletion and both runtime-retention scenarios ([CI run](https://github.com/kagent-dev/kagent/actions/runs/35600363146/job/106334757083)). That head also passed the relevant service race tests. Live E2E for this UI/proto/index follow-up is pending CI.
- Earlier SQL benchmarks covered one history with 100,000 checkpoints (42.5 ms median), 1,024 histories with 100 checkpoints each (71.7 ms), and 1,024 histories with one checkpoint each (7.0 ms). These are warm PostgreSQL execution timings from `619cad47`, before removing the duplicate local scan; they exclude API overhead. Query/index optimization experiments remain separate from this PR.
- The migration-immutability check intentionally fails because migration 1 is the pre-release baseline; the guard is unchanged.

### DIFF
--- a/go/core/internal/database/checkpoints.go
+++ b/go/core/internal/database/checkpoints.go
@@ -57,8 +57,10 @@ func (c *Client) ForkAgentInstance(ctx context.Context, checkpointID, userID, re
 		if err != nil {
 			return fmt.Errorf("get checkpoint runtime revision: %w", err)
 		}
+		// Serialize new lineage with deletion of any checkpoint in its inherited prefix.
+		// NO KEY UPDATE leaves foreign-key checks free to read the history.
 		sourceContextID, err := queryOne(ctx, tx, `
-			SELECT context_id FROM a2a_context WHERE id = $1
+			SELECT context_id FROM a2a_context WHERE id = $1 FOR NO KEY UPDATE
 		`, pgx.RowTo[uuid.UUID], checkpoint.SourceHistoryID)
 		if err != nil {
 			return fmt.Errorf("get checkpoint context: %w", err)
@@ -384,15 +386,28 @@ func checkpointSnapshot(row agentInstanceCheckpointRow) *AgentInstanceTaskSnapsh
 	}
 }
 
-// ListAgentInstanceCheckpoints returns an owner's READY checkpoints for the source
-// instance in ascending ID order after afterID, up to limit. Retained checkpoints remain
-// listable after the source instance is deleted.
+// ListAgentInstanceCheckpoints returns an owner's READY checkpoints in the instance's
+// history and inherited prefixes, in ascending ID order after afterID, up to limit.
+// Checkpoint fields retain their source provenance. Locally created checkpoints remain
+// listable after the instance is deleted.
 func (c *Client) ListAgentInstanceCheckpoints(ctx context.Context, instanceID, userID, afterID string, limit int) ([]*apiv1alpha1.Checkpoint, error) {
 	rows, err := queryMany(ctx, c.db, `
+		WITH RECURSIVE lineage (history_id, boundary) AS (
+		    SELECT history_id, NULL::bigint FROM agent_instance WHERE id = $1 AND user_id = $2
+		    UNION
+		    SELECT c.source_history_id, c.history_sequence
+		    FROM lineage l
+		    JOIN a2a_context h ON h.id = l.history_id
+		    JOIN agent_instance_checkpoint c ON c.id = h.source_checkpoint_id
+		    WHERE h.user_id = $2 AND c.user_id = $2
+		)
 		SELECT id, source_instance_id, user_id, request_id, head_task_id, history_sequence, snapshot_atespace,
 		    snapshot_uri, snapshot_content_scope, tag_uid, state, data, source_history_id, prepared_revision,
-		    source_name FROM agent_instance_checkpoint
-		WHERE source_instance_id = $1
+		    source_name FROM agent_instance_checkpoint c
+		WHERE (source_instance_id = $1 OR EXISTS (
+		    SELECT 1 FROM lineage l WHERE l.history_id = c.source_history_id
+		      AND (l.boundary IS NULL OR c.history_sequence <= l.boundary)
+		))
 		  AND user_id = $2
 		  AND state = 'READY'
 		  AND (NULLIF($3::text, '') IS NULL OR id > NULLIF($3::text, '')::uuid)
@@ -417,8 +432,9 @@ func (c *Client) ListAgentInstanceCheckpoints(ctx context.Context, instanceID, u
 
 // BeginDeleteAgentInstanceCheckpoint marks an owned READY checkpoint DELETING and returns
 // the snapshot and tag identity for external cleanup. Retrying a DELETING checkpoint is
-// allowed. A fork reference, another lifecycle state, or a missing/unowned checkpoint
-// returns ErrNotFound.
+// allowed even if new forks have since inherited the boundary. An inherited history
+// reference blocks starting deletion; another state or a missing/unowned checkpoint
+// also returns ErrNotFound.
 func (c *Client) BeginDeleteAgentInstanceCheckpoint(ctx context.Context, id, userID string) (*AgentInstanceTaskSnapshot, string, error) {
 	var snapshot *AgentInstanceTaskSnapshot
 	var tagUID string
@@ -426,6 +442,12 @@ func (c *Client) BeginDeleteAgentInstanceCheckpoint(ctx context.Context, id, use
 		row, err := lockCheckpoint(ctx, tx, id, false, userID, nil)
 		if err != nil {
 			return notFoundOr(err)
+		}
+		// Take the same history lock as ForkAgentInstance before checking its descendants.
+		if _, err := queryOne(ctx, tx, `
+			SELECT id FROM a2a_context WHERE id = $1 FOR NO KEY UPDATE
+		`, pgx.RowTo[uuid.UUID], row.SourceHistoryID); err != nil {
+			return fmt.Errorf("lock checkpoint history: %w", err)
 		}
 		checkpoint, err := toAgentInstanceCheckpoint(row)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


iplay88keys:
Overall, I think this approach makes sense given the planned follow-up to stop copying event history. Through the investigation I did on this, Codex helped come up with a few areas where we can likely simplify and harden this PR:

- Retain the originating instance ID on `agent_history` so deleted instances can still resolve their full lineage. This would remove the local-checkpoint fallback and its partial results after deletion.
- Enforce that parent and child histories share the same owner/context and that the cutoff identifies a real parent-history event.
- Validate the listing query against a single history with many checkpoints and deeply nested forks, since the current synthetic test distributes checkpoints across many histories.
- Add a deterministic concurrency test for fork versus checkpoint deletion.

EItanya:
@iplay88keys addressed the lineage and concurrency points:

- `agent_history` now retains a unique originating `instance_id`. Listing starts there directly, so deleting an instance preserves its complete local/inherited checkpoint listing and removes the fallback. Tests cover pagination and ownership after deletion, including a fork with no local checkpoints.
- Composite foreign keys enforce matching parent/child owner and context, and bind the instance to its history's instance ID, owner, and context. After further design review, cutoff validation stays in the fork transaction instead of a reverse history-to-event foreign key. Tests verify missing/wrong-history cutoffs and mismatched task boundaries reject the fork, leave no durable instance/history, and allow the same request to succeed after repair. Future event GC must preserve inherited boundaries and coordinate with forks.
- Added deterministic fork/deletion tests for both operation orders, covering the same checkpoint and deletion of an earlier checkpoint while forking a later one. They wait for an actual PostgreSQL lock wait and verify checkpoint visibility, retention, and cleanup retries.

We also already benchmarked the dense-history/deep-chain cases before this update:

| Histories visited | Checkpoints per history | Median SQL time | p95 |
| --- | --- | --- | --- |
| 1 | 100,000 | 42.5 ms | 43.7 ms |
| 1,024 | 100 | 71.7 ms | 75.2 ms |
| 1,024 | 1 | 7.0 ms | 7.4 ms |

PostgreSQL 18.6, 2-CPU/4-GiB container, warm cache, 30 measured executions after three warmups, generic prepared plans, 51 rows fetched for a 50-checkpoint page. Depth includes the current history. The fixture used synthetic 256-byte checkpoint payloads and checked expected IDs; these timings exclude API/protobuf overhead and do not measure cold storage, concurrent writes, or bloat. They predate this update's removal of the duplicate local scan. Separate index/pagination experiments improved dense cases, but those optimizations are not included here.

Database/migration and relevant service tests passed with `-race`; E2E compilation and Go lint passed. Added live E2E coverage for listing deleted forks; live execution has not been verified for the latest update. Migration changes remain in the pre-release baseline, with the immutability guard unchanged.

go/core/internal/database/checkpoints.go
```diff
@@ -392,20 +397,37 @@ func checkpointSnapshot(row agentInstanceCheckpointRow) *AgentInstanceTaskSnapsh
 	}
 }
 
-// ListAgentInstanceCheckpoints returns an owner's READY checkpoints for the source
-// instance in ascending ID order after afterID, up to limit. Retained checkpoints remain
-// listable after the source instance is deleted.
+// ListAgentInstanceCheckpoints returns an owner's READY checkpoints in the instance's
+// history and inherited prefixes, in ascending ID order after afterID, up to limit.
+// Checkpoint fields retain their source provenance. The complete lineage remains
```
iplay88keys, line 402: We should also document this on the proto. `agent_instance_id` now means the checkpoint's originating instance which could differ from the instance being listed.

go/core/internal/database/checkpoints.go
```diff
@@ -425,8 +447,9 @@ func (c *Client) ListAgentInstanceCheckpoints(ctx context.Context, instanceID, u
 
 // BeginDeleteAgentInstanceCheckpoint marks an owned READY checkpoint DELETING and returns
 // the snapshot and tag identity for external cleanup. Retrying a DELETING checkpoint is
-// allowed. A fork reference, another lifecycle state, or a missing/unowned checkpoint
-// returns ErrNotFound.
+// allowed even if new forks have since inherited the boundary. An inherited history
+// reference blocks starting deletion; another state or a missing/unowned checkpoint
```
iplay88keys, line 451: Since inherited checkpoints are now returned by List, the UI exposes rename and delete for them. Delete will fail while rename updates the originating checkpoint globally. Those actions should probably be gated using something like checkpoint.agent_instance_id == current instance ID.

go/core/pkg/migrations/core/000001_initial.sql
```diff
@@ -86,14 +101,16 @@ CREATE TABLE agent_instance_checkpoint (
...
     data                   BYTEA       NOT NULL,
-    source_history_id      UUID        NOT NULL REFERENCES a2a_context(id) ON DELETE RESTRICT,
+    source_history_id      UUID        NOT NULL REFERENCES agent_history(id) ON DELETE RESTRICT,
     prepared_revision      TEXT        REFERENCES runtime_revision(revision) ON DELETE RESTRICT,
     CHECK (snapshot_content_scope IN ('FULL', 'DATA')),
     CHECK (state IN ('CREATING', 'READY', 'FAILED', 'DELETING')),
     UNIQUE (user_id, request_id)
 );
 CREATE INDEX agent_instance_checkpoint_list_idx
     ON agent_instance_checkpoint (source_instance_id, id);
```
iplay88keys: This seems unnecessary now that the checkpoint listing no longer filters by `source_instance_id`. The remaining lookups appear to be for `state = 'CREATING'`, which the partial index below already covers:

```
CREATE UNIQUE INDEX agent_instance_checkpoint_one_creating_idx
    ON agent_instance_checkpoint (source_instance_id)
    WHERE state = 'CREATING';
```

go/core/internal/database/checkpoints.go
EItanya: Addressed in 5c703e05. The proto now documents the originating instance ID on `Checkpoint.agent_instance_id` and the local/inherited semantics of `ListCheckpoints`, including listing after deletion. Regenerated the Go and TypeScript artifacts; protobuf lint, compatibility, and generation checks passed.

go/core/internal/database/checkpoints.go
EItanya: Addressed in 5c703e05. Both the transcript controls and snapshot details hide rename/delete unless the checkpoint's originating instance matches the current instance. Viewing and forking remain available. Updated the mock backend to return inherited prefixes with original provenance, and replaced the old zero-inherited-mark expectation. All 12 checkpoint browser tests passed in Chromium and Firefox, including inherited action gating and controls on a new local checkpoint.

go/core/pkg/migrations/core/000001_initial.sql
EItanya: Removed `agent_instance_checkpoint_list_idx` in 5c703e05. The remaining source-instance filters require `state = 'CREATING'` and retain the existing partial index. The PostgreSQL suite, inline-statement prepare checks, and migration Up/Down tests passed with the race detector.

go/core/internal/database/checkpoints.go
```diff
@@ -449,9 +478,11 @@ func (c *Client) BeginDeleteAgentInstanceCheckpoint(ctx context.Context, id, use
 			SET state = 'DELETING', data = $3
 			WHERE agent_instance_checkpoint.id = $1 AND agent_instance_checkpoint.user_id = $2
 			  AND agent_instance_checkpoint.state IN ('READY', 'DELETING')
-			  AND NOT EXISTS (
-			      SELECT 1 FROM agent_instance i WHERE i.pinned_checkpoint_id = agent_instance_checkpoint.id
-			  )
+			  AND (agent_instance_checkpoint.state = 'DELETING' OR NOT EXISTS (
```
mesutoezdil, line 481: dropped the pinned-live-instance guard here, tag gets deleted before the fk catches it?

go/core/pkg/migrations/core/000001_initial.sql
```diff
@@ -65,13 +65,28 @@ CREATE TABLE agent_template_harness_pair (
...
     ON agent_template_harness_pair (namespace, agent_template_name, harness_name);
 
-CREATE TABLE a2a_context (
-    id         UUID        PRIMARY KEY,
-    user_id    TEXT        NOT NULL CHECK (user_id <> ''),
-    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    context_id UUID        NOT NULL,
-    CONSTRAINT a2a_context_binding_key UNIQUE (id, context_id)
+-- Each row owns one history branch; forks preserve the public A2A context_id.
+CREATE TABLE agent_history (
```
mesutoezdil, line 69: new migration instead of editing 000001 directly?

PR #2784, opened 2026-09-10
fix(adk): reject inputResponses that do not match the current input request

### LINKED ISSUE #2761: tasks/update returns success and discards the answer when inputResponses uses an unexpected key
> **Disclosure:** this issue was written by an AI agent (Claude) while documenting the
> human-in-the-loop flow over MCP, and reviewed by a human before filing.

## Summary

`tasks/update` answers a paused task by looking up `params.InputResponses` under a key
derived from the task. On a key miss it returns `{"resultType": "complete"}` having done
nothing. The client sees success, the answer is discarded, and the task stays
`input_required` with no indication anything went wrong.

## The code

```go
// go/core/internal/mcp/tasks.go:171-175
key := inputRequestKey(task)
inputResponse := params.InputResponses[key]
if inputResponse == nil {
	return &completeTaskResult{ResultType: "complete"}, nil    // <-- silent no-op
}
```

Note the contrast two lines further down: a value of the *wrong type* under the right key
is a proper error.

```go
response, ok := inputResponse.(*mcp.ElicitResult)
if !ok || response == nil {
	return nil, invalidParams(fmt.Errorf("input response %q must be an elicitation result", key))
}
```

So a malformed answer is reported and a misaddressed answer is not.

The expected key is not guessable from the elicitation alone in every case:

```go
// go/core/internal/mcp/tasks.go:337-342
func inputRequestKey(task *a2atype.Task) string {
	if task.Status.Message != nil && task.Status.Message.ID != "" {
		return task.Status.Message.ID
	}
	return string(task.ID)
}
```

## Reproduction

1. Drive an agent to pause with `ask_user` (a prompt instructing the agent to use its
   `ask_user` tool reliably produces one; it is registered unconditionally at
   `go/adk/pkg/agent/agent.go:205`).
2. `tasks/get` the paused task and read `inputRequests`, a map keyed by request ID.
3. Call `tasks/update` with `inputResponses` keyed by something else, for example
   `not-a-real-key`, carrying a valid ElicitResult.

Observed: the reply is `{"resultType": "complete"}`, and a follow-up `tasks/get` shows the
task still `input_required`. No error is returned and nothing indicates the answer was
dropped. Verified on a kind cluster against controller `v0.10.0-rc1-112-g30a8ffe4`.

## Expected

An `inputResponses` map that contains no entry for the task's current input request is
rejected with `invalidParams`, naming the expected key, in the same way a wrong-typed value
already is. A success envelope should not be returned for a write that did not happen.

## Note

`resultType` is about the envelope rather than the run, so a client cannot use it to detect
this either: a task paused on a question reports `"resultType": "complete"` alongside
`"status": "input_required"`. The only reliable signal is `status`, which makes the silent
success especially easy to mistake for a completed answer.

wangzhengzhuo05:
## Description
`tasks/update` answers a paused task by looking up `params.InputResponses` under a key derived from the task. When the map contains no entry for the tasks current input request, the handler returned `{"resultType": "complete"}` after doing nothing: the client saw success, the answer was silently discarded, and the task stayed `input_required` with no error signal.

This change makes a key-miss reject the request with an `invalidParams` error that names the expected key — the same handling the wrong-typed-value branch already applies — instead of returning a success envelope for a write that did not happen.

## Why
A success envelope must not be returned for an answer that was not applied. The only reliable signal today is the task `status`, which made it easy to mistake a dropped answer for a completed one.

## How
`go/core/internal/mcp/tasks.go` — in `updateTask`, the `inputResponse == nil` branch now returns `invalidParams(fmt.Errorf("inputResponses has no entry for the current input request %q", key))` instead of `&completeTaskResult{ResultType: "complete"}`. No other path changes: the not-input-required early return still returns `complete` (correctly, since there is nothing to answer), and the wrong-type branch is untouched.

## Tests
Added `TestTaskUpdateRejectsMismatchedInputResponsesKey` in `go/core/internal/mcp/tasks_test.go`. It drives `updateTask` with an `InputResponses` map keyed by a non-matching key and asserts:
- the error is non-nil and is a `jsonrpc.Error` with `Code == CodeInvalidParams`,
- the message names the expected key,
- no message is dispatched (`gateway.reply == nil`).

`go test ./core/internal/mcp/` passes.

## Mutation check
Reverting only the fixed branch (restoring the silent `{resultType:"complete"}` return) makes the new test fail with `updateTask() succeeded with mismatched inputResponses key`, confirming the test genuinely guards the fix.

Fixes #2761

### DIFF
--- a/go/core/internal/mcp/tasks.go
+++ b/go/core/internal/mcp/tasks.go
@@ -171,7 +171,7 @@ func (h *Handler) updateTask(ctx context.Context, _ *mcp.ServerSession, params *
 	key := inputRequestKey(task)
 	inputResponse := params.InputResponses[key]
 	if inputResponse == nil {
-		return &completeTaskResult{ResultType: "complete"}, nil
+		return nil, invalidParams(fmt.Errorf("inputResponses has no entry for the current input request %q", key))
 	}
 	response, ok := inputResponse.(*mcp.ElicitResult)
 	if !ok || response == nil {
--- a/go/core/internal/mcp/tasks_test.go
+++ b/go/core/internal/mcp/tasks_test.go
@@ -146,6 +146,43 @@ func TestTaskUpdateContinuesA2ATask(t *testing.T) {
 	}
 }
 
+func TestTaskUpdateRejectsMismatchedInputResponsesKey(t *testing.T) {
+	message := a2atype.NewMessage(a2atype.MessageRoleAgent, a2atype.NewTextPart("Which database?"))
+	gateway := &fakeGateway{task: &a2atype.Task{
+		ID: testTaskID, ContextID: testInstanceID,
+		Status: a2atype.TaskStatus{State: a2atype.TaskStateInputRequired, Message: message},
+	}}
+	h := &Handler{gateway: gateway}
+	ref, err := encodeTaskReference(taskReference{
+		InstanceID: testInstanceID, TaskID: testTaskID,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = h.updateTask(authContext(), nil, &updateTaskParams{
+		ParamsBase: taskParamsBase(),
+		TaskID:     ref,
+		InputResponses: mcp.InputResponseMap{"not-a-real-key": &mcp.ElicitResult{
+			Action: "accept", Content: map[string]any{"response": "PostgreSQL"},
+		}},
+	})
+	if err == nil {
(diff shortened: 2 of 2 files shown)


wangzhengzhuo05:
Closing this to keep the tracker clean \u2014 it has been sitting without review for several days. This is no reflection on the change itself: the fix and its local verification still stand. If the issue is still present and you would like the patch revisited, say the word and I will rebase it onto current main and re-open. Thanks for your time.

Issue #2728, opened 2026-09-07
Go declarative runtime silently ignores `spec.declarative.context.compaction` while being the CRD default


davidkarlsen:
## Summary

As of 0.10.0 the Agent CRD defaults `spec.declarative.runtime` to `go`, so any agent that omits the field runs on the Go ADK. The Go runtime does not implement `spec.declarative.context.compaction` — for long-running tool-using agents that field is what bounds session history growth. An operator upgrading to 0.10.0 without an explicit `runtime` therefore loses compaction, and the only signal is a status condition that no UI or printer column surfaces.

## Evidence (all at tag `v0.10.0`, commit `ae008d1e`)

1. **`go` is the CRD default** — `kagent.dev_agents.yaml:13236` `default: go` (same at `kagent.dev_sandboxagents.yaml:10893`). Note `go/api/v1alpha2/agent_types.go:325-336` (`EffectiveDeclarativeRuntime`) still documents and implements a Python fallback, which is unreachable for objects created through the API.

2. **Compaction is unimplemented on Go** — `go/core/internal/controller/reconciler/reconciler.go:946-948`:

   ```go
   // Context compression: Not yet implemented in Go runtime
   if decl.Context != nil && decl.Context.Compaction != nil {
       unsupported = append(unsupported, "context compression/compaction (not implemented in Go runtime)")
   }
   ```

   `grep -rn Compaction go/adk` returns **zero** functional hits, while the controller compiles the config regardless (`translator/agent/compiler.go:265-302`) — the field is delivered to the runtime and dropped.

3. **The gap is soft-failed** — `reconciler.go:341-350` sets `UnsupportedFeatures=True`; the agent still reconciles Accepted and serves traffic, just without compaction. Good that a signal exists; it is invisible in the UI and in `kubectl get agents`.

## Scope

1. **Implement `context.compaction` in the Go runtime.** The upstream primitives are not in place yet — in `google/adk-go`, [#298 "ADR-010: Native SDK Session History Compaction"](https://github.com/google/adk-go/issues/298) (design) and [#1001 "Add EventCompaction support to EventActions and compaction filtering in contents_processor.go"](https://github.com/google/adk-go/issues/1001) (implementation) are both **open** as of 2026-09-07; see also [discussion #589](https://github.com/google/adk-go/discussions/589). Either consume that work when it lands, or implement session-history compaction kagent-side for the Go runtime.

[...]

Issue #1866, opened 2026-05-14
[FEATURE] Add prompt caching toggle for Anthropic / Bedrock models in ModelConfig


kimsehwan96:
### 📋 Prerequisites

- [x] I have searched the [existing issues](./issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)

### 📝 Feature Summary

Allow users to opt into Anthropic prompt caching (cache_control markers) on a per-ModelConfig basis.

### ❓ Problem Statement / Motivation

When kagent talks to Anthropic (Direct API or Bedrock-Claude), the agent loop accumulates conversation history and re-sends system messages + tools on every turn. Anthropic's prompt caching offers a 90% read discount (25% write surcharge), but it requires the caller to mark cacheable blocks explicitly `cache_control: {type: "ephemeral"}`. kagent's model adapters don't inject these markers anywhere system, tools, or history so caching never engages.

OpenAI handles caching automatically (prefix-based, no markers required), so kagent users on OpenAI already benefit. Anthropic / Bedrock-Claude users effectively pay full price on every turn.

### 💡 Proposed Solution

```yaml
  spec:
    provider: Anthropic
    anthropic:
      promptCache:
        enabled: true        # default false (backward compat)
        ttl: ephemeral       # ephemeral (5m) | extended (1h)
```

Semantics per provider:
- Anthropic / Bedrock-Claude: model adapter injects cache_control markers at sensible defaults (system message last block + tools end). Honors ttl.
- OpenAI: no-op (automatic caching already covers it); option accepted but logged as informational so a single user-facing API works across providers.

### 🔄 Alternatives Considered

- LiteLLM inject_cache_control_breakpoints at the gateway layer works, but only for users who run a gateway in front of kagent, and the kagent layer has no visibility into whether caching is active.

### 🎯 Affected Service(s)

Multiple services / System-wide

### 📚 Additional Context

- Anthropic prompt caching: https://platform.claude.com/docs/en/build-with-claude/prompt-caching
- OpenAI automatic prompt caching: https://platform.openai.com/docs/guides/prompt-caching

### 🙋 Are you willing to contribute?

- [x] I am willing to submit a PR for this feature

PR #2788, opened 2026-09-10
feat(models): add prompt caching for the Anthropic provider (promptCaching, cacheTTL)

### LINKED ISSUE #2787: [FEATURE] Prompt caching for the Anthropic provider (cache_control on system prompt and tools), like Bedrock's promptCaching
### 📋 Prerequisites

### 📝 Feature Summary

Add prompt caching for the **Anthropic** model provider (direct API), mirroring the `promptCaching` / `cacheTTL` knobs that #1940 added for Bedrock: a `cache_control` breakpoint on the system prompt and on the tool definitions (optionally the last message), so the stable prefix of every agent loop call is billed as a cache read instead of fresh input.

A declarative agent re-sends its full history on every model call. With the Anthropic provider nothing is marked cacheable today (`go/adk/pkg/models/anthropic_adk.go` and the Python Anthropic path send no `cache_control`; `promptCaching` lives on `BedrockConfig` only), so the whole prefix is paid at full input price on every call.

Measured on kagent 0.10.1 with `claude-sonnet-4-6` through a gateway that logs `gen_ai.usage.*` per call (162 calls over 24 h, three agents):

- `cache_read.input_tokens = 0` and `cache_creation.input_tokens = 0` on **every** call.
- One incident-investigation session of a Python (`kagent-app`) agent: 119 model calls, input growing from 4,423 to **190,809 tokens** per call, average 93 k — i.e. roughly 11 M input tokens billed for one session, most of it the same prefix over and over. A Go (`golang-adk`) agent in the same window: 41 calls, 3.6 k → 93 k.

Anthropic's prompt caching would turn most of that into cache reads (system prompt + tools are identical across the loop; with a breakpoint on the last message the conversation prefix is too). Anthropic added the feature to the Messages API in 2024 and it is a one-line addition per request (`cache_control: {"type": "ephemeral"}` on a content block / tool definition; `ttl: "1h"` optional).

### 💡 Proposed Solution

- `AnthropicConfig` gains `promptCaching: bool` (default false, like Bedrock) and `cacheTTL: "5m" | "1h"`.
- Go runtime (`anthropic_adk.go`): when enabled, set `CacheControl` on the last system text block and on the last tool definition; optionally on the last user message content block (the usual "moving breakpoint" pattern for agent loops).
- Python runtime: same on the `anthropic` client requests.
- Surface `cache_read_input_tokens` / `cache_creation_input_tokens` into `usage_metadata.cached_content_token_count` (relates to #2669 / #2670) so the effect is visible in the UI's usage tab.
- Docs: a note next to the Bedrock `promptCaching` docs.

### 🔄 Alternatives Considered

- Enabling it unconditionally: cheaper for everyone, but the Bedrock knob is opt-in, and a minimum cacheable prefix applies per model; keeping the same opt-in shape is consistent.
- Compaction (#2728 / `spec.declarative.context.compaction`) reduces the prefix instead of caching it; both are needed — compaction bounds growth, caching removes the per-call repeat cost.

### 📚 Additional Context
**Prior request:** #1866 asked for exactly this (Anthropic *and* Bedrock) and was closed for inactivity after #1940 shipped the Bedrock half; this issue re-opens the Anthropic half with measurements.

kagent 0.10.1, Go runtime `golang-adk` and Python `kagent-app`, Anthropic provider via an OpenAI/Anthropic-compatible gateway (`baseUrl` set, `/v1/messages`). Numbers above are from the gateway's `gen_ai.usage.input_tokens`, `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.cache_creation.input_tokens` fields.

teemow:
Closes #2787. Picks up the Anthropic half of #1866, whose Bedrock half shipped in #1940.

## Problem

A declarative agent re-sends its full history on every model call. With the Anthropic provider nothing is marked cacheable today: `promptCaching` exists only on `BedrockConfig`, and neither runtime sends `cache_control`, so the whole prefix (system prompt, tool definitions, previous turns) is billed at full input price on every call.

Measured on kagent 0.10.1 on an internal installation, through a gateway that logs `gen_ai.usage.*` per model call (162 calls over 24 h, three agents, `claude-sonnet-4-6` via the Anthropic provider): `cache_read.input_tokens` and `cache_creation.input_tokens` were 0 on every call. One incident-investigation session of a Python (`kagent-app`) agent made 119 model calls whose input grew from 4,423 to 190,809 tokens per call (average 93 k), i.e. roughly 11 M input tokens billed for one session, most of it the same prefix over and over; a Go (`golang-adk`) agent in the same window went 3.6 k → 93 k over 41 calls. Anthropic's prompt caching turns that repeat into cache reads at a fraction of the input price. Full numbers in #2787.

## Change

- `spec.anthropic.promptCaching` (bool, default `false`) and `spec.anthropic.cacheTTL` (`"5m"` | `"1h"`, default `"5m"`) on `ModelConfig`, with the same names, defaults and doc shape as the Bedrock fields, wired through `adk.Anthropic` (`prompt_caching`, `cache_ttl`) to both runtimes. CRDs regenerated (`make controller-manifests`).
- When enabled, every Messages API request carries three `cache_control` breakpoints: the last tool definition, the last system prompt block, and the last content block of the latest conversation turn. Anthropic renders tools → system → messages and caches the prefix up to each breakpoint, so each call of an agent loop reads its whole previous history from the cache and writes only the new turn. That is three of the four breakpoints Anthropic allows; the conversation breakpoint skips thinking blocks, which Anthropic refuses to cache.
- **Go runtime** (`go/adk/pkg/models/anthropic_adk.go`): the request shaping moves into `buildAnthropicParams` so it can be unit-tested, `markAnthropicCacheBreakpoints` sets the markers, and usage now maps `cache_read_input_tokens` to `CachedContentTokenCount` and folds cache read/creation tokens into `PromptTokenCount` (the GenAI shape treats cached as a breakdown of prompt, which is also how google-adk maps Anthropic usage on the Python side), so the UI's usage view shows the effect. Streaming and non-streaming.
- **Python runtime** (`KAgentAnthropicLlm`): wraps the SDK client's `messages` resource with `PromptCachingMessages`, which marks the same three breakpoints before each `create` call. That is the one seam google-adk's `AnthropicLlm` hands its request through on both the streaming and the non-streaming path, so the request builder is not duplicated. google-adk ≥ 2.x already maps `cache_read_input_tokens` into `cached_content_token_count`; a regression guard for that is added.
- **Claude harness compiler**: treats the CRD-defaulted `cacheTTL: "5m"` like the Bedrock one (without this every `spec.anthropic` block would now fail the "no options beyond baseUrl" check, because the CRD defaults the field whenever the block is present) and keeps rejecting `promptCaching: true`, since Claude Code caches natively.
- **Helm**: `providers.anthropic.config.promptCaching` / `cacheTTL` documented in `values.yaml`; the chart already renders `config` into the provider block.

Design notes: opt-in like the Bedrock knob. A per-model minimum cacheable prefix applies (1024–4096 tokens; below it the markers are silently ignored) and 5-minute cache writes cost 1.25× normal input, so a prefix has to be read at least once to pay off. One knob enables all three breakpoints: the moving conversation breakpoint is what makes agent loops cheap, and a separate switch for it would be extra API surface for no realistic configuration. This is complementary to compaction (#2728): compaction bounds the growth of the prefix, caching removes the per-call repeat cost.

## Testing

- Go: `go test ./adk/... ./api/... ./core/internal/translator/...`, `make -C go lint`. New `anthropic_adk_test.go` asserts the wire body (as the SDK serializes it) carries exactly three markers at the expected positions for the default, `"5m"` and `"1h"` TTLs, none when disabled, the trailing-tool-result and no-tools/no-system cases, the usage fold, and drives both the SSE streaming and the non-streaming path through an `httptest` server. Plus agent wiring and config deserialization, translator wiring, and Claude compiler accept (`cacheTTL: "5m"`) / reject (`promptCaching`) cases.
- Python: `pytest packages/kagent-adk/tests/unittests` (228 passed; `test_grpc_health` fails locally only because port 8081 is taken on my machine). New tests for `mark_prompt_cache_breakpoints` (positions, thinking blocks skipped, no mutation of caller objects), the client wrapping, and end-to-end through google-adk's request builder with the SDK client mocked, including the usage-mapping guard.
- Against the Anthropic API (`claude-sonnet-4-6`, ~2.7 k-token system prompt + 2 tools, three consecutive agent-loop calls alternating non-streaming/streaming, both runtimes, values are `prompt_token_count / cached_content_token_count` as reported by the runtime):

  | runtime | promptCaching | call 1 | call 2 | call 3 |
  |---|---|---|---|---|
  | Go | off | 2739 / 0 | 2751 / 0 | 2763 / 0 |
  | Go | on, 5m | 2739 / 0 (cache write) | 2751 / 2736 | 2763 / 2748 |
  | Python | off | 2740 / 0 | 2756 / 0 | 2772 / 0 |
  | Python | on, 5m | 2740 / 0 (cache write) | 2756 / 2737 | 2772 / 2753 |

  After the first call ~99 % of every prompt is served from the cache and surfaces as cached tokens.
- Not run here: a kind-based run of an `AgentTemplate` on `main` (the kind flow on `main` has Substrate disabled by default). The kind end-to-end run was done on the `release/v0.10.x` backport, #2789, which runs agents as Deployments: Go and Python declarative agents with the built-in Kubernetes tools (~14 k-token prefix), three turns through the controller's A2A endpoint — every model call after the first reads ~99.5 % of its prompt from the cache (e.g. Go 15665 / 14257 cached on the tool-result follow-up, 16337 / 16266 on turn 3), reported in the A2A usage metadata. Table in #2789.

Follow-ups, not in this PR: the UI model form exposes neither this nor Bedrock's `promptCaching`; a docs note for kagent-dev/website `supported-providers/anthropic.md` once this lands.

### DIFF
--- a/go/adk/pkg/agent/agent.go
+++ b/go/adk/pkg/agent/agent.go
@@ -317,6 +317,8 @@ func CreateLLM(ctx context.Context, m adk.Model) (adkmodel.LLM, error) {
 			Temperature:     m.Temperature,
 			TopP:            m.TopP,
 			TopK:            m.TopK,
+			PromptCaching:   m.PromptCaching,
+			CacheTTL:        m.CacheTTL,
 		}
 		return models.NewAnthropicModel(ctx, cfg)
 
--- a/go/adk/pkg/agent/agent_test.go
+++ b/go/adk/pkg/agent/agent_test.go
@@ -57,7 +57,9 @@ func TestConfigDeserialization_Anthropic(t *testing.T) {
 		"model": {
 			"type": "anthropic",
 			"model": "claude-sonnet-4-20250514",
-			"base_url": "https://api.anthropic.com"
+			"base_url": "https://api.anthropic.com",
+			"prompt_caching": true,
+			"cache_ttl": "1h"
 		},
 		"description": "test agent",
 		"instruction": "you are helpful"
@@ -75,6 +77,38 @@ func TestConfigDeserialization_Anthropic(t *testing.T) {
 
 	if anthropic.Model != "claude-sonnet-4-20250514" {
 		t.Errorf("model name = %q, want %q", anthropic.Model, "claude-sonnet-4-20250514")
+	}
+	if !anthropic.PromptCaching {
+		t.Error("PromptCaching = false, want true")
+	}
+	if anthropic.CacheTTL != "1h" {
+		t.Errorf("CacheTTL = %q, want %q", anthropic.CacheTTL, "1h")
+	}
+}
+
+// TestCreateLLM_AnthropicPromptCaching verifies the prompt caching knobs reach
+// the Anthropic model, which is where the cache_control markers are set.
+func TestCreateLLM_AnthropicPromptCaching(t *testing.T) {
+	t.Setenv("ANTHROPIC_API_KEY", "test-key")
+	m := &adk.Anthropic{
+		BaseModel:     adk.BaseModel{Type: "anthropic", Model: "claude-sonnet-4-6"},
+		PromptCaching: true,
+		CacheTTL:      "1h",
+	}
+
+	llm, err := CreateLLM(context.Background(), m)
+	if err != nil {
+		t.Fatalf("CreateLLM: %v", err)
+	}
+	am, ok := llm.(*models.AnthropicModel)
+	if !ok {
+		t.Fatalf("model is %T, want *models.AnthropicModel", llm)
+	}
+	if !am.Config.PromptCaching {
+		t.Error("Config.PromptCaching = false, want true")
+	}
+	if am.Config.CacheTTL != "1h" {
+		t.Errorf("Config.CacheTTL = %q, want %q", am.Config.CacheTTL, "1h")
 	}
 }
 
--- a/go/adk/pkg/models/anthropic.go
+++ b/go/adk/pkg/models/anthropic.go
@@ -34,6 +34,15 @@ type AnthropicConfig struct {
 	Temperature *float64
 	TopP        *float64
 	TopK        *int
+	// PromptCaching, when true, marks the last tool definition, the last system
+	// prompt block and the last block of the latest conversation turn with a
+	// cache_control breakpoint so Anthropic bills the stable prefix of an agent
+	// loop as a cache read instead of fresh input. See markAnthropicCacheBreakpoints.
+	PromptCaching bool
+	// CacheTTL selects the cache retention window when PromptCaching is on.
+	// "" or "5m" uses the API's default 5-minute cache; "1h" opts into the
+	// 1-hour cache, which is billed at a higher cache-write rate. See anthropicCacheControl.
+	CacheTTL string
 }
 
 // AnthropicModel implements model.LLM for Anthropic Claude models.
--- a/go/adk/pkg/models/anthropic_adk.go
+++ b/go/adk/pkg/models/anthropic_adk.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"iter"
+	"slices"
 	"strings"
 
 	"github.com/anthropics/anthropic-sdk-go"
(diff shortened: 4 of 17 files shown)


EItanya (commented):
🤖 AI-generated review.

No correctness issues found at `f5679c2a`. One optional simplification:

1. **Reuse Python ADK’s existing caching support** — [_anthropic.py:81](https://github.com/kagent-dev/kagent/blob/f5679c2af173f9bee46a4ffee548a51b01daee02/python/packages/kagent-adk/src/kagent/adk/models/_anthropic.py#L81). The locked `google-adk 2.8.0` already generates these three breakpoints through `LlmRequest.cache_config`. Mapping the settings to `ContextCacheConfig` could remove the custom messages wrapper and marker logic. I verified identical wire requests for both TTLs, with streaming enabled and disabled. Check support at the declared minimum ADK version before switching.

Validation passed: four affected Go packages (`adk/pkg/models`, `adk/pkg/agent`, `core/internal/translator/adkconfig`, and `core/internal/translator/claude`), all 24 Python Anthropic tests, and an additional SDK comparison probe using mocked HTTP. No live Anthropic or cluster tests run.

teemow:
Looked at it. Kept the wrapper for now: ADK's Anthropic breakpoints are not in 2.6.2 (our floor), `ContextCacheConfig` is app-wide and experimental, and it only caches from the second turn. Happy to switch once we raise the floor.

PR #2790, opened 2026-09-10
feat(adk): honour context compaction in the Go runtime

teemow:
## Problem

Since 0.10.0 the Agent CRD defaults `spec.declarative.runtime` to `go`, and the Go runtime never read `spec.declarative.context.compaction`. The controller compiles the field into the agent's `config.json` (`context_config.compaction`) and the Python runtime consumes it, but a Go agent grew its prompt with every turn regardless of what the operator configured; the only signal was the `UnsupportedFeatures` condition (#2728). Measured on an internal installation running 0.10.1: an agent session grew from 4k to 190k input tokens per model call over 119 calls, and every Go agent there ran without compaction because the runtime ignored the field.

`google.golang.org/adk/v2` v2.3.0 (2026-08-31) ships native compaction (`session/compaction`: sliding window, tail retention, LLM summarizer) and the `runner.Config.Compaction` hook, so the runtime can consume it instead of carrying its own port of the algorithm, which is what #1971 and #2025 attempted before the primitives existed.

## Change

- **Dependency:** `google.golang.org/adk/v2` v2.2.0 → v2.3.0. The OpenTelemetry exporters and contrib instrumentation move to the matching v1.45.0 / v0.21.0 / v0.70.0 set, which the new ADK version requires for `otel/log`. otel/log v0.21.0 hands log bodies as attribute.Value; the telemetry test on main
follows that in the same commit. No other kagent code needed changes for the bump.
- **Runtime:** `agent.CompactionConfig` translates `context_config.compaction` into a `compaction.Config` and `runner.CreateRunnerConfig` hands it to the ADK runner. `compaction_interval` / `overlap_size` drive the sliding window, `token_threshold` / `event_retention_size` tail retention; either strategy is off when its settings are unset, so an agent without `context_config` runs exactly as before.
- **Half-configured settings** are ignored with a warning rather than rejected: `token_threshold` without `event_retention_size` (and the reverse), `overlap_size` without an interval. That mirrors the Python runtime, where adk-python runs tail retention only when both settings are present. What remains is validated at startup with the runner's own rules, so a configuration the runner would refuse (negative values, no strategy at all) fails fast with the reason instead of on the first turn.
- **Summarizer:** `summarizer_model` builds a `compaction.LLMSummarizer` through the same model factory as the agent (`CreateLLM`), with the model's generation config and a 60 s bound per call (the same bound the runner applies to its default summarizer). A `prompt_template` gets the tool-name instruction the Python runtime appends (#1907), so summaries keep namespaced tool names; a template without a summarizer model runs on the agent's own model. With neither, the runner summarizes with the agent's model and ADK's default prompt, matching the Python runtime.
- **Observability:** startup logs `context compaction enabled` with the resolved values and the summarizer model, or `context compaction not configured`, so an operator can see from the agent log what the runtime does with the field.
- **API:** `main` had no field for compaction in the v1alpha3 API (the 0.10 line has `spec.declarative.context`), so nothing produced `context_config`. `spec.kagent.compaction` on the `Harness` adds it, next to `spec.kagent.memory`: `compactionInterval` / `overlapSize` for the sliding window, `tokenThreshold` / `eventRetentionSize` for tail retention, and an optional `summarizer` with a `modelConfigRef` (a `ModelConfig` in the Harness namespace) and a `promptTemplate`. It is runtime policy rather than portable agent behaviour (specific to the ADK runner; the Claude and Codex harnesses have nothing to apply it to), which is why it lives on the Harness and not on the `AgentTemplate`. The CRD rejects at admission what the runtime would refuse at startup: no strategy, `tokenThreshold` without `eventRetentionSize` (and the reverse), `overlapSize` without `compactionInterval`, a `promptTemplate` without `{conversation_history}`.
- **Controller:** the kagent compiler translates the Harness's compaction policy into `context_config` in `config.json` of every root agent the Harness runs. A summarizer `ModelConfig` other than the agent's own is resolved from the Harness namespace like the memory model, so its credentials and egress join the revision and a change to it compiles a new revision; the agent's own model stays implicit because the runtime already summarizes with it by default.

Compaction records travel on `EventActions.Compaction`; the local DurableDir session store persists `Actions` as a whole (ADK's `session/database`), so a summary survives an actor restart and the same range is not summarized again.

The 0.10 line, where the field already exists on the Agent CRD and the reconciler flagged it as unsupported, is #2791.

## Testing

- `go test ./adk/pkg/agent/`: `TestCompactionStrategies` (table: each strategy alone and combined, unset settings, the three half-configured cases, no strategy, negative values), `TestCompactionConfig` (not configured → nil; default summarizer; dedicated summarizer model; custom prompt on the agent model; prompt without `{conversation_history}` rejected; prompt without any model rejected; the `context_config` block as the controller compiles it), `TestCompactionSlidingWindowFiresAfterInterval` (a scripted model drives the ADK runner: no record after turn 1, one record after turn 2 whose content is the summarizer's reply, and turn 3's request shows the summary and none of the compacted turns), `TestCompactionTailRetentionFiresOverTokenThreshold`.
- `go test ./core/internal/translator/...`: `TestApplyCompaction` (not configured → no `context_config`; all four settings and a prompt on the agent model → mapped with no `summarizer_model`; a dedicated summarizer `ModelConfig` → its model in `config.json`, its credential in the environment, its host in the egress, itself in the provenance models; a missing summarizer `ModelConfig` → error).
- `go test ./api/v1alpha3/` (envtest): the five new CEL cases reject each half configuration with its message, and the fully configured Harness is accepted.
- `make -C go test`: all packages pass.
- `make -C go lint`: 0 issues.

## End-to-end

`TestAgentInstanceContextCompaction` (in the `test-e2e` job) runs the whole chain with `mockllm` as the summarizer: a clone of the suite's `kagent` Harness with `compactionInterval: 2` and a dedicated summarizer `ModelConfig`, admitting a plain `AgentTemplate`; both models point at one mock LLM behind a recording proxy on the test host; a default header on each `ModelConfig` tells the two models apart. Three turns through the public A2A API, then the recorded requests show what the runtime sent: the summarizer model was called exactly once, after the second turn, with both turns in the transcript and the custom prompt; the second turn's request still carried the first turn raw; the third turn's request carries the summary and none of the four compacted messages. Run before pushing against a local kind cluster with Substrate 0.0.26 and the controller and Go ADK images built from this branch: passes in 4.9 s.

The real-model run (kind, images built from the branch, Anthropic `claude-haiku-4-5`, a five-turn tool-using conversation: turn 5 needs 2,524 prompt tokens with compaction against 50,247 without) was done on the 0.10 backport #2791. The runtime code is the same on both branches apart from the logging package (`log/slog` here, logr on 0.10).

Refs #2728 (closed by #2791 for the 0.10 line; on `main` the field returns with the v1alpha3 API).


EItanya (commented):
Awesome work! Do you think you can add an e2e test with `mockllm` for this?

teemow:
Added in e5238f0f: `TestAgentInstanceContextCompaction` in `go/core/test/e2e` runs the whole chain with mockllm as the summarizer. The template has `compactionInterval: 2` and a dedicated summarizer `ModelConfig`; both point at one mock LLM behind a recording proxy on the test host, and a default header per `ModelConfig` tells the two models apart. After three turns the recorded requests show the summarizer called exactly once (after turn 2, with both turns in the transcript and the custom prompt), turn 2's request still carrying turn 1 raw, and turn 3's request carrying the summary and none of the four compacted messages.

`main` had no way to configure compaction on the v1alpha3 `AgentTemplate`, so d734c31b adds `spec.context.compaction` (sliding window, tail retention, optional `summarizer.modelConfig` / `promptTemplate`), CEL rules that reject the half configurations at admission, the translator mapping into `context_config`, and status warnings where the setting is not applied (Shared children, Claude and Codex harnesses). Details are in the description.

Passes locally against a kind cluster with Substrate 0.0.26 and the controller and Go ADK images built from the branch; the workflow run for the new head is waiting for approval.

teemow:
Let me know if the api change is too much and I'll remove it again.

go/api/v1alpha3/agenttemplate_types.go
```diff
@@ -207,6 +270,11 @@ type AgentTemplateSpec struct {
 	// +kubebuilder:validation:MaxItems=20
 	// +optional
 	Plugins []PluginBundle `json:"plugins,omitempty"`
+	// Context configures how the runtime manages the conversation context,
+	// such as compacting older events. The kagent harness applies it to the
+	// root agent; other harnesses report it as ignored.
+	// +optional
+	Context *AgentTemplateContextSpec `json:"context,omitempty"`
```
supreme-gg-gg, line 277: I'm wondering if we should place this in `KagentHarness` instead of `AgentTemplate`. 

It seems to me that the split we've been using is `AgentTemplate` for portable agent behaviour and the harness variant for runtime-specific policy for a reusable class of agents. The compaction setting here is very specific to ADK shape, not supported in Codex / Claude, and even if later we add support for other harness, they will behave differently across runtimes. Thus, this sounds more like a harness policy than portable template config. A precedence here would be `KagentHarnessMemory`.

EItanya: I actually agree with this 1000%, great catch @supreme-gg-gg

teemow: Agreed, moved. `spec.kagent.compaction` on the Harness next to `memory` (ff2f9910); the AgentTemplate has no context field any more. The e2e test clones the kagent Harness with the policy.

go/core/test/e2e/interaction_test.go
```diff
@@ -1054,10 +1054,45 @@ func startForkMemoryMock(t *testing.T) string {
...
+	URL string
+
+	mu       sync.Mutex
+	requests []recordedModelRequest
+}
+
+type recordedModelRequest struct {
+	Header http.Header
+	Body   []byte
+}
```
EItanya, line 1085: I'm ok merging this for now, but can you please create a follow-up to add this to `mockllm`, I don't think it belongs here

EItanya: Actually I lied, can you please add this to `mockllm` first

EItanya: Actually one last thing, I don't think we really need this right now. As long as the mock passes it means that the interaction worked exactly as expected?

teemow: Not quite. mockllm only matches the last message and does not record or count calls.
Without the recorder the test also passes when the runtime ignores the policy: turns 1-3
still match their mocks, the summarizer mock is just never hit. The recorder is what checks
that the summarizer was called once with both turns, and that turn 3 carries the summary
instead of the raw turns.

It is the same proxy you added for the checkpoint test in #2754, just shared by both tests.
I would keep it here and open a mockllm PR that records requests. Then both proxies can go.
OK with that?

EItanya:
e2e tests are failing

teemow:
Rebasing on your #2844 fixed it

supreme-gg-gg (commented):
Hey @teemow can you resolve the merge conflicts? I took another look and this seems about ready

EItanya (approved):

EItanya:
In general can please try to not use force pushes in PRs, it makes review much much harder

PR #2793, opened 2026-09-10
fix(adk): only compact MCP App results that carry a render payload

### LINKED ISSUE #2716: go/adk: MakeMCPAppModelResultCallback blanks non-UI tool results, same gap #2579 fixed in Python
#2579 fixed this on the Python kagent-adk side. `MakeMCPAppModelResultCallback` / `compactMCPAppModelResponse` in `go/adk/pkg/agent/mcp_apps.go` has the identical bug and wasn't touched there, since it's a separate PR concern. dimetron asked on that review thread for the Go side to be tracked separately so Python and Go don't drift on MCP App semantics, so opening this.

The callback gates purely on tool name: `appToolNames[part.FunctionResponse.Name]`, populated once per session in `mcp_ui.go`'s `agentVisibleToolFilter` from each tool's declared `_meta.ui.resourceUri` at `ListTools` time. Once a tool is in that set, every non-error result from it gets its content replaced by `mcpAppRenderedNotice`, even a call that returned plain text with no `_meta.ui` and no `structuredContent` at all, because `compactMCPAppModelResponse` only branches on `IsError` before doing the unconditional collapse.

Reproduced with a Go test against current HEAD (b8d53d3): called the callback with a `jenkins_monitor_build` result (an app-declared tool) carrying only `content: [{type: text, text: "No build found for job demo/1: it was deleted."}]`, no `_meta.ui`, no `structuredContent`. Expected the text to pass through since there's nothing to compact. Got the canned render notice instead.

Same fix shape as #2579: check the actual result for a UI resource (or `structuredContent`) before compacting, not just whether the tool is capable of producing one.

wangzhengzhuo05:
### Summary

`MakeMCPAppModelResultCallback` replaced the result of every MCP App-capable tool with the canned render notice, even when an individual call returned an ordinary text-only result (no `structuredContent`, no `_meta.ui`). That hid real content from the model.

The callback now collapses a successful result only when the result itself carries a render payload — a non-nil `structuredContent` or a `_meta.ui.resourceUri` — and passes plain results through unchanged. This mirrors the Python-side fix in #2579 so Go and Python stop drifting on MCP App semantics.

### What changed

- `go/adk/pkg/mcp/mcp_ui.go`: add an exported `HasUIResource(meta)` helper that reuses the existing `parseMCPUIMetadata` as the single source of truth for `_meta.ui` parsing.
- `go/adk/pkg/agent/mcp_apps.go`: gate the success-path compaction on `result.StructuredContent == nil && !kagentmcp.HasUIResource(result.Meta)`; return the response unchanged when there is no render payload. The `isError` path keeps its existing behaviour.
- `go/adk/pkg/agent/mcp_apps_test.go`: add a regression test using the issue's repro — an app-declared tool returning only `content: [{type: text, text: "No build found for job demo/1: it was deleted."}]` must reach the model unchanged.

### Why not just branch on `IsError`

A tool declaring `_meta.ui.resourceUri` is *capable* of rendering an app; a given call may still return plain content (e.g. "not found"). Deciding from the tool name alone conflates capability with an actual render payload, which is what the regression covers.

### Verification

- `go test ./adk/pkg/agent/... -count=1` — pass (4 MCP App tests: 3 pre-existing + 1 new)
- `go test ./adk/pkg/mcp/... -count=1` — pass
- `go build ./...` — pass
- Mutation check: with the new guard disabled, `TestMakeMCPAppModelResultCallbackPassesThroughPlainResultFromAppTool` fails exactly as expected (the plain text is replaced by the render notice). Restored afterwards.

Fixes #2716

### DIFF
--- a/go/adk/pkg/agent/mcp_apps.go
+++ b/go/adk/pkg/agent/mcp_apps.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
+	kagentmcp "github.com/kagent-dev/kagent/go/adk/pkg/mcp"
 	"google.golang.org/adk/v2/agent"
 	"google.golang.org/adk/v2/agent/llmagent"
 	adkmodel "google.golang.org/adk/v2/model"
@@ -61,6 +62,15 @@ func compactMCPAppModelResponse(response map[string]any) map[string]any {
 		// diagnose and recover; only drop the heavy structured payload.
 		result.StructuredContent = nil
 		return encodeCallToolResult(result, response)
+	}
+
+	// Only collapse when the result actually carries an MCP App render payload.
+	// A tool can be App-capable (declared `_meta.ui.resourceUri`) yet return an
+	// ordinary result for a given call — plain text with no `structuredContent`
+	// and no `_meta.ui`. Replacing that with the render notice would hide real
+	// content from the model, so pass it through unchanged.
+	if result.StructuredContent == nil && !kagentmcp.HasUIResource(result.Meta) {
+		return response
 	}
 
 	// On success, collapse the render payload into a terminal directive so the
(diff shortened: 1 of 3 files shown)


wangzhengzhuo05:
Closing this to keep the tracker clean \u2014 it has been sitting without review for several days. This is no reflection on the change itself: the fix and its local verification still stand. If the issue is still present and you would like the patch revisited, say the word and I will rebase it onto current main and re-open. Thanks for your time.

PR #2795, opened 2026-09-10
fix(sts): deliver the exchanged STS token on MCP tool calls

ricardomiguel-os:
## Problem

The STS token exchange succeeds and the exchanged token is cached, but it never leaves the pod. MCP tool calls go out with **no `Authorization` header at all**, so the upstream server sees no identity instead of the user delegated to the agent.

`sessionIDFromContext` recovered the session with a direct type assertion:

```go
ctx.(interface{ SessionID() string })
```

A type assertion inspects only the context object in hand. `createTransport` always builds its `http.Client` with a non-zero `Timeout` (`defaultTimeout` is 30m), and `http.Client` re-wraps the request context in a deadline context whenever `Timeout > 0` (`net/http.setRequestCancel` → `context.WithDeadline`). By the time `RoundTrip` runs, the context in hand is that wrapper, not ADK's `ToolContext`, so the assertion never matches and `HeaderProvider` returns `nil`.

No configuration avoids this: the timeout is non-zero even when unset.

The neighbouring `propagateToken` path keeps working because it uses `a2asrv.CallContextFrom`, a `Value` lookup, and `Value` lookups traverse the parent chain.

## Fix

1. `a2a/executor.go` stamps the session ID as a context value, beside the bearer token it already stamps.
2. `sts/plugin.go` prefers that `Value` lookup in `sessionIDFromContext`, keeping the type assertion as a fallback so callers still holding ADK's `ToolContext` are unaffected.
3. `models/base.go` gives `contextKey` a name. `BearerTokenKey` and `SessionIDKey` would otherwise both be pointers to a zero-size struct, and Go may place those at the same address, so the two keys compare equal and the session stamp overwrites the bearer token. The end-to-end test caught this by seeing the session ID arrive at the STS as `subject_token`.

The token cache key, its write side and its eviction are untouched. `NewTokenPropagationPlugin`'s signature is unchanged.

Requests with no user in context, such as the toolset discovery performed at startup, still get no header. That behaviour is asserted in the new test rather than changed.

## Evidence

Red/green on the end-to-end test, reverting only the `Value` lookup in `sessionIDFromContext`:

```
--- FAIL: TestSTSExchangedTokenReachesMCPTool
    invocation[0..5] Authorization = ""
    want: "Bearer EXCHANGED-STS-TOKEN"
```

With the fix: `go test -race -skip 'TestE2E.*' ./adk/...` green across all packages, `golangci-lint run` reports 0 issues, `go build ./...` and `gofmt` clean.

## Tests

- `mcp/sts_injection_test.go` — end-to-end, nothing stubbed but the LLM: real `runner.Runner`, `llmagent`, `mcptoolset`, the real `CreateToolsets`/`createTransport` (so a real `http.Client` with the deadline wrapper), real `sts.TokenPropagationPlugin` performing a real exchange against an `httptest` STS, and a real MCP server recording every `Authorization` it receives.
- `sts/plugin_test.go` — session recovery precedence: session as a value, the same wrapped in a deadline context, session only via `SessionID()`, no session, session present with no cached token, and value-beats-method.
- `a2a/executor_test.go` — the executor stamps the session ID, and the two context keys coexist without overwriting each other.


ricardomiguel-os:
Folded into #2803, which contains this change in full plus the LLM transport.

Keeping them apart did not help review: #2803 is based on `main`, so its diff already showed everything here, and a cross-fork PR cannot be based on a branch in my fork. One PR, one diff.

PR #2796, opened 2026-09-10
fix: skip unchanged MCP tool writes to Postgres

supreme-gg-gg:
## Summary

- Cache the last persisted MCP tool list "snapshot" in the reconciler and skip `StoreToolServer` / `RefreshToolsForServer` when nothing changed.
- Controllers still poll on the existing refresh interval (which is still configurable through #2480), but idle Postgres is no longer written every cycle.

## Test plan

- [x] Unit tests for snapshot equality, description/tool changes, and cache evict
- [x] On a cluster: confirm periodic `successfully registered remote MCP server` logs, and that `tool` / `toolserver.updated_at` and `n_tup_upd` stay at the first write

### DIFF
--- a/go/core/internal/controller/reconciler/reconciler.go
+++ b/go/core/internal/controller/reconciler/reconciler.go
@@ -1181,6 +1181,7 @@ func (a *kagentReconciler) upsertToolServerForRemoteMCPServer(ctx context.Contex
 		return tools, nil
 	}
 
+	a.evictToolSnapshot(toolServer.Name, toolServer.GroupKind)
 	if _, err := a.dbClient.StoreToolServer(ctx, toolServer); err != nil {
 		return nil, fmt.Errorf("failed to store toolServer %s: %w", toolServer.Name, err)
 	}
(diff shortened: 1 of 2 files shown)


EItanya (commented):
🤖 AI-generated review.

1. **Invalidate the cached snapshot before attempting changed writes** — [reconciler.go:1184–1189](https://github.com/kagent-dev/kagent/blob/f423a4dad551f48b872ff40e2a5578414e7bbe0c/go/core/internal/controller/reconciler/reconciler.go#L1184-L1189). `StoreToolServer` commits separately from the tool refresh. If snapshot A is cached, writing B updates the server row, and the refresh fails, A remains cached. Rolling back to A then skips persistence indefinitely, leaving B’s metadata in the database. An uncertain commit outcome can also leave stale tools. Evict the snapshot before attempting the writes; remember it only after both succeed.

Reproduced through real MCP discovery with injected database failures. Existing reconciler tests pass; both regression cases and the full package pass with the one-line fix applied through a scratch overlay. No live PostgreSQL testing was performed.

EItanya (approved):

PR #2798, opened 2026-09-10
checkpoints updates

Charlesthebird:
Removing this in favor of the inline checkpoint deletion


Charlesthebird:
The round trip against a real cluster: snapshots taken in a chat, then deleted together from the page.

https://github.com/user-attachments/assets/ed4b53fe-5ed5-44f3-8e20-84d2ef46cf72

- Three checkpoints taken from the composer in a conversation.
- The page lists every one across all conversations, not just that chat's.
- Three picked and deleted together: one request each, sequential, under one `toast.promise` — a failure does not stop the run and the toast says how it went.

Recorded on the branch before the split, on the same code this PR carries.

---
*🤖 written by Claude*

PR #2799, opened 2026-09-11
feat(substrate): render BYO extraContainers in the SandboxAgent ActorTemplate

carlochessamp:
Supersedes #2691 and #2750 — this is the same change, combined into one PR so it can be reviewed and landed as a unit. Both earlier PRs are closed in favor of this one.

## Problem

`byo.deployment.extraContainers` (and the declarative `deployment.extraContainers`) render into the translator's pod template and work on the plain Deployment path, but on substrate `buildSandboxAgentActorTemplate` picked out only the kagent container and built a one-element `Containers` slice — every extra container was silently dropped. Nothing fails: the CRD accepts them, the ActorTemplate CRD supports up to ten containers, the sidecars are just gone at runtime.

Concrete case: an agent plus a credential-brokering forward proxy on `127.0.0.1` — on substrate the proxy never starts and the agent loses all outbound API access.

## Fix

Render the pod template's remaining containers as additional ActorTemplate containers, appended after the agent container:

- **Index-based agent lookup** (`findKagentContainerIndex`) — the agent container stays `Containers[0]`, so the durableDir session-store mount (`applyDurableDirSessionStore`) remains agent-private.
- **Image must be digest-pinned** (`pinImageRef`, per container — snapshots key on the image).
- **Command must be explicit** — command and args are concatenated; the combined command must be non-empty (atelet copies `Command` verbatim into the OCI `Process.Args` with no shell and no image-entrypoint fallback).
- **Env is fail-closed**: literals and `secretKeyRef` survive; `envFrom` and any other `valueFrom` source (`configMapKeyRef`, `fieldRef`, ...) are **rejected with a named-container error** instead of silently dropped — a sidecar whose credential source vanished mid-flight is worse than a rejected apply.
- **VolumeMounts are rejected** — ActorTemplate volumes only support the single `durableDir` already consumed by the agent's `/data` session store.
- **HTTP readiness probes map to `Readyz`**, so actor readiness keeps gating on the sidecar; named ports resolve against the extra container's own `containerPorts` (the translator only registers the agent container's `http` port). Resources, container ports, and TCP/exec probes have no ActorTemplate equivalent and are ignored.
- **At most 10 containers** (the CRD's `MaxItems`).
- The extras feed `actorTemplateShapeHash`, so adding/removing a sidecar fans out blue-green to a new template + actors like any spec change.

## Resolution of the two approaches

#2691 contributed the structure (index refactor, Readyz mapping, named-port resolution, 10-container cap, shape-hash assertion); #2750 contributed the fail-closed env policy and the concatenated command/args emptiness check. Two semantic deltas from #2691 as originally written, both toward #2750's stricter behavior:

- `configMapKeyRef`/other `valueFrom` env is now **rejected** rather than silently dropped;
- the command check runs on the **concatenated** command+args, so an args-only sidecar is accepted instead of being required to set `command`.

Scoped to the SandboxAgent path; the AgentHarness substrate path has no extra-container field today. Based on the current `release/v0.10.x` tip (both earlier PRs were behind).

## Testing

- `TestBuildSandboxAgentActorTemplateExtraContainers`: happy path (command/args concat, literal + `secretKeyRef` env, agent stays container 0 with the durableDir mount, readyz mapped, no mounts), named readiness-port resolution, and the failure modes (unpinned image, missing command/args, `envFrom`, `configMapKeyRef`, `fieldRef`, volume mounts, unresolvable named port, >10 containers) — plus the shape-hash template-name change.
- `go test ./core/pkg/sandboxbackend/...` green (4 packages), `go vet` clean.

### DIFF
--- a/go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
+++ b/go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
@@ -64,10 +65,11 @@ func (p *Lifecycle) buildSandboxAgentActorTemplate(
 	wpKey types.NamespacedName,
 	podTemplate corev1.PodTemplateSpec,
 ) (*atev1alpha1.ActorTemplate, error) {
-	kagentContainer := findKagentContainer(podTemplate.Spec.Containers)
-	if kagentContainer == nil {
+	kagentIdx := findKagentContainerIndex(podTemplate.Spec.Containers)
+	if kagentIdx < 0 {
 		return nil, fmt.Errorf("pod template is missing the kagent container")
 	}
+	kagentContainer := &podTemplate.Spec.Containers[kagentIdx]
 	image, err := pinImageRef(kagentContainer.Image)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


go/core/pkg/sandboxbackend/substrate/agent_lifecycle.go
```diff
@@ -176,16 +186,133 @@ func applyDurableDirSessionStore(spec *atev1alpha1.ActorTemplateSpec) {
...
+		for _, e := range c.Env {
+			if e.ValueFrom != nil && e.ValueFrom.SecretKeyRef == nil {
+				return nil, fmt.Errorf("extra container %q env %q: ActorTemplates support literal values and secretKeyRef only", c.Name, e.Name)
+			}
+		}
+		ec := atev1alpha1.Container{
+			Name:    c.Name,
+			Image:   image,
+			Command: command,
+			Env:     actorTemplateEnvFromPodEnv(c.Env),
```
mesutoezdil, line 260: kagent container's env can silently drop the same way?

PR #2800, opened 2026-09-11
fix(translator): honour a BYO harness's configured PORT so the card and runtime agree

### LINKED ISSUE #2758: byo harness injects no PORT but advertises the agent on :80, so a BYO image built with go/adk/pkg/app is READY and unreachable
> **Disclosure:** this issue was written by an AI agent (Claude) while verifying what a
> bring-your-own-agent guide must tell readers, and reviewed by a human before filing.

## Summary

The `byo` harness compiler injects no `PORT`, but the agent card it builds advertises the
agent at `http://127.0.0.1:80`. kagent's own BYO helper (`go/adk/pkg/app`) defaults to
**8080**. So a BYO image built the documented way reports **`READY`** and then fails every
invoke, because the gateway dials a port nothing is listening on.

The instance being `READY` is what makes this bad: readiness is probed on a *different*
port that succeeds regardless, so nothing surfaces the mismatch.

## Environment

kagent `b8d53d39` (chart `kagent-v0.10.0-rc1-129-gb8d53d39`), substrate 0.0.25, kind, Kubernetes 1.35.

## Reproduction

Built a BYO image from `go/core/test/byoa2a/main.go` with the one line `Port: "80"` removed,
so `app.New` falls back to `PORT` and then to its 8080 default:

```go
application, err := app.New(app.AppConfig{
    AgentCard: a2atype.AgentCard{Name: "byoport", Version: "v1", ...},
    AppName: "byoport", Logger: logger,       // no Port set
}, executor{})
```

Harness and AgentTemplate:

```yaml
apiVersion: kagent.dev/v1alpha3
kind: Harness
metadata: {name: byoport, namespace: kagent}
spec:
  byo: {}
  allowedAgentTemplates: {selector: {matchLabels: {kagent.dev/harness: byoport}}}
  workload:
    image: localhost:5001/byoport@sha256:ea596db3dac8da570980143210efeb2b47bcfb0a3afc5aa0f325a6063c5cf009
    command: ["/byoport"]
  substrate:
    workerPoolRef: {name: kagent-default}
    snapshotPolicy: {location: s3://ate-snapshots/kagent}
```

Result — the template goes fully ready:

```
Accepted=True(Accepted) ResolvedRefs=True(Resolved) Compatible=True(Compatible) Ready=True(Ready)
```

The AgentInstance is created `READY`, and the invoke fails:

```
$ kagent create agent-instance --agent-template byoport-agent --harness byoport
| 01a08301-8cd0-72c8-818f-26c7490ce37d | byoport-agent | byoport | READY | ... |

$ kagent invoke --agent-instance 01a08301-8cd0-72c8-818f-26c7490ce37d --task "hello"
Error: invoke AgentInstance: upstream call failed: Connect: tunnel failed
```

Adding `PORT` to the Harness fixes it with **no image change**:

```bash
kubectl patch harness byoport -n kagent --type=merge \
  -p '{"spec":{"env":[{"name":"PORT","value":"80"}]}}'
```

```
$ kagent invoke --agent-instance <new id> --task "hello"
BYO agent response
```

That is the proof: same image, same command, only `PORT` differs.

## Why it happens

- `go/core/internal/translator/kagent/compiler.go:73` injects `corev1.EnvVar{Name: "PORT", Value: "80"}`.
- `go/core/internal/translator/byo/compiler.go` injects **no** `PORT`. Its environment is
  only `compiled.Environment` plus `adkconfig.HarnessEnvironment(harness)`, i.e. the
  Harness's own `spec.env`.
- `go/core/internal/translator/byo/compiler.go:64` nonetheless builds the card with
  `SupportedInterfaces: [{URL: "http://127.0.0.1:80", ...}]`.
- `go/adk/pkg/app/app.go:23,161-165` — `cfg.Port` -> `PORT` env -> `defaultPort = "8080"`.

The e2e image passes only because `go/core/test/byoa2a/main.go:40` hardcodes `Port: "80"`.
The other in-repo reference, `go/adk/examples/byo/main.go`, documents `PORT` as
"default 8080" and would not work unpinned.

## Why readiness does not catch it

`go/core/internal/substrate/actor_template.go:69-71` probes `/readyz` on port **8081**, and
`go/adk/pkg/a2a/server/server.go:134` serves that on `:8081` independently of `PORT`. So the
health port is satisfied while the A2A port is wrong.

## Expected

One of:

1. `byo/compiler.go` injects `PORT=80` the way the kagent compiler does, so the card and the
   runtime agree by construction; or
2. the card is built from the port the image actually uses, with a documented field for it; or
3. the mismatch is rejected or surfaced as a condition instead of presenting as `READY`.

Option 1 matches existing behaviour for the other harnesses and needs no API change.

## Adjacent, not tested

`byo/compiler.go:64` sets `ProtocolBinding: TransportProtocolGRPC` while `app.New`
advertises `a2atype.NewAgentInterface("/", TransportProtocolJSONRPC)` (`app.go:180`).
Whether the gateway cares was not tested here.

wangzhengzhuo05:
### What

The `byo` harness compiler advertised the agent card at `http://127.0.0.1:80` while injecting no `PORT`, so a BYO image built with `go/adk/pkg/app` listened on its `8080` default, reported `READY`, and then failed every invoke.

The port is now taken from the harness's own configuration: whatever `PORT` the harness sets in `spec.env` is what both the runtime receives and the card advertises. `PORT=80` is only injected when the harness configures no usable port.

### Why

`go/core/internal/translator/byo/compiler.go` built the card with a hard-coded `http://127.0.0.1:80`, but its environment was only `compiled.Environment` plus `adkconfig.HarnessEnvironment(harness)`. `go/adk/pkg/app` resolves the listen port as `cfg.Port` -> `PORT` env -> default `8080`, so the card and the runtime disagreed. Readiness is probed on `:8081` independently of `PORT`, so nothing surfaced the mismatch and the instance stayed `READY` while unreachable.

BYO is an opaque harness, so the compiler must not force a port it cannot know (review feedback on this PR). At the same time the card and the runtime must not be able to disagree, so the advertised port is derived from the environment that is actually rendered.

### How

- Build the environment as `adkconfig.DedupeEnv(compiled.Environment + HarnessEnvironment(harness))`.
- `configuredPort` reads the last `PORT` entry in that environment (the same last-wins semantics `DedupeEnv` uses). A value in `1..65535` is the harness's port, and nothing is injected.
- Otherwise the default `80` is appended, matching the built-in `kagent` compiler for an unconfigured BYO harness.
- `agentTemplateCard` now takes the resolved port and builds `http://127.0.0.1:<port>`.
- Environment is resolved before `BuildProvenance` so the provenance hashes the same environment that ships.

No API or CRD change.

### Tests

- `TestCompilePortResolution` covers: no `PORT` -> exactly one `PORT=80` and card `http://127.0.0.1:80`; `spec.env PORT=8080` -> exactly one `PORT=8080` and card `http://127.0.0.1:8080`; `PORT=banana` -> falls back to exactly one `PORT=80` and card `http://127.0.0.1:80`.
- `TestCompileOpaqueImage` unchanged in intent.
- `cd go && go test ./core/internal/translator/byo/...` passes; `cd go && go build ./...` passes; `gofmt -l` clean.
- Mutation check: pinning `configuredPort` to always report "not configured" makes the `PORT=8080` case fail, so the regression test guards the behaviour rather than the implementation.

This change was implemented with AI assistance (OpenCode + muse-spark 1.3) and reviewed and verified locally; I understand it and take responsibility for it.

Fixes #2758

### DIFF
--- a/go/core/internal/translator/byo/compiler.go
+++ b/go/core/internal/translator/byo/compiler.go
@@ -13,6 +13,7 @@ import (
 	v2translator "github.com/kagent-dev/kagent/go/core/internal/translator"
 	"github.com/kagent-dev/kagent/go/core/internal/translator/adkconfig"
 	"istio.io/istio/pkg/kube/krt"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // Compiler translates resolved inputs into a BYO A2A runtime revision.
@@ -38,7 +39,7 @@ func (c *Compiler) Compile(ctx context.Context, input *v2translator.HarnessInput
 	if err != nil {
 		return nil, fmt.Errorf("convert agent card: %w", err)
 	}
-	environment := adkconfig.DedupeEnv(append(compiled.Environment, adkconfig.HarnessEnvironment(harness)...))
+	environment := adkconfig.DedupeEnv(append(append(compiled.Environment, adkconfig.HarnessEnvironment(harness)...), corev1.EnvVar{Name: "PORT", Value: "80"}))
 	provenance, err := c.config.BuildProvenance(ctx, harness, compiled.Templates, compiled.Models, environment)
 	if err != nil {
 		return nil, fmt.Errorf("build revision provenance: %w", err)
--- a/go/core/internal/translator/byo/compiler.go
+++ b/go/core/internal/translator/byo/compiler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"strconv"
 	"strings"
 
 	a2atype "github.com/a2aproject/a2a-go/v2/a2a"
@@ -15,6 +16,24 @@ import (
 	"istio.io/istio/pkg/kube/krt"
 	corev1 "k8s.io/api/core/v1"
 )
+
+// defaultPort is injected only when the harness configures no usable PORT.
+const defaultPort = "80"
+
+// configuredPort returns the port the harness configured for itself via PORT.
+func configuredPort(environment []corev1.EnvVar) (string, bool) {
+	for i := len(environment) - 1; i >= 0; i-- {
+		if environment[i].Name != "PORT" {
+			continue
+		}
+		port, err := strconv.Atoi(environment[i].Value)
+		if err != nil || port < 1 || port > 65535 {
+			return "", false
+		}
+		return environment[i].Value, true
+	}
+	return "", false
+}
 
 // Compiler translates resolved inputs into a BYO A2A runtime revision.
 type Compiler struct{ config *adkconfig.Builder }
@@ -35,11 +54,16 @@ func (c *Compiler) Compile(ctx context.Context, input *v2translator.HarnessInput
 	if err != nil {
 		return nil, fmt.Errorf("marshal agent config: %w", err)
 	}
-	card, err := pbconv.ToProtoAgentCard(agentTemplateCard(template))
+	environment := adkconfig.DedupeEnv(append(compiled.Environment, adkconfig.HarnessEnvironment(harness)...))
+	port, ok := configuredPort(environment)
+	if !ok {
+		port = defaultPort
+		environment = adkconfig.DedupeEnv(append(environment, corev1.EnvVar{Name: "PORT", Value: defaultPort}))
+	}
+	card, err := pbconv.ToProtoAgentCard(agentTemplateCard(template, port))
 	if err != nil {
 		return nil, fmt.Errorf("convert agent card: %w", err)
 	}
-	environment := adkconfig.DedupeEnv(append(append(compiled.Environment, adkconfig.HarnessEnvironment(harness)...), corev1.EnvVar{Name: "PORT", Value: "80"}))
 	provenance, err := c.config.BuildProvenance(ctx, harness, compiled.Templates, compiled.Models, environment)
 	if err != nil {
 		return nil, fmt.Errorf("build revision provenance: %w", err)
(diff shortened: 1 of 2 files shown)


EItanya (dismissed):

EItanya:
I approved this, but actually I realized that I'm not sure it makes sense. BYO is meant to be opaque harness, so I'm not sure we should auto inject that variable, maybe the user can do it via config?

wangzhengzhuo05:
Thanks for the review — agreed, and reworked.

BYO is opaque, so the harness's own configuration now wins:

- a `PORT` set in `spec.env` is honoured verbatim — nothing is injected, and the card advertises `http://127.0.0.1:<that port>`, so the card and the runtime cannot disagree;
- `PORT=80` is only injected when the harness configures no usable port, which keeps the unconfigured BYO case from #2758 working out of the box and matches what the built-in `kagent` compiler already does.

Covered by `TestCompilePortResolution`:
- no `PORT` -> injected `PORT=80`, card `:80`
- `PORT=8080` in `spec.env` -> single `PORT=8080`, card `:8080` (the case you flagged)
- `PORT=banana` -> falls back to `PORT=80`, card `:80`

`go test ./core/internal/translator/byo/...` and `go build ./...` are green, and the `PORT=8080` case fails if the resolution is stubbed out. No API or CRD change.

EItanya:
Sorry for changing my mind here, but I actually think that this requires an API change. Part of any BYO harness integration is the port on which they serve A2A. Therefore I think that port should be a setting in the BYO section of Harness. What do you think? That would also allow us to avoid reading the `PORT` env in the comiler

go/core/internal/translator/byo/compiler.go
```diff
@@ -13,8 +14,27 @@ import (
...
+// defaultPort is injected only when the harness configures no usable PORT.
+const defaultPort = "80"
+
+// configuredPort returns the port the harness configured for itself via PORT.
+func configuredPort(environment []corev1.EnvVar) (string, bool) {
+	for i := len(environment) - 1; i >= 0; i-- {
+		if environment[i].Name != "PORT" {
+			continue
+		}
+		port, err := strconv.Atoi(environment[i].Value)
```
mesutoezdil, line 29: credentialref port falls through to the 80 default here?

wangzhengzhuo05:
Thanks for the second look — and no worries at all about changing your mind, I think you're right.

If the A2A port is part of what defines a BYO harness integration, then it belongs in the `Harness` spec as an explicit field, and the translator/compiler should stop inferring it from `PORT` in the environment. That is a different change from what this PR does, so it does not make sense to keep this sitting in your review queue while that design question is open.

Closing it. For the record, what this branch contained (in case any of it is worth folding into the API-shaped version): it made the inference path and the rendered card agree on a single port — a `PORT` set in `spec.env` was honoured verbatim and the card advertised exactly that port, with `PORT=80` injected only when nothing was configured — plus a regression test covering the three cases (configured port, unconfigured, non-default). The underlying complaint from #2758 was really that the card and the runtime could advertise different ports, which the `port` field in the BYO section would also solve, just more honestly.

Thanks for the careful review — the push-back was the most useful part of this thread.

PR #2803, opened 2026-09-11
feat(sts): send the exchanged STS token on LLM and MCP calls

ricardomiguel-os:
Supersedes #2795, which this branch contains in full.

## Problem

The STS exchange succeeds and the exchanged token is cached, but it never leaves the pod. MCP tool calls go out with no `Authorization` header, and LLM calls go out with the caller's own token. Either way the upstream sees the wrong identity instead of the user delegated to the agent.

Two causes:

- **MCP.** `sessionIDFromContext` recovered the session with a type assertion on the context in hand. `http.Client` re-wraps the request context whenever `Timeout > 0`, which is always (`defaultTimeout` is 30m), so by `RoundTrip` the assertion never matches and `HeaderProvider` returns nil. No configuration avoids it.
- **LLM.** `PassthroughToken` only ever read `BearerTokenKey`, so the exchanged token had no way in.

## Fix

The executor stamps the session ID and the token provider on each request context, beside the bearer token it already stamps. Both transports then resolve the same way: the exchanged token for the session, else the caller's own.

- `sessionIDFromContext` prefers a `Value` lookup, which traverses the parent chain, keeping the type assertion as a fallback for callers still holding ADK's `ToolContext`.
- `PassthroughToken` gains the exchanged-token step. Its signature is unchanged, so `models/openai.go`, `models/anthropic.go` and `embedding/embedding.go` pick it up unedited.
- `contextKey` gets a name. The keys were pointers to a zero-size struct, which Go may place at the same address, making two distinct keys compare equal.
- Python mirrors this in `resolve_passthrough_token`, shared by `LLMPassthroughPlugin` and by `KAgentEmbedding`, which has no `callback_context` and reads the ContextVars instead.

The token cache, its key, its write side and its eviction are untouched.

## What deliberately gets no token

- **A request presenting no caller token.** The cache is per session and outlives the run, so a later turn arriving without an `Authorization` header would otherwise be served with the token exchanged for whoever sent an earlier turn. Identity and credential are separate headers here, `x-user-id` and `Authorization`, so a request can carry one without the other. The exchanged token replaces the caller's, it never stands in for its absence.
- **Startup toolset discovery**, which has no user in context.
- **Propagate-only mode** (`KAGENT_PROPAGATE_TOKEN=true`, no `STS_WELL_KNOWN_URI`). No exchange happens there: the plugin caches the caller's raw token, and `extractJWTExpiry` returns 0 for a non-JWT, which `HasExpired` reads as never expiring. The LLM path reads that same token from the request context anyway, always for the request being served.
- **No STS configured at all.** Nothing is stamped and the caller's token is returned, exactly as before.

A ServiceAccount-token fallback was considered and left out: with passthrough on and no inbound token it would send the pod's Kubernetes token to whatever `baseUrl` points at, including a public provider.

## For operators

`KAGENT_STS_RESOURCE` / `KAGENT_STS_AUDIENCE` scope the exchanged token to a backend. It now goes to the model endpoint as well as MCP, so those values must cover both.

## Tests

- `mcp/sts_injection_test.go` and `models/llm_sts_injection_test.go` — end-to-end, nothing stubbed but the upstream: a real runner, `llmagent`, the real transports with their real 30m timeouts, and a real `TokenPropagationPlugin` exchanging against an `httptest` STS.
- `sts/plugin_test.go`, `models/passthrough_token_test.go`, `a2a/executor_test.go` and the Python equivalents — session recovery precedence, the resolution chain, the no-caller-token cases, and the context keys coexisting without overwriting each other.

Red/green, reverting only the exchanged-token step:

```
--- FAIL: TestSTSExchangedTokenReachesMCPTool
    invocation[0] Authorization = "", want "Bearer EXCHANGED-STS-TOKEN"
--- FAIL: TestSTSExchangedTokenReachesLLM
    llm request[0] Authorization = "Bearer RAW-CALLER-TOKEN", want "Bearer EXCHANGED-STS-TOKEN"
```

Reverting only the caller-token gate:

```
--- FAIL: TestHeaderProvider_RecoversSessionID/cached_token_but_no_caller_token_on_this_request
    Authorization header = "Bearer EXCHANGED-STS-TOKEN", want ""
```

Green:

```
go build ./...                             clean
go vet ./adk/...                           clean
gofmt -l adk/                              clean
go test -race -skip 'TestE2E.*' ./adk/...  all packages ok
ruff check / ruff format --check           clean
pytest kagent-adk/tests/unittests          415 passed, 1 skipped
```

## Unrelated, noticed while here

In the Python runtime `set_passthrough_key` writes the token onto the model object, and ADK caches the resolved runner, so the model is built once and shared for the life of the process. Two concurrent requests can cross tokens. This predates the change and is not touched by it, but an exchanged token carries more authority than a caller token, so the same bug gets a worse outcome. The Go path is unaffected: there the token travels per request in the context. Happy to open a separate issue.


go/adk/pkg/models/base.go
```diff
@@ -77,24 +77,67 @@ func withConnectTimeout(base http.RoundTripper, connectTimeout time.Duration) (h
...
+// ExchangedTokenProviderKey is the context key for storing an
+// ExchangedTokenProvider. The A2A executor stamps it alongside the bearer token
+// and session ID.
+var ExchangedTokenProviderKey = &contextKey{name: "exchanged-token-provider"}
+
+// ExchangedTokenProvider returns the STS-exchanged token cached for a session.
+// It is an interface because the sts package imports this one, so the dependency
+// can only run the other way through a context value.
+type ExchangedTokenProvider interface {
+	GetTokenForSession(sessionID string) string
```
mesutoezdil, line 97: #2459 removes this method, would this break the build?

ricardomiguel-os: Thanks @mesutoezdil, for the heads-up.

Yes, it would break the build.
Do you know if it's expected to have #2459 reviewed and merged soon?

Or I can change the interface here from `GetTokenForSession(sessionID)` to `GetToken(ctx)` first. Passing the request instead of the session ID means the lookup keeps working after the rekey, so depending on the PR merge order, it will be a simple change rather than breaking the build

EItanya (commented):
🤖 AI-generated review.

I found **one issue I’d fix before merging** at `7efd5114`.

1. **A later request can use another caller’s cached credentials.** The new [Go LLM lookup](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/go/adk/pkg/models/base.go#L134) and [Python equivalent](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/python/packages/kagent-adk/src/kagent/adk/_bearer_token.py#L77) select tokens using only the session ID. Requiring a nonempty caller token doesn’t establish ownership of that cached credential.

   Reproduced with sequential requests on one session: Alice’s request sends `DELEGATED-ALICE`; Bob’s subsequent request also sends `DELEGATED-ALICE`, without another STS exchange.

   The [repaired Go MCP lookup](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/go/adk/pkg/sts/plugin.go#L285) also introduces a propagate-only regression: after the caller token changes, the cached first token overwrites the current request’s token.

   **Recommended fix:** integrate the caller-aware cache changes from #2459/#2460 and update these consumers accordingly. Propagate-only MCP should read the current request’s bearer directly.

Focused Go checks and **73 Python tests passed**; three additional regression probes reproduced the issue. No cluster E2E was run.

EItanya:
🤖 AI-generated follow-up.

The [`ExchangesTokens()` check in main](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/go/adk/cmd/main.go#L193) feels like a boundary worth tightening. The typed-nil guard is valid Go, but `main` also has to understand which plugin modes produce credentials suitable for LLM calls. Meanwhile, the model selects the session cache entry, and the executor carries the provider through request context.

I’d keep the existing STS plugin and make its provider contract explicitly request-aware:

```go
type ExchangedTokenProvider interface {
    ExchangedToken(ctx context.Context) (string, bool)
}
```

The implementation would own mode handling, caller identity, cache selection, and expiry. Illustratively, with helper names standing in for the implementation:

```go
func (p *TokenPropagationPlugin) ExchangedToken(
    ctx context.Context,
) (string, bool) {
    if p.integration == nil {
        return "", false // No exchanged credential in propagate-only mode.
    }

    bearer := auth.BearerTokenFromContext(ctx)
    if bearer == "" {
        return "", false
    }

    // Select by session + hash of the current caller's bearer,
    // and validate credential expiry.
    return p.cachedTokenForRequest(ctx, bearer)
}
```

`BeforeRunCallback` would use the same caller-aware identity when populating the cache; exchange timing can remain there. This fits the cache fixes discussed in #2459/#2460.

Then inject this provider into the model and MCP clients at construction:

- **LLM:** when API-key passthrough is enabled, use the exchanged credential if available; otherwise use the current request’s bearer, preserving this PR’s fallback behavior.
- **MCP:** retain the existing ordering: forward the current caller’s authorization when propagation is enabled, override it if an exchanged credential exists for this request, then apply static headers with their existing precedence.
- **Propagate-only mode:** forward the current request’s bearer directly. There is no need to cache the raw bearer.
- **Executor:** continue carrying request identity in context, with the provider held as an explicit dependency of its consumers.

The shared operation is the exchanged-credential lookup. Each transport keeps its existing caller-token forwarding rules, so this does not accidentally make MCP forward a raw bearer after an exchange failure when propagation is disabled.

The builder can normalize an absent plugin to a genuinely nil interface once. That leaves ordinary dependency wiring in `main`, while the STS component owns the credential semantics. Moving only the current conditional into a helper would leave the underlying ownership issue in place.

EItanya:
🤖 AI-generated follow-up.

There are similar cleanups in the Python implementation. I’d keep this PR’s cleanup focused on explicit dependencies, one owner for exchanged-token lookup, and unambiguous request inputs.

1. **Inject the provider into its consumers.**  
   [`set_exchanged_token_provider()`](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/python/packages/kagent-adk/src/kagent/adk/_bearer_token.py#L35) changes process-wide state from CLI startup. Instead, pass the provider into `LLMPassthroughPlugin` and `KAgentEmbedding`, threading the latter through the app’s memory-service construction. Each application then owns its dependency.

2. **Give the STS plugin a caller-aware lookup.**  
   The equivalent contract could be:

   ```python
   class ExchangedTokenProvider(Protocol):
       def exchanged_token(
           self, *, session_id: str, bearer_token: str
       ) -> str | None: ...
   ```

   The plugin owns mode handling, caller-aware cache keys, and expiry. Both the LLM path and MCP header provider use that lookup. Currently, Python’s MCP header provider reads the cache directly, separately from the new LLM lookup. The cache write path must use the same caller-aware identity, consistent with #2460.

3. **Separate request-context access from token resolution.**  
   Currently, `resolve_passthrough_token(inbound_token=None)` means “read the ContextVar,” making an explicitly absent credential indistinguishable from an omitted argument. Make those inputs required. Illustratively:

   ```python
   def resolve_passthrough_token(
       provider: ExchangedTokenProvider | None,
       *,
       inbound_token: str | None,
       session_id: str | None,
   ) -> str | None:
       if not inbound_token:
           return None

       if provider is not None and session_id:
           exchanged = provider.exchanged_token(
               session_id=session_id,
               bearer_token=inbound_token,
           )
           if exchanged:
               return exchanged

       return inbound_token
   ```

   The LLM callback supplies its request values explicitly. Embeddings read the task-local ContextVars and supply those values explicitly. ContextVars remain useful for background tasks. This helper preserves the LLM fallback behavior; MCP should share the exchanged-token lookup while retaining its own caller-token forwarding rules.

4. **Separately, remove credential mutation from model objects.**  
   `set_passthrough_key()` changes model state and invalidates cached clients. Request-specific credentials should be supplied when making the outbound call. That deserves a separate change with concurrency coverage; it predates this PR and is not an additional regression finding here.

ricardomiguel-os:
> 🤖 AI-generated review.
> 
> I found **one issue I’d fix before merging** at `7efd5114`.
> 
> 1. **A later request can use another caller’s cached credentials.** The new [Go LLM lookup](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/go/adk/pkg/models/base.go#L134) and [Python equivalent](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/python/packages/kagent-adk/src/kagent/adk/_bearer_token.py#L77) select tokens using only the session ID. Requiring a nonempty caller token doesn’t establish ownership of that cached credential.
>    Reproduced with sequential requests on one session: Alice’s request sends `DELEGATED-ALICE`; Bob’s subsequent request also sends `DELEGATED-ALICE`, without another STS exchange.
>    The [repaired Go MCP lookup](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/go/adk/pkg/sts/plugin.go#L285) also introduces a propagate-only regression: after the caller token changes, the cached first token overwrites the current request’s token.
>    **Recommended fix:** integrate the caller-aware cache changes from #2459/#2460 and update these consumers accordingly. Propagate-only MCP should read the current request’s bearer directly.
> 
> Focused Go checks and **73 Python tests passed**; three additional regression probes reproduced the issue. No cluster E2E was run.

Thanks @EItanya,
You are right. Which option do you suggest I follow?

1. Store a hash of the caller bearer on `TokenCacheEntry`, both readers return nothing on mismatch. Simple change, explicitly a stopgap that #2459/#2460 supersede wholesale.
2. Queue 2803 behind #2459/#2460 and pick up `cacheKey{sessionID, subject}`.
3. Incorporate the PR #2459 changes into this one.

#2459 is the right model — keying on subject survives token rotation, mine doesn't. 
Do you know where PR #2459 stands?

EItanya:
> https://github.com/kagent-dev/kagent/pull/2459 is the right model — keying on subject survives token rotation, mine doesn't.
Do you know where PR https://github.com/kagent-dev/kagent/pull/2459 stands?

I am going to merge that PR now, let's use that approach

ricardomiguel-os:
> > #2459 is the right model — keying on subject survives token rotation, mine doesn't.
> > Do you know where PR #2459 stands?
> 
> I am going to merge that PR now, let's use that approach

Thanks @EItanya 

> Recommended fix: integrate the caller-aware cache changes from https://github.com/kagent-dev/kagent/pull/2459 / https://github.com/kagent-dev/kagent/pull/2460 and update these consumers accordingly. Propagate-only MCP should read the current request’s bearer directly.

I've implemented the bot changes; now it's just missing the Python counterpart (https://github.com/kagent-dev/kagent/pull/2460). Should I wait for it and rebase?

ricardomiguel-os:
> 🤖 AI-generated follow-up.
> 
> The [`ExchangesTokens()` check in main](https://github.com/kagent-dev/kagent/blob/7efd51140d99d57d57080cbfd51c9de7ff3c1006/go/adk/cmd/main.go#L193) feels like a boundary worth tightening. The typed-nil guard is valid Go, but `main` also has to understand which plugin modes produce credentials suitable for LLM calls. Meanwhile, the model selects the session cache entry, and the executor carries the provider through request context.
> 
> I’d keep the existing STS plugin and make its provider contract explicitly request-aware...

- `ExchangedToken(ctx) (string, bool)` on the plugin owns mode handling, session recovery, caller identity and expiry.
- It is the only read of the token cache. `HeaderProvider` is now a thin wrapper over it, so the LLM and MCP paths cannot drift apart.
- `ExchangesTokens()` is gone, and with it the condition in `main`. Only the typed-nil guard remains.
- Propagate-only no longer caches the raw bearer.
- MCP ordering is unchanged: propagate forwards, an exchanged credential overrides, static headers still win.

ricardomiguel-os:
> 🤖 AI-generated follow-up.
> 
> There are similar cleanups in the Python implementation. I’d keep this PR’s cleanup focused on explicit dependencies, one owner for exchanged-token lookup, and unambiguous request inputs...

1. **Injection.** `set_exchanged_token_provider` and the module global are gone. The provider is passed into `LLMPassthroughPlugin` and `KAgentEmbedding`, the latter threaded through `KAgentApp` → `KagentMemoryService`.
2. **One caller-aware lookup.** `exchanged_token(*, session_id, bearer_token)` on the plugin. `header_provider` goes through it, so the MCP path gained three behaviours it did not have: the caller-token requirement, an expiry check, and returning nothing in propagate-only mode. It also no longer caches the raw subject token.
3. **Required inputs.** `resolve_passthrough_token(provider, *, inbound_token, session_id)`, exactly your signature. Embeddings read the ContextVars and pass them in.
4. **`set_passthrough_key`.** Agreed it is separate. Do you want me to open an issue or even a tentative PR for this?

EItanya (commented):
🤖 AI-generated follow-up review.

Rechecked `92512a0d8f95dc6d5430f1b91bdc86ef89019e88`. The Go caller-isolation fix is working: the original regression probe now sends `DELEGATED-ALICE` for Alice and `DELEGATED-BOB` for Bob, with separate exchanges. Python provider injection and the required explicit resolver inputs are also implemented.

Two correctness issues remain:

1. **Python still shares cached credentials between callers.** [`exchanged_token()`](https://github.com/kagent-dev/kagent/blob/92512a0d8f95dc6d5430f1b91bdc86ef89019e88/python/packages/agentsts-adk/src/agentsts/adk/_base.py#L175) checks that `bearer_token` is nonempty, but still selects the cache entry using only `session_id`; the write path is also session-only. Repeating the sequential Alice/Bob probe still returns `DELEGATED-ALICE` for both callers, with only one STS exchange. This is the outstanding Python half of the original finding. Integrate the caller-aware cache changes from #2460 and update both readers and writers before merging.

2. **The Python cleanup regresses propagate-only MCP authentication.** With `KAGENT_PROPAGATE_TOKEN=true`, no STS configuration, and the default `allowed_headers=None`, [the STS header provider now returns no authorization header](https://github.com/kagent-dev/kagent/blob/92512a0d8f95dc6d5430f1b91bdc86ef89019e88/python/packages/agentsts-adk/src/agentsts/adk/_base.py#L169). The Python MCP builder only forwards the raw header when `authorization` is explicitly allowlisted; it does not automatically forward it because propagation is enabled. A probe constructing the production agent and MCP header provider returns `{}` on this head, while the same probe passes on the previously reviewed `7efd5114`. Explicitly allowlisting `authorization` also passes. Replace the removed raw-token caching with direct forwarding of the current request's bearer in propagate-only mode, while preserving the rule that exchange failure must not enable raw-token forwarding when propagation is disabled.

The remaining **Go design follow-up** is separate from those correctness issues: credential lookup and mode handling now belong to the plugin, but [the executor still stores the provider in context](https://github.com/kagent-dev/kagent/blob/92512a0d8f95dc6d5430f1b91bdc86ef89019e88/go/adk/pkg/a2a/executor.go#L474), and model consumers retrieve it there. The requested construction-time injection into those consumers is still outstanding. Request identity can remain in context; the provider should be their explicit dependency.

The pre-existing Python `set_passthrough_key()` model-mutation issue remains separate, as agreed.

Validation: focused Go tests passed, including the original two-caller LLM regression probe; **78 existing Python tests passed**. Additional Python probes reproduced both issues above. The Python isolation probe mocks STS exchange, and the forwarding probe checks the constructed header provider without an external MCP server. No cluster E2E was run.

go/adk/pkg/models/base.go
ricardomiguel-os: I will resolve this conversation since the origin PR was merged and I've rebased from main

ricardomiguel-os:
Thanks for the input @EItanya

1. Should I wait for this merge https://github.com/kagent-dev/kagent/pull/2460? It has open comments from another reviewer.
2. I've fixed this on https://github.com/kagent-dev/kagent/pull/2803/changes/079f98cd8c3040fe1179dd02e19a2fbbd438a629

Related to the design decision of having the provider in the context, I made a change to give it to the consumers when they are built.

Issue #2762, opened 2026-09-08
Computed EgressDestinations are stored but never applied, and the OTLP endpoint is absent from them


Rachael-Graham:
> **Disclosure:** this issue was written by an AI agent (Claude) while documenting kagent's
> egress and observability surfaces, and reviewed by a human before filing.

## Summary

kagent computes an egress hostname allowlist per revision and writes it to the database and
nowhere else. Nothing applies it to the Actor, even though Substrate now implements an
`ActorEgressPolicy` API that could consume it. Separately, the OTLP exporter endpoint is
never added to that list, so enabling enforcement later would silently break tracing and
audit logging.

Filing these together because they are one gap with one landing zone, but they can be split
if preferred.

## 1. The computed list has no consumer

Every harness compiler sets it:

- `go/core/internal/translator/kagent/compiler.go` — `EgressDestinations: slices.Compact(compiled.Egress)`
- `go/core/internal/translator/byo/compiler.go`, `claude/compiler.go`, `codex/compiler.go` — same

`go/core/internal/controller/reconciler.go` then stores it on the revision row. A full
`grep -rn 'EgressDestinations' --include='*.go' go` shows consumers only under
`internal/database/`, `internal/translator/`, and that one reconciler line. Nothing under
`internal/substrate/` reads it, so it reaches the `ActorTemplate` nowhere and is recorded
metadata rather than an allowlist.

`grep -rn 'EgressPolicy\|ActorEgressPolicy' --include='*.go' go` returns **zero hits**,
while substrate exposes `Get`/`Create`/`Update`/`DeleteActorEgressPolicy` with a real
server-side implementation (`cmd/ateapi/internal/controlapi/egress_policy.go`).

Note substrate's `EgressPolicy` is documented as default-deny — "A request is denied when no
rule matches" — so whichever component starts creating policies decides the failure mode for
everything not on the list.

## 2. The OTLP endpoint is missing from the list

`agentConfigDestinations` in `go/core/internal/translator/adkconfig/builder.go` derives
destinations from HTTP tools, SSE tools, and the model provider's hosts.
`CompileSkillResources` (`translator/skills.go`, `appendArtifactSourceDestination`) adds the
registry or repository host per skill source. The OTel exporter endpoint is never added.

That is harmless only while the list is unenforced. Once it is enforced, an Actor loses its
route to the collector and both documented observability paths break with **no error and no
failure message** — no traces, no audit records, nothing pointing at egress.

## Expected

[...]

PR #2805, opened 2026-09-11
feat(observability): add OTEL traces and logs to harnesses

### LINKED ISSUE #2760: otel.logging Helm settings never reach agent runtimes, so the documented audit-logging path emits nothing
> **Disclosure:** this issue was written by an AI agent (Claude) while verifying kagent's
> audit-logging documentation, and reviewed by a human before filing.

## Summary

`otel.logging.enabled=true` in the Helm chart configures the controller only. The
controller forwards a fixed five-name allowlist of **tracing** variables to agent runtimes
and no logging variable is on it, so the Helm setting is inert for an AgentInstance even
though the GenAI prompt and reply events are emitted by the runtime.

Verified both ways on a live cluster: the Helm path emits nothing, and the same
configuration applied through `Harness.spec.env` works.

## The allowlist

```go
// go/core/internal/translator/otel.go:13-19
var otelEnvNames = []string{
	"OTEL_TRACING_ENABLED",
	"OTEL_EXPORTER_OTLP_ENDPOINT",
	"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
	"OTEL_EXPORTER_OTLP_PROTOCOL",
	"OTEL_EXPORTER_OTLP_TRACES_PROTOCOL",
}
```

`OtelEnvFromProcess()` copies only those names out of the controller's environment, and it
is called from one place, `go/core/internal/translator/kagent/compiler.go:77`.

Meanwhile the chart does set the logging variables on the controller —
`helm/kagent/templates/controller-configmap.yaml` renders `OTEL_LOGGING_ENABLED` and the
`OTEL_EXPORTER_OTLP_LOGS_*` family. Confirmed on a running cluster:

```
$ kubectl get configmap kagent-controller -n kagent -o jsonpath='{.data}' | ...
OTEL_LOGGING_ENABLED = false
OTEL_TRACING_ENABLED = true
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT = http://jaeger.jaeger.svc.cluster.local:4317
```

So the value exists on the controller and is never forwarded to the runtime that would act
on it.

## Workaround that works, and why

Setting the variables on `Harness.spec.env` does reach the runtime. It is authoritative
precisely *because* logging variables are off the allowlist: `HarnessEnvironment(harness)`
is appended before `OtelEnvFromProcess()` and `DedupeEnv` keeps the last write, so the
controller overrides `spec.env` for the five tracing variables but not for any logging
variable.

That also means the workaround inverts if logging variables are ever added to the
allowlist, so it is not a stable contract to document.

## Expected

Either the logging variables join the forwarded allowlist so `otel.logging.*` reaches agent
runtimes, or the chart documents that the setting is controller-only and names the supported
per-Harness path.

## Note on a second, separate gate

The Go ADK elides log message content unless
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true`. The same variable has the
**opposite** default for traces, where the span payload is emitted unless the value is
explicitly `"false"`. That asymmetry appears deliberate and is not part of this report.

## Related

#1273 fixed an OTel logging problem in the 0.7.x Python runtime. This is a different path:
the v1alpha3 controller translator forwarding environment to Substrate Actors.

supreme-gg-gg:
Adds correlated OpenTelemetry tracing and logging across the Kagent, Codex, and Claude harnesses. Refactors shared tracing configuration and setup code. Builds on top of previous V2 observability works like #2663 and #2786. Also closes #2760 and partially addresses #2762.

### Key Changes

- Resolve and validate OTLP trace and log settings consistently across all harness compilers.
- Propagate telemetry configuration, service identity, and collector egress into runtime revisions.
- Configure Codex and Claude native trace and log exporters with content capture enabled.
- Propagate W3C trace context through Codex JSON-RPC requests and Claude process environment.
- Initialize shared Go tracing in the native harness wrappers and move common tracing code out of the ADK package.
- Flush request spans before terminal and input-required events cause an Actor to pause.

### Notes

- Traces and logs are independently enabled through `OTEL_TRACING_ENABLED` and `OTEL_LOGGING_ENABLED`
- Invalid or missing enabled-signal endpoints and unsupported protocols now fail harness compilation.
- Span attributes for Claude Code is [documented here](https://code.claude.com/docs/en/monitoring-usage#span-attributes) and available events / logs is [documented here](https://code.claude.com/docs/en/monitoring-usage#events)
- Events / logs emitted by Codex is [documented here](https://learn.chatgpt.com/docs/config-file/config-advanced#what-gets-emitted). There is no official docs for Codex's span hierarchy and attributes
- Codex and Claude Code force flushes traces when their process exists before the gateway Suspends the instance. The adapter process will also flush its own traces before it responds with a terminal event as per the changes in `go/adk/pkg/a2a/server/server.go`. However, during a pending HITL, since we do not quit the Codex or Claude process, it is not guaranteed that all traces are delivered immediately.
- For both Codex and Claude, native spans after HITL completes remain in the original request trace because the in-flight process carries the same trace context (through env var or JSON-RPC). The resume A2A trace may contain only controller and Go wrapper spans.


supreme-gg-gg:
This recording shows the traces and logs for ADK, Claude, and Codex harness in Grafana


https://github.com/user-attachments/assets/07ba44ad-a55a-4f09-8260-eae197e222b1

go/pkg/tracing/tracing.go
```diff
@@ -0,0 +1,141 @@
...
+
+func environmentValue(name, fallback string) string {
+	if value := strings.TrimSpace(os.Getenv(name)); value != "" {
+		return value
+	}
+	return fallback
+}
+
+// ForceFlush exports spans buffered by the global tracer provider.
+func ForceFlush(ctx context.Context) error {
```
mesutoezdil, line 93: this drops log flushing, the exact fix in #2772

supreme-gg-gg: This doesn't "drop" log flushing since we didn't have it before, I'd leave it to #2772 and one of us can rebase on the other. Note that this log flushing also does not apply to codex and claude harness in this PR, it only applies to ADK agents

go/adk/pkg/a2a/server/tracing.go
```diff
@@ -0,0 +1,41 @@
...
+	case *a2atype.Task:
+		state = payload.Status.State
+	default:
+		return nil
+	}
+	if !state.Terminal() && state != a2atype.TaskStateInputRequired && state != a2atype.TaskStateAuthRequired {
+		return nil
+	}
+	span := trace.SpanFromContext(ctx)
+	span.End()
```
krisztianfekete, line 59: Since this ends otelhttp's server span, so some attributes are likely lost. You can start your own child span in `Before` and end that one instead, or capture the status yourself before ending.

go/core/internal/translator/claude/compiler.go
```diff
@@ -70,10 +79,43 @@ func (c *Compiler) Compile(ctx context.Context, input *v2translator.HarnessInput
...
+			logsExporter = "otlp"
+		}
+		environment = append(environment,
+			corev1.EnvVar{Name: "CLAUDE_CODE_ENABLE_TELEMETRY", Value: "1"},
+			corev1.EnvVar{Name: "CLAUDE_CODE_ENHANCED_TELEMETRY_BETA", Value: "1"},
+			corev1.EnvVar{Name: "OTEL_TRACES_EXPORTER", Value: tracesExporter},
+			corev1.EnvVar{Name: "OTEL_METRICS_EXPORTER", Value: "none"},
+			corev1.EnvVar{Name: "OTEL_LOGS_EXPORTER", Value: logsExporter},
+			corev1.EnvVar{Name: "OTEL_LOG_USER_PROMPTS", Value: "1"},
+			corev1.EnvVar{Name: "OTEL_LOG_TOOL_DETAILS", Value: "1"},
```
krisztianfekete: These are all off by default, so maybe we should wire these up to Helm, so it's opt-in? We should do this for all similar stuff as well (outside of this PR).

go/core/internal/translator/claude/compiler.go
```diff
@@ -70,10 +79,43 @@ func (c *Compiler) Compile(ctx context.Context, input *v2translator.HarnessInput
...
+		)
+		if traceConfig.Enabled {
+			environment = append(environment, corev1.EnvVar{Name: "OTEL_LOG_TOOL_CONTENT", Value: "1"})
+		}
+		if logConfig.Enabled {
+			environment = append(environment,
+				corev1.EnvVar{Name: "OTEL_LOG_ASSISTANT_RESPONSES", Value: "1"},
+				corev1.EnvVar{Name: "OTEL_LOG_RAW_API_BODIES", Value: "1"},
```
krisztianfekete: These should potentially be opt-in as well, controlled separately from prompt logging as I'd expect this to be quite noisy/expensive?

go/core/internal/translator/otel.go
```diff
@@ -1,31 +1,149 @@
...
+	}
+	switch protocol {
+	case "grpc", "http/protobuf":
+	default:
+		return signalConfig{}, fmt.Errorf("unsupported OTLP %s protocol %q", signal, protocol)
+	}
+
+	parsed, err := url.Parse(endpoint)
+	if err != nil || (parsed.Scheme != "http" && parsed.Scheme != "https") || parsed.Hostname() == "" || parsed.User != nil || parsed.RawQuery != "" || parsed.Fragment != "" {
+		return signalConfig{}, fmt.Errorf("OTLP %s endpoint must be an absolute HTTP(S) URL without credentials, query, or fragment", signal)
```
krisztianfekete: Can we check this at startup, and log a warning if it's wrong instead?

go/harness/codex/internal/adapter/adapter.go
```diff
@@ -158,6 +180,14 @@ func renderConfig(cfg config.Config, codexHome string) ([]byte, error) {
 			},
 		}
 	}
+	if cfg.Telemetry != nil {
+		native.Otel = &nativeOtel{
+			LogUserPrompt: cfg.Telemetry.CaptureContent,
+			Environment:   "kagent",
```
krisztianfekete: What's the Codex default here?

go/core/internal/translator/codex/compiler.go
```diff
@@ -82,12 +96,29 @@ func (c *Compiler) Compile(ctx context.Context, input *v2translator.HarnessInput
...
 		environment = append(environment, envVar)
 	}
+	template, harness := input.Root.Template, input.Harness
+	environment = append(environment,
+		corev1.EnvVar{Name: env.KagentName.Name(), Value: template.Name + "-" + harness.Name},
+		corev1.EnvVar{Name: env.KagentNamespace.Name(), Value: template.Namespace},
+		corev1.EnvVar{Name: preResponseTraceFlushEnv, Value: "true"},
+	)
+	environment = append(environment, traceConfig.Environment()...)
+	environment = append(environment, logConfig.Environment()...)
```
krisztianfekete: I think we have to set `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` so we are fully consistent with our wrappers, don't we? Likely in both Codex and Claude.

go/core/internal/translator/codex/compiler.go
```diff
@@ -71,7 +84,8 @@ func (c *Compiler) Compile(ctx context.Context, input *v2translator.HarnessInput
 	}
 	environment := append(providerEnvironment, mcp.environment...)
 	for _, variable := range input.Harness.Spec.Env {
-		if _, reserved := ownedEnvironment[variable.Name]; reserved || strings.HasPrefix(variable.Name, mcpCredentialPrefix) {
+		_, reserved := ownedEnvironment[variable.Name]
+		if reserved || strings.HasPrefix(variable.Name, mcpCredentialPrefix) || strings.HasPrefix(variable.Name, otelEnvironmentPrefix) {
```
krisztianfekete: We have to pass some `OTEL_` env vars to not break functionality (I am thinking mainly about the export intervals, but maybe even the endpoints?)

go/core/internal/translator/otel.go
```diff
@@ -1,31 +1,149 @@
...
+}
+
+// LogConfig is the controller-owned log export configuration compiled into
+// each runtime revision.
+type LogConfig struct {
+	Enabled  bool
+	Endpoint string
+	Protocol string
+	hostname string
+}
```
krisztianfekete: Maybe this could be a `SignalConfig` instead?

krisztianfekete (commented):
Thanks @supreme-gg-gg, this is looking really good, added some comments/questions!

cc. @dhaifley w/r/t the flushing problems you were working on recently.

go/harness/codex/internal/adapter/adapter.go
supreme-gg-gg: The default value for `environment` should be "dev" here: https://learn.chatgpt.com/docs/config-file/config-advanced#observability-and-telemetry; I'll remove this `kagent` setting

go/core/internal/translator/codex/compiler.go
supreme-gg-gg: Yea I'll set them for both harness

go/core/internal/translator/claude/compiler.go
supreme-gg-gg: For this and the following comments, I think we can have two helm values:

1. one for including details like user prompt, tool, assistant response etc. which will apply to relevant settings for both traces and logs, defaulted to false (an alternative could be to have them configurable for traces and logs separately, but I assume most user would want both since traces would not contain all the details and would result in very granular settings)
2. the other one for logging API bodies, since that one could be noisy and rarely useful compared to the first one

go/core/internal/translator/codex/compiler.go
supreme-gg-gg: I don't see an option to configure service name in the docs unfortunately. We're already propagating `traceparent` and `tracestate` so that would give us trace correlation. Maybe we can just let each native CLI own their service identity?

go/adk/pkg/a2a/server/tracing.go
EItanya, line 59: **Early span completion permanently loses HTTP response attributes.**

This ends `otelhttp`'s server span before its middleware records response metadata. I confirmed the concern already raised here with a regression probe: adding an assertion for HTTP status 200 to `TestSpansExportedBeforeResponseBodyCloses` fails at this head because the exported server span contains only request attributes. The same probe passes when substituting the base revision of `server.go`. The later middleware updates cannot modify an already-ended span.

Please use an explicitly owned request span for the early flush and leave HTTP span completion to `otelhttp`.

EItanya (commented):
🤖 AI-generated review. Validated the HTTP response attribute loss discussed below.

supreme-gg-gg:
@krisztianfekete I've addressed your comments in https://github.com/kagent-dev/kagent/pull/2805/commits/ad7cf7f76b343312f8ef2481c4b2bbd95aa40058, thanks for the review!

go/core/internal/translator/codex/compiler.go
krisztianfekete: Most SDKs should honor `OTEL_RESOURCE_ATTRIBUTES` by defaukt, can you please check if that's the case here?

krisztianfekete (commented):
Looks good to me! I am a bit paranoid about potential flushing problems still, can we maybe have an e2e test to ensure that part os robust as well?

go/core/internal/translator/codex/compiler.go
supreme-gg-gg: I tried setting this for both harness but I don't seem to see the custom attributes. For Codex I think this is a known issue https://github.com/openai/codex/issues/30987, not too sure why it's not working for Claude

supreme-gg-gg:
@krisztianfekete I've added a flush-traces E2E test in https://github.com/kagent-dev/kagent/pull/2805/commits/57cf9a775aee1ff9c4a73fc022907e867bd9e4d5, can you take a look?

.github/workflows/ci.yaml
```diff
@@ -167,6 +167,30 @@ jobs:
           export KAGENT_E2E_CLI="$PWD/core/bin/kagent-local"
           go test -v github.com/kagent-dev/kagent/go/core/test/e2e -failfast -shuffle=on -parallel 4
 
+      - name: Run completed-chat trace flush e2e test
```
EItanya: Why does this require a separate test run?

supreme-gg-gg: I'll merge this into the step above

EItanya (approved):

PR #2806, opened 2026-09-12
fix(adk): return tool errors to model

### LINKED ISSUE #2794: [BUG] Python runtime fails the whole turn when the model calls a tool name that is not registered; the Go runtime hands the error back to the model
## Description

On the Python declarative runtime (`kagent-app`), a model that emits a function call whose name is not in `tools_dict` kills the entire A2A turn: google-adk's `_get_tool` raises `ValueError("Tool '<name>' not found. …")`, no `on_tool_error_callback` is registered that turns it into a response, the exception escapes `handle_function_call_list_async`, and `_agent_executor` publishes the task as `failed`. The user sees a failed turn carrying the raw ADK error text; the model never gets the chance to correct itself.

The Go runtime (`golang-adk`) does not have this failure mode: `google.golang.org/adk` `handleFunctionCalls` builds a `FunctionResponse` `{"error": "tool '<name>' not found. Available tools: …"}` for an unknown name and the loop continues, so the model retries with a registered tool.

### How it happened in practice

An agent whose tools are the meta-tools of an MCP aggregator (`filter_tools`, `describe_tool`, `call_tool`, …) had called `call_tool(name="x_kubernetes_list", …)` three times successfully. On the fourth step the model emitted `x_kubernetes_list` as the function name itself. Log from `kagent-app:0.10.1-full`:

```
kagent_adk.kagent.adk._agent_executor - ERROR - Error handling A2A request: Tool 'x_kubernetes_list' not found.
Available tools: list_tools, describe_tool, list_core_tools, filter_tools, call_tool, list_resources, describe_resource, get_resource, list_prompts, describe_prompt, get_prompt, ask_user, skills, bash, read_file, write_file, edit_file
...
  File ".../google/adk/flows/llm_flows/functions.py", line 505, in _execute_single_function_call_async
    raise tool_error
  File ".../google/adk/flows/llm_flows/functions.py", line 491, in _execute_single_function_call_async
    tool = _get_tool(function_call, tools_dict)
  File ".../google/adk/flows/llm_flows/functions.py", line 1049, in _get_tool
    raise ValueError(error_msg)
```

Every proxy/aggregator tool pattern makes this likely: the model sees the real tool names as strings in results and eventually calls one directly. Observed rate on one installation: 1 direct call in about 460 proxied calls over 14 days, and each one is a dead turn.

## Expected behaviour

A tool name the model made up is the model's mistake to correct. The runtime should hand it back as a function response error, the way the Go runtime does and the way a failing MCP tool already does (`McpTool.run_async` returns `{"error": …}`), instead of failing the task.

## Root cause and upstream status (google-adk)

- kagent `release/v0.10.x` pins `google-adk>=1.38.0,<2` (lock 1.38.0); `main` pins `>=2.6.2,<3` (lock 2.8.0). In both, `_execute_single_function_call_async` re-raises when no `on_tool_error_callback` returns a response (`functions.py` L491–505 in 1.38.0).
- google/adk-python fixed exactly this on 2026-08-26: `f3377937 fix: report an unknown tool name to the model instead of raising` (plus `f842a15c fix: balance callback lifecycle for hallucinated tool calls`, 2026-08-31). Both commits are only on `release/candidate` (2.x). They are not in v2.8.0, not in v1.39.1 and not on the `v1` branch, so the 1.x line that kagent 0.10.x uses will not receive them.

## Proposed fix

1. `kagent-adk` registers a plugin whose `on_tool_error_callback` returns `{"error": str(error)}` (the shape `McpTool.run_async` and the Go runtime use), added to the plugin list built in `python/packages/kagent-adk/src/kagent/adk/cli.py` (both the `static` and the `run` paths). This also covers every other exception a tool raises; #2727 describes the same failure class for `ask_user`. google-adk's built-in `ReflectAndRetryToolPlugin` (since 1.17) is an alternative with a retry cap and reflection guidance, at the cost of a different response shape.
2. Later, once a google-adk release carries f3377937, `main` picks it up through its `<3` range; `release/v0.10.x` still needs (1).

Related: #2727 (same failure class, `ask_user` raising), #1061 (earlier report of this error, closed as a model bug).

## Version

kagent 0.10.1 (`kagent-app:0.10.1-full`), `spec.declarative.runtime: python`, google-adk 1.38.0, Anthropic provider. The Go runtime (`golang-adk`, google.golang.org/adk/v2 v2.1.0) is not affected.

tabish-malik13:
## Summary

- add a fallback ADK plugin that returns unhandled tool errors to the model as structured function responses
- preserve custom tool error handlers by running the fallback last
- use the configured plugin set in local CLI test runs as well as served agents
- add coverage for unknown tool calls and plugin ordering

Fixes #2794

## Testing

- `ruff format --check`
- `ruff check`
- `pytest packages/kagent-adk/tests -q` (404 passed, 1 skipped)

### DIFF
--- a/python/packages/kagent-adk/src/kagent/adk/_a2a.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_a2a.py
@@ -36,6 +36,7 @@ from ._agent_executor import A2aAgentExecutor, A2aAgentExecutorConfig
 from ._lifespan import LifespanManager
 from ._memory_service import KagentMemoryService
 from ._token import KAgentTokenService
+from ._tool_error_plugin import ToolErrorPlugin
 from .types import AgentConfig
 
 logger = logging.getLogger(__name__)
@@ -86,7 +87,9 @@ class KAgentApp:
         self.app_name = app_name
         self.agent_card = agent_card
         self._lifespan = lifespan
-        self.plugins = plugins if plugins is not None else []
+        # Custom handlers get the first chance to recover. The final fallback turns
+        # unhandled tool errors into responses the model can reason about.
+        self.plugins = [*(plugins or []), ToolErrorPlugin()]
         self.stream = stream
         self.agent_config = agent_config
 
@@ -248,9 +251,9 @@ class KAgentApp:
         )
 
         root_agent = self.root_agent_factory()
+        adk_app = App(name=self.app_name, root_agent=root_agent, plugins=self.plugins)
         runner = Runner(
-            agent=root_agent,
-            app_name=self.app_name,
+            app=adk_app,
             session_service=session_service,
             artifact_service=InMemoryArtifactService(),
         )
--- a/python/packages/kagent-adk/src/kagent/adk/_tool_error_plugin.py
+++ b/python/packages/kagent-adk/src/kagent/adk/_tool_error_plugin.py
@@ -1,0 +1,23 @@
+"""Convert tool failures into responses that an agent can reason about."""
+
+from typing import Any
+
+from google.adk.plugins import BasePlugin
+from google.adk.tools import BaseTool, ToolContext
+
+
+class ToolErrorPlugin(BasePlugin):
+    """Return tool errors to the model instead of aborting the turn."""
+
+    def __init__(self):
+        super().__init__(name="tool_error")
+
+    async def on_tool_error_callback(
+        self,
+        *,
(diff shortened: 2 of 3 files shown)


EItanya:
Thanks for the fix. How did you decide on this solution, I'd be surprised if adk-python didn't have a better builtin solution to this problem

Issue #2746, opened 2026-09-08
Recover interrupted checkpoint operations and allow failed checkpoint cleanup


EItanya:
Follow-up from the review of #2742, separate from revision GC scheduling in #2744.

A checkpoint that reaches `FAILED` still retains `prepared_revision`, but `BeginDeleteAgentInstanceCheckpoint` only accepts `READY` and `DELETING`. The public delete path therefore reports NotFound for the failed record instead of cleaning it up. Retiring the template and deleting its source instance cannot release that revision.

An interrupted `CREATING` reservation is more disruptive: it retains the revision and blocks task admission and source-instance lifecycle changes. Retrying the original create request can resume it, but there is no automatic recovery when the caller disappears. A failed compensation can also leave the reservation in `CREATING`.

Implement recovery in the checkpoint lifecycle owner:

- Allow failed checkpoints to be deleted through the public API, including the case where tag creation never succeeded or compensation already removed the tag.
- Resume or compensate abandoned creation and deletion after restart. Coordinate recovery with in-flight create requests so a slow, live tag copy cannot be reaped solely because it is old.
- Keep revision/context references until the lifecycle owner has established that cleanup is complete. Revision GC must continue to count every checkpoint row; filtering out `FAILED` or `CREATING` would bypass the safety invariant.
- Preserve request idempotency, tag identity checks, and fork-reference protection. Do not hold database locks across Substrate calls.

Acceptance coverage: PostgreSQL and real Substrate tests for failed tag creation, failed compensation, lost responses, and restart between reservation/tag creation/finalization; deleting a failed checkpoint must release its references, abandoned reservations must stop blocking their source instance, and a concurrent live create or fork must remain safe.

Issue #2744, opened 2026-09-08
Add runtime revision GC backlog and failure metrics


EItanya:
Follow-up to #2742, which now implements independent runtime revision GC under the existing controller leader election: startup and one-minute periodic sweeps, per-candidate deadlines, failure isolation, durable deletion recovery, and preparation polling while a digest is deleting.

This issue tracks the remaining cleanup observability work:

- Expose the pending runtime cleanup count and oldest pending cleanup age.
- Count cleanup failures without revision or template labels that create unbounded metric cardinality.
- Retain revision-specific context in error logs and document how operators distinguish a growing backlog from a persistently failing Substrate deletion.
- Verify metrics recover after successful retries and controller restarts. A failed candidate must remain visible while other candidates are collected.

The lifecycle E2E in #2742 now checks eventual ActorTemplate and golden-actor cleanup through the public status API after releasing the last checkpoint, without a later template change. Controller tests cover startup discovery, retries, failure isolation, deadlines, and cancellation. PostgreSQL tests retain the reference-acquisition and UID-safety coverage.

A concrete operational case from validation is a Substrate deletion returning `Internal: ... runsc state: exit status 128` after an API/DNS outage. Its durable deletion marker survives restart. GC must report this backlog while continuing unrelated preparation and cleanup. Checkpoint-specific recovery remains separate in #2746.

PR #2808, opened 2026-09-13
feat: add runtime revision GC backlog and failure metrics

gargsajal9:
## Summary

Addresses [#2744](https://github.com/kagent-dev/kagent/issues/2744) by adding visibility into runtime-revision garbage collection: pending cleanup count and failures grouped by a fixed set of stages.

## Metrics Implementing

- `kagent_runtime_revision_gc_pending`: count from the existing discovery query.
- `kagent_runtime_revision_gc_failures_total{stage}`: failed attempts, with `discovery` and `collection` stages.

## Testing

- Coverage includes unit tests, real PostgreSQL/migration checks. Existing reference-safety tests passed; the new observation-race regression was removed with the observer.
- Recovery evidence: repeated same-object deletion failures did not block unrelated cleanup; removing the fault allowed pending count to return to zero.

## Scope and review focus

This PR registers collectors only -  the metrics listener remains disabled, with export/authentication deferred

Checkpoint recovery in [#2746](https://github.com/kagent-dev/kagent/issues/2746) is outside scope. Feedback is requested on bounded metric dimensions.

## AI assistance

GitHub Copilot assisted with implementation and test coverage

I confirm that I reviewed the changes.

### DIFF
--- a/docs/architecture/runtime-revision-cleanup.md
+++ b/docs/architecture/runtime-revision-cleanup.md
@@ -1,0 +1,131 @@
+# Runtime revision cleanup
+
+Runtime revision garbage collection runs in the controller holding the leader
+lease. It discovers work at startup and every minute, independently of template
+preparation. Each discovery, backlog observation, and candidate attempt has a
+one-minute deadline. Sweeps are serial: a slow candidate delays later work, but a
+failed candidate does not prevent other candidates from being attempted.
+
+An eligible revision has no active desired or last-successful pair reference,
+AgentInstance reference, or checkpoint reference. All checkpoint states retain
+their references. After claiming a revision for deletion, GC preserves the row
+and ActorTemplate identity until Substrate cleanup and database finalization
+succeed. Do not manually remove these references or deletion markers to silence
+an alert.
+
+## Enable authenticated metrics
+
+The controller metrics listener is disabled by default. The existing Helm
+settings enable the dedicated metrics Service:
+
+```yaml
+controller:
+  metrics:
+    enabled: true
+    bindAddress: ":8443"
+    secureServing: true
+    service:
+      port: 8443
+```
+
+The chart sets `METRICS_BIND_ADDRESS` and `METRICS_SECURE`. Outside Helm, an
+unset/empty bind address or `"0"` disables the listener. Secure serving defaults
+to true; an invalid `METRICS_SECURE` value fails startup. An explicit false
+enables unauthenticated HTTP and should only be used on an appropriately
+restricted development listener.
+
+Secure scrapes use HTTPS and a Kubernetes bearer token. The controller's
+existing metrics-auth ClusterRole permits TokenReviews and SubjectAccessReviews.
+Bind `<fullname>-metrics-reader` to the scraper's ServiceAccount to allow
+`GET /metrics`, and configure the scraper to trust the listener certificate.
+Do not disable certificate verification or authentication as an operational
+workaround. No ServiceMonitor is created by this feature.
+
+GC metrics use the registry served by the controller manager, not the public
+gRPC/MCP application port. A successful HTTP response alone is not evidence
+that the GC metrics are present.
+
+## Metrics and timestamps
+
+| Metric | Type | Meaning |
+| --- | --- | --- |
+| `kagent_runtime_revision_gc_pending` | Gauge | Eligible persisted revisions in the latest successful backlog observation, including unattempted candidates and incomplete deletions. |
+| `kagent_runtime_revision_gc_oldest_pending_age_seconds` | Gauge | Time since the oldest candidate in that observation was first durably discovered as eligible. Zero for an observed empty backlog. |
+| `kagent_runtime_revision_gc_failures_total{stage}` | Counter | Failed operations, separated by a fixed stage label. Counts attempts, not unique revisions. |
+| `kagent_runtime_revision_gc_last_successful_backlog_observation_timestamp_seconds` | Gauge | Unix timestamp of the active collector's latest successful observation; zero until its first successful read. |
+
+Backlog gauges have no application labels. Failure stages are limited to
+`discovery`, `backlog_observation`, `begin_deletion`, `get_actor_template`,
+`uid_check`, `delete_actor_template`, and `finalize`. No metric labels contain
+revision IDs, template names, namespaces, UIDs, or error text. Prometheus can
+still attach its ordinary deployment, pod, and scrape-target labels.
+
+`cleanup_pending_since` records the first durable discovery of a continuous
+eligible period. This is a lower bound on cleanup waiting time, not revision
+creation time and not the exact time its last reference was released. Reference
+reacquisition clears it atomically, so a later eligible period starts afresh.
+Retries and controller restarts preserve it. Existing deletion claims retain
+their original `deleted_at` age during migration; historical unclaimed rows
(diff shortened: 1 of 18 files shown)


gargsajal9:
looking for early review

EItanya (commented):
🤖 AI-generated review, reviewed and approved by me before posting.

Could we simplify this first version to two metrics?

- `kagent_runtime_revision_gc_pending`: count from the existing discovery query.
- `kagent_runtime_revision_gc_failures_total{stage}`: failed attempts, with `discovery` and `collection` stages.

These would already show whether cleanup is falling behind or repeatedly failing, with existing logs providing revision-specific details.

I’d defer oldest pending age. Making it durable adds a migration, writes to reference-acquisition paths, and transaction/locking logic for an observability feature. We can add it separately if count trends and failure rates prove insufficient.

For backlog updates, reuse discovery at the start and refresh once at the end of each sweep. The current full observation after every candidate introduces quadratic database work precisely when the backlog grows. If discovery fails, retain the previous count and increment the discovery failure counter.

That would keep this focused on two standard Prometheus collectors and the authenticated listener, without schema or lifecycle changes.

gargsajal9:
> 🤖 AI-generated review, reviewed and approved by me before posting.
> 
> Could we simplify this first version to two metrics?
> 
> * `kagent_runtime_revision_gc_pending`: count from the existing discovery query.
> * `kagent_runtime_revision_gc_failures_total{stage}`: failed attempts, with `discovery` and `collection` stages.
> 
> These would already show whether cleanup is falling behind or repeatedly failing, with existing logs providing revision-specific details.
> 
> I’d defer oldest pending age. Making it durable adds a migration, writes to reference-acquisition paths, and transaction/locking logic for an observability feature. We can add it separately if count trends and failure rates prove insufficient.
> 
> For backlog updates, reuse discovery at the start and refresh once at the end of each sweep. The current full observation after every candidate introduces quadratic database work precisely when the backlog grows. If discovery fails, retain the previous count and increment the discovery failure counter.
> 
> That would keep this focused on two standard Prometheus collectors and the authenticated listener, without schema or lifecycle changes.

I have updated the PR incorporated changes suggested by you. Please have a look again!

go/core/pkg/app/app.go
```diff
@@ -335,6 +345,21 @@ func Run(ctx context.Context, opts Options) error {
 	return group.Wait()
 }
 
+func controllerMetricsOptions() (metricsserver.Options, error) {
+	secure, err := strconv.ParseBool(env("METRICS_SECURE", "true"))
```
EItanya: Env vars should all be registered in the env package

go/core/pkg/app/app.go
```diff
@@ -335,6 +345,21 @@ func Run(ctx context.Context, opts Options) error {
...
+	if err != nil {
+		return metricsserver.Options{}, fmt.Errorf("parse METRICS_SECURE: %w", err)
+	}
+	options := metricsserver.Options{
+		BindAddress:   env("METRICS_BIND_ADDRESS", "0"),
```
EItanya: Same comment as above

go/core/test/e2e/README.md
```diff
@@ -43,6 +43,35 @@ input continuation, and cancellation.
...
+
+`TestRuntimeRevisionGCMetricsLifecycle` adds an actual authenticated metrics
+scrape to the existing public runtime-revision lifecycle scenario. It requires
+a dedicated installation with secure controller metrics enabled, an initially
+empty cleanup backlog, and a direct HTTPS URL for the active GC leader:
+
+```bash
+KAGENT_E2E_METRICS_URL=https://<active-controller>:8443/metrics
+KAGENT_E2E_METRICS_TOKEN_FILE=<path-to-scraper-bearer-token>
+KAGENT_E2E_METRICS_CA_FILE=<path-to-trusted-metrics-ca>
```
EItanya: Testing/adding secure metrics support feels separate from the feature you're trying to implement, can you do that in a follow-up to keep scope down?

go/core/internal/controller/runtime_revision_gc.go
```diff
@@ -58,22 +66,46 @@ func (r *RuntimeRevisionGC) Start(ctx context.Context) error {
...
+		return
+	}
+	revisions, err := r.discover(ctx)
 	if err != nil {
-		logging.FromContext(ctx).ErrorContext(ctx, "failed to list unreferenced runtime revisions", "error", err)
 		return
 	}
 	// ponytail: sweeps are serial; add bounded workers if slow deletions delay reclamation.
 	for _, candidate := range revisions {
 		if ctx.Err() != nil {
```
EItanya, line 74: I feel like we're making this check ALL over the place, can we potentially reduce the number of times?

docs/architecture/runtime-revision-cleanup.md
```diff
@@ -0,0 +1,120 @@
...
+failed candidate does not prevent other candidates from being attempted.
+
+An eligible revision has no active desired or last-successful pair reference,
+AgentInstance reference, or checkpoint reference. All checkpoint states retain
+their references. After claiming a revision for deletion, GC preserves the row
+and ActorTemplate identity until Substrate cleanup and database finalization
+succeed. Do not manually remove these references or deletion markers to silence
+an alert.
+
+## Enable authenticated metrics
```
EItanya: Same comment about decoupling secure metrics from this feature

go/core/pkg/app/app.go
gargsajal9: Removed form this PR. Will open follow up issue and PR

go/core/pkg/app/app.go
gargsajal9: Removed form this PR. Will open follow up issue and PR

go/core/internal/controller/runtime_revision_gc.go
gargsajal9: Reduced checks

docs/architecture/runtime-revision-cleanup.md
gargsajal9: Removed form this PR. Will open follow up issue and PR

go/core/test/e2e/README.md
gargsajal9: Removed form this PR. Will open follow up issue and PR

gargsajal9:
@EItanya csn you review again?

krisztianfekete (changes requested):
Thanks for this. FYI #2929 is about to land and moves the controller's metrics onto OpenTelemetry (we are changing our approach and preferring OTel for metrics as well) with one MeterProvider with a Prometheus reader on the existing /metrics registry + OTLP export. Once it merges, could you rebase and switch these to the OTel metrics API?

High level workflow for adding new metrics is this, but we'll document this soon:
 - Names go through the Weaver registry in telemetry/registry, and the policies expect `kagent.runtime_revision.gc.pending (Int64Gauge, unit {revision})` and `kagent.runtime_revision.gc.failures (Int64Counter, unit {failure})`, with stage registered as `kagent.gc.stage`.  The `/metrics` path still renders them as `kagent_runtime_revision_gc_*`.
 - Please inject the MeterProvider and pin the shape with a `ManualReader` and `telemetrytest.FindMetric`, like the shape tests in #2929.
 - Note that the metrics endpoint is already served since #2886, via `controller.metrics.enabled`

 The bounded discovery/collection stage set looks right to me.

PR #2811, opened 2026-09-13
fix: record GenAI tool-call arguments and result on execute_tool spans

### LINKED ISSUE #2734: execute_tool spans carry no gen_ai.tool.call.arguments or .result, so tool calls render empty in OTel-native backends
## Summary

kagent's `execute_tool` spans identify the tool but carry none of the exchange. They include
`gen_ai.tool.name`, `gen_ai.tool.call.id`, `gen_ai.tool.call.type` and `gen_ai.tool.description`, and no
`gen_ai.tool.call.arguments` or `gen_ai.tool.call.result`.

Backends that follow the OTel GenAI semantic conventions read those two attributes for a tool span's
input and output, so the span renders as a tool call with an empty body. The information is not lost, but
it is only reachable indirectly: the assistant's tool request appears in the model span's
`gen_ai.output.messages`, and the tool's result appears in the *next* model call's
`gen_ai.input.messages`.

## Environment

- kagent 0.9.9 (Helm chart), agent on Amazon Bedrock (Nova Pro)
- OpenTelemetry Collector, exporting to a GenAI-aware backend
- Python auto-instrumentation injected by the OpenTelemetry Operator

## What we observe

One tool-using interaction produces this tree:

```
invocation                      no gen_ai content
└── invoke_agent <agent>        no gen_ai content
    ├── call_llm                no gen_ai content
    │   └── generate_content    gen_ai.input.messages + gen_ai.output.messages  (the tool REQUEST)
    │       └── execute_tool X  gen_ai.tool.name/.call.id/.call.type/.description only
    └── call_llm                no gen_ai content
        └── generate_content    gen_ai.input.messages contains the tool RESULT
```

So the one span named for the tool is the one span that cannot show what the tool did.

## Why it matters

Anyone reading a trace looks first at the span named after the action. In our case that is a workshop
where people are learning to inspect agent behaviour, and the tool span is exactly where they look and
exactly where nothing is. More generally, an operator debugging "what did the agent actually pass to this
tool, and what came back" has to reconstruct it by opening two different model spans and diffing their
message arrays, which is not obvious and does not survive being explained once.

We tried to fill the gap at the collector and could not: a transform copying `tool_call_args` /
`tool_response` onto the span matched nothing, because those attributes are not present on it in any
form. Correlating across spans (reading the next model call's input) is not something OTTL can express,
since it operates per span.

## Suggested change

Set the two attributes on the `execute_tool` span at the point the tool is invoked and returns:

- `gen_ai.tool.call.arguments` — the arguments passed to the tool
- `gen_ai.tool.call.result` — what the tool returned

Both are in the OTel GenAI conventions for tool spans, and the values are already in hand at that point
in the call path.

Two things worth considering alongside it, if useful:

1. **Errors.** A tool that fails currently leaves `status.code` unset, so a failed call is
   indistinguishable from a successful one in the trace. Setting an error status when the tool result is
   an error would make failures findable.
2. **Content capture.** If these attributes should be gated the way message content is
   (`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`), gating them the same way would keep the
   behaviour predictable for operators who deliberately disable content capture.

Happy to test a change against our setup and report back; we run this daily across a fleet of clusters
and can reproduce on demand.

TigerkidYang:
## Problem

`execute_tool` spans identify the tool (`gen_ai.tool.name`, `.call.id`, `.description`) but carry the exchange only under ADK's own `gcp.vertex.agent.tool_call_args` / `gcp.vertex.agent.tool_response` keys. Backends that follow the OTel GenAI semantic conventions read `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result`, so the one span named after the tool renders with an empty body and operators reconstruct the call from the surrounding model spans (or run a collector transform, as the reporter does today).

Both runtimes are affected: `adk-go` v2.2.0 (`internal/telemetry/telemetry.go`) and `google-adk` 2.8.0 (`telemetry/tracing.py`) only emit the `gcp.vertex.agent.*` keys.

## Fix

Both ADKs run after-tool callbacks inside the `execute_tool` span context, so the mapping is done from kagent's own callbacks without touching ADK:

- **Go** (`go/adk`): new `telemetry.SetToolCallAttributes(ctx, args, result, err)` called from `makeAfterToolCallback`. Payload capture reuses `marshalSpanPayload`, so it honours `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=false` (empty `{}`) and the existing 32 KiB truncation. A tool error also sets the span status to `Error`. (adk-go already sets error status from a `result["error"]`, so this only matters when a callback chain produces an error without one.)
- **Python** (`kagent-adk`): new `_tool_span.make_tool_span_attributes_callback()` registered as `after_tool_callback` on the kagent `Agent`. Content capture follows ADK's `TelemetryConfig` lookup (per-request `capture_message_content`, otherwise `ADK_CAPTURE_MESSAGE_CONTENT_IN_SPANS`), matching how ADK gates its own `gcp.vertex.agent.*` payloads. Error status is not touched here because `google-adk` 2.8.0 already sets it in `trace_tool_call`.

The `gcp.vertex.agent.*` attributes are left in place; this adds the semconv keys alongside them.

## Tests

- Go: `TestSetToolCallAttributes_OnActiveSpan`, `_ErrorSetsStatus`, `_EmitsEmptyPayloadWhenContentCaptureDisabled` in `go/adk/pkg/telemetry`. `go vet` + `go test -race ./adk/pkg/telemetry/ ./adk/pkg/agent/` pass.
- Python: `tests/unittests/test_tool_span.py` (5 tests: attributes recorded, non-dict result wrapped, env switch off, per-request `TelemetryConfig` off, no-op without a recording span). `ruff check` / `ruff format --check` clean. The `kagent-adk` unit suite has 12 pre-existing failures on my Windows machine (TLS cert fixtures and env-var materialization) that are identical with and without this change.

Not verified against a live cluster; @peopleforrester offered to test on their fleet, which would be the ideal confirmation.

Fixes #2734

---
This change was prepared with Claude Code assistance; the diff was reviewed and the tests above were run locally before opening the PR.


EItanya:
@krisztianfekete

python/packages/kagent-adk/src/kagent/adk/_tool_span.py
```diff
@@ -0,0 +1,60 @@
...
+GEN_AI_TOOL_CALL_ARGUMENTS = "gen_ai.tool.call.arguments"
+GEN_AI_TOOL_CALL_RESULT = "gen_ai.tool.call.result"
+
+
+def _should_capture_content(tool_context: ToolContext) -> bool:
+    """Mirrors ADK's telemetry config lookup: per-request RunConfig first, then env defaults."""
+    invocation_context = getattr(tool_context, "_invocation_context", None)
+    run_config = getattr(invocation_context, "run_config", None)
+    config = getattr(run_config, "telemetry", None) or TelemetryConfig()
+    return bool(config.should_add_content_to_legacy_spans)
```
mesutoezdil: why gate gen_ai keys with the legacy-spans capture flag?

TigerkidYang: Fair point — these are semconv `gen_ai.*` keys, so they should follow the OTel content switch, not ADK's legacy-span knob. Switched `_should_capture_content` to `TelemetryConfig.should_add_content_to_experimental_spans` (the span-bearing `SPAN_ONLY` / `SPAN_AND_EVENT` modes of `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, per-request `capture_message_content` respected), which is also what the issue asked for.

Note this makes the Python side opt-in by default (OTel default is `NO_CONTENT`), whereas the Go side keeps kagent's existing `marshalSpanPayload` convention (same env var, capture unless `false`). Happy to align Go to opt-in too if you'd prefer the two runtimes to match.

PR #2820, opened 2026-09-14
fix(ui): ten fixes in the agent rail and the schedules empty state

Charlesthebird:
This PR:
- Fixes an issue with the agent details left sidebar app extension items not always being rendered.
- Fixes content layout shift issues on the chat page so that loading feels smoother.
- The rest are minor cosmetic fixes

--- 

🤖  Written by claude from here:

Ten fixes in the agent rail and the schedules list. Each is independent, and none changes behaviour for a build with no extension installed.

| # | Was | Now |
|---|---|---|
| 1 | `agentLinks.details` was built and never read, so "Agent Details" always led to the pair-keyed page | `agentHref` follows the declared link |
| 2 | `newChatHref` would have followed it | Still the agent's own address — `/agents/:id/new` serves nothing |
| 3 | Contributions got only a conversation id, so an entry vanished on the pair-only pages | `ExtensionAgentRailItemProps` also carries `pair` |
| 4 | A collapsed rail cancelled a hardcoded `space(6)` gap | Measures the row's own `column-gap` |
| 5 | `ConversationStateDot` drew a state that no longer varies | Removed, with the optimistic chain that fed it |
| 6 | Agent Details, the card's second line and select-all arrived with the record, moving the rail | Drawn in place, inert, before it lands |
| 7 | The rail's sticky offset assumed this app's own header | Reads `--agent-rail-sticky-top`, same default |
| 8 | Schedules said "No schedules yet" and always reserved an 800px scroll floor | Says what it means; the floor needs rows |
| 9 | New chat stood 2px shorter as a button than as the link it replaces | It keeps the row's own border, so the rail does not change height |
| 10 | The example extension declared no `agentLinks` | It does, which is what gives these points browser coverage |

## Two worth a sentence

**4** — `AgentChatPage` uses `gap: theme.space(2)`; the other three rail surfaces use `space(6)`. The rail cancelled 24px of an 8px gap, dragging the gutter past the left edge of the page and under the sidebar.

**7** — a shell that replaces the header starts its content below its own chrome, so the default offset landed twice. The rail's wrapper is tall enough that it does not take a sticky offset while the short gutter does, so the two drifted apart rather than moving together.

## Testing

```
yarn typecheck        clean
yarn lint             0 errors
yarn test             467 passed
yarn test:pw          121 passed (chromium + chromium-with-extension)
```

Each new assertion fails with its own fix reverted. Firefox was not exercised; it is not installed on the machine this was written on.

---
*🤖 written by Claude*

PR #2825, opened 2026-09-15
test(ui): one lifecycle journey per resource, and bugs fixes found during rewrite

### LINKED ISSUE #2700: [UI] Playwright e2e tests should be one test per resource CRUD
Our UI playwright tests don't need to test every case. We should have a single test file per resource (mcp/model/prompt/etc) that covers the CRUD flow. This way if the resource fails, we see the entire flow and where it fails in the video output (this should result in one video per resource CRUD sequence).

We could also use this opportunity to go over and remove extra Playwright tests that are unnecessary. This would help cut down the test time.

toreysoloio:
- merge related test files, reorganized tests into a more CRUD-oriented flow, and reorganized test files/folders a bit
- merged/removed various additional tests to decrease test time
- added more helper functions for some some and/or flacky actions
- updated scheduling components/tests to use testid instead of detecting elements via page copy
- added various `useInvalidate...()` helpers in components to make sure lists are promptly updated after CRUD actions

---
*🤖 written by Claude (start)*

The Playwright suite had a spec per behaviour, so a resource that broke gave you several short videos and none of them showed that the state one test established is the state the next acted on. It is now one journey per resource — create, read it back, edit it, delete it, with its empty and failure states among the steps — which is one video per resource and one place to look.

Closes #2700

## The suite, before and after

| | before | after |
|---|---|---|
| spec files | 33 | 25 |
| tests | 239 | 157 |
| wall clock, `yarn test:pw` | 2.3m | 2.1m |
| failing tests per full run | 2 | 0–1 |

Measured on one machine: two runs of `main`, fourteen of this branch. Worth reading honestly — cutting a third of the tests bought about 9% of the wall clock, not a third, because the suite is parallel and bounded by its slowest specs rather than by how many there are.

## Flakes, all Firefox

Each was diagnosed by instrumenting the failing click rather than by inference. Rates are per-test, under three times the normal parallelism — a stress harness rather than CI.

| | before | after |
|---|---|---|
| `chat-scroll-bottom` never dismissed | 1 / 8 | 0 / 20 |
| rename did not apply | 1 / 40 | 0 / 24 |
| share link never appeared | 0 / 40 | 0 / 24 |
| conversation rename / agent delete | 3 / 12 | 0 / 22 |
| template edit could not be saved | 1 / 5 full runs | 0 / 30 |

Two causes. The scroll-bottom click was dispatched at the button's centre and landed on the panel behind it, because the button had moved between the coordinates being taken and the event going out — a layout bug, fixed in the page. The rest were clicks computed while antd was still zooming a dialog in: the rename click landed 238px left and 224px below a button that had not moved at all. Playwright's stability check compares two animation frames, and a starved main thread can serve both from the same frame of the animation; `pressOnce` compares samples on a clock instead.

The share-link row is the weak one — it never reproduced before the change either, so it carries no evidence. It is the same shape as the rename and gets the same treatment on that reasoning alone.

### One flake this branch did not fix

Eleven of the fourteen full runs were clean. The other three each failed the same test, `chat/questions.spec.ts` — a `chat-send` click whose reply never arrives. It is **not from this branch**: it reproduces identically with this branch's chat-page fix reverted, which is the control that settles it. It passes 20/20 in isolation and needs full-suite contention to appear, so it is a pre-existing flake this suite is now quiet enough to notice rather than a new one. Untouched here deliberately — speculative hardening is what caused the one regression this branch had to back out.

### Merged with main

`main` is merged in, including #2820, which lands on four of the same files. Worth knowing for the chat page in particular: that PR also fixes layout shift there, on the rail's contents, while this one fixes the right-hand column — different elements, same class of bug, no overlap. Its `--agent-rail-sticky-top` change is applied to both wrappers this branch restructured. Two full runs since the merge, both clean.

## Five product bugs the journeys found

1. The model provider picker listed a provider once per model rather than once, and logged a duplicate React key.
2. Creating a model, MCP server or harness returned to a list still showing the set without it, which reads as a create that silently failed.
3. A failed list re-read after a *successful* create was reported as a failed create, on seven pages — beside a Try again that re-posts and comes back 409.
4. A failed harness read rendered as an empty tab: "there are none" where the truth was "could not be read".
5. The chat page gated its whole right-hand column on the instance read, so the conversation shifted sideways when the record landed — 20px at the collapse breakpoint, the panel's full width above it.

Agent templates and schedules then turned out to have the same cache staleness as (2): templates had no invalidation hook at all and its three writers refreshed a key the page they navigate to does not hold, and schedules had one wired only to delete. Both are fixed here.

## Testing

```bash
yarn test:pw          # chromium + firefox, no cluster needed
```

For the product fixes, with `ENABLE_MOCK_UI=true yarn dev`:

1. Open an agent conversation in a wide window — it no longer jumps sideways as it finishes loading.
2. Create a model at `/models/new` — the new row is in the list you land on.
3. Create an agent template, then delete an agent from its page — neither leaves a stale row behind.
4. Open `/agents?mock=error` and pick the Harnesses tab — it says the read failed instead of showing an empty table.
5. Open `/models/new` and the provider picker — each provider appears once.

---
*🤖 written by Claude (end)*


ui/src/pages/AgentTemplateDetailsPage.tsx
```diff
@@ -224,9 +217,10 @@ export function AgentTemplateDetailsPage() {
       });
       // Re-read before leaving edit mode, so the read-only view below is the saved
       // copy. The other order shows the values that were just replaced, which reads
-      // as a save that did not take.
+      // as a save that did not take. The sweep reaches the list and the agents derived
+      // from it, which this page's own read does not.
       await template.refresh();
-      await templates.refresh();
+      await invalidateTemplates();
```
Charlesthebird, line 225: **Level:** 🔴 High · Blocking

The line above this one is bug (3) from your own description, unfixed on this page: `await template.refresh()` sits inside the save's `try`, and `ApiResource.refresh()` deliberately rethrows (`useApiResource.ts:64-66`), so a failed re-read after a **successful** update renders "Could not save the agent template" and never runs `setEditingRef(undefined)` — the page stays in edit mode telling the user a save that worked did not.

```suggestion
      await template.refresh().catch(() => {});
      await invalidateTemplates();
```

---
*🤖 written by Claude*

ui/src/pages/agents/HarnessesTab.tsx
```diff
@@ -57,6 +57,10 @@ export function HarnessesTab() {
   const namespaces = useNamespaces();
   const harnesses = useHarnessesAcrossNamespaces(namespaces.data?.map((row) => row.name));
 
+  // `ListHarnesses` needs a namespace, so a failed namespace read means the harness read
```
Charlesthebird, line 65: **Level:** 🔴 High · Blocking

The same bug again at `:112`, in `remove()`: `await harnesses.refresh()` is inside the `try`, so a successful `removeHarness` followed by a failed re-read sets `failure` and the tab reports a delete that did happen as one that did not.

```suggestion
      await harnesses.refresh().catch(() => {});
```
(applied at `:112`, not here — this is the nearest changed line.)

---
*🤖 written by Claude*

Charlesthebird (commented):

---

The restructuring is sound and the flake diagnoses hold up. Three things block merge, and they are all the same bug: a re-read that fails after a write that **succeeded** is reported as the write failing. This PR fixes it for creates and leaves it on two of its own paths.

The other one worth reading before the rest: **the cache sweep does nothing on the five pages that create-then-navigate.** SWR's filtered `mutate` only refetches a key that has a mounted subscriber — `startRevalidate` returns the cached value untouched when `EVENT_REVALIDATORS[key]` is empty (`swr@2.5.1`, `dist/config-context-ext53wcz.mjs:276-288`). The destination list is unmounted at that instant, so the sweep resolves immediately with the stale set and the list's own mount revalidation does the work it would have done anyway. The hooks do real work where the reader is already on screen; the comments claim more than that.

<details>

- ✅ **Provider dedupe** — `models.providers` really does concatenate stock and configured providers, so same-type duplicates were real; `ModelForm.tsx:176-193` keeps the first of each and drops nothing the form reads.
- ✅ **Harness read failure** — `namespaces.error ?? harnesses.error` is the right fix: `useHarnessesAcrossNamespaces` keys `null` without namespaces, so its own error never fires. `AgentsPage.tsx:124` and `AgentTemplatesPage.tsx:120` already had this; the tab was the one page missing it.
- ✅ **Layout shift** — both right-hand wrappers hold their width from the first frame, and the 1440px collapse breakpoint is untouched.
- ✅ **Cache key shapes** — every new hook's predicate matches its readers: `models.list`/`models.get` (correctly excluding the provider catalogue), `mcpServers.list`/`tools.list`, `harnesses.list`/`listAll`, `agentTemplates.list`/`listAll`/`get`.
- ✅ **Models error coverage** — all four steps of the deleted `models-errors.spec.ts` land in the journey at `models.spec.ts:283-320`.
- ✅ **Shell coverage** — `shell-chrome.spec.ts`'s two tests carry into `app-shell.spec.ts` intact.
- ⚠️ **Required-field coverage** — ported per-resource, except one assertion that was narrowed silently. Inline.
- ⚠️ **The suite's own conventions** — `README.md` says assert against the list rather than a toast, and press a dialog's button with `pressOnce`. New code breaks both. Inline.

</details>

---
*🤖 written by Claude*

ui/src/api/hooks/useInvalidateModels.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export function useInvalidateModels(): () => Promise<void> {
+  const { mutate } = useSWRConfig();
+
+  return useCallback(async () => {
+    await mutate(
+      (key) =>
+        Array.isArray(key) &&
+        typeof key[0] === "string" &&
+        CONFIGURATION_KEYS.includes(key[0]),
```
Charlesthebird: **Level:** 🟠 Medium · Blocking

This sweep is a no-op on the five pages that call it and then navigate. SWR's filtered `mutate` only refetches a key with a mounted subscriber — `startRevalidate` returns `get().data` untouched when `EVENT_REVALIDATORS[key]` is empty (`swr@2.5.1`, `dist/config-context-ext53wcz.mjs:276-288`) — and the destination list is unmounted at that moment, so `await` resolves on the stale set and the list still renders it until its own mount revalidation lands.

Worth knowing that the journeys do not discriminate: `toHaveCount(1)` auto-retries past the mount revalidation, so `models.spec.ts` step 6 passes with these hooks removed. Either keep the destination's read subscribed on the writer page, or drop "so the list lands showing it" from the five comments. The sweep does work where the reader is already on screen (`AgentTemplatesPage.tsx:225`).

---
*🤖 written by Claude*

ui/src/pages/AgentTemplatesPage.tsx
```diff
@@ -218,13 +220,15 @@ export function AgentTemplatesTab() {
                   row.name,
                 )
               }
-              onDeleted={templates.refresh}
+              /* The agents list reads templates under a key of its own, so refreshing
+                 this list alone leaves an agent there that nothing runs. */
+              onDeleted={invalidateTemplates}
```
Charlesthebird, line 225: **Level:** 🟠 Medium · Not Blocking

The delete path has bug (3) too, and this PR does not reach it: `DeleteResourceButton.tsx:78` calls `onDeleted()` inside the same `try` as `onDelete()`, so a failed re-read after a successful delete shows "Could not delete …" with `duration: Infinity`.

Six handlers can trigger it, all passing an `ApiResource.refresh` that rejects — `ModelsPage.tsx:172`, `McpServersPage.tsx:229`, `PromptsPage.tsx:163`, `UnmappedConversationsPage.tsx:138,166`, `AgentPage.tsx:385`. One guard in `DeleteResourceButton` covers all of them:

```js
      await onDelete();
      setOpen(false);
      toast.success(`Deleted ${kind} ${name}`);
      await onDeleted().catch(() => {});
```

---
*🤖 written by Claude*

ui/src/pages/ModelNewPage.tsx
```diff
@@ -13,9 +13,13 @@ import { apiClient, type CreateModelConfigRequest } from "@/api";
  */
 export function ModelNewPage() {
   const navigate = useNavigate();
+  const invalidateModels = useInvalidateModels();
 
   async function createModel(payload: CreateModelConfigRequest): Promise<void> {
     await apiClient.models.create(payload);
+    // Before navigating, so the list lands showing it. Swallowed: the write has
+    // already succeeded, and the list reports its own read failure on arrival.
+    await invalidateModels().catch(() => {});
```
Charlesthebird, line 22: **Level:** 🟠 Medium · Not Blocking

This guard cannot fire. `mutate(filter)` with no data never rejects — the revalidator catches into hook state and returns (`swr@2.5.1`, `dist/use-swr-emll9s78.mjs:435-464`) — and with no mounted subscriber it does not even fetch. Same at `McpServerNewPage.tsx:154`, `HarnessNewPage.tsx:94`, `AgentTemplateNewPage.tsx:67`, `PromptNewPage.tsx:22`, `PromptEditPage.tsx:61`, `ScheduledRunForm.tsx:87`, `AgentPage.tsx:726`.

`ModelEditPage.tsx:40-41` is the one place the guard is load-bearing, because `ApiResource.refresh()` does reject. Either drop the seven dead ones and their "Swallowed:" comments, or say they are belt-and-braces — a comment describing a failure mode that cannot happen is the kind this repo treats as a defect.

---
*🤖 written by Claude*

ui/playwright/tests/models/models.spec.ts
```diff
@@ -56,33 +77,247 @@ test("models: the list loads and renders each configuration", async ({ page }) =
...
 
-    await page.getByRole("button", { name: "Refresh" }).click();
+    await clickRefresh(page);
     await expect
       .poll(() => operationCalls(page, rpc.listModelConfigs), { timeout: 10_000 })
       .toBeGreaterThan(before);
 
+    // A refresh usually returns the same rows, so a successful one is otherwise
+    // indistinguishable from a button that did nothing.
+    await expect(page.getByText("Models refreshed")).toBeVisible();
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

`README.md:205` in this PR: "Assert against the list a user would read, not against a toast." `App.tsx:92` mounts `<Toaster>` with no `toastOptions`, so this one lives 2000ms — on the contended runs the config at `playwright.config.ts:118-130` describes, that is the transient-element flake the team convention rules out.

The RPC-count poll six lines up already proves the re-read happened. Same assertion at `prompts.spec.ts:119`, `mcp-servers.spec.ts:147` and `dashboard.spec.ts:29`, all four of them new here.

---
*🤖 written by Claude*

ui/playwright/helpers/resource.ts
```diff
@@ -0,0 +1,266 @@
...
+ * Any previous confirmation is waited out first, because a dismissed popconfirm stays
+ * visible while it animates away — a caller that has just pressed "Keep" would
+ * otherwise find that closing dialog and click the Delete inside it.
+ *
+ * The trigger is clicked exactly once: it is a toggle, so a retry closes what the
+ * first press opened. The `toBeVisible` between the two is what makes a Delete aimed
+ * at nothing fail here rather than several assertions downstream.
+ */
+export async function confirmDelete(page: Page, name: string): Promise<void> {
+  const open = confirmation(page);
```
Charlesthebird, line 235: **Level:** 🟠 Medium · Not Blocking

`README.md:194` says press a dialog's button with `pressOnce`, and this is the highest-traffic instance of exactly the click `pressOnce` was written for — a popconfirm's OK, mid-zoom, from six lifecycle specs. `agent-page.spec.ts:474` wraps the identical click in `pressOnce`; the shared helper the other six route through still does not.

```suggestion
  await pressOnce(open.getByRole("button", { name: "Delete" }));
```

---
*🤖 written by Claude*

ui/playwright/helpers/resource.ts
```diff
@@ -0,0 +1,266 @@
...
+ * argument for why a second click is safe.
+ */
+export async function pressOnce(button: Locator): Promise<void> {
+  let previous: string | undefined;
+  await expect(async () => {
+    const where = JSON.stringify(await button.boundingBox());
+    const held = where === previous;
+    previous = where;
+    expect(held, `still arriving, at ${where}`).toBe(true);
+  }).toPass({ timeout: 10_000, intervals: [100] });
```
Charlesthebird, line 68: **Level:** 🟠 Medium · Not Blocking

`boundingBox()` returns `null` for an element with no box, and `JSON.stringify(null)` is `"null"` — so two consecutive nulls compare equal and this loop calls it stable on the first pair. A button inside a closed `.ant-modal-wrap` (antd sets `display: none`) passes the wait in 100ms and falls straight through to the check this helper exists to replace.

```suggestion
  await expect(button).toBeVisible();
  let previous: string | undefined;
  await expect(async () => {
    const box = await button.boundingBox();
    const where = JSON.stringify(box);
    const held = box !== null && where === previous;
    previous = where;
    expect(held, `still arriving, at ${where}`).toBe(true);
  }).toPass({ timeout: 10_000, intervals: [100] });
```

---
*🤖 written by Claude*

ui/playwright/tests/agents/agent-page.spec.ts
```diff
@@ -518,8 +510,10 @@ test("agents: conversations can be picked and deleted together from the table to
   });
 
   await test.step("3. and deleting says what goes with it", async () => {
-    await page.getByTestId("delete-1 selected").click();
-    const prompt = page.locator(".ant-popover:visible");
+    const prompt = confirmation(page);
+    await pressUntil(page.getByTestId("delete-1 selected"), () =>
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

`resource.ts:72-73` names this exact case as the first of `pressUntil`'s three preconditions: "Retrying a control that toggles undoes the press that worked — an antd popconfirm trigger is one." `delete-1 selected` is the trigger `<Button>` wrapped by `<Popconfirm>` (`DeleteResourceButton.tsx:109,129`), so a second press closes what the first opened.

```suggestion
    await page.getByTestId("delete-1 selected").click();
    await expect(prompt).toBeVisible();
```

---
*🤖 written by Claude*

ui/eslint.config.mjs
```diff
@@ -38,4 +38,66 @@ export default tseslint.config(
       "@typescript-eslint/no-namespace": "off",
     },
   },
+
+  /*
+   * The browser suite's conventions, enforced rather than only written down.
+   *
+   * `playwright/README.md` states these; a convention nothing checks is one that
```
Charlesthebird, line 45: **Level:** 🟠 Medium · Not Blocking

`playwright/live/` is outside this glob and already breaks both rules: `live/pages.spec.ts:1`, `live/substrate.spec.ts:1` and `live/agent-lifecycle.spec.ts:1` import `{ test, expect }` straight from `@playwright/test`, and `live/agent-lifecycle.spec.ts:63,88` reach for `.ant-select-item-option` and `.ant-popconfirm`. `live/schedules.spec.ts` uses the fixture, so the folder disagrees with itself.

Widening to `playwright/**/*.spec.ts` needs those four call sites fixed in the same change.

---
*🤖 written by Claude*

ui/playwright/tests/models/models.spec.ts
```diff
@@ -56,33 +77,247 @@ test("models: the list loads and renders each configuration", async ({ page }) =
...
+     * key is write-only, so the configuration comes back carrying no credential to
+     * seed the draft from, and the form opens on the mode that matches what it was
+     * given rather than on the one it was created with.
+     */
+    await page
+      .getByTestId("model-auth-type")
+      .getByText("API key", { exact: true })
+      .click();
+    await expectRequired(page, {
+      marked: [],
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

`marked: []` drops an assertion the old suite made at this same step — `forms/required-fields.spec.ts:98` had `marked: ["Name", "Namespace"]` here, and `ModelForm.tsx:470,502` still marks both unconditionally, so `identityLocked` disables the inputs without clearing the mark. Nothing in `DEFERRED.md` records the narrowing.

```suggestion
      marked: ["Name", "Namespace"],
```
Worth making `expectRequired` throw when both lists are empty, so the next one cannot pass silently.

---
*🤖 written by Claude*

ui/src/api/hooks/useInvalidateHarnesses.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * A sweep rather than one `refresh()`: the tab reads `harnesses.listAll` and a scoped
+ * caller `harnesses.list`, so refreshing one means reconstructing the other's key.
+ *
+ * Resolves once the re-reads have landed, so a caller can await it before navigating.
+ */
+export function useInvalidateHarnesses(): () => Promise<void> {
+  const { mutate } = useSWRConfig();
+
+  return useCallback(async () => {
```
Charlesthebird: [nit]

Seven of these now, identical but for the predicate. One `useInvalidateKeys(match)` in a single file would replace about 150 lines, and each resource keeps its prefix as a one-line constant where it is used.

---
*🤖 written by Claude*

ui/playwright/DEFERRED.md
```diff
@@ -133,15 +108,9 @@ Two things worth keeping from writing it, because both cost time and neither is
...
-MCP server does *not* appear on its list until the reader presses Refresh: `AgentNewPage`
-refreshes its list after a create and `PromptNewPage`, `ModelNewPage` and
-`McpServerNewPage` do not. The first draft of this journey was written against prompt
-libraries, and failing there is how that was found. The same asymmetry exists on the branch
-this one was ported from, so it is a shipping defect rather than a regression, and it is
-deliberately not fixed here — that would widen a large port. **The journey for those three
-belongs in the change that fixes them**, where it is what proves the fix, rather than
-sitting red in the suite describing a known bug.
+**Every resource reads its create back through its list now**, not only harnesses — see
+*No longer deferred: the lifecycle half* above, which also records the create-cache defect
```
Charlesthebird: [nit]

This points at a heading that the same diff renamed — it is "Covered: form validation, and every resource's lifecycle" at `:33` now.

```suggestion
*Covered: form validation, and every resource's lifecycle* above, which also records the create-cache defect
```

---
*🤖 written by Claude*

Charlesthebird (commented):

---

Four more from a second pass over the same diff. Nothing here blocks merge; the first is a regression the layout fix introduces on its own error path.

<details>
<summary>✅ What I checked</summary>

- ✅ **Filter-race coverage** — the two-write clobber the old `list-filters.spec.ts` reproduced is asserted between writes at `models.spec.ts:113-119`, so the URL round-trip is still pinned.
- ✅ **Schedules history** — expand-all, search, no-match and clear all survive the port at `schedules.spec.ts:153-185`.
- ✅ **Sweep predicates** — SWR hands the filter the unserialized key, so `Array.isArray(key) && typeof key[0] === "string"` is correct rather than lucky.
- ✅ **Empty-state scroll fix** — dropping `scroll={{ x: 800 }}` when there are no rows is real; the width floor did leave a scrollbar with nothing to scroll to.

</details>

---
*🤖 written by Claude*

ui/src/pages/AgentChatPage.tsx
```diff
@@ -735,80 +735,79 @@ export function AgentChatPage() {
...
-            {/* Slides rather than vanishing, for the same reason the rail does: an
-                unmounted panel makes the transcript jump its whole width in one frame,
-                which reads as a layout fault rather than as something closing. */}
+        {/* Slides rather than vanishing, for the same reason the rail does: an
+            unmounted panel makes the transcript jump its whole width in one frame,
+            which reads as a layout fault rather than as something closing. */}
+        <div
+          css={{
+            flexShrink: 0,
+            width: isContextOpen ? 248 : 0,
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

The error path loses the space it used to give back. `isContextOpen` defaults true on a wide window (`:350-353`), so when the instance read fails or 404s, `instance.data` stays undefined and this wrapper sits at 248px containing nothing — while the toggle above is `visibility: hidden`, so the reader cannot collapse it either. Before this change the whole block was `null` and the transcript took the width.

```suggestion
            width: isContextOpen && !instance.error ? 248 : 0,
```

---
*🤖 written by Claude*

ui/playwright/tests/models/models.spec.ts
```diff
@@ -56,33 +77,247 @@ test("models: the list loads and renders each configuration", async ({ page }) =
...
+
+    await expect(rowNamed(page, CREATED)).toHaveCount(0, { timeout: 30_000 });
+    // "Gone" has to mean that one rather than the read: a list that failed to reload
+    // is also a list the row is missing from.
+    await expect(dataRows(page)).toHaveCount(SEEDED.length);
+    await expect(rowNamed(page, "default-model-config")).toHaveCount(1);
   });
 
-  await test.step("5. an empty result says so instead of showing a bare table", async () => {
+  await test.step("11. an empty result says so instead of showing a bare table", async () => {
```
Charlesthebird, line 279: **Level:** 🟠 Medium · Not Blocking

Empty, failure and retry are the last three steps in all six resource journeys — `mcp-servers.spec.ts:210`, `prompts.spec.ts:378`, `agent-templates.spec.ts:396`, `harnesses.spec.ts:186`, `schedules.spec.ts:445`. A journey is ordered, so one broken create costs that resource its empty-state and failure coverage entirely, and those are the steps least likely to be broken by the same change.

`README.md` justifies the position by the fixture reset, but that argument permits putting them first just as well — the scenario reload starts a fresh backend either way.

---
*🤖 written by Claude*

ui/playwright/helpers/resource.ts
```diff
@@ -0,0 +1,266 @@
...
+ * control that genuinely needed pressing twice — a real product bug it can no longer
+ * catch, and the first place to look if a "click does nothing the first time" report
+ * arrives.
+ */
+export async function pressUntil(
+  button: Locator,
+  settled: () => Promise<unknown>,
+  // Half the thirty-second test budget, so that when this is what failed, this is
+  // what says so: at thirty the test expired first and reported its own timeout.
+  timeout = 15_000,
```
Charlesthebird, line 102: **Level:** 🟡 Low · Not Blocking

The budget does not bound what the comment says it bounds. `toPass` checks its deadline *between* attempts, and no `actionTimeout` is set in `playwright.config.ts`, so Playwright's default of no timeout applies — a click blocked by an overlay hangs inside one attempt and expires at the test timeout, which is the failure the comment above says this prevents.

```suggestion
  timeout = 15_000,
): Promise<void> {
  await expect(async () => {
    if (await button.isVisible()) await button.click({ timeout: 5_000 });
```

---
*🤖 written by Claude*

ui/src/pages/ModelEditPage.tsx
```diff
@@ -35,10 +35,10 @@ export function ModelEditPage() {
     if (!namespace || !name) return;
 
     await apiClient.models.update(namespace, name, payload);
-    // Refreshed before navigating, so the list lands showing the new values rather
-    // than the previous ones for a moment.
-    await models.refresh();
-    await model.refresh();
+    // Before navigating, so the list lands showing the new values. Swallowed: the
+    // write has already succeeded, and the list reports its own read failure.
+    await models.refresh().catch(() => {});
```
Charlesthebird, line 42: **Level:** 🟡 Low · Not Blocking

This page still holds `useModels()` only to have something to call `refresh()` on — the pattern `useInvalidateModels` was added to retire, in its own words ("so a writer does not have to subscribe to a list it never renders just to have something to refresh"). It fetches the whole list on a page that never draws it.

Since the destination *is* unmounted here, keeping `refresh()` is defensible — it is the one call on these pages that genuinely fetches before navigating. Worth a comment saying that, or the list subscription looks like the oversight the four new hooks exist to fix.

---
*🤖 written by Claude*

ui/src/pages/AgentTemplateDetailsPage.tsx
toreysoloio: Guarded the re-read so a failed one no longer reports a save that succeeded as failed, in `386eede2`.


---
*🤖 written by Claude*

ui/src/pages/agents/HarnessesTab.tsx
toreysoloio: Same guard applied at `:112` in `386eede2`.


---
*🤖 written by Claude*

ui/src/api/hooks/useInvalidateModels.ts
toreysoloio: Confirmed against the installed source, and by removing the sweep from `HarnessNewPage` — `harnesses.spec.ts` passes 8/8 without it, so the journeys do not discriminate either. Dropped the "so the list lands showing it" claim from all eight sites in `386eede2`; the sweep still does what it says where a reader is mounted.


---
*🤖 written by Claude*

ui/eslint.config.mjs
toreysoloio: Confirmed the gap — three `live/` specs import from `@playwright/test` and `agent-lifecycle.spec.ts` reaches for both classes. Leaving the glob as it is: `live/` runs against a cluster, so widening it means editing four call sites in a suite this change cannot run, and the shared fixture's console guard is written around the mock backend rather than a real one. Happy to take it as a follow-up that can actually be verified.


---
*🤖 written by Claude*

ui/src/pages/AgentTemplatesPage.tsx
toreysoloio: Guarded `onDeleted()` inside `DeleteResourceButton` in `75cfbb10`, which covers all six handlers.


---
*🤖 written by Claude*

ui/src/pages/ModelNewPage.tsx
toreysoloio: Kept the guards and stopped claiming they carry weight — the comments now read "Guarded, not load-bearing" (`75cfbb10`).


---
*🤖 written by Claude*

ui/playwright/tests/models/models.spec.ts
toreysoloio: Dropped all four in `75cfbb10` — models and mcp-servers already counted the RPC, and prompts and dashboard now do rather than losing the claim.


---
*🤖 written by Claude*

ui/playwright/helpers/resource.ts
toreysoloio: `confirmDelete` presses the OK through `pressOnce` in `75cfbb10`; 150/150 across the six lifecycle specs at ten workers.


---
*🤖 written by Claude*

ui/playwright/helpers/resource.ts
toreysoloio: `boundingBox()` measures `null` for an element with no box and `JSON.stringify(null)` is `"null"`, so two of those compared equal — added the visibility wait and the `box !== null` guard in `75cfbb10`.


---
*🤖 written by Claude*

ui/playwright/tests/agents/agent-page.spec.ts
toreysoloio: Clicked once with the assertion after it in `75cfbb10` — it is the popconfirm's own trigger, as you say.


---
*🤖 written by Claude*

ui/playwright/tests/models/models.spec.ts
toreysoloio: Restored `marked: ["Name", "Namespace"]`, and `expectRequired` now throws when both lists are empty so the next one cannot pass silently (`75cfbb10`).


---
*🤖 written by Claude*

ui/src/api/hooks/useInvalidateHarnesses.ts
toreysoloio: Collapsed onto one `useInvalidateKeys(names)` with a predicate each, 205 lines to 80, in `75cfbb10`.


---
*🤖 written by Claude*

ui/playwright/DEFERRED.md
toreysoloio: Repointed at the current heading in `75cfbb10`.


---
*🤖 written by Claude*

ui/src/pages/AgentChatPage.tsx
toreysoloio: The panel holds no width when the instance read failed, in `75cfbb10`.


---
*🤖 written by Claude*

ui/playwright/helpers/resource.ts
toreysoloio: Confirmed no `actionTimeout` is configured, so the click had none — bounded it at 5s in `75cfbb10`.


---
*🤖 written by Claude*

ui/src/pages/ModelEditPage.tsx
toreysoloio: Added a comment saying exactly that in `75cfbb10`: this is the one page where `refresh()` earns the subscription, because a sweep finds no mounted subscriber for the list before navigating.


---
*🤖 written by Claude*

ui/playwright/tests/models/models.spec.ts
toreysoloio: Agreed on the coupling, and that the fixture-reset argument does not choose an end — recorded it in `DEFERRED.md` in `edf85cbf` rather than reshuffling six specs at review time. Noted there that moving them first trades one coupling for another, and that the fix which actually removes it is a second `test` per file, which costs `conventions.test.ts` its one-spec-one-test rule and so belongs in its own change.


---
*🤖 written by Claude*

ui/playwright/DEFERRED.md
Charlesthebird, line 1: We should clean up this file and address these "deferred" items. Added an issue here: https://github.com/kagent-dev/kagent/issues/2828

Charlesthebird (approved):

PR #2826, opened 2026-09-15
fix(ui): keep long-running agent streams open and expose MCP requireApproval

supreme-gg-gg:
Stop applying the 30s unary gRPC-Web deadline to chat streams, so long-running agents stay connected. Streams use the existing `useChat` idle timeout instead (30min idle timer). Also adds a “Require approval” checkbox on each MCP binding in the agent template form, and persist `requireApproval` on the template spec.


supreme-gg-gg:
The UI for HITL:

Screenshot 2026-09-15 at 12 09 27 PM

Charlesthebird (approved):
It has some merge conflicts, but LGTM! 🎉

PR #2840, opened 2026-09-16
fix(ui): let a remote MCP server save, and the fixes around it

Charlesthebird:
---
*🤖 written by Claude (start)*

Adding a remote MCP server failed with `Invalid ToolServer resource`, and an agent bound to a UI-created MCP server would not start.

| Fix | Cause |
|---|---|
| A remote MCP server saves | `wrap` let `undefined`-valued keys through. protobuf-es turns those into kind-less Struct entries, which Go rejects without naming a field. Fixed for all five write paths. |
| Command servers are no longer offered to an agent | The picker listed every tool server but wrote `kind: RemoteMCPServer`. The CRD's enum accepts that; the compiler resolves nothing but a real one. |
| Streamable HTTP is the default, and says why | The form defaulted to SSE. Codex rejects a server whose protocol is set to SSE, so a tooltip now names that beside the checkbox. |
| Three extension points on the agent rail | A chat row's menu, a marker on the row, and a gutter action. All three render nothing with no extension installed. |
| A pressed table row differs from a hovered one | `rowHoverBg` was `color.border`, a quiet token. On dark, hover and press measured 1.107:1 apart. Hover is now the theme's foreground at a fifth opacity. |
| The live schedules spec runs again | The form moved to its own routes, so every `editor.getByLabel(...)` waited on a dialog and timed out. |

## Testing

1. Tool Servers → New. Streamable HTTP is already ticked, and the info icon says why.
2. Save `http://blah.com/mcp`. The server is created; discovery then fails, because the host is dead.
3. Save a second one with Connection timeout cleared.
4. Agent Templates → New, add an MCP tool. A command server is greyed out.
5. Hover then press a row on any list. The two look different, in both themes.
6. Open a chat row's menu in the rail. No blank strip under Chat details.
7. Schedules loads and shows its empty state.

---
*🤖 written by Claude (end)*


Charlesthebird:
Recorded against a live Kind cluster on this branch — the real controller, no mock mode.


https://github.com/user-attachments/assets/aa4c89f1-b3ac-4b40-9ffc-c22cd293edbd



---
*🤖 written by Claude*

Charlesthebird:
The checkbox now says why it matters. Light and dark.

tooltip-light
tooltip-dark


---
*🤖 written by Claude*

Charlesthebird:
Row hover and press, before and after, both themes.

| theme | | before | after |
|---|---|---|---|
| dark | rest → hover | 1.346:1 | 1.826:1 |
| dark | hover → press | 1.107:1 | 1.377:1 |
| light | rest → hover | 1.405:1 | 1.527:1 |
| light | hover → press | 1.202:1 | 1.334:1 |

Sampled from the rendered pixel, not `getComputedStyle` — the press paints a 30% alpha over a gradient. Hover and press were never indistinguishable on dark, but they sat at the same luminance, which is the half a colour-blind reader cannot use. The light hover is now neutral rather than lavender, which is a visible change to every clickable table.

hover-before-after


---
*🤖 written by Claude*

peterj (approved):

EItanya (approved):

PR #2841, opened 2026-09-16
ci: add nightly releases with shared publishing actions

EItanya:
Add nightly builds at 02:00 UTC, with manual dispatch on the default branch. Scheduled runs skip unchanged commits; manual runs force a rebuild. This follows [agentgateway's nightly publishing pattern](https://github.com/agentgateway/agentgateway/blob/main/.github/workflows/nightly.yml).

Nightly and tagged releases share composite actions for image publishing, Helm publishing, and CLI/chart builds. Nightlies publish commit-specific versions (`0.0.0-alpha.g<commit>`) to the existing GHCR repositories.

CLI binaries and checksums for Linux/macOS amd64 and arm64, plus Windows amd64, are uploaded with chart archives to the workflow run as a `nightly-<version>` Actions artifact. Nightlies do not publish to PyPI or create GitHub releases. Tagged releases retain their existing destinations, with the chart attachment path corrected to `dist/`.

Validation:
- Local smoke checks for dispatch, skip/retry, version normalization, image tags, and real Helm rendering.
- Actionlint, allowing the existing Blacksmith runner label; shell syntax and local action input checks.
- `git diff --check`.

Registry publishing has not been exercised locally.


the-mentor:
Awesome work 💪
I'll test it out and report back 🤞

PR #2843, opened 2026-09-16
fix(adk): stop embedding the system prompt in the pod agent card

### LINKED ISSUE #2549: Agent pod serves ADK-generated agent card embedding the full system prompt, diverging from the curated a2aConfig card
> **Disclosure:** this issue was written by an AI agent (Claude) during an internal
> evaluation of kagent + agentgateway, and reviewed by a human before filing.
> If you don't find it useful, feel free to close it.

### Environment

kagent v0.10.0-rc2, Go ADK runtime, kind cluster. (rc3 changelog shows nothing
related, so this should still apply.)

### Summary

An agent pod's own A2A endpoint (`:8080/.well-known/agent-card.json`) serves a
**generated** agent card whose first skill (`name: model`) embeds the agent's
**entire system message** in its `description`: every operating rule, tool-usage
instruction, and behavioral constraint. The controller serves the curated
`a2aConfig` card for the same agent, so the two surfaces diverge.

### Repro

```sh
# leaking generated card (pod / Service):
curl <agent-svc>:8080/.well-known/agent-card.json

# curated a2aConfig card (controller):
curl kagent-controller:8083/api/a2a/kagent/<agent>/.well-known/agent-card.json
```

The first returns a card with the full system prompt inside
`skills[0].description`; the second returns the short, curated skills authored
in the Agent CR's `a2aConfig`.

### Why this seems worth fixing

We understand the card shape is inherited from Google ADK's `AgentCardBuilder`,
which deliberately embeds instructions so client agents can pick delegation
targets. Two kagent-level aspects still look defective:

1. **Unsafe default on the gateway-facing port.** The raw card is served on the
   same port kagent labels `appProtocol: kgateway.dev/a2a`, i.e. the surface
   gateway integrations are expected to front. Discovery is usually granted more
   broadly than invocation, and system prompts often embed sensitive business
   logic (fee policy, escalation criteria, fraud thresholds). Anyone who can
   *list* the agent can read how it is instructed to behave, which is useful
   reconnaissance for prompt injection.
2. **Two divergent cards for one agent.** kagent's own docs treat the curated
   `a2aConfig` card as the discovery surface, but the generated card is what any
   direct or gateway-fronted caller actually receives.

### Suggested direction

Serve the curated `a2aConfig` card from the pod endpoint too (or at minimum
offer a flag to suppress instruction embedding in the generated card).

### Workaround

Front agent endpoints with a gateway and deny
`GET /.well-known/agent-card.json` by policy, serving discovery only from the
controller's curated surface.

TigerkidYang:
## Problem

The agent pod's `/.well-known/agent-card.json` embeds the agent's **entire system prompt**: `EnrichAgentCard` unconditionally replaces the curated card's `skills` with `adka2a.BuildAgentSkills(agent)`, and adk-go's `buildDescriptionFromInstructions` joins the agent description with the full instruction into the first skill's description. Anyone who can discover the agent can read how it is instructed to behave (fee policies, escalation rules, …), and the pod card diverges from the curated card the controller serves. The reporter re-confirmed this on 0.10.0 stable.

## Fix

`EnrichAgentCard` no longer touches `skills` at all. The compiler-generated card is authoritative (it is also what the a2a-gateway serves from the pinned runtime revision), so the pod card now carries exactly the skills the compiler put there and nothing built from the ADK agent. The function still fills a missing description, the HITL extension and the default interface. `adka2a` is no longer imported here.

The pod still serves `/.well-known/agent-card.json` to direct-to-pod callers (the reporter's agentgateway setup, and the 0.10.x line), which is where this closes the leak.

Delegation is not affected: the remote-A2A tools in both runtimes (`go/adk/pkg/tools/remote_a2a_tool.go`, `kagent-adk/_remote_a2a_tool.py`) read only the card `Description`, and the reporter confirmed empirically that routing between sub-agents never depended on skills.

## Tests

`agentcard_test.go`: an agent with an instruction and a sub-agent. A card without skills stays without skills and no instruction text reaches it; compiler-generated skills and description are kept verbatim; the description / HITL / default-interface enrichment still applies. `go vet`, `go test -run TestEnrichAgentCard ./adk/pkg/a2a/` and `go test ./adk/pkg/app/` pass.

## Backport

The same function exists unchanged on `release/v0.10.x`, where gateway-fronted callers hit the pod card directly; happy to open the backport once this lands.

Fixes #2549

---
Prepared with Claude Code assistance; the diff was reviewed and the tests above were run locally before opening the PR.


go/adk/pkg/a2a/agentcard.go
```diff
@@ -36,6 +40,34 @@ func EnrichAgentCard(card *a2atype.AgentCard, agent adkagent.Agent) {
 	}
 }
 
+// deriveAgentSkills builds a skill list from the agent's own description and
+// its sub-agents. It deliberately ignores instructions and tools: the former
+// is the system prompt, the latter is already reachable through tool listing.
```
supreme-gg-gg: what's "tool listing"? I don't recall such a mechanism in A2A

supreme-gg-gg (commented):
The a2a-gateway now serves the public Agent Card from the pinned runtime revision and invokes the Actor directly over gRPC without discovering its card. I think we should we remove runtime skill enrichment instead of maintaining a custom `deriveAgentSkills` implementation. It appears unnecessary on this path and loses upstream support for workflow/custom agents and tool-derived skills. The compiler-generated card should remain authoritative.

go/adk/pkg/a2a/agentcard.go
TigerkidYang: You are right, A2A has no such mechanism. The wording was inaccurate, and that code is gone now (see below).

TigerkidYang:
Thanks @supreme-gg-gg, agreed. With the gateway serving the card from the pinned revision, the compiler-generated card should stay authoritative, and a custom `deriveAgentSkills` is not worth maintaining. Pushed 1 commit that drops skill enrichment entirely: `EnrichAgentCard` no longer touches `skills` and only fills a missing description, the HITL extension and the default interface.

The pod still serves `/.well-known/agent-card.json` to direct-to-pod callers (the reporter's agentgateway setup, and 0.10.x), so this is what closes the leak there. Tests now assert that a card without skills stays without skills and that compiler-generated skills are kept verbatim. PR description updated to match.

supreme-gg-gg (approved):

PR #2847, opened 2026-09-16
feat(ui): name snapshots, and put fork, rename and delete on their mark

Charlesthebird:
---
*🤖 written by Claude (start)*

"Checkpoint" is "Snapshot" in the chat. A snapshot carries a name, its mark on the transcript carries Fork, Rename and Delete, and pressing the mark opens the record behind it.

| Change | Detail |
|---|---|
| A snapshot has a name | `Checkpoint.name`, generated on creation as `<agentInstanceId>-<headTaskId>`. |
| Renaming one | `UpdateCheckpointName`, mirroring `UpdateAgentInstanceName`. An empty name restores the generated default, so a fork always has something to be called. |
| A fork takes the snapshot's name | The chat no longer renames after forking, because `ForkAgentInstance` copies the checkpoint's name onto the new conversation. |
| The mark carries its actions | Fork, Rename and Delete sit under it; pressing the mark itself opens the record, which holds the id, turn, state and age. |
| Two extension points | `app_agents_agentChat_snapshotDivider_actions` beside those controls, and `app_agents_agentChat_snapshotDetails_footer` at the foot of the record. Both take `{ snapshotId, instanceId, label }`. |

The name lives in the `data` payload with the rest of the checkpoint, and the `source_name` column is gone. Nothing filters or sorts checkpoints by name, so a column earned it nothing and cost a second place for the value to live.

## Testing

1. Chat with an agent for two turns, then take a snapshot from the composer.
2. The mark reads `Snapshot "<instance id>-<turn id>"`, clipped if it is long. Hovering offers to open the record; the full name is on the record itself.
3. Send a third turn and take a second snapshot.
4. Rename the first one from its mark, reload the page, and press the mark. The new name is on the record.
5. Fork from that mark. The new conversation opens named after the snapshot, holding only the turns above it.
6. Check both themes: the mark is a tinted card, and Rename's outline stays legible on each.

---
*🤖 written by Claude (end)*


Charlesthebird:
https://github.com/user-attachments/assets/93614ad8-3896-4a24-bcdb-29c233bc76af



---
*🤖 written by Claude*

Charlesthebird:
Claude was trying to update the `source_name` column on the checkpoint to be called `name`. I told it to reuse `source_name` so that we don't have migrations or schema changes in this PR.

Charlesthebird:
https://github.com/user-attachments/assets/bf06d4c1-a83a-49bd-ae6b-e711aabefeb5



---
*🤖 written by Claude*

Charlesthebird:
https://github.com/user-attachments/assets/98da09ce-3b32-4678-8492-828abf04544e



---
*🤖 written by Claude*

EItanya (commented):
🤖 AI-generated review.

Reviewed the Go changes at `90dadcde8969951d167ea25f90a0c2cc9bba0ff8` for correctness, code cleanliness, and structure. The RPC/service/store boundaries look sensible; the main simplification is in checkpoint persistence.

1. **Use `Checkpoint.Name` as the single source of truth and remove `source_name`.**

   [The new serializer](https://github.com/kagent-dev/kagent/blob/90dadcde8969951d167ea25f90a0c2cc9bba0ff8/go/core/internal/database/checkpoints.go#L472) clones the protobuf, removes `Name`, and reconstructs it from `SourceName` on reads. `SourceName` now means the checkpoint's own name, requiring a lengthy comment explaining the misleading field name.

   Breaking schema changes are allowed right now, so preserving this column does not justify introducing a second persistence convention. The “so rename is one UPDATE” explanation also does not require this arrangement: [the existing instance rename](https://github.com/kagent-dev/kagent/blob/90dadcde8969951d167ea25f90a0c2cc9bba0ff8/go/core/internal/database/agent_instances.go#L202) already updates the protobuf with one UPDATE under a transaction. No query filters or sorts checkpoints by name.

   Please persist `Checkpoint.Name` in the protobuf, follow the existing transactional rename pattern, and have forks inherit the decoded checkpoint's name. Remove `source_name`/`SourceName`, `marshalCheckpoint`, the read-side name override, and the comments explaining those exceptions. That keeps one representation without compatibility branches or duplicate storage.

2. **Preserve source-payload validation when removing the source-name lookup.**

   [Removing the `toAgentInstance(instance)` call](https://github.com/kagent-dev/kagent/blob/90dadcde8969951d167ea25f90a0c2cc9bba0ff8/go/core/internal/database/checkpoints.go#L196) also removes validation of the stored source protobuf.

   I reproduced this against PostgreSQL by creating an instance and a valid completed task/snapshot, then deliberately replacing only the instance payload with the invalid byte `0xff`. `GetAgentInstance` rejects the payload, but checkpoint reservation succeeds and commits a `CREATING` checkpoint. Restoring just the decoding guard makes the probe pass.

   This concerns already-corrupt durable data; ordinary renaming does not cause the corruption. Please preserve that validation independently of reading the source name.

3. **Make the default-name tests independent of the implementation helper.**

   [The creation and reset assertions](https://github.com/kagent-dev/kagent/blob/90dadcde8969951d167ea25f90a0c2cc9bba0ff8/go/core/internal/database/client_agent_instance_test.go#L328) compute their expectations with the same `defaultCheckpointName` helper used by production. Returning an incorrect constant from that helper would leave those assertions green. Use explicit expected strings to verify the naming contract. Also cover rename rejecting `CREATING`, `FAILED`, and `DELETING` checkpoints.

The transport handler, authorization policy, service delegation, and MCP field addition look appropriately scoped; no additional abstractions are needed there.

Validation: all five relevant Go packages passed, including PostgreSQL persistence and inline SQL preparation checks:

```sh
go test ./core/internal/service/checkpoint ./core/internal/grpcserver ./core/internal/mcp
go test ./core/internal/database ./core/pkg/migrations
```

The additional corruption probe fails on the reviewed head and passes with only the removed validation restored. I did not run live-cluster E2E tests, UI tests, or local lint/generation checks.

Charlesthebird:
https://github.com/user-attachments/assets/312bbc21-1e48-4e14-b4ca-0bd9995854ad



---
*🤖 written by Claude*

Charlesthebird:
All three done in [`4f7afeb5`](https://github.com/kagent-dev/kagent/commit/4f7afeb5).

1. `Checkpoint.Name` now persists in the payload. The `source_name` column is gone. Rename follows `UpdateAgentInstanceName` under one transaction. Forks inherit the decoded checkpoint's name. `marshalCheckpoint` and the read-side override are deleted.
2. `toAgentInstance(instance)` is back in reservation. A test writes `0xff` and expects rejection.
3. Expectations are literal strings now. Rename rejection covers CREATING, FAILED and DELETING.


---
*🤖 written by Claude*

Charlesthebird:
Re-recorded against `f75b14c8`. Step 2 now matches the build: hovering the mark offers to open the record, and the full name is on the record rather than in a tooltip. Supersedes the earlier reel.


https://github.com/user-attachments/assets/8428c07d-7242-4a51-ae49-1fbce9d241a1



---
*🤖 written by Claude*

EItanya (approved):

PR #2848, opened 2026-09-16
fix(gateway): preserve task cancellation when runtime streams close

supreme-gg-gg:
## Issue

Task cancellation could race with delivery of the runtime’s terminal `CANCELED` event. If the gateway closed the runtime connection first, public task streams received:

```
rpc error: code = Canceled desc = grpc: the client connection is closing
```

This was discovered when cancellation related E2E tests such as `TestE2EClaudeMockActiveTaskCancellation` flaked even though cancellation succeeded.

## Fix

Register the pending terminal cancellation with the task ingester before closing the runtime stream. The ingester now publishes and persists `CANCELED` when cancellation-triggered closure wins the race, while preserving genuine stream errors when no cancellation is pending.

## Validation

- Focused gateway test repeated 100 times
- Race detector repeated 10 times
- Claude cancellation E2E verified against the patched controller

```
go test ./core/test/e2e -run '^TestE2EClaudeMockActiveTaskCancellation$' -count=10 -v -timeout=15m
=== RUN   TestE2EClaudeMockActiveTaskCancellation
=== PAUSE TestE2EClaudeMockActiveTaskCancellation
=== CONT  TestE2EClaudeMockActiveTaskCancellation
--- PASS: TestE2EClaudeMockActiveTaskCancellation (23.78s)
=== RUN   TestE2EClaudeMockActiveTaskCancellation
=== PAUSE TestE2EClaudeMockActiveTaskCancellation
=== CONT  TestE2EClaudeMockActiveTaskCancellation
...
PASS
ok      github.com/kagent-dev/kagent/go/core/test/e2e   204.160s
```


EItanya (commented):
🤖 AI-generated review.

The structural concern here is that cancellation is a protocol spread across two owners: `taskRun` owns persistence, but `Gateway.CancelTask` controls the sequence that makes it safe. There is also a regression caused by that split.

1. **Preserve the runtime's cancellation result.**

   [gateway.go:287](https://github.com/kagent-dev/kagent/blob/2c5e3813a8aa6e589edbe38f94118e9b059f53b0/go/core/internal/a2agateway/gateway.go#L287) installs a synthetic `CANCELED` event before calling the runtime. When stream closure wins, the ingester persists that event, and `CancelTask` returns before incorporating the RPC response.

   A focused probe reproduced this with a valid canceled task carrying a status message: the message survives with the pre-PR production files and disappears on this head. Keep the early cancellation intent, but resolve it using the validated runtime result; synthesize cancellation only when needed.

2. **Make cancellation an operation owned by `taskRun`.**

   [gateway.go:282–326](https://github.com/kagent-dev/kagent/blob/2c5e3813a8aa6e589edbe38f94118e9b059f53b0/go/core/internal/a2agateway/gateway.go#L282) now requires the caller to register a fallback, coordinate runtime locks, close ingress, and wait on `run.done`. Correctness depends on remembering that ordering.

   Could we put that sequence behind a cancellation operation on the existing run? Its contract should express cancellation and completion; locking, connection closure, and fallback consumption should stay internal. Likewise, `terminalFallback a2atype.Event` is broader than the requirement—explicit cancellation state would communicate the invariant better. This needs no new service or interface.

3. **Optional: clarify the event-processing helper.**

   [task_run.go:85](https://github.com/kagent-dev/kagent/blob/2c5e3813a8aa6e589edbe38f94118e9b059f53b0/go/core/internal/a2agateway/task_run.go#L85) calls the closure `persist`, but it also closes connections, quiesces compute, publishes events, updates state, and decides whether ingestion continues. Sharing that path is sensible; the name and boolean result obscure its responsibility.

Validation at `2c5e3813a8aa6e589edbe38f94118e9b059f53b0`: the gateway package tests passed, and the three cancellation tests passed 10 repetitions under the race detector. The review-only regression probe used separate streaming and cancellation clients with a fake store/workflow; it failed on this head and passed with the pre-PR production files. No local E2E tests were run.

Issue #2851, opened 2026-09-17
[BUG] Zero-argument tool calls serialize as "null" after session reload in OpenAI-shaped adapters


alexdarbyshire:
### 📋 Prerequisites

- [x] I have searched the [existing issues](https://github.com/kagent-dev/kagent/issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

App Service

### 🚦 Impact/Severity

Blocker

### 🐛 Bug Description

Zero-argument tool succeeds within one task, fail on follow-up task after history is reloaded
 
- Affects inference backends which require function call arguments to be an object 
- OpenAI adapters serializes nil `FunctionCall.Args` as the string `"null"` instead of `"{}"`
- Observable on agent substrate suspend/resume (which triggers replay path)

Underlying JSON reprex: [googleapis/go-genai#920](https://github.com/googleapis/go-genai/issues/920). Filed upstream issue proposing `omitempty` be replaced with `omitzero`

Related persistence-related nil-argument problem addressed for Bedrock in [#2222](https://github.com/kagent-dev/kagent/pull/2222)

### 🔄 Steps To Reproduce

1. Use the Go ADK OpenAI Chat Completions adapter with persistent session history and an object-validating backend such as oMLX/ollama.
2. Have the agent call a zero-argument tool, such as `list_skills`, with model-generated arguments `"{}"`.
3. Allow the tool result and the task to complete.
4. Send a follow-up message in the same context, causing a new task to load the persisted history. No suspend/resume is required.
5. The reloaded empty arguments are nil and the adapter serializes them as `"null"` on replay.

The underlying transformation is:

```text
model arguments "{}"
  → non-nil empty FunctionCall.Args map
  → persistence omits args (json:"args,omitempty")
  → reload restores Args as nil
  → json.Marshal(nil map) produces null
  → provider arguments string "null"
```

### 🤔 Expected Behavior

Zero-argument tool call replayed with `"arguments":"{}"`, including after session reload.

### 📱 Actual Behavior

Zero-argument tool call replayed with `"arguments":"null"`

Inference backend oMLX responds with HTTP Status code 422: Unprocessable content

[...]

PR #2852, opened 2026-09-17
fix(adk): guard nil tool-call args and coerce to empty object in OpenAI-compatible adapters

alexdarbyshire:
# Issue
- Nil tool-call args serialize as `"null"`
- Occurs after session reload where reloaded session contains tool calls with no args (`omitempty` in upstream go-genai drops an empty map)
- oMLX returns a 422, ollama returns a 400 to unexpected "null" string

Fixes #[2851](https://github.com/kagent-dev/kagent/issues/2851)

# Fix
- Add nil guard clauses to OpenAI compatible adapters, coerce to empty object

# Tests
- Add regression test for each adapter

# Notes
- Considered abstracting in accordance with STYLE.md. Went with in-line guard for readability
- AI used to write/wrangle code 

## SAP AI 
- added defensively for OpenAI compatibility (they support BYO Model)
- Untested if SAP AI model proxy coerces "null" string to {} as several other providers do

# References
- Same fix as Bedrock [#2222](https://github.com/kagent-dev/kagent/pull/2222)
- Equivalent guard in [adk-go's OpenAI adapter](https://github.com/google/adk-go/blob/f7e16e0226d8a255e39543df2fdfc0e15041de51/model/openaimodel/request.go#L259-L261).


supreme-gg-gg (dismissed):

go/adk/pkg/models/openai_adk.go
```diff
@@ -269,7 +269,11 @@ func genaiContentsToOpenAIMessages(contents []*genai.Content, config *genai.Gene
 			toolCalls := make([]openai.ChatCompletionMessageToolCallUnionParam, 0, len(functionCalls))
 			var toolResponseMessages []openai.ChatCompletionMessageParamUnion
 			for _, fc := range functionCalls {
-				argsJSON, _ := json.Marshal(fc.Args)
+				args := fc.Args
```
supreme-gg-gg, line 272: Can you refactor this to a shared until in `models/base.go` since this is used across openai, openai responses, sapaicore, and bedrock

alexdarbyshire: Extracted nonNilFunctionCallArgs into models/base.go and updated the four callers in [0efab2f6](https://github.com/kagent-dev/kagent/pull/2852/commits/0efab2f6bdd71bf39c53a02310a1f5314974fd05) :)

supreme-gg-gg (approved):

PR #2856, opened 2026-09-17
fix(adk): serialize nil FunctionCall.Args as {} not null for OpenAI adapters (#2851)

### LINKED ISSUE #2851: [BUG] Zero-argument tool calls serialize as "null" after session reload in OpenAI-shaped adapters

boxcee-interview:
Closes #2851

## What was wrong
A zero-argument tool call succeeds in one task, then fails on the follow-up task after history is reloaded. Persistence drops empty function-call args (`FunctionCall.Args` is `json:"args,omitempty"`), so a reloaded call carries a nil `Args` map. The OpenAI chat-completions and Responses adapters then serialized that nil map as the string `"null"` instead of `"{}"`. Object-validating backends (oMLX/ollama) reject it with HTTP 422.

## What the fix does
Coerce a nil `Args` map to an empty map before marshaling, in the two OpenAI adapters that serialize function-call arguments:
- `go/adk/pkg/models/openai_adk.go` (chat completions)
- `go/adk/pkg/models/openai_responses.go` (Responses API)

This mirrors the identical nil-to-empty coercion already present in the Bedrock adapter (`bedrock.go:686`). Added two regression tests (one per adapter) asserting a nil-args call serializes to a JSON object, not `null`.

## How to Test
Docker, `golang:1.27`, module root `go/`:
```
go test ./adk/pkg/models/ -run 'NilFunctionCallArgs' -v
```
Both tests PASS. Full package: `go test ./adk/pkg/models/` → ok.
```
=== RUN   TestGenaiContentsToOpenAIMessages_NilFunctionCallArgs
--- PASS: TestGenaiContentsToOpenAIMessages_NilFunctionCallArgs (0.00s)
=== RUN   TestGenaiContentsToResponsesInput_NilFunctionCallArgs
--- PASS: TestGenaiContentsToResponsesInput_NilFunctionCallArgs (0.00s)
```


supreme-gg-gg:
This is a duplicate of #2852, pls check for existing PRs before opening a new one next time

PR #2857, opened 2026-09-17
fix: serialize AgentInstance lifecycle execution

EItanya:
Concurrent lifecycle retries can issue duplicate Actor mutations, and delayed callers can act after a newer operation. Keep the current lifecycle generation and non-expiring executor claim on the AgentInstance row. Claim and completion check that identity; superseded observers return an explicit conflict without touching the runtime.

A successful joined caller observes current instance state while its generation remains current. Remove the separate operation-history table, serialized input/result copies, and historical lifecycle result retention. Create/Fork retries preserve instance identity and return current state without provisioning again.

Deletion leaves a DELETED tombstone that reserves the owner/request ID, revokes shares, and releases runtime revision and checkpoint pins. Fork source identity survives checkpoint deletion; a generated foreign-key column pins that checkpoint only while the instance is live. Normal instance/task/share access excludes tombstones, and reused creation request IDs cannot recreate deleted compute.

Read-only preparation may retry before execution is claimed. Ambiguous runtime failures and failed database completion retain the claim and resource pins, blocking conflicting lifecycle operations including Delete. Preserve pinned fork creation and per-Actor egress policy setup before publishing READY.

Part 2 follows #2792. Shared A2A/checkpoint execution and multi-replica observation remain later work; live Substrate settlement has not been verified. This requires a clean database and no overlapping older lifecycle binaries.

Validation:
- Core internal/package suites, including SQL preparation and migration Up/Down.
- PostgreSQL-backed lifecycle and store race tests covering exclusive claims, stale/ABA completions, delayed observers, ambiguous mutations, failed database completion, and egress failures.
- Tombstone request reuse, task/share access, concurrent share creation/deletion, and checkpoint/revision pin cleanup.
- Authenticated lifecycle service integration with real PostgreSQL and controlled Actor responses.
- Go lint and diff checks.

The unreleased schema is consolidated in migration 1. The existing CI migration-immutability check rejects that edit; no forward migration or policy bypass is included.

Design and implementation tracker are updated locally and remain uncommitted.


go/core/internal/database/agent_instance_operations.go
```diff
@@ -0,0 +1,321 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// InstanceOperation identifies one admitted lifecycle operation. Instance is its
+// original input; Result and Failure describe its retained, immutable outcome.
+// TODO: add bounded pruning of settled outcomes after at least 24 hours;
+// pending or uncertain operations must never expire.
+type InstanceOperation struct {
+	ID                 uuid.UUID
+	SourceCheckpointID *uuid.UUID // Pinned fork intent; nil for ordinary creation.
```
supreme-gg-gg: Do you think we should have a `Kind` field here so we don't need to read `Instance.Kind` every time, so we won't mix the instance input with operation identity?

supreme-gg-gg (approved):

go/core/internal/database/agent_instance_operations.go
```diff
@@ -0,0 +1,321 @@
...
+	Failure            string
+}
+
+// BeginAgentInstanceOperation admits or joins one lifecycle operation under the
+// instance lock. Joined callers retain the same identity, but must win Claim before
+// touching the runtime. Delete may supersede unclaimed work; claimed work blocks
+// conflicting operations indefinitely until its executor records completion.
+// Callers authorize access. Missing instances return ErrNotFound, except that a
+// completed Delete can return its retained result to an already authorized caller.
+func (c *Client) BeginAgentInstanceOperation(ctx context.Context, instanceID string, kind apiv1alpha1.AgentInstanceOperation) (*InstanceOperation, error) {
```
supreme-gg-gg, line 30: Could some of the lifecycle comparisons be extracted into named helpers, which would make it easier to read and follow? Perhaps adding some inline comments would suffice.

go/core/internal/database/agent_instance_operations.go
EItanya: Added `Kind` from the existing persisted column and switched admission, completion, and dispatch to use it. `Instance` remains the admitted input snapshot.

go/core/internal/database/agent_instance_operations.go
EItanya: Named the compound conditions and documented completed retries, Delete supersession, and preparation failure versus claimed completion inline. Straightforward comparisons stay local; no additional helpers.

PR #2858, opened 2026-09-17
feat(harness): export executor construction and health config

xytian315:
## Summary

- Export harness executor construction so another program can start the Claude and Codex harnesses with its own startup steps while reusing kagent's adapter and A2A server.
- Add configurable health probe paths and handler to A2A server


go/adk/pkg/a2a/server/health.go
```diff
@@ -13,4 +13,8 @@ func RegisterHealthEndpoints(mux *http.ServeMux) {
 	})
 	mux.Handle("/health", handler)
 	mux.Handle("/healthz", handler)
+	mux.HandleFunc("GET /ping", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"status":"Healthy"}`))
+	})
```
EItanya: Can we just make the health endpoint configrable?

go/adk/pkg/a2a/server/server.go
```diff
@@ -84,7 +84,7 @@ func NewA2AServer(agentCard a2atype.AgentCard, executor a2asrv.AgentExecutor, lo
 		switch {
 		case strings.HasPrefix(r.URL.Path, "/grpc.health.v1.Health/"):
 			return false
-		case r.URL.Path == "/health", r.URL.Path == "/healthz", r.URL.Path == a2asrv.WellKnownAgentCardPath:
+		case r.URL.Path == "/health", r.URL.Path == "/healthz", r.URL.Path == "/ping", r.URL.Path == a2asrv.WellKnownAgentCardPath:
```
EItanya: Can this be part of the health configuration?

go/harness/runtime/a2a/deferred.go
```diff
@@ -0,0 +1,106 @@
...
+	"os"
+	"sync"
+
+	a2atype "github.com/a2aproject/a2a-go/v2/a2a"
+	"github.com/a2aproject/a2a-go/v2/a2asrv"
+)
+
+// Deferred wraps an executor whose build may have to wait for dir to exist.
+// Some hosts attach the durable directory only when the first request arrives.
+type Deferred struct {
```
EItanya: This doesn't feel like it needs to be in core, can't we make the points available to use this runtime elsewhere while using our adapter?

go/harness/codex/cmd/main.go
```diff
@@ -23,8 +27,11 @@ import (
 const (
 	configEnv    = "KAGENT_CONFIG_JSON"
 	agentCardEnv = "KAGENT_AGENT_CARD_JSON"
-	dataDir      = "/data"
-	privatePort  = "80"
+	dataDirEnv   = "KAGENT_DURABLE_DIR"
+	portEnv      = "PORT"
+
+	defaultDataDir     = "/data"
+	defaultPrivatePort = "80"
```
EItanya: These are shared with claude, any reason they aren't in our env registry, also should probably be `KAGENT_PORT` for consistency.

EItanya (commented):
biggest comment is about the deferred runner, feels potentially orthogonal, and functionality that could be injected rather than live here

EItanya:
🤖 AI-generated follow-up.

The latest revision addresses all four earlier comments: health paths and the handler are configurable, tracing uses that configuration, deferred startup is removed in favor of exported executor constructors, and the proposed environment overrides are removed.

One small follow-up in `registerHealthEndpoints`: `HealthPaths` promises literal exact paths, but passing them directly to `http.ServeMux.Handle` gives them ServeMux pattern semantics. I reproduced this with `/ping/`: `GET /ping/child` returns `200 OK` from the health handler and exports a trace span, since the tracing filter uses exact equality. Could we enforce exact matching or validate the supported path forms so routing and tracing agree, with a test covering this? The intended `/ping` configuration works correctly.

Focused server, app, Claude/Codex harness, and runtime tests passed at `efe6477a`.

go/adk/pkg/a2a/server/health.go
```diff
@@ -4,13 +4,21 @@ import (
...
-	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusOK)
-		_, _ = w.Write([]byte("OK"))
-	})
-	mux.Handle("/health", handler)
-	mux.Handle("/healthz", handler)
+	registerHealthEndpoints(mux, defaultHealthPaths(), defaultHealthHandler)
+}
+
+func registerHealthEndpoints(mux *http.ServeMux, paths []string, handler http.Handler) {
```
supreme-gg-gg, line 20: Duplicate health paths could make `http.ServeMux.Handle` panic, we can add a check here to be safe (or in `server.go` below where you checked for mux patterns)?

supreme-gg-gg (commented):
The changes LGTM overall

EItanya (approved):

Issue #2828, opened 2026-09-15
UI: track the deferred Playwright coverage in DEFERRED.md


Charlesthebird:
[`ui/playwright/DEFERRED.md`](https://github.com/kagent-dev/kagent/blob/main/ui/playwright/DEFERRED.md) records what the browser suite deliberately skips, rather than committing skipped specs. It has never had a tracking issue. Nothing surfaces an entry when its blocker lands.

| Blocked on | Entries |
|---|---|
| A proto change | Server-side paging, searching, sorting; auto-titling; conversation search scope |
| A surface that doesn't exist | Agent-form validation; an agent's readiness reason end to end |
| A fixture that doesn't exist | A template labelled for the only harness |
| A real controller | Streaming; the suspending stage; tool approval |
| A convention amendment | A broken create aborts that resource's failure steps |

Each entry gets closed or re-argued when its blocker moves.

---
*🤖 written by Claude*

Charlesthebird:
Re-opening since the linked PR is being reverted

Charlesthebird:
Goal is to do this, but not add more time in CI

PR #2874, opened 2026-09-18
test(ui): run the live Playwright suite in CI against the built UI image

### LINKED ISSUE #2657: [FEATURE] Run the live UI Playwright suite in CI against the built UI image
## Problem

CI's `ui-tests` job proves the UI is self-consistent, not that it works. It runs `yarn typecheck`, `yarn lint`, `yarn test`, and `yarn test:pw` — and `test:pw` is the mock suite, served by MSW with no backend at all. A backend contract change cannot fail it.

#2653 is what that looks like in practice: #2638 stopped populating three fields the Substrate page renders, every check stayed green, and it was found by hand.

A live suite already exists and CI never runs it:

- `ui/playwright/live/` holds `pages.spec.ts`, `agent-lifecycle.spec.ts`, and `substrate.spec.ts`, run by `yarn test:pw:live`.
- No workflow references `test:pw:live` or `UI_LOOP_LIVE`.

Live mode also does not test the shipped artifact. `LIVE_COMMAND` in `ui/playwright.config.ts` is `yarn dev`, so the suite hits a Vite dev server proxying to the cluster — not the nginx image an operator runs.

## Proposal

Run the live suite in CI against the real built UI image.

- Serve the image from `ui/Dockerfile`, not `yarn dev`, so the run exercises what ships: nginx addressing and the runtime `window.environmentVariables` rewrite, which no dev-server run touches.
- Point the suite at the port-forwarded `svc/kagent-ui` rather than a Vite port. `playwright.config.ts` is already written for this — the comment above `LIVE_COMMAND` names a probed port-forward as the case that replaces that line.
- Reuse what exists rather than building a second path: `scripts/setup-cluster/setup-cluster.sh` already builds `ui/Dockerfile`, sets it on `deploy/kagent-ui`, and port-forwards the service on 8080. The `test-e2e` job already stands up a Kind cluster with `helm-install`, and `build (ui)` already builds the image.

## Keep chat on the mock backend

Chat stays where it is. Its journeys need deterministic streaming deltas, tool call and result ordering, cancellation, and a failed turn with retry — none of which a real model gives you reliably, and all of which the mock suite already asserts.

The live directory has no chat spec today, and it should not gain one. Live coverage is for the surfaces where a controller response is the thing under test.

## Notes

- `ui/playwright/DEFERRED.md` already anticipates this: the `cleanup.spec.ts` entry is deferred "while the suite runs on the mock backend" and says to revisit "if the suite gains a live-backend mode".
- Worth adding once it runs: assertions on the Actor Templates columns, which is the gap #2653 fell through.
- The mock suite keeps its job. It is fast, runs both engines, and covers error paths a real cluster will not produce on demand. This adds a lane rather than replacing one.

---
*🤖 written by Claude*

toreysoloio:
---
*🤖 written by Claude (start)*

The browser suite only ever ran against fixtures, so nothing checked the app an operator actually gets. `playwright/live/` now runs in the `test-e2e` job against the image built from `ui/Dockerfile` — nginx, the SPA fallback and the runtime config — on the same cluster as the Go end-to-end tests.

Closes #2657
Refs #2828

## What runs where

`tests/` is mock-only and `live/` needs a cluster; `shared/` is the journeys that assert the same thing against both, one create-read-edit-delete per resource. That split is the point: a spec in `shared/` may not name a fixture or a `?mock=` scenario, which `conventions.test.ts` enforces.

## The product fix in here

Every list page drew its summary on `!error && !isLoading`, and SWR reports `isLoading` false on the first paint — so a page that had not yet asked the controller anything rendered "0 of 0 harnesses" for about 600ms. The six summaries now wait for `data`.

## The deferred coverage in #2828

| Blocked on | Entry | Where it stands |
|---|---|---|
| A proto change | Server-side paging, searching, sorting | open — substrate's landed in #2771; models, tool servers and prompts still answer with everything |
| | Auto-titling; conversation search scope | open |
| A surface that doesn't exist | Agent-form validation | open — there is no agent form |
| | An agent's readiness reason, end to end | open — but the record half is covered now, and had no spec at all before this branch |
| A fixture that doesn't exist | A template labelled for the only harness | open — checked against a one-harness cluster instead |
| A real controller | Tool approval | **done** — the controls shipped in #2714 and nothing drove them; `tests/chat/approvals.spec.ts` does |
| | Streaming; the suspending stage | open, and misfiled: the live lane exists now, so what is missing is a model that can answer |
| A convention amendment | A broken create aborts that resource's failure steps | open |

Also closed: a turn parked without the HITL extension, and three claims in `DEFERRED.md` that had gone stale — including one section saying the extension-point specs were never started, two paragraphs below the list of what they cover.

## Testing

Mock suite, no cluster needed:

1. `cd ui && yarn test:pw`

Live, against a real backend:

2. `scripts/setup-cluster/setup-cluster.sh` — it leaves the UI on `localhost:8080`.
3. `cd ui && UI_LOOP_LIVE_URL=http://localhost:8080 yarn test:pw:live`
4. Check the cluster afterwards: `kubectl get agenttemplate,harness,modelconfig -A | grep e2e-live` should find nothing, since each journey removes what it made.

In CI, the `test-e2e` job runs the same lane against the LoadBalancer address and uploads `playwright-report/` on failure.

---
*🤖 written by Claude (end)*


ui/playwright/shared/agent-templates/agent-templates.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+    await remove.click();
+
+    /*
+     * Either branch is legitimate: the count comes from `status.harnesses` while step 5
+     * waited on `admittingHarnesses`, two fields filled by different work. The mock
+     * suite pins the wording; what a cluster shows is that the sentence is computed
+     * from real state at all, rather than coming back blank.
+     */
+    await expect(page.getByTestId("template-delete-consequence")).toContainText(
+      /built from this template|no agent was ever built from it/,
```
Charlesthebird, line 301: **Level:** 🟠 Medium · Blocking

"The mock suite pins the wording" is no longer true: this PR removed the old step 11 checks ("1 agent is built from this template", "keep working", "no new one can be started") and nothing else asserts them. Restore that step in `tests/agent-templates/`, or drop the claim here.

---
*🤖 written by Claude*

ui/playwright/shared/schedules/schedules.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+      page
+        .getByRole("dialog", { name: `Delete schedule ${CREATED}?`, exact: true })
+        .getByRole("button", { name: "Delete", exact: true }),
+      () => expect(page).toHaveURL(/\/schedules(\?|$)/),
+    );
+    detailURL = undefined;
+
+    await expect(page.getByRole("link", { name: CREATED, exact: true })).toHaveCount(0, {
+      timeout: READ_TIMEOUT,
+    });
```
Charlesthebird, line 216: **Level:** 🟠 Medium · Not Blocking

This absence check runs before the list has drawn, so a delete that failed server-side still passes with zero rows. Wait for the list to settle first:

```suggestion
    await expect(
      page
        .getByTestId("schedules-empty")
        .or(page.locator('[data-testid="schedules-table"] tbody tr.ant-table-row'))
        .first(),
    ).toBeVisible({ timeout: READ_TIMEOUT });
    await expect(page.getByRole("link", { name: CREATED, exact: true })).toHaveCount(0, {
      timeout: READ_TIMEOUT,
    });
```

---
*🤖 written by Claude*

ui/playwright/tests/schedules/schedules.spec.ts
```diff
@@ -255,184 +237,7 @@ test("schedules: a schedule is created, read, run, changed and deleted", async (
...
-   *
-   * They were split out on the reasoning that a `page.goto` resets the fixture backend
-   * and would throw away the schedule the lifecycle is holding. True, and beside the
-   * point once they run *last*: by here the schedule has been deleted and there is
-   * nothing left to lose. `models`, `mcp-servers` and `prompts` all end the same way,
-   * and this file reading differently from them was the contradiction rather than the
-   * reset.
-   */
-  await test.step("17. a read failure is not an empty list", async () => {
+  await test.step("11. a read failure is not an empty list", async () => {
```
Charlesthebird, line 255: **Level:** 🟠 Medium · Not Blocking

Old steps 12–16 were deleted without a shared replacement: the list row's "Weekly on…" cadence, rename-from-list not duplicating, both Keep paths, "Stops future executions.", and the other rows surviving a delete. Port them here (they need the seeded list), or record them in `DEFERRED.md` and fix `README.md:106`.

---
*🤖 written by Claude*

ui/playwright/shared/models/models.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+      timeout: READ_TIMEOUT,
+    });
+    /*
+     * The identity and the provider are the ref and what the ref means, so an edit
+     * changes neither. Asserted here because it is the boundary between "edit" and
+     * "make a new one", and a form that quietly allowed it would write a resource
+     * nothing else in the cluster points at.
+     */
+    await expect(page.getByTestId("model-name")).toBeDisabled();
+    await expect(page.getByTestId("model-model").locator("input")).toBeDisabled();
```
Charlesthebird, line 140: **Level:** 🟠 Medium · Not Blocking

The old step 7 `expectRequired(... unmarked: ["API key (leave blank to keep existing)"])` is gone, so nothing checks that editing a model doesn't demand the key again. Add it back here or in `tests/models/`.

---
*🤖 written by Claude*

ui/playwright/shared/models/models.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    await page.getByTestId("model-name").fill(CREATED);
+    // The namespace is half the ref, so one created without it is addressed as
+    // `/name` and never appears on the list. `kagent` is where both backends put
+    // things.
+    await selectOption(page, "model-namespace", "kagent");
+    await page.getByTestId("model-api-key").fill("sk-not-a-real-key");
+
+    await page.getByTestId("model-submit").click();
+    await page.waitForURL(/\/models(\?|$)/, { timeout: READ_TIMEOUT });
+    created = true;
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

`created` flips only after the redirect, so a create that landed but redirected slowly fails the test and leaks the resource. The sweep already checks for the row, so set it before submitting; same in prompts:79, harnesses:122 and agent-templates:168.

```suggestion
    created = true;
    await page.getByTestId("model-submit").click();
    await page.waitForURL(/\/models(\?|$)/, { timeout: READ_TIMEOUT });
```

---
*🤖 written by Claude*

ui/playwright/tests/prompts/prompts.spec.ts
```diff
@@ -183,17 +178,20 @@ test("prompts: a library is created, read, changed and deleted", async ({
...
+    /*
+     * And that is where this step stops. Creating the library, reading it back, adding a
+     * fragment and deleting it runs against both backends from
+     * `playwright/shared/prompts/prompts.spec.ts`. Left here it was the same journey
+     * driven twice, and the count it used to assert is stronger there, where it is
+     * relative to whatever the list already held.
+     *
+     * Navigated rather than submitted, so the steps below start on the list with a draft
+     * abandoned — which is also the state a reader leaves behind when they change their
+     * mind, and nothing else covers it.
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

`loadPage` is a `page.goto`, which reloads the module-level mock state, so an abandoned draft that wrongly saved would be wiped and this still passes. Drop the coverage claim:

```suggestion
     * Navigated rather than submitted, so the steps below start on the list.
```

---
*🤖 written by Claude*

ui/playwright.config.ts
```diff
@@ -153,7 +222,7 @@ export default defineConfig({
     ? [
         {
           name: LIVE_PROJECT,
-          testDir: "./playwright/live",
+          testMatch: [/live\/.*\.spec\.ts$/, SHARED_SPECS],
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

Playwright matches `testMatch` regexes against the absolute path, so a checkout under e.g. `/home/olive/` puts every `tests/` spec in the live lane. Globs resolve against `testDir`; do the same for `SHARED_SPECS` (:38) and the mock lanes (:239, :255).

```suggestion
          testMatch: ["live/**/*.spec.ts", "shared/**/*.spec.ts"],
```

---
*🤖 written by Claude*

ui/src/pages/AgentsPage.tsx
```diff
@@ -459,7 +459,11 @@ export function AgentsTab() {
           ]}
           trailing={
             <Space size={8}>
-              {!loadFailure && !templates.isLoading ? (
+              {/* `data !== undefined` as well as `!isLoading`: SWR runs its fetcher in an effect, so
+              the first paint reports "not loading" on a page that has not asked yet, and this
+              counted "0 of 0" for the 600ms before the answer arrived. A count of nothing is a
+              claim, and until the read lands this page has not earned it. */}
+              {!loadFailure && !templates.isLoading && templates.data !== undefined ? (
```
Charlesthebird, line 463: **Level:** 🟡 Low · Not Blocking

An empty namespace list gives a `null` SWR key, so `data` stays undefined and the summary never renders (same in `AgentTemplatesPage` and `HarnessesTab`); resolving `[]` to an empty result in the hooks fixes it. The comment above also blames SWR's effect, but the flash came from `useApiResource` reporting `isLoading: false` for a held-back key, and it's pasted into six pages.

---
*🤖 written by Claude*

.github/workflows/ci.yaml
```diff
@@ -178,6 +189,107 @@ jobs:
...
+          UI_IP="$(kubectl get svc -n kagent kagent-ui -o jsonpath='{.status.loadBalancer.ingress[0].ip}')"
+          test -n "$UI_IP"
+          export UI_LOOP_LIVE_URL="http://${UI_IP}:8080"
+          echo "UI_LOOP_LIVE_URL: $UI_LOOP_LIVE_URL"
+          # `globalSetup` refuses the run if that address answers with fixtures, or is a
+          # dev server rather than the built image.
+          yarn test:pw:live
+
+      - name: Upload live browser report
+        if: ${{ failure() && steps.install-kagent.outcome == 'success' }}
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

With 2 CI retries, a test that fails then passes leaves no trace. Upload on any finished run; `if-no-files-found: ignore` already covers a clean one.

```suggestion
        if: ${{ !cancelled() && steps.install-kagent.outcome == 'success' }}
```

---
*🤖 written by Claude*

Charlesthebird (commented):

---

**Blocks merge:** the shared agent-templates spec claims the mock suite still pins the delete-dialog wording, but this PR deleted that pin. The rest is non-blocking. The main theme is coverage lost when mock steps moved to `shared/`: schedules, the model edit form, and one vacuous absence check.

<details><summary>⚠️ What I checked</summary>

- ✅ UI image pushed to the kind registry and pulled with `pullPolicy=Always`; LoadBalancer IP read after `helm --wait`
- ✅ Port 8080 matches the chart service and nginx `listen`
- ✅ `mockServiceWorker.js` 404s in `nginx.conf:134` and is deleted in the Dockerfile
- ✅ `proto-generate` runs once before the parallel builds
- ✅ Live step has no `continue-on-error`; fork PRs get no secrets
- ✅ Live `workers: 1` / `fullyParallel: false` override the top-level config
- ✅ Every moved spec lands in at least one lane
- ✅ Cleanup deletes only `throwawayName` resources (pid + timestamp)
- ✅ No `waitForTimeout`, toast or `networkidle` waits in `shared/`, `live/` or new `tests/` specs
- ✅ Mock HITL payload uses the same URI-keyed shape as `a2aGrpcChatClient.ts:457`
- ✅ Empty and error states kept for harnesses, models, prompts, schedules, agent-templates
- ⚠️ Coverage parity between old mock steps and new shared journeys
- ⚠️ Absence checks wait for a positive list signal first

</details>

---
*🤖 written by Claude*

ui/playwright/live/substrate.spec.ts
```diff
@@ -18,26 +19,29 @@ import { expectNoLoadFailure, liveRoutes, loadLive } from "./helpers/live";
...
-  await expectNoLoadFailure(page);
+  await loadApp(page, liveRoutes.substrate);
 
   await test.step("1. the page is there rather than an error", async () => {
     await expect(page.getByTestId("substrate-actors-card")).toBeVisible({
       timeout: 60_000,
     });
+    // After the card, not after `loadApp`: the alerts are counted once, and a page that
+    // has not read anything yet has none — see `expectNoLoadFailure`.
+    await expectNoLoadFailure(page);
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

The actors card renders before any read lands, so this runs once before an error can appear and never again. Move it to the end of step 3.

---
*🤖 written by Claude*

ui/playwright/DEFERRED.md
```diff
@@ -1,208 +1,156 @@
...
+- **A cleanup spec.** Each mock test gets a fresh browser context, and each live spec
+  deletes what it made in a `test.afterEach`. A run killed outright still leaks, which
+  is why `throwawayName` puts the process and a timestamp in every name: anything
+  matching `e2e-live-*` in `kagent` is litter and safe to remove by hand. A sweep spec
+  stays the wrong shape for that, being one bad selector away from deleting real work.
+- **Agent-create validation.** There is no agent form. An agent is an `AgentTemplate`
+  paired with a `Harness`, materialised by admission — `router/routes.ts` records why
+  there is no `agentNew` and no `agentEdit`. Nothing creates one, so there is nothing
+  to validate. The create-and-read-back property that lived in the removed agent-create
+  spec is now `tests/harnesses/harnesses.spec.ts`.
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

`tests/harnesses/harnesses.spec.ts` now creates nothing; the create-and-read-back lives in `shared/harnesses/` and `shared/agent-templates/`. The PR body cites the same wrong file.

---
*🤖 written by Claude*

ui/playwright/conventions.test.ts
```diff
@@ -136,6 +160,51 @@ describe("playwright layout", () => {
...
+    /*
+     * `shared/` runs in both suites, so a scenario query there is either ignored by a
+     * cluster or — worse — read as a claim the run cannot make. The narrowness is the
+     * folder's whole value: assertions true on both backends are the weakest ones, and
+     * a lifecycle drifting in here would quietly cost the mock suite its precision.
+     */
+    for (const spec of sharedSpecs()) {
+      const source = readFileSync(spec, "utf8");
+      expect(
+        /mock=|withScenario|scenario:|helpers\/mockCalls/.test(source),
```
Charlesthebird, line 182: **Level:** 🟡 Low · Not Blocking

This only catches `?mock=` and scenario helpers, so a shared spec importing a fixture (e.g. `instances.ready`) passes here and fails live. The README says it catches fixtures too, so widen the check or narrow the README.

---
*🤖 written by Claude*

ui/playwright/shared/agent-templates/agent-templates.spec.ts
toreysoloio: Restored in `6c8873c4`: `tests/agent-templates/` step 11 asserts the populated sentence — the count, "keep working" and "no new one can be started" — alongside the empty branch it already had.


---
*🤖 written by Claude*

ui/playwright/shared/schedules/schedules.spec.ts
toreysoloio: Applied in `6c8873c4`; the absence now waits for the empty state or a first row before it means anything.


---
*🤖 written by Claude*

ui/playwright/tests/schedules/schedules.spec.ts
toreysoloio: Ported in `6c8873c4`: the row's derived cadence, both confirmation shapes and their Keep, and that a delete takes one row and leaves the others. Rename-from-list is not ported — there is no rename control on this page.


---
*🤖 written by Claude*

ui/playwright/shared/models/models.spec.ts
toreysoloio: Added in `6c8873c4` as `tests/models/` step 6, which switches to the inline key first since every seeded model authenticates by secret reference.


---
*🤖 written by Claude*

ui/playwright/shared/models/models.spec.ts
toreysoloio: Applied in `6c8873c4` across all four specs.


---
*🤖 written by Claude*

ui/playwright/tests/prompts/prompts.spec.ts
toreysoloio: Claim dropped in `6c8873c4`.


---
*🤖 written by Claude*

ui/playwright.config.ts
toreysoloio: Globbed all four in `6c8873c4`; both lanes verified unchanged at 96 mock and 13 live.


---
*🤖 written by Claude*

ui/src/pages/AgentsPage.tsx
toreysoloio: Fixed in `6c8873c4`: both namespace hooks key on the array so `[]` resolves instead of holding back, and the explanation moved into `useApiResource` with the right cause. The held-back key also explains why an earlier attempt to measure the flash found nothing.


---
*🤖 written by Claude*

.github/workflows/ci.yaml
toreysoloio: Applied in `6c8873c4`.


---
*🤖 written by Claude*

ui/playwright/live/substrate.spec.ts
toreysoloio: Moved to the end of step 3 in `6c8873c4`.


---
*🤖 written by Claude*

ui/playwright/DEFERRED.md
toreysoloio: Corrected in `6c8873c4` to `shared/harnesses/` and `shared/agent-templates/`, and in the PR body.


---
*🤖 written by Claude*

ui/playwright/conventions.test.ts
toreysoloio: Widened in `6c8873c4` to reject a shared spec importing a named fixture, mutation-tested against `instances`.


---
*🤖 written by Claude*

Charlesthebird (approved):

PR #2877, opened 2026-09-18
fix(ui): give checkpoint delete cancel buttons unique data-testids (#2871)

### LINKED ISSUE #2871: [BUG] Flaky chat checkpoints spec: two confirmations share a Cancel label
## Affected service

UI — `ui/playwright/tests/chat/checkpoints.spec.ts`

## Description

`chat: the mark names itself, carries its controls, and opens its record` is flaky. It fails on the first attempt and passes on retry, so the run is reported green with `2 flaky` and the failure is easy to miss.

The failing step is `and the mark's own delete asks first, then takes it`:

```
Error: locator.click: Test timeout of 30000ms exceeded
    113 | await expect(page.getByText("Delete this snapshot?")).toBeVisible();
  > 114 | await page.getByRole("button", { name: "Cancel" }).click();
    115 | await expect(dividers(page)).toHaveCount(1);
      at ui/playwright/tests/chat/checkpoints.spec.ts:114:56
```

## Cause

Two components render a confirmation with the title `Delete this snapshot?`:

- `ui/src/components/chat/CheckpointDivider.tsx` — the divider's own delete
- `ui/src/components/chat/SnapshotDetailsModal.tsx` — the dialog's delete

The step immediately before presses `Escape` to close the dialog and asserts `snapshot-details-body` has gone. It then opens the divider's confirmation and reaches for `getByRole("button", { name: "Cancel" })` across the whole page. While the dialog is still closing, that label matches a control that is on its way out and never becomes clickable, so the click waits out the full timeout.

Both confirmations already give their **ok** button a `data-testid` — `chat-checkpoint-delete-confirm-${checkpointId}` and `snapshot-details-delete-confirm`. Neither gives one to its **cancel** button, which is why the spec has to fall back to a label that is not unique.

The same page-wide lookup appears twice in the spec, at lines 114 and 213, for the two different confirmations.

## Suggested fix

Give each cancel button the `data-testid` its ok button already has, and have the spec use it:

```tsx
// CheckpointDivider.tsx
cancelButtonProps={{ "data-testid": `chat-checkpoint-delete-cancel-${checkpointId}` }}

// SnapshotDetailsModal.tsx
cancelButtonProps={{ "data-testid": "snapshot-details-delete-cancel" }}
```

That keeps the assertion deterministic rather than raising the timeout, and it matches the pattern the confirm buttons already set.

## Reproduction

Run the chat suite on Chromium. It reproduces intermittently and more readily on a Linux CI runner than on a laptop, since the flake is a race with the dialog's close animation.

---
*🤖 written by Claude*

boxcee-interview:
Closes #2871

## What was wrong

Two Popconfirm dialogs on the same chat page both title their confirmation "Delete this snapshot?" — the divider's own delete (CheckpointDivider.tsx) and the snapshot record dialog's delete (SnapshotDetailsModal.tsx). The ok buttons of each already carry a data-testid; neither cancel button did, so the spec had to look up the cancel by role+name across the whole page. While the record dialog is still mid-close-animation after Escape, the page-wide Cancel lookup resolves to the control on its way out, which never becomes clickable — the click waits out the 30s timeout and the suite reports flaky.

## What the fix does

Gives each cancel button the data-testid its ok button already sets, and points the spec at it:

- CheckpointDivider.tsx: cancelButtonProps data-testid chat-checkpoint-delete-cancel-
- SnapshotDetailsModal.tsx: cancelButtonProps data-testid snapshot-details-delete-cancel
- checkpoints.spec.ts lines 114 and 213: getByTestId(...) instead of getByRole("button", { name: "Cancel" })

## How to Test

cd ui && yarn test:pw — the previously flaky test "chat: the mark names itself, carries its controls, and opens its record" now targets the divider's own cancel button, and "chat: a snapshot is deleted from its record" targets the dialog's own cancel. Verified in an isolated container (node:24, yarn 4.9.0, Playwright chromium): yarn lint clean (0 errors), yarn typecheck clean, and the checkpoints spec passing on chromium — including both affected tests.


ui/playwright/tests/chat/checkpoints.spec.ts
```diff
@@ -111,7 +111,7 @@ test("chat: the mark names itself, carries its controls, and opens its record",
 
     await line.getByTestId(`chat-checkpoint-delete-${id}`).click();
     await expect(page.getByText("Delete this snapshot?")).toBeVisible();
-    await page.getByRole("button", { name: "Cancel" }).click();
+    await page.getByTestId(`chat-checkpoint-delete-cancel-${id}`).click();
```
Charlesthebird: **Level:** 🔴 High · Blocking

Measured on `main` at this line: one `Cancel` in the DOM, one Popconfirm, zero `.ant-modal-wrap` — the old lookup was never ambiguous and this resolves to the same node. That Cancel is 0px wide at that instant, still animating in, which is what `pressOnce` (`ui/playwright/helpers/resource.ts:56`) exists for. The suggestion needs `import { confirmation, pressOnce } from "../../helpers/resource";` at the top, which a suggestion block can't add.

```suggestion
    await pressOnce(confirmation(page).getByTestId(`chat-checkpoint-delete-cancel-${id}`));
```

*🤖 written by Claude*

ui/playwright/tests/chat/checkpoints.spec.ts
```diff
@@ -111,7 +111,7 @@ test("chat: the mark names itself, carries its controls, and opens its record",
...
     await expect(dividers(page)).toHaveCount(1);
```
Charlesthebird, line 120: **Level:** 🟠 Medium · Blocking

This already holds before the click, so it passes whether Cancel landed, missed onto the backdrop, or left the Popconfirm open — a missed click then surfaces four lines down as "the divider was not deleted", which is the shape of the reported failure. Its sibling at :214 asserts the dismissal.

```suggestion
    await expect(page.getByText("Delete this snapshot?")).toBeHidden();
    await expect(dividers(page)).toHaveCount(1);
```

*🤖 written by Claude*

ui/playwright/tests/chat/checkpoints.spec.ts
```diff
@@ -210,7 +210,7 @@ test("chat: a snapshot is deleted from its record, and stays deleted", async ({
     await openSnapshot(page, page.getByTestId(`chat-checkpoint-mark-${id}`));
     await page.getByTestId("snapshot-details-delete").click();
     await expect(page.getByText("Delete this snapshot?")).toBeVisible();
-    await page.getByRole("button", { name: "Cancel" }).click();
+    await page.getByTestId("snapshot-details-delete-cancel").click();
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

Three sibling lookups keep the page-wide shape: `checkpoints.spec.ts:165` (`Save`), `turns.spec.ts:59` (`Retry`), and `agent-page.spec.ts:155` (`Cancel`, in a file already using `pressOnce` nine lines later). The `Save` one failed once here under worker contention; 30 clean repeats were green, so treat it as a lead rather than a proven flake.

*🤖 written by Claude*

Charlesthebird (commented):

---

The stated root cause is measurably wrong, so the flake survives. Instrumented on `main` at the exact failing line, `getByRole("button", { name: "Cancel" })` resolves to **one** element, one Popconfirm is mounted, and **zero** `.ant-modal-wrap` remain — the closing snapshot modal this PR blames is already gone, and `SnapshotDetailsModal.tsx:88` passes `footer={null}` so it never had a `Cancel` to contribute. The new testid lands on that same single node, so the locator swap is a no-op.

What the same probe does show: that lone Cancel measures **0px wide** while reporting `display: inline-flex; visibility: visible`. The Popconfirm is still animating in — exactly the case `pressOnce` (`ui/playwright/helpers/resource.ts:56`) was written for, "Playwright can compute a click's coordinates while one is still moving". `checkpoints.spec.ts` is the only file in `tests/chat/` importing neither it nor `confirmation()`.

The two `cancelButtonProps` additions are harmless and match the `okButtonProps` pattern already there — keep them as stable hooks, just not as the fix.

<details>

- ✅ `cancelButtonProps` takes the same `ButtonProps` as the adjacent `okButtonProps`, so `"data-testid"` typechecks by precedent.
- ✅ No timeouts were added — the rule against buying a pass with a timeout holds.
- ✅ Both new testids are unique on the page: the divider's is keyed by `checkpointId`, and only one `SnapshotDetailsModal` mounts at a time (`AgentChatPage.tsx:852`).
- ✅ Nothing overlays the Cancel button at that moment — `elementFromPoint` at its centre returns the button itself.
- ❌ The claimed second `Cancel` does not exist: one Cancel, one Popconfirm, zero modal wrappers, measured at the failing line.
- ❌ Three sibling lookups with the same shape are left unfixed — `checkpoints.spec.ts:165`, `turns.spec.ts:59`, `agent-page.spec.ts:155`, the last in a file that already uses `pressOnce` nine lines later.

</details>

---
*🤖 written by Claude*

ui/src/components/chat/SnapshotDetailsModal.tsx
```diff
@@ -227,6 +227,7 @@ export function SnapshotDetailsModal({
             okText="Delete"
             okButtonProps={{ danger: true, "data-testid": "snapshot-details-delete-confirm" }}
             cancelText="Cancel"
+            cancelButtonProps={{ "data-testid": "snapshot-details-delete-cancel" }}
```
Charlesthebird, line 230: [nit]

Worth keeping as a stable hook, but it earns nothing here: `footer={null}` on line 88 means this modal contributes no competing `Cancel`, so the lookup at spec:213 was already unambiguous.

*🤖 written by Claude*

boxcee-interview:
Thanks — the animation is the actual failure mode, agreed. Pushed e37761b: the cancel-button testids stay as stable hooks, the two Cancel presses in checkpoints.spec.ts and the Save press in the rename step now go through pressOnce, and the same treatment went to the three sibling raw dialog clicks you flagged (checkpoints.spec.ts:165, agents/agent-page.spec.ts:155). All four tests pass in isolation on chromium; the rail-menu duplicate test is the one that flakes intermittently in this environment, unrelated to these lines.

ui/playwright/tests/chat/checkpoints.spec.ts
Charlesthebird: `pressOnce` plus the import is the fix. The `confirmation()` scoping in my suggestion isn't needed now the testid is unique.


---
*🤖 written by Claude*

ui/playwright/tests/chat/checkpoints.spec.ts
Charlesthebird: Still open. `toHaveCount(1)` holds before the click as well, so it passes whether or not Cancel landed — `pressOnce` makes a miss unlikely, not detectable.


---
*🤖 written by Claude*

ui/playwright/tests/chat/checkpoints.spec.ts
Charlesthebird: `agent-page.spec.ts:155` and `checkpoints.spec.ts:165` are handled. `turns.spec.ts:59` still takes a raw click; it was the weakest of the three, so this stays open only as a marker.


---
*🤖 written by Claude*

ui/src/components/chat/SnapshotDetailsModal.tsx
Charlesthebird: Withdrawn — the testid now carries the `pressOnce` at :214, so it earns its place.


---
*🤖 written by Claude*

Charlesthebird (approved):

boxcee-interview:
Pushed 7ae6425 addressing the two open points:

- checkpoints.spec.ts:118 — the cancel now asserts the popconfirm actually closed: await expect(page.getByText("Delete this snapshot?")).toBeHidden() after the pressOnce, so a missed press no longer passes on the divider count alone.
- turns.spec.ts:59 — the Retry click now routes through pressOnce (weakest of the three siblings, now handled).

All four touched tests pass in isolation on chromium in a clean container; the one remaining failure (composer checkpoint, checkpoints.spec.ts:24) also fails on the unmodified base in the same container, so it is pre-existing and unrelated to these lines.

PR #2879, opened 2026-09-19
fix(ui): wait for schedule deletion redirect in live test

### LINKED ISSUE #2850: Live schedules spec asserts a detail page that deleting redirects away from
## `ui/playwright/live/schedules.spec.ts` asserts a detail page the delete redirects away from

The teardown of `live: schedule configuration persists through the browser and controller` deletes the schedule it created, then reads the detail page:

```ts
await page.getByRole("dialog").getByRole("button", { name: "Delete", exact: true }).click();
await expect(page.getByText("This schedule was deleted. Its execution history is retained.")).toBeVisible();
await expect(page.getByRole("button", { name: "Run", exact: true })).toBeDisabled();
await page.getByRole("link", { name: "Back", exact: true }).click();
```

None of those three survive the delete. `ScheduledRunsPage.tsx` hands its delete button:

```ts
onDeleted={async () => { await invalidate(); await navigate(paths.scheduledRuns); }}
```

so the page navigates to `/schedules` and the alert, the `Run` button and the `Back` link go with it.

### Which assertion fails is a race

It depends on whether `invalidate()` resolves before the assertion runs.

| Where | Fails at |
|---|---|
| Linux CI runner | `Run` — `element(s) not found` |
| macOS, local cluster | `Back` — `locator.click` timed out after 120s |

Both end on the schedules list, which is the screenshot the run attaches.

### Why CI does not catch it

This is a live spec, so it needs a cluster. Reproduced on both platforms above against a kind cluster at `bb0cba62`.

### Fix

Assert what the delete does — the redirect, and that the schedule is gone from the list:

```ts
await expect(page).toHaveURL(/\/schedules$/);
await expect(page.getByRole("link", { name, exact: true })).toHaveCount(0);
```

That keeps what the teardown is for. The alternative is to drop the redirect and leave the reader on a tombstoned detail page, which is what the spec was written against — either way the two should agree.

---
*🤖 written by Claude*

Dragonzz27:
Fixes #2850.

## Summary
Wait for the redirect to the schedules list after deletion, then assert that the deleted schedule is absent. Do not inspect or click controls on the detail page that navigation is unmounting. Production navigation and retained-history behavior are unchanged.

## Verification
From `ui/`:
- `yarn typecheck` — passed.
- `yarn exec eslint playwright/live/schedules.spec.ts --max-warnings 0` — passed.
- `yarn test` — 524 tests passed.
- Chromium schedules lifecycle suite — passed.
- Reproduced the old cleanup sequence against the actual UI with the in-browser fixture backend: the `Back` link detaches during navigation and times out. The same browser probe with the redirect assertion passes; both that probe and the existing schedule lifecycle test passed (2/2).

The live cluster suite was not run: this local environment has no configured kagent cluster. The browser probe verifies the UI navigation race, not controller persistence. Existing retained-history coverage in the fixture suite is unchanged.

Implemented with AI assistance; all changes were reviewed and validated locally.


ui/playwright/live/schedules.spec.ts
```diff
@@ -63,9 +63,9 @@ test("live: schedule configuration persists through the browser and controller",
       await page.goto(detailURL);
       await page.getByRole("button", { name: `Delete schedule ${name}`, exact: true }).click();
       await page.getByRole("dialog").getByRole("button", { name: "Delete", exact: true }).click();
-      await expect(page.getByText("This schedule was deleted. Its execution history is retained.")).toBeVisible();
-      await expect(page.getByRole("button", { name: "Run", exact: true })).toBeDisabled();
-      await page.getByRole("link", { name: "Back", exact: true }).click();
+      // Deleting leaves the detail page; its retained-history controls are no
+      // longer mounted. Wait for navigation before checking the list.
+      await expect(page).toHaveURL(/\/schedules(?:\?.*)?$/);
       await expect(page.getByRole("link", { name, exact: true })).toHaveCount(0);
```
Charlesthebird, line 69: **Level:** 🔴 High · Blocking

This assertion cannot fail, and after the edit it is the only thing standing between a working delete and a leaked schedule. `page.goto(detailURL)` on line 63 starts a fresh document, and `useInvalidateKeys` "reaches what is on screen, and only that" (`src/api/hooks/useInvalidateKeys.ts:19`), so the list key is never cached: the freshly mounted list renders zero rows while it reads, and this passes on the first poll. It would pass just as happily against a controller that kept the schedule.

There is no DOM state that separates "loading" from "empty" here — against `?mock=slow` the loading table renders one `.ant-table-placeholder` row with `schedules-empty` visible — so the read itself has to be waited for. Attach it before the click, since the response can land before the URL assertion resolves:

```ts
const listRead = page.waitForResponse((r) => r.url().includes("ScheduledRunService/ListScheduledRuns"));
await page.getByRole("dialog").getByRole("button", { name: "Delete", exact: true }).click();
await expect(page).toHaveURL(/\/schedules(?:\?.*)?$/);
await listRead;
await expect(page.getByTestId(`schedule-link-${name}`)).toHaveCount(0);
```

The list polls every 10s (`src/pages/ScheduledRunsPage.tsx:32`), so a missed first read cannot hang the wait. `getByTestId` rather than the link role, to match the list's own handle.

---
*🤖 written by Claude*

ui/playwright/live/schedules.spec.ts
```diff
@@ -63,9 +63,9 @@ test("live: schedule configuration persists through the browser and controller",
       await page.goto(detailURL);
       await page.getByRole("button", { name: `Delete schedule ${name}`, exact: true }).click();
       await page.getByRole("dialog").getByRole("button", { name: "Delete", exact: true }).click();
-      await expect(page.getByText("This schedule was deleted. Its execution history is retained.")).toBeVisible();
-      await expect(page.getByRole("button", { name: "Run", exact: true })).toBeDisabled();
-      await page.getByRole("link", { name: "Back", exact: true }).click();
+      // Deleting leaves the detail page; its retained-history controls are no
```
Charlesthebird, line 66: [nit]

Only the alert is about retained history — `Run` is the page's primary action, disabled because the schedule is gone (`src/pages/ScheduledRunsPage.tsx:113,167`) — and the second sentence restates the line under it.

```suggestion
      // Deleting navigates back to the list, so the detail page's own controls are gone.
```

---
*🤖 written by Claude*

Charlesthebird (commented):

---

**What blocks merge:** the test no longer verifies that the schedule is gone from the schedules page. Line 69 survived the edit but can never fail — it runs against a table that is still loading — so the assertion the issue asked for is present in form only. Details inline.

The rest of the change is right, and stronger than the issue credits it for: the removed `Back` click was a hard failure rather than a race, since the list page has no Back link. `toHaveURL` also stands in for "the delete RPC succeeded" — `onDeleted` fires only after `onDelete()` resolves (`src/components/table/DeleteResourceButton.tsx:75`), so a failed delete leaves the page put and fails the assertion.

One thing found while checking the obvious fix, outside this PR: `dataRows` in `playwright/live/helpers/live.ts:59` does not exclude antd's empty-state row. `filter({ hasNot })` looks for a matching *descendant*, and `.ant-table-placeholder` is the `tr` itself, so an empty table counts as one data row. Measured against `?mock=slow`: 1 during load, 0 with `tbody tr:not(.ant-table-placeholder)`. That is why gating on `dataRows(page).first()` — as `live/pages.spec.ts:50` does — is not a fix here.

<details><summary>⚠️ What I checked</summary>

- ✅ The regex matches `paths.scheduledRuns` = `/schedules` (`src/router/routes.ts:94`), no other route ends that way and no basename is configured; its optional query group mirrors `playwright/tests/schedules/schedules.spec.ts:362`.
- ✅ The detail page always unmounts on a successful delete: `onDeleted` runs `invalidate()` then `navigate(paths.scheduledRuns)` (`src/pages/ScheduledRunsPage.tsx:258`).
- ✅ The tombstone alert is not dead UI — it still renders on direct navigation to a deleted schedule and on the detail page's 10s poll (`src/pages/ScheduledRunsPage.tsx:108,192`).
- ✅ Tombstone coverage is not lost: the fixture suite asserts the deleted note and that Run, Pause and Edit are all disabled (`playwright/tests/schedules/schedules.spec.ts:458`), plus absence from the list at step 20.
- ✅ The live project is Chromium-only with `expect: { timeout: 30_000 }` (`playwright.config.ts:146`), so the redirect has room against a real cluster.
- ✅ The teardown still deletes the schedule it created (lines 63-65 untouched).
- ❌ Nothing verifies the schedule is absent from the list it redirects to: measured against `?mock=slow`, a loading schedules table renders zero rows and a visible `schedules-empty`, so an absence assertion passes before the list has been read.
- ⚠️ No CI Build run exists for `54a15621`; only DCO, labels and Snyk reported, so the typecheck and unit-test results in the description are unverified here.

</details>

---
*🤖 written by Claude*

PR #2890, opened 2026-09-21
feat: support structured output in JSON Schema over A2A DataPart 

supreme-gg-gg:
## Summary

> This feature is only supported in Kagent's own harness, not yet in Claude / Codex

- Add inline and ConfigMap-backed output schemas to ADK agent templates.
- Validate schemas against a portable provider-compatible JSON Schema profile.
- Support structured output with tools across OpenAI, Anthropic, Bedrock, and Gemini through ADK.
- Validate final responses and publish them as JSON A2A artifacts with schema metadata.
- Add unit and E2E coverage for tool calls and partial streaming responses.
- Document the structured-output design and provider behavior.
- Relevant UI changes to configure output mode, display structured output, etc

## Testing

- Unit and E2E tests
- Manual testing on OpenAI (Chat completion + Responses), Gemini (API), Claude (Bedrock)
- Tested with both tool using and structured output


supreme-gg-gg:
First image: UI for using tool call with structured output

Screenshot 2026-09-21 at 10 12 44 PM

Second image: AgentTemplate details page showing the agent's structured output schema

Screenshot 2026-09-21 at 10 13 15 PM

Third image: Gemini models might (depending on version) use an additional tool to set structured output. This is done internally by ADK and is a known workaround for some older model that does not support tool calling with structured output.

Screenshot 2026-09-21 at 10 39 38 PM

go/adk/pkg/a2a/consts.go
```diff
@@ -13,6 +13,7 @@ const (
 	A2ADataPartMetadataIsLongRunningKey     = "is_long_running"
 	A2ADataPartMetadataTypeFunctionCall     = "function_call"
 	A2ADataPartMetadataTypeFunctionResponse = "function_response"
+	OutputSchemaSHA256MetadataKey           = "kagent.dev/output-schema-sha256"
```
EItanya: We used to have a `KagentMetadata` fn which prepended this key, do we still have that?

go/adk/pkg/a2a/executor.go
```diff
@@ -57,12 +72,19 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
...
+		rootName = runnerConfig.Agent.Name()
+	}
 	builtin := adka2a.NewExecutor(adka2a.ExecutorConfig{
 		RunnerConfig:       runnerConfig,
 		RunConfig:          runConfig,
 		A2APartConverter:   a2aPartConverter,
-		GenAIPartConverter: genAIPartConverter,
+		GenAIPartConverter: structuredOutputPartConverter(output, rootName),
 		AfterEventCallback: func(ctx adka2a.ExecutorContext, event *adksession.Event, processed *a2atype.TaskArtifactUpdateEvent) error {
+			if processed == nil {
```
EItanya: How can this be `nil`?

go/adk/pkg/a2a/executor.go
```diff
@@ -57,12 +72,19 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
 	if cfg.SessionService != nil {
 		runnerConfig.SessionService = cfg.SessionService
 	}
+	rootName := ""
+	if runnerConfig.Agent != nil {
+		rootName = runnerConfig.Agent.Name()
+	}
```
EItanya: Can `rootName` be empty?

go/adk/pkg/models/openai_adk.go
```diff
@@ -172,6 +172,16 @@ func generateContentChatCompletions(
 		}, params.Messages...)
 	}
 	applyOpenAIConfig(&params, m.Config)
+	if schema, err := structuredOutputSchema(req.Config); err != nil {
+		yield(nil, err)
+		return
+	} else if schema != nil {
+		params.ResponseFormat = openai.ChatCompletionNewParamsResponseFormatUnion{
+			OfJSONSchema: &shared.ResponseFormatJSONSchemaParam{JSONSchema: shared.ResponseFormatJSONSchemaJSONSchemaParam{
+				Name: "kagent_output", Schema: schema, Strict: param.NewOpt(true),
```
EItanya: should `kagent_output` be a `const`? Always what is this name used for?

go/adk/pkg/models/openai_responses.go
```diff
@@ -36,6 +36,14 @@ func generateContentResponses(
 		params.Instructions = param.NewOpt(instructions)
 	}
 	applyOpenAIResponsesConfig(&params, m.Config)
+	if schema, err := structuredOutputSchema(req.Config); err != nil {
+		yield(nil, err)
+		return
+	} else if schema != nil {
+		format := responses.ResponseFormatTextConfigParamOfJSONSchema("kagent_output", schema)
```
EItanya: Same question about the `name`

go/adk/pkg/outputschema/project.go
```diff
@@ -0,0 +1,140 @@
...
+// Project converts a canonical kagent output schema to the genai.Schema used
+// by Go ADK and also provides the concrete Go ADK compatibility check:
+// recursive references and schemas whose expanded form exceeds the configured
+// safety limits are rejected here.
+//
+// The canonical JSON Schema remains authoritative for provider requests that
+// accept raw JSON Schema and for final response validation. genai.Schema lacks
+// const and additionalProperties, and represents enum values as strings, so
+// those constraints may be enforced only by the provider and final validator.
+func Project(raw json.RawMessage) (*genai.Schema, error) {
```
EItanya: The name `project` here is confusing, very LLM sounding. Can we come up with something clearer?

go/api/v1alpha3/agenttemplate_types.go
EItanya: Your PR didn't do this, but this feels like it should be a shared type, not specific to `AgentTemplate`

go/api/v1alpha3/agenttemplate_types.go
```diff
@@ -199,6 +201,16 @@ type AgentTemplateSpec struct {
...
+	// OutputSchema constrains successful terminal output when this template is
+	// compiled as the root agent.
+	// +optional
+	// +kubebuilder:validation:Type=object
+	// +kubebuilder:pruning:PreserveUnknownFields
+	OutputSchema *apiextensionsv1.JSON `json:"outputSchema,omitempty"`
+	// OutputSchemaFrom references a JSON Schema stored as JSON in a
+	// same-namespace ConfigMap key.
+	// +optional
+	OutputSchemaFrom *AgentTemplateConfigMapKeyReference `json:"outputSchemaFrom,omitempty"`
```
EItanya: Not saying this isn't useful, but unless this has been explicitly asked for, can we start without it to reduce scope, it's strictly additive anyway.

go/adk/pkg/a2a/executor.go
```diff
@@ -76,7 +98,7 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
 				}
 				processed.Artifact.SetMeta(apia2a.TimelinePositionMetadataKey, position.UTC().Format(time.RFC3339Nano))
 			}
-			return nil
+			return transformStructuredOutput(output, rootName, event, processed)
```
EItanya, line 104: 🤖 AI-generated review.

**Refusals can report success with no structured result.** Upstream ADK invokes `AfterEventCallback` only when conversion produces a nonempty artifact. Both OpenAI adapters turn provider refusals into empty content without an error, so this validation is never reached. I reproduced this with mocked Chat Completions and Responses refusals through the real adapters, ADK LLM agent, runner, and executor: both returned `TASK_STATE_COMPLETED` with zero artifacts.

Please require a validated root result before successful terminal completion, while preserving input-required and error transitions. Moving the nil check alone will not fix this because the callback itself is skipped.

go/adk/pkg/a2a/executor.go
```diff
@@ -86,7 +108,89 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
...
+	if err := output.schema.Validate(value); err != nil {
+		return fmt.Errorf("output_validation_failed: root agent output does not conform to its schema")
+	}
+	if processed == nil || processed.Artifact == nil {
+		return fmt.Errorf("output_validation_failed: root agent produced no result artifact")
+	}
+	part := a2atype.NewDataPart(value)
+	part.MediaType = "application/json"
+	part.SetMeta(OutputSchemaSHA256MetadataKey, output.sha256)
+	processed.Artifact.Parts = a2atype.ContentParts{part}
```
EItanya: 🤖 AI-generated review.

**Existing remote-agent and CLI consumers discard this JSON answer.** Replacing the final text with a `DataPart` leaves `extractTextFromTask` in `go/adk/pkg/tools/remote_a2a_tool.go` and `PartsText` in `go/core/cli/internal/a2a/assembler.go` reading only `part.Text()`.

For a completed task containing the structured answer `{"answer":4}`, I reproduced an empty remote-tool `Result` with no error and empty default CLI output. Please update both consumers to preserve JSON DataParts, for example by serializing their data, so callers can actually receive the new result.

go/adk/pkg/a2a/executor.go
```diff
@@ -86,7 +108,89 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
...
+}
+
+func transformStructuredOutput(output *structuredOutput, rootName string, event *adksession.Event, processed *a2atype.TaskArtifactUpdateEvent) error {
+	if output == nil || event.Author != rootName {
+		return nil
+	}
+	if !event.IsFinalResponse() {
+		return nil
+	}
+	value, err := finalStructuredOutput(event)
```
EItanya: 🤖 AI-generated review.

**Incomplete provider responses can pass as successful structured output.** `IsFinalResponse()` does not check the provider's finish reason. With the real OpenAI adapter, a mocked response containing `{"answer":4}` and `finish_reason: "length"` becomes `FinishReasonMaxTokens`, but this path still emits a structured artifact and completes the task because the JSON happens to validate.

Please check the provider completion reason before accepting the answer, so incomplete or refused responses fail as required by the documented contract.

ui/src/api/chat/a2aGrpcChatClient.ts
```diff
@@ -195,13 +208,24 @@ function toPart(part: A2APart): ChatPart | undefined {
...
+function dataKindOf(
+  data: Record<string, unknown>,
+  mediaType: string,
+  metadata: Record<string, unknown> | undefined,
+): ChatDataPart["dataKind"] {
+  if (
+    mediaType.split(";", 1)[0]?.trim().toLowerCase() === "application/json" &&
+    typeof metadata?.[OUTPUT_SCHEMA_SHA256_METADATA_KEY] === "string"
+  ) {
+    return "structured_output";
```
EItanya, line 227: 🤖 AI-generated review.

**The downstream filter still hides valid structured output containing `"name": "ask_user"`.** This correctly identifies the part as `structured_output`, but `visibleParts` below unconditionally drops every DataPart whose `data.name` is `"ask_user"`, including schema-constrained user data.

I reproduced an empty transcript for a completed JSON result describing a tool with that name; the same payload with `name: "example"` remains visible. Please exempt `structured_output` from the internal-tool filter, or restrict that filter to tool traffic.

go/core/internal/translator/outputschema.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const maxOutputSchemaBytes = 64 << 10
+
+// portableOutputSchemaMetaSchema describes the JSON Schema subset that every
+// supported kagent ADK integration can consume. It is deliberately itself a
+// JSON Schema so keyword forms and recursive schema locations are validated by
+// jsonschema-go rather than by a parallel handwritten schema implementation.
+const portableOutputSchemaMetaSchema = `{
```
EItanya, line 25: We will probably need to document this and its implications so consumers can see it

go/adk/pkg/a2a/consts.go
supreme-gg-gg: That function uses the `kagent_` prefix for ADK function metadata these are public A2A result metadata like `kagent.dev/timeline-position`

go/adk/pkg/a2a/executor.go
supreme-gg-gg: It can't be empty, I've added validation

go/adk/pkg/models/openai_adk.go
supreme-gg-gg: Yes I've made it a const and shared across providers. This is used to identify the output schema with a name

go/adk/pkg/a2a/consts.go
EItanya: Maybe we should still have functions to make sure they use the correct prefix. Also maybe the prefix should be kagent.dev/a2a?

go/adk/pkg/a2a/executor.go
supreme-gg-gg: I've added this validation to the executor, now if it doesn't emit a structured result (e.g. from refusals) it will fail the task with task state failed rather than completed with zero artifacts. The model adapters currently don't handle refusals, but having this validation in the executor acts like a final guard since there are other events not necessarily refusals that will result in zero artifact (since ADK allows it).

go/adk/pkg/a2a/executor.go
```diff
@@ -77,19 +101,105 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
...
+func transformStructuredOutput(output *structuredOutput, rootName string, event *adksession.Event, processed *a2atype.TaskArtifactUpdateEvent) error {
+	if output == nil || event.Author != rootName {
+		return nil
+	}
+	if !event.IsFinalResponse() {
+		return nil
+	}
+	if event.FinishReason != genai.FinishReasonStop {
+		return fmt.Errorf("output_validation_failed: root agent did not complete structured output")
+	}
```
Charlesthebird: **Level:** 🔴 High · Blocking

An older Gemini model routes structured output through ADK's `set_model_response` tool, and `createFinalModelResponseEvent` builds that final event with no finish reason (`adk@v2.3.0 internal/llminternal/base_flow.go:708`). It arrives as `""` and fails here, though ADK itself treats `""`, `Unspecified` and `Stop` as normal completion (`internal/telemetry/genaimessages.go:351`).

```suggestion
	switch event.FinishReason {
	case "", genai.FinishReasonUnspecified, genai.FinishReasonStop:
	default:
		return fmt.Errorf("output_validation_failed: root agent did not complete structured output")
	}
```

---
*🤖 written by Claude*

go/adk/pkg/a2a/executor.go
```diff
@@ -77,19 +101,105 @@ func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
...
 	}
 }
 
+func transformStructuredOutput(output *structuredOutput, rootName string, event *adksession.Event, processed *a2atype.TaskArtifactUpdateEvent) error {
+	if output == nil || event.Author != rootName {
+		return nil
+	}
+	if !event.IsFinalResponse() {
+		return nil
+	}
```
Charlesthebird: **Level:** 🔴 High · Blocking

`IsFinalResponse()` returns true whenever `LongRunningToolIDs` is non-empty (`adk@v2.3.0 session/session.go:225`), which is exactly a HITL approval request — it carries a function call and no text, so `finalStructuredOutput` fails the run. The guard above does not catch it: OpenAI, Anthropic and Bedrock all map tool-use to `FinishReasonStop` (`openai_adk.go:43`, `anthropic_adk.go:29`, `bedrock.go:850`).

```suggestion
	if !event.IsFinalResponse() || len(event.LongRunningToolIDs) > 0 || event.Actions.SkipSummarization {
		return nil
	}
```

---
*🤖 written by Claude*

go/adk/pkg/models/openai_adk.go
```diff
@@ -172,6 +172,16 @@ func generateContentChatCompletions(
 		}, params.Messages...)
 	}
 	applyOpenAIConfig(&params, m.Config)
+	if schema, err := structuredOutputSchema(req.Config); err != nil {
+		yield(nil, err)
+		return
+	} else if schema != nil {
+		params.ResponseFormat = openai.ChatCompletionNewParamsResponseFormatUnion{
+			OfJSONSchema: &shared.ResponseFormatJSONSchemaParam{JSONSchema: shared.ResponseFormatJSONSchemaJSONSchemaParam{
+				Name: structuredOutputName, Schema: schema, Strict: param.NewOpt(true),
```
Charlesthebird: **Level:** 🟠 Medium · Blocking

Strict mode requires `additionalProperties: false` on every object and every property listed in `required`, but the portable profile leaves both optional, so a schema the controller accepts gets a 400 on first invocation. Either enforce those two rules in the profile or drop `Strict` (same in `openai_responses.go:43`).

---
*🤖 written by Claude*

go/adk/pkg/outputschema/convert.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+	for _, candidate := range schema.AnyOf {
+		value, err := c.convert(candidate, next, depth+1)
+		if err != nil {
+			return nil, fmt.Errorf("convert output anyOf: %w", err)
+		}
+		converted.AnyOf = append(converted.AnyOf, value)
+	}
+	for _, value := range schema.Enum {
```
Charlesthebird, line 128: **Level:** 🟠 Medium · Not Blocking

Non-string enum members are skipped and `const` survives only when it is a string, so a mixed enum `["a", 1]` reaches the provider as `["a"]` — narrower than the contract; `additionalProperties` is dropped entirely. Gemini gets only this conversion and there is no text fallback, so a schema using those keywords fails validation at the end of an otherwise good run.

---
*🤖 written by Claude*

go/api/v1alpha3/harness_types.go
```diff
@@ -237,6 +237,10 @@ type HarnessCapabilities struct {
 	InputRequired bool `json:"inputRequired"`
 	// +required
 	Approvals bool `json:"approvals"`
+	// StructuredOutput reports whether the pinned adapter can enforce a root
+	// JSON output contract and emit it as an A2A DataPart.
+	// +required
+	StructuredOutput bool `json:"structuredOutput"`
```
Charlesthebird, line 243: **Level:** 🟠 Medium · Not Blocking

This is `+required` on an already-shipped API version, so re-applying an existing Harness manifest now fails. Nothing reads it either — the compiler gate is `harnessType(...) != HarnessTypeKagent` and the form checks `harness.runtime !== "kagent"` — so it is worth either reading the capability in both places or dropping the field.

---
*🤖 written by Claude*

go/core/internal/translator/outputschema.go
```diff
@@ -0,0 +1,155 @@
...
+	"$schema":"https://json-schema.org/draft/2020-12/schema",
```
Charlesthebird, line 26: **Level:** 🟡 Low · Not Blocking

`additionalProperties: false` on `$defs/schema` applies to the root through the `allOf`, and `$schema` is not in the property list, so a schema that opens with the line every published JSON Schema opens with is rejected as an unknown keyword. Allowing `$schema` and `$id` would save a confusing first attempt.

---
*🤖 written by Claude*

docs/architecture/structured-output.md
```diff
@@ -0,0 +1,82 @@
...
+at runtime; in that case the Task fails.
+
+The serialized schema is limited to 64 KiB. Conversion to the Go ADK schema is
+also limited to 32 levels and 1,000 expanded schema nodes; these bounds include
+nodes expanded through local `$defs` references. Schemas that exceed any limit
+fail compilation.
+
+The semantic resolution code is in
+[`go/core/internal/translator/outputschema.go`](../../go/core/internal/translator/outputschema.go).
+Runtime validation and A2A conversion are owned by the Go and Python kagent ADK
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

The PR ships no Python implementation, so this reads as a contract the Python runtime already honours.

---
*🤖 written by Claude*

ui/src/components/chat/StructuredOutputCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+        borderRadius: theme.radius.md,
+        background: theme.color.bgElevated,
+        padding: theme.space(3),
+        display: "grid",
+        gap: theme.space(2),
+      }}
+    >
+      <div css={{ display: "flex", alignItems: "center", gap: theme.space(2) }}>
+        <FileJson size={15} css={{ color: theme.color.textMuted }} />
+        <Text css={{ fontWeight: 600, fontSize: 13 }}>Structured result</Text>
```
Charlesthebird, line 35: [nit]

The schema digest reaches the client in `part.metadata` but never appears on screen, so the card says the answer is JSON without saying which contract it was held to.

---
*🤖 written by Claude*

Charlesthebird (commented):

---

Two things block merge, both in `transformStructuredOutput`. The finish-reason guard added this round rejects an empty reason, which is what ADK's `set_model_response` path produces — the Gemini-with-tools case in the third screenshot. Separately, a HITL approval request still reaches the transform, where it has no text to validate and fails the task.

<details>

- ✅ `AfterEventCallback` fires only for a non-nil A2A event, and its error becomes a failed task — `adk@v2.3.0 server/adka2a/v2/executor.go:366-375`.
- ✅ The CEL rule making `outputSchema` and `outputSchemaFrom` mutually exclusive is on `AgentTemplateSpec`.
- ✅ `outputSchemaFrom` is added to the provenance ConfigMap set, so editing the ConfigMap recompiles the revision.
- ✅ Cycle detection uses pointer identity after `$ref` resolution, so `A → B → A` is caught.
- ✅ The canonical form re-marshals through `map[string]any`, so the digest is stable across key order.
- ⚠️ `IsFinalResponse()` short-circuits to true on `LongRunningToolIDs` and `SkipSummarization` — `adk@v2.3.0 session/session.go:225`.
- ⚠️ ADK builds the `set_model_response` final event with no finish reason — `adk@v2.3.0 internal/llminternal/outputschema_processor.go:67-76`.

</details>

---
*🤖 written by Claude*

Charlesthebird (commented):
The UI changes here LGTM - thanks for the screenshots! I also kicked off an automated review.

And I added https://github.com/kagent-dev/kagent/issues/2894 as a follow-up issue since the chat message part renderers might be a good place for extension points.

go/adk/pkg/a2a/executor.go
supreme-gg-gg: I've found this myself as well and I'll add a fix

go/api/v1alpha3/harness_types.go
supreme-gg-gg: This comment is not entirely valid because `status.capabilities` is optional controller-derived status, so existing Harness manifests are not required to provide `structuredOutput`, and current controller doesn't populate that record. I think we should retain this as capability contract and once we start to populate it we can migrate runtime-type checks to use capability

go/adk/pkg/a2a/consts.go
supreme-gg-gg: I agree, I'll add both. I also think it's worth doing a follow-up to review the metadata before we do a stable release so we can still centralize and change them (e.g. we will want to change the timeline position above as well, and some legacy ADK metadata are no longer necessary, and probably renaming the `adk_` and `kagent_` ones)

supreme-gg-gg:
Thanks for the review @Charlesthebird and I've also verified now structured output works with HITL:

https://github.com/user-attachments/assets/3bacadc5-b518-4d91-beff-088533e24047

go/api/v1alpha3/agenttemplate_types.go
supreme-gg-gg: Ok I've also refactored this in https://github.com/kagent-dev/kagent/pull/2890/commits/1dbd99c942d9c247133e354a1fecc850d7aeb354

go/adk/pkg/a2a/executor.go
```diff
@@ -35,21 +40,33 @@ type KAgentExecutorConfig struct {
...
+	sha256 string
 }
 
 var _ a2asrv.AgentExecutor = (*KAgentExecutor)(nil)
 
-// NewKAgentExecutor creates a KAgentExecutor from config.
-func NewKAgentExecutor(cfg KAgentExecutorConfig) *KAgentExecutor {
+// NewKAgentExecutor creates a KAgentExecutor from config. It returns an error if
+// the configured output schema cannot be prepared for runtime validation.
+func NewKAgentExecutor(cfg KAgentExecutorConfig) (*KAgentExecutor, error) {
```
mesutoezdil, line 65: #2896 also rewrites this callback, conflicts on merge

supreme-gg-gg: Hey @mesutoezdil can you avoid leaving this type of comments since they're not too constructive. Merge conflicts are natural and authors are expected to rebase on top, this simply adds more noise during reviews. Unless the changes across PRs are going in opposite directions, in which case it might be worthwhile to bring the author's attention to it and drive some discussion, otherwise this type of comment don't add any value, as in the case here and in a few other PRs that you left similar comments before.

mesutoezdil: fair point, noted 👍🏽 thx for feedback

Charlesthebird (approved):

go/adk/pkg/a2a/consts.go
supreme-gg-gg: I have a follow up for A2A metadata in #2902 if you want to take a quick look first, not fully ready for review yet

EItanya (approved):

Issue #2270, opened 2026-07-16
[BUG] kagent invoke fails with "cannot unmarshal object into []*errordetails.Typed" for agents without SupportedInterfaces


codechamp1:
### 📋 Prerequisites

- [x] I have searched the [existing issues](https://github.com/kagent-dev/kagent/issues) to avoid creating a duplicate
- [x] By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/kagent-dev/kagent/blob/main/CODE_OF_CONDUCT.md)
- [x] I am using the latest version of the software
- [x] I have tried to clear cache/cookies or used incognito mode (if ui-related)
- [x] I can consistently reproduce this issue

### 🎯 Affected Service(s)

Multiple services / System-wide issue

### 🚦 Impact/Severity

Blocker

## Bug Description

`kagent invoke` fails with a JSON-RPC internal error for all declarative agents. The controller selects the wrong A2A transport version when an agent's card does not advertise `SupportedInterfaces`, causing a JSON unmarshalling panic when the agent returns any error response.

### 🔄 Steps To Reproduce

1. Install kagent 0.10.0-beta7
2. Create a declarative agent via `kubectl apply`
3. Run:
   ```
   kagent invoke --agent <agent-name> --task "hello"
   ```

### 🤔 Expected Behavior

The agent responds to the task normally, same as when invoked via the UI.

### 📱 Actual Behavior

```
ERROR client/client.go:302  JSON-RPC error in SSE event for request : jsonrpc error -32603:
failed to decode response: json: cannot unmarshal object into Go struct field
Error.error.data of type []*errordetails.Typed
```

The invocation always fails. The UI works fine for the same agent.

### 💻 Environment

- kagent version: 0.10.0-beta7
- Controller image: `ghcr.io/kagent-dev/kagent/controller:0.10.0-beta7`
- Kubernetes: GKE
- OS: macOS

## Possible Root Cause

**File:** `go/core/internal/a2a/a2a_registrar.go`
**Function:** `filterInterfacesByVersion`

The controller registers a v0.3 compat transport (`a2av0.NewJSONRPCTransport`) to talk to agent pods, but it is only selected when the agent card advertises `SupportedInterfaces` with `ProtocolVersion: "0.3"`. Declarative agents created via `kubectl apply` do not advertise `SupportedInterfaces`, so `filterInterfacesByVersion` returns unversioned interfaces and the v1 transport is selected instead.

[...]

PR #2891, opened 2026-09-22
fix(cli): resolve two stacked bugs breaking 'kagent invoke'

vainokurula-e:
## Summary

`kagent invoke` is currently broken end-to-end on `release/v0.10.x` due to two independent, stacked bugs in `go/core/cli/internal/cli/agent/invoke.go`. This PR fixes both.

### Bug 1: `a2a-go/v2` fails to decode `error.data` (relates to #2270)

`a2a-go/v2 v2.3.1` panics/fails when a JSON-RPC error response's `data` field is a single object instead of `[]*errordetails.Typed`. Bumping to `v2.5.0` resolves this (verified via `go.sum` diff against `main`, which already carries `v2.5.0`).

### Bug 2: empty `messageId` rejected by the server (untracked upstream)

`InvokeCmd` builds the outgoing `protocol.Message` as a raw struct literal:

```go
Message: protocol.Message{
    Kind:      protocol.KindMessage,
    Role:      protocol.MessageRoleUser,
    ContextID: sessionID,
    Parts:     []protocol.Part{protocol.NewTextPart(task)},
},
```

This never sets `MessageID`, leaving it as `""`. The A2A spec requires a non-empty `messageId`, so the controller rejects every request with:

```
-32602 invalid params: message ID is required
```

Root cause was confirmed by capturing raw HTTP traffic between the CLI and controller with a logging proxy, which showed `"messageId":""` in the outgoing request and the exact server-side rejection reason.

Fix: use the existing `protocol.NewMessageWithContext(...)` helper, which generates a valid ID via `protocol.GenerateMessageID()` internally. This mirrors what the rewritten CLI on `main` (`go/core/cli/internal/commands/agentinstance/invoke.go`, using `a2atype.NewMessage(...)`) already does correctly, so this brings `release/v0.10.x` in line with `main`'s behavior for this code path.

## Verification

Both fixes were built into a local controller image + CLI binary and tested end-to-end against a real kAgent deployment (Minikube + Vertex AI-backed agent). Before the fix: every `kagent invoke` call failed. After: full successful invocations returning `"status":{"state":"completed"}`.

## Notes

- Bug 1 is likely the same issue tracked in #2270, though that issue's discussion also raises a separate transport-version-selection theory that this PR does not address.
- Bug 2 does not appear to be tracked anywhere upstream; happy to file a separate issue if useful for tracking.


vainokurula-e:
Updated per CONTRIBUTING.md review:
- Added DCO `Signed-off-by` trailers to all 3 commits (one was missing).
- Added a regression test (`TestInvokeCmd_MessageHasID`) covering the exact defect — asserts `NewMessageWithContext` always produces a non-empty `MessageID`.
- Rebased onto the current tip of `release/v0.10.x` (no conflicts; the two new commits there are unrelated UI fixes).

Still marked as draft while I do a final self-review before requesting review.

